PR #20, opened 2018-11-09
chore: setup CircleCI

thymikee:
Let's kick the CI in


package.json
```diff
@@ -2,5 +2,9 @@
   "private": true,
   "workspaces": [
     "packages/*"
-  ]
+  ],
+  "scripts": {
+    "test": "cd packages/local-cli && yarn jest",
```
Esemesek: If we have to `cd` into the directory to run tests and flow-check, maybe we should go back after?
```sh
cd packages/local-cli && yarn jest && cd -
```

thymikee: nah, we only fire this on CI and it works in a container. but I'll add it

PR #26, opened 2018-11-10
Add (very basic) Yarn Plug 'n Play support

empyrical:
This pull request adds very basic support for Yarn PnP in the RN-CLI. It does not yet solve all parts (like the link, native build, bundle, eject) but what this does do is:

 * `global-cli`: If `pnpapi` is present, it will use that to look for the project's `react-native` package instead of looking in `node_modules`. I also tweaked it to look for `react-native-local-cli/cli.js` if it exists.
 * `local-cli`: `findReactNativePath` has been adjusted to support PnP

### DIFF
--- a/packages/global-cli/index.js
+++ b/packages/global-cli/index.js
@@ -45,6 +45,14 @@ var execSync = require('child_process').execSync;
 var chalk = require('chalk');
 var prompt = require('prompt');
 var semver = require('semver');
+var pnp = null;
+
+try {
+  pnp = require('pnpapi');
+} catch(e) {
+  // Not in PnP
+}
+
 /**
  * Used arguments:
  *   -v --version - to print current version of react-native-cli and react-native dependency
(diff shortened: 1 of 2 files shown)


empyrical:
@grabbou Hmm, hold off on merging this. I think I might turn this into a bigger PR (Metro PnP support might be something we can do without Metro needing do do stuff on their end)

grabbou:
Sure. We should rebase this when my other PR gets merged as I removed `findReactNativePath` entirely.

grabbou:
@empyrical, now that things have been merged, do you feel like rebasing your PR and working on it one more time?

What would be the best way to reproduce it locally and test the new workflow?

thymikee:
Also is there a way the `pnpapi` package could be open sourced? cc @arcanis

empyrical:
@grabbou I'm going to close this and make another one, as this is way out of date now. When I try this again, I can make another pull request

Now that this project has CircleCI set up, I could make tests in that pull request as well.

@thymikee The `pnpapi` package is open sourced - when a project's `.pnp.js` has been installed, all calls to `require('pnpapi')` are redirected to the `.pnp.js` file.

https://github.com/yarnpkg/yarn/blob/fb03788279278084e81aac9cde40151a3134761b/src/util/generate-pnp-map-api.tpl.js#L606-L610

Also - when you run a command through Yarn, like running `yarn run react-native` or `yarn node`, it will automatically check if PNP is being used in your project. If it does, it will call node with the `--require` argument, like so:

```sh
node --require /path/to/.pnp.js
```

(to be more precise, it sets this option by setting the `NODE_OPTIONS` environment variable, so hypothetically any node subprocesses you spin up might also work with PNP)

This will allow Node to automatically try and use the PNP file as a require hook. So that is why the `pnpapi` "module" works here, even though the `pnpapi` module on NPM is basically just a placeholder (possibly to prevent abuse)

What I need to experiment with is: when you install `react-native-cli` globally with Yarn, will it still automatically find a `.pnp.js` file? If not, `global-cli` will need to look for and require `.pnp.js` directly rather than just requiring `pnpapi`

thymikee:
Thank you @empyrical. Don't bother about `react-native-cli`, it's gonna be killed soon anyway :)

PR #28, opened 2018-11-15
fix: Fix opening dev tools

Esemesek:
## The issue
Opening `dev tools` was causing opening of `chrome` browser on a wrong URL.

## Current behavior
Dev tools are opened on default `port` and `host`. Default `port` is `8081` and `host` is `localhost`.
If you want to change the default behavior, you can do it by providing `--host` and `--port` cmd arguments.

### DIFF
--- a/packages/local-cli/server/middleware/getDevToolsMiddleware.js
+++ b/packages/local-cli/server/middleware/getDevToolsMiddleware.js
@@ -12,8 +12,8 @@ const launchChrome = require('../util/launchChrome');
 
 const {exec} = require('child_process');
 
-function launchChromeDevTools(host, post, args = '') {
-  var debuggerURL = `http://${host}:${post}/debugger-ui${args}`;
+function launchChromeDevTools(host, port, args = '') {
+  var debuggerURL = `http://${host}:${port}/debugger-ui${args}`;
   console.log('Launching Dev Tools...');
   launchChrome(debuggerURL);
 }
--- a/packages/local-cli/server/middleware/getDevToolsMiddleware.js
+++ b/packages/local-cli/server/middleware/getDevToolsMiddleware.js
@@ -12,8 +12,8 @@ const launchChrome = require('../util/launchChrome');
 
 const {exec} = require('child_process');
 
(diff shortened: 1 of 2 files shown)


packages/local-cli/server/middleware/getDevToolsMiddleware.js
```diff
@@ -12,8 +12,8 @@ const launchChrome = require('../util/launchChrome');
 
 const {exec} = require('child_process');
 
-function launchChromeDevTools(host, args = '') {
-  var debuggerURL = 'http://' + host + '/debugger-ui' + args;
+function launchChromeDevTools(host, port, args = '') {
+  var debuggerURL = `http://${host}:${port}/debugger-ui${args}`;
```
empyrical: (nit) you could probably turn this into a `const` while this part is being modified

grabbou: Oh, this "suggested change" is nice!

grabbou: @Esemesek please resolve this one and we will ship!

grabbou:
Thank you for sending this PR! 

Please let me know how was your experience contributing to this project. Was it easy to get it all set up? How did you test it inside `react-native` project? Anything we can do better?

packages/local-cli/server/middleware/getDevToolsMiddleware.js
thymikee: @grabbou it's fixed in https://github.com/react-native-community/react-native-cli/pull/28/commits/6a727c3567ac210c60e45a5ac71c05eb9a58c201

grabbou: Interesting why it doesn't say "dimissed" then

thymikee: Github UI changed a bit and I think it doesn't auto-fold the outdated conversations, unless you mark them so, it just adds the `Outdated` label at the top of the diff.

PR #45, opened 2018-12-12
Add repository URLs to package.json files

fson:
This makes it easier to find the new repo of these packages through npm (it's becomes a link at https://www.npmjs.com/package/react-native-cli).


grabbou:
Thank you very much!

PR #52, opened 2018-12-19
Add ESLint configuration

grabbou:
This PR adds ESLint configuration with Prettier and automatically fixes most of the issues. Some of the changes had to be done manually.

ESLint is now part of our testing pipeline.

3 warnings remaining for disabled tests, to be implemented in upcoming PRs.

I excluded "debugger-ui" folder from ESLint based on historical issues we had with running ESLint in Haul on that folder. This runs in browser context "as-is" without transpilation and should not be changed.

"templates" folder will be gone in favour of the one present in React Native master (to be done as a next PR).


grabbou:
Requesting  @Esemesek and @thymikee review.

packages/global-cli/index.js
```diff
@@ -36,15 +36,12 @@
 // /!\ DO NOT MODIFY THIS FILE /!\
 // ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-'use strict';
```
thymikee: removing `'use strict'` requires adding `@babel/plugin-transform-strict-mode` to babel config.

grabbou: Gotcha. Going to update that.

grabbou: ESLint configuration is set to assume we are dealing with ES6 modules (implied strict mode as per spec) where in fact, we are not. Gotta change that and Babel plugin is a good one too.

packages/local-cli/core/__fixtures__/projects.js
```diff
@@ -34,4 +34,4 @@ const withPods = {
   ios: ios.pod,
 };
 
-module.exports = {flat, nested, withExamples, withPods};
```
thymikee: so we go with non-fb style of extra spacing in destructuring statements?

packages/local-cli/core/findPlugins.js
```diff
@@ -91,13 +89,11 @@ const findPluginsInFolder = folder => {
...
     Object.keys(pjson.dependencies || {}),
-    Object.keys(pjson.devDependencies || {}),
+    Object.keys(pjson.devDependencies || {})
   );
 
   return deps.reduce((acc, pkg) => {
-    let commands = acc.commands;
-    let platforms = acc.platforms;
-    let haste = acc.haste;
+    let { commands, platforms } = acc;
```
thymikee: `haste` could be retrieved here

grabbou: Yes, but this would cause another error: using "let" where "haste" is not re-assigned. Wanted to avoid double-destructing statement for the sake of it.

packages/local-cli/core/__fixtures__/projects.js
grabbou: Back to the roots! https://github.com/rnpm/rnpm/blob/master/src/config/android/index.js#L110

On a serious note: I don't have preference here. Just sticking to whatever ESLint preset is recommending. I guess we can always change that in case we agree on a general preset for all RNCommunity projects - right now, it's pretty much up to project maintainers.

grabbou:
Thanks @thymikee - that's right! Luckily, most of the changes were performed automatically by ESLint, so I am quite confident they will not cause major issues.

packages/local-cli/bundle/buildBundle.js
```diff
@@ -10,8 +10,6 @@
 import type { ContextT } from '../core/types.flow';
 import type { CommandLineArgs } from './bundleCommandLineArgs';
 
-('use strict');
```
satya164: You probably shouldn't remove `'use strict'` if you want your scripts to run in strict mode as you're not using ES6 modules.

thymikee: This is what's used: `@babel/plugin-transform-strict-mode`. Does it matter which sourceType we use when having that?

thymikee: Yea, I requested it :D glad you're up to date now :)

PR #56, opened 2018-12-19
Fix metro config when running from source

### LINKED ISSUE #68: assetExts passed to `react-native start` as a command line argument is ignored
I found this bug in React Native v0.57 and even though the Metro config handling in `react-native start` has changed a bit since then, it looks like this bug exists in `react-native-local-cli` as well.

It seems that we need to handle `assetExts` here, the same way `sourceExts` are handled: https://github.com/react-native-community/react-native-cli/blob/6de63fed2c931ba15296b3d51d85332030d741fd/packages/local-cli/server/runServer.js#L55

grabbou:
Metro depends on some private React Native files when starting the server. We are using "require.resolve", which does not work when we actually run CLI from within React Native master (there's no "react-native" package in "node_modules").

I wanted to avoid hacking around "require.resolve" and paths to support that particular use-case. Doing that might become brittle in the future when we add PnP support etc., where "node_modules" awareness might become a problem.

I consider this a temporary hack and hopefully we can do better in the future, e.g. by creating a special package, such as: "react-native-bundler-compat", so that other bundles, such as Webpack can also use these files.

To make this working in React Native, the following "metro.config.js" has to be created:
```
module.exports = {
  serializer: {
    getModulesRunBeforeMainModule: () => [
      require.resolve('./Libraries/Core/InitializeCore'),
    ],
    getPolyfills: require('./rn-get-polyfills'),
  },
  resolver: {
    hasteImplModulePath: require.resolve('./jest/hasteImpl'),
  },
  transformer: {
    assetRegistryPath: require.resolve(
      './Libraries/Image/AssetRegistry'
    ),
  },
};
```
to overwrite the default paths.

Also, fixes #68 as we don't overwrite config ourselves but pass it straight to Metro.

### DIFF
--- a/packages/local-cli/util/loadMetroConfig.js
+++ b/packages/local-cli/util/loadMetroConfig.js
@@ -54,8 +54,7 @@ const getDefaultConfig = (root: string) => {
       getModulesRunBeforeMainModule: () => [
         require.resolve('react-native/Libraries/Core/InitializeCore'),
       ],
-      getPolyfills: (...args) =>
-        require('react-native/rn-get-polyfills')(...args),
+      getPolyfills: () => require('react-native/rn-get-polyfills')(),
     },
     server: {
       port: process.env.RCT_METRO_PORT || 8081,
@@ -80,40 +79,6 @@ export type ConfigOptionsT = {
 };
 
 /**
- * Overwrites Metro configuration with options.
- *
- * This ensures that options are always taking precedence over other
- * configuration present.
- */
-const overwriteWithOptions = (config: ConfigT, options: ConfigOptionsT) => {
-  if (options.maxWorkers) {
-    config.maxWorkers = options.maxWorkers;
-  }
-
-  if (options.port) {
-    config.server.port = options.port;
-  }
-
-  if (options.reporter) {
-    config.reporter = options.reporter;
-  }
-
-  if (options.resetCache) {
-    config.resetCache = options.resetCache;
-  }
-
-  if (options.watchFolders) {
-    config.watchFolders = options.watchFolders;
-  }
-
-  if (options.sourceExts && options.sourceExts !== config.resolver.sourceExts) {
-    config.resolver.sourceExts = options.sourceExts.concat(
-      config.resolver.sourceExts
-    );
-  }
-};
-
-/**
  * Loads Metro Config and applies `options` on top of the resolved config.
  *
  * This allows the CLI to always overwrite the file settings.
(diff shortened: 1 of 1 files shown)


packages/local-cli/util/loadMetroConfig.js
```diff
@@ -135,6 +132,29 @@ module.exports = async function load(
...
+   * statically. They can't be a function, unline `serializer.getModulesRunBeforeMainModule`.
+   *
+   * We only resolve React Native files if no other value was provided by the user.
+   *
+   * This is to allow using CLI inside React Native source code, where there is no
+   * "react-native" module in "node_modules".
+   *
+   * See "react-native/metro.config.js"
+   */
+  if (!config.resolver.hasteImplModulePath) {
```
grabbou: The idea is when React Native provides its "defaults" in "metro.config.js", this code will not execute and hence, not cause an error.

I was a bit confused coming up with a good documentation for that code (lines 135-145) - if anyone feels like this can be described in a better way, please comment.

thymikee: Couldn't find the `react-native/metro.config.js` file you're talking about, can you link to it?

packages/local-cli/util/loadMetroConfig.js
```diff
@@ -135,6 +132,29 @@ module.exports = async function load(
...
+    config.resolver.hasteImplModulePath = require.resolve(
+      'react-native/jest/hasteImpl'
+    );
+  }
+
+  if (!config.transformer.assetRegistryPath) {
```
Kudo: metro seems have a default assetRegistryPath, so this check seems not enough.
```js
if (!config.transformer.assetRegistryPath || config.transformer.assetRegistryPath === 'missing-asset-registry-path')
```

My env & use case:
1. current RN master code (https://github.com/facebook/react-native/commit/5957c8b370fd44fd0439461dee1e7d9eab806866)
2. react-native-local-cli@1.0.0-alpha.5
3. Apply this PR's patch with some minor change, that replace `require('react-native/foo')` to `require('../../../foo')`.
In case the path is `/path/to/react-native/node_modules/react-native-local-cli/util/loadMetroConfig.js` 
Patch file at https://gist.github.com/Kudo/a51f7992d6d1405330a3dcbb1d9e90fd
4. Running ./scripts/packager.sh and test with RNTester Android

grabbou:
Removed custom code to overwrite config with `argv` as it is already part of Metro: https://github.com/facebook/metro/blob/8ee697c49d2ed78e0e12ae3b6969b8e94b9630e4/packages/metro-config/src/loadConfig.js#L241

grabbou:
Applied some minor changes - the code is ready to go.

packages/local-cli/util/loadMetroConfig.js
```diff
@@ -49,13 +49,12 @@ const getDefaultConfig = (root: string) => {
...
         ...plugins.haste.providesModuleNodeModules,
       ],
-      hasteImplModulePath: require.resolve('react-native/jest/hasteImpl'),
     },
     serializer: {
       getModulesRunBeforeMainModule: () => [
         require.resolve('react-native/Libraries/Core/InitializeCore'),
       ],
-      getPolyfills: require('react-native/rn-get-polyfills'),
+      getPolyfills: () => require('react-native/rn-get-polyfills')(),
```
thymikee: why this change? we don't want to pass any args?

grabbou: I've checked Metro configuration and `rn-get-polyfills` and the function does not receive/accept any arguments: https://github.com/facebook/react-native/blob/master/rn-get-polyfills.js#L12

Kudo:
Thanks @grabbou, hope to release a new version of local-cli.
RN Android master branch code broke on metro upgrade breaking change for a while.

PR #58, opened 2018-12-20
fix: Re-enable some unit tests

### LINKED ISSUE #54: Few unit tests are disabled
There are 3 unit tests that are disabled for historical reasons. Unfortunately. I don't really remember what was the original reason.

Let's re-enable them (might need some little adjustments).

Esemesek:
Re-enable skipped tests.
Fixes #54

### DIFF
--- a/packages/local-cli/link/__tests__/ios/removeProjectFromProject-test.js
+++ b/packages/local-cli/link/__tests__/ios/removeProjectFromProject-test.js
@@ -35,9 +35,4 @@ describe('ios::addFileToProject', () => {
     const file = removeProjectFromProject(project, filePath);
     expect(project.pbxFileReferenceSection()[file.fileRef]).not.toBeDefined();
   });
-
-  xit('should remove file from PBXContainerProxy', () => {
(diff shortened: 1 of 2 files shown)


packages/local-cli/link/__tests__/ios/removeProjectFromProject-test.js
```diff
@@ -35,9 +35,4 @@ describe('ios::addFileToProject', () => {
     const file = removeProjectFromProject(project, filePath);
     expect(project.pbxFileReferenceSection()[file.fileRef]).not.toBeDefined();
   });
-
-  xit('should remove file from PBXContainerProxy', () => {
-    // todo(mike): add in .xcodeproj after Xcode modifications so we can test extra
```
thymikee: Maybe worth marking this test as a todo? In jest 24 with circus runner you can use test.todo for that

Esemesek: You are probably right. I didn't know about that feature. 👍

thymikee:
Closed and reopened to trigger CI as I've just turned on CI checks from forks

PR #59, opened 2018-12-29
Prevent crash in findMatchingSimulator

zubko:
`findMatchingSimulator` could crash if the simulator's name is custom (renamed in Xcode) for ex: `iPhone6v11`.

After this fix it doesn't crash and the correct custom named simulator is chosen.

### DIFF
--- a/packages/local-cli/runIOS/findMatchingSimulator.js
+++ b/packages/local-cli/runIOS/findMatchingSimulator.js
@@ -31,7 +31,7 @@ function findMatchingSimulator(simulators, simulatorString) {
   const parsedSimulatorName = simulatorString
     ? simulatorString.match(/(.*)? (?:\((.*)?\))?/)
     : [];
-  if (parsedSimulatorName[2] !== undefined) {
+  if (parsedSimulatorName && parsedSimulatorName[2] !== undefined) {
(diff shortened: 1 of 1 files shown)


packages/local-cli/runIOS/findMatchingSimulator.js
```diff
@@ -31,7 +31,7 @@ function findMatchingSimulator(simulators, simulatorString) {
   const parsedSimulatorName = simulatorString
     ? simulatorString.match(/(.*)? (?:\((.*)?\))?/)
     : [];
-  if (parsedSimulatorName[2] !== undefined) {
+  if (parsedSimulatorName && parsedSimulatorName[2] !== undefined) {
```
grabbou: This is interesting. 

On line 31, we make sure it's either a matched value or an array - it shouldn't be "undefined" at all.

Could you provide more details when this would be null, or causing a crash?

grabbou: Ah, it confused me:
```
simulatorString
    ? simulatorString.match(/(.*)? (?:\((.*)?\))?/)
    : [];
```
Thought we check on `simulator.match` instead xd

PR #63, opened 2019-01-08
Fix launching chrome dev tools

janicduplessis:
We didn't pass the port to the middleware manager which is needed by the `getDevToolsMiddleware` to launch chrome dev tools.


grabbou:
Awesome @janicduplessis, thank you for this PR! Just in time for the release!

Issue #48, opened 2018-12-15
Make package name required for "link"


grabbou:
When you don't pass a package name to the "react-native link", it will try to look for all packages to link from your node_modules.

This is causing us a lot of issues, such as linking wrong packages or the ones that someone doesn't want to link (e.g. decided to use a different way of linking that is out of scope for this command).

We should simply stick to requiring package name.

grabbou:
Please leave a comment if you are using `react-native link` to link your project assets and describe your use case. This behaviour will be also deprecated since the name of the package is going to be required.

ltcaosj:
I would like to be able to link my assets to iOS and Android projects.
I can do it with RN 0.57.8, but I cannot with RN 0.59.x
<img width="558" alt="link_assets1" src="https://user-images.githubusercontent.com/5376380/53001585-9355fe00-345d-11e9-877a-3472b36b5f42.png">
<img width="427" alt="link_assets" src="https://user-images.githubusercontent.com/5376380/53001591-97821b80-345d-11e9-927c-219adc944141.png">

Esemesek:
Yup. We plan to introduce a new command or flag that will link only project assets. For now, you should be able to link project assets with `yarn link`, but it is deprecated.

ltcaosj:
thanks @Esemesek

sijad:
currently it's not possible to link root assets with RN 59 rc2.

I've changed `() => linkAssets(platforms, project, assets)` to ` () => linkAssets(platforms, project, { assets }),` in [linkAll](https://github.com/react-native-community/react-native-cli/blob/master/packages/cli/src/link/linkAll.js#L47)
and it worked, but I'm not sure if it's proper fix.

Esemesek:
@sijad This is a bug. `linkAll` should link root assets. I'll fix that in a minute.

kennetpostigo:
I use it to link my font assets

Esemesek:
@kennetpostigo in the future there will be a separate command to link project assets.

baranonen:
I think that when "react-native link" is entered, it must ask if the user really wants to link all libraries. This way, it will be better for the people who installed more than one library and wants to link all of them, or if the user is new to React Native, this will be easier for him/her.

fbartho:
We have `react-native link` setup in our PostInstall script so that when we update our dependencies any changes to the linking instructions get applied correctly.

We do not have Cocoapods, so we don't want there to be dependencies that are partially linked.

[...]

PR #64, opened 2019-01-08
imp: Make package name a required argument in link

Esemesek:
Make `packageName` required in `link` command.
Fixes: #48

### DIFF
--- a/packages/local-cli/link/__tests__/__snapshots__/getDependencyConfig-test.js.snap
+++ b/packages/local-cli/link/__tests__/__snapshots__/getDependencyConfig-test.js.snap
@@ -1,41 +1,19 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`getDependencyConfig should filter out invalid react-native projects 1`] = `
-Array [
-  Object {
-    "assets": Array [],
-    "commands": Object {},
-    "config": Object {
-      "android": Object {
-        "sampleAndroidKey": "",
-      },
-      "ios": Object {
-        "sampleiOSKey": "",
-      },
-    },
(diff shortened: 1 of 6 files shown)


thymikee:
What about linking assets like fonts? Is this still supported somehow?
Here's example of usage: https://medium.com/react-native-training/react-native-custom-fonts-ccc9aacf9e5e#bcc2
cc @grabbou

grabbou:
Yes, it is still supported. To be honest, I wonder if we should keep it. Linking only a single dependency but, at the same time, linking all the assets, looks to me like a really badly designed behaviour.

I wonder whether we should:
- allow running `link` without a package name that would only link assets from the project itself (like in the document you presented). The downside of this direction is that it's confusing for those who were running `react-native link` without a name before.
- allow running `link` with the project name, e.g `react-native link HelloWorld` where HelloWorld is the `name` inside `package.json` of the project. That would link assets only, since native files are already linked.

ferrannp:
Definitely linked assets is needed. 

> Linking only a single dependency but, at the same time, linking all the assets

That's weird yes, unless the single dependency has assets.

I think something explicit is better, just like:

```sh
react-native link --assets
```

thymikee:
Or linking assets (or whatever there is in the config file) would be performed when no name is passed.

grabbou:
> Or linking assets (or whatever there is in the config file) would be performed when no name is passed.

Problem with this is that it's an implicit breaking change. People running `react-native link` and not being aware of this change, will report problems of native files not being linked.

Since we are forcing the name, I guess the better approach is to require the name of the project itself or... create another command (probably unwanted).

PR #67, opened 2019-01-09
fix: --appFolder should only install specific app on Android.

### LINKED ISSUE #32: --appFolder on Android not really working as expected
Let's say you have more than 1 app in your android directory. When you do:

```bash
react-native run-android --appFolder someAppFolder
```

It is installing all apps. This is because under the hood the cli is running `./gradlew installDebug`. (or whatever variant). Instead it should be:

```bash
./gradlew :someAppFolder:installDebug
```

ferrannp:
Fixes #32.

Extracting bunch of functions to its own files too + adding tests.

For not being a breaking change, default behavior (without passing `--appFolder`) will be just to call `installDebug`.

### DIFF
--- a/packages/local-cli/runAndroid/runOnAllDevices.js
+++ b/packages/local-cli/runAndroid/runOnAllDevices.js
@@ -16,8 +16,8 @@ const adb = require('./adb');
 const tryRunAdbReverse = require('./tryRunAdbReverse');
 const tryLaunchAppOnDevice = require('./tryLaunchAppOnDevice');
 
-function getInstallCommand(appFolder, install = 'install') {
-  return appFolder ? `${appFolder}:${install}` : install;
+function getCommand(appFolder, command) {
+  return appFolder ? `${appFolder}:${command}` : command;
 }
 
 function runOnAllDevices(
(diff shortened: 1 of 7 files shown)


thymikee:
Mind adding a test for that?

ferrannp:
Is there any test for this whole file?

thymikee:
Dunno, but it's worth to add it anyway, would make for an amazing contribution :)

ferrannp:
Ok, gonna give it a chance. Then try to focus on other super urgent issues before release 😄

packages/local-cli/runAndroid/runOnAllDevices.js
```diff
@@ -0,0 +1,117 @@
...
+  packageName,
+  adbPath
+) {
+  try {
+    const gradleArgs = [];
+    if (args.variant) {
+      gradleArgs.push(
+        `${
+          args.appFolder
+        }:install${args.variant[0].toUpperCase()}${args.variant.slice(1)}`
```
grabbou: This change has a minor breaking change that we should be aware of. I am not sure which approach is the desired one - we should probably check it and decide what is the way we want to keep it.

`args.appFolder` is an optional CLI flag that defaults to "app". Prior to this PR, we were not using this option at all, making it impossible to change the app to install on the device, when multiple are defined. 

The command executed by Gradle was always e.g: "installDebug", where in reality, it should've been "app:installDebug", since the "appFolder" defaults to "app". 

That means, after this PR is merged, all the default executions will contain "app:". This might be breaking for people that were depending on this undocumented bug.

I wonder whether we should stick to it, or remove the default "app" value of "appFolder" and make it either "appFolder:installDebug" or "installDebug", depending on whether it is passed to the CLI.

janicduplessis: Yea I think it would be better to default to keep running `installDebug` if `appFolder` is not specified.

ferrannp: Seems reasonable, I will make that change.

PR #72, opened 2019-01-10
fix: no need from cd - on package.json so we can pass custom arguments

ferrannp:


thymikee:
This will change your working directory after running tests.
Time to setup Jest `projects` and run tests from the root ;)

ferrannp:
@thymikee it doesn't locally. Isn't that the default behavior of `npm:scripts` ?

thymikee:
If it works, then cool, I'm not aware of such feature :D

grabbou:
@thymikee it depends. Each command on CircleCI runs in its own session, so it always starts from the "workingDirectory". If we have a separate "run" phase for "npm test", it should be good regardless of the "npm scripts" behaviour.

grabbou:
I am going to merge it and we should fix them.

PR #77, opened 2019-01-10
feat: be able to use link for specific --platforms only

### LINKED ISSUE #25: Make link work for specified platform only
Basically
```sh
react-native link --platform=ios
```

ferrannp:
Closes #25.

Default behavior is still to link everything. We can also use:

```sh
--platforms=android
--platforms=ios
--platforms=android,ios
// etc
```


ferrannp:
`master` has broken tests right now (related to `link`).

thymikee:
@ferrannp can you rebase? Flow was fixed in #79

packages/local-cli/link/__tests__/link-test.js
```diff
@@ -226,9 +226,76 @@ describe('link', () => {
...
-    link(['react-native-blur'], context).then(() => {
+    link(['react-native-blur'], context, {}).then(() => {
       expect(registerNativeModule.mock.calls).toHaveLength(1);
       done();
     });
   });
+
+  it('should link only for specific platforms if --platforms is used', async () => {
+    jest.setMock('../getProjectDependencies', () => ['react-native-maps']);
+    jest.setMock('../../core/getPackageConfiguration', () => ({
```
thymikee: can you move `jest.mock` or `jest.doMock` instead?

packages/local-cli/link/__tests__/link-test.js
```diff
@@ -226,9 +226,76 @@ describe('link', () => {
...
+    assertPlaftormsCalledTimes(1, 0);
+
+    await link(
+      ['react-native-gradient'],
+      { root: '/' },
+      { platforms: ['ios'] }
+    );
+    assertPlaftormsCalledTimes(0, 1);
+
+    await link(
```
thymikee: this link call is a duplicate of https://github.com/react-native-community/react-native-cli/pull/77/files?w=1#diff-f71ce3341c36eb9ac6eefd5606744ff6R273, can you remove it?

packages/local-cli/link/link.js
```diff
@@ -99,16 +99,23 @@ const linkAssets = (platforms, project, dependency) => {
...
  * @param args [packageName]
  */
-function link([rawPackageName]: [string], ctx: ContextT) {
+function link([rawPackageName]: [string], ctx: ContextT, opts: FlagsType) {
   let platforms;
   let project;
   try {
     platforms = getPlatforms(ctx.root);
+    if (opts.platforms) {
+      platforms = pick(platforms, opts.platforms);
```
thymikee: would be better to move this logic into `getPlatforms` so we don't have to overwrite

packages/local-cli/link/link.js
```diff
@@ -157,4 +164,12 @@ module.exports = {
   description:
     'links native dependencies for specified package (updates native build files)',
   name: 'link <packageName>',
+  options: [
+    {
+      command: '--platforms [list]',
+      description:
+        'If you want to link dependencies only for specific platforms',
```
thymikee:
```suggestion
        'scope link command to certain platforms (comma-separated)',
```

packages/local-cli/link/link.js
```diff
@@ -157,4 +164,12 @@ module.exports = {
...
+      parse: (val: string) => val.split(','),
```
thymikee: would be cool to also lowercase the items, so stuff like this also passes:
```
link --platforms iOS
link --platforms ios,Android
```

packages/local-cli/link/link.js
```diff
@@ -99,16 +99,23 @@ const linkAssets = (platforms, project, dependency) => {
   log.info('Assets have been successfully linked to your project');
 };
 
+type FlagsType = {
+  platforms: string,
```
thymikee:
```suggestion
  platforms: Array<string>,
```

packages/local-cli/link/__tests__/link-test.js
ferrannp: That works everywhere except here:

```js
jest.setMock('../../core/getPlatforms', () => ({
  android: { linkConfig: genericAndroidLinkConfig },
  ios: { linkConfig: genericIOSLinkConfig },
}));
```

Changing to `doMock` makes the test fail, do you know why? By the way, all the other tests from this file use also `setMock` so not sure that might be affecting the others somehow...

packages/local-cli/link/__tests__/link-test.js
ferrannp: Oh you are right! One of them should be empty array.

packages/local-cli/link/link.js
ferrannp: So that was my first idea, pass a second argument to `getPlatforms` to somehow blacklist some platforms. On second though, maybe another file will need these `getPlatforms` and it might be better it stays with same functionality to always return all? Then the consumer (like `link`) can decide what to do with it. You know, to know overcomplicate the function.

packages/local-cli/link/__tests__/link-test.js
thymikee: I'd say get rid of as many `setMock` as you can. It's a legacy API and shouldn't be used (I'm not even sure what it does actually, this project is first time I encounter it in the wild)

packages/local-cli/link/link.js
thymikee: I think it would be perfectly valid for `getPlatforms` to decide which platforms to return based on "options" argument :)

packages/local-cli/link/__tests__/link-test.js
ferrannp: Done 💪

PR #79, opened 2019-01-10
fix: Fix flow erros

Esemesek:
Fixing flow errors. In some places, I removed `@flow` annotation because its module was never entirely typed, and adding types would require much more effort, than this simple fix.

EDIT: Also removed `dependencies` command since it exists already in `metro`.
PR: https://github.com/facebook/metro/pull/335
Renaming cmd in `metro`: https://github.com/facebook/metro/pull/339


thymikee:
Merging to unbreak master, but the comments remain valid and I'd appreciate some discussion/followup :)

PR #81, opened 2019-01-11
fix: android duplicates in linked packages

### LINKED ISSUE #43: Fix duplicate Android packages being linked
Recent commit updated Android part of the link mechanism to use "implementation" instead of "compile" to align with what Gradle recommends: https://github.com/react-native-community/react-native-cli/commit/c6c34a2a6e0d5898a62bc10b7c2936cadba3e071

Unfortunately, it breaks the detection of an already installed packages.

The way "link" detects an installed package on Android is by looking for "compile(XXX)" inside `build.gradle`. We don't have a sophisticated tool for parsing and reading Gradle files, hence we use a really simple RegExp to look for informations that we want. It was always meant to be a temporary solution, so we are going to work on resolving that in the nearest future.

Meanwhile, what happens is that now, after the update, "link" will look for "implementation(XXX)" string instead to indicate that the given package is already linked. 

This means that when "link" command is run on an existing codebase that still uses "compile" (e.g. just updated to a new project), it will not treat such project as installed and will add it once again, this time using `implementation(XXX)`. This results with duplicate packages. 

We should teach `link/android/isInstalled.js` to return "true" in both scenarios when "compile" and "implementation" are present. Both are valid right now in Gradle v3 (with "compile" being obsolete).

We should also make sure that "unlink" can handle both cases.

CC: @thymikee @dulmandakh (original reviewer of the PR)

thymikee:
Fixes #43

- Improved the Regex to be more forgiving in terms of spacing and `implementation`/`compile` names with variants and configs
- add more thorough test
- cleanup makeBuildPatch test to not test implementation details
- cleanup link test so it doesn't output to stdout unnecessarily
- tested on a project I had issues with duplicates (resolved with this diff)

cc @ferrannp

### DIFF
--- a/packages/local-cli/link/__tests__/android/isInstalled-test.js
+++ b/packages/local-cli/link/__tests__/android/isInstalled-test.js
@@ -19,11 +19,20 @@ const projectConfig = {
 };
 
 describe('android::isInstalled', () => {
-  it('should return true when project is already in build.gradle', () => {
-    expect(isInstalled(projectConfig, 'test')).toBeTruthy();
-    expect(isInstalled(projectConfig, 'test2')).toBeTruthy();
-  });
-
-  it('should return false when project is not in build.gradle', () =>
-    expect(isInstalled(projectConfig, 'test3')).toBeFalsy());
+  test.each([
+    ['test-impl', true],
+    ['test-impl-config', true],
+    ['test-impl-config-spaces', true],
+    ['test-impl-debug', true],
+    ['test-impl-abc', true],
+    ['test-compile', true],
+    ['test-compile-debug', true],
+    ['test-compile-abc', true],
+    ['test-not-there-yet', false],
+  ])(
+    'properly detects if %p project is already in build.gradle',
+    (project, isPresent) => {
+      expect(isInstalled(projectConfig, project)).toBe(isPresent);
+    }
+  );
 });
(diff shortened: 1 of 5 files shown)


packages/local-cli/link/__tests__/android/isInstalled-test.js
```diff
@@ -19,11 +19,20 @@ const projectConfig = {
...
+    ['test-impl', true],
+    ['test-impl-config', true],
+    ['test-impl-config-spaces', true],
+    ['test-impl-debug', true],
+    ['test-impl-abc', true],
+    ['test-compile', true],
+    ['test-compile-debug', true],
+    ['test-compile-abc', true],
+    ['test-not-there-yet', false],
+  ])(
```
ferrannp: dope

grabbou:
Awesome!

matt-block:
@grabbou thanks for making me aware of this PR!

These changes are solid but it doesn't seem to address the `unlink` issue. `makeBuildPatch` still returns an "implementation" patch string, so a previously linked "compile" dependency would not be successfully unlinked.

Let's say we want to unlink a "compile" dependency that was linked way back. When the [`revokePatch`](https://github.com/react-native-community/react-native-cli/blob/master/packages/cli/src/commands/link/android/patches/revokePatch.js) function in called in [`unregisterNativeModule.js#L41`](https://github.com/react-native-community/react-native-cli/blob/master/packages/cli/src/commands/link/android/unregisterNativeModule.js#L41) it will look for an "implementation" string to replace with an empty string but, since we have a "compile" dependency, our dependency entry in the Gradle file will never be found and thus it would not be removed.

PR #84, opened 2019-01-11
build: use template from react-native

Esemesek:
Remove `templates` from `react-native-cli` and use template from `react-native` repo.

Tests:
Used https://github.com/verdaccio/verdaccio to create local distributions of `react-native` and `react-native-local-cli` and initialized project using those packages


grabbou:
We will need to update path here: https://github.com/react-native-community/react-native-cli/blob/master/packages/local-cli/upgrade/upgrade.js#L143 too.

grabbou:
I was unable to get it up & running locally, so let's go through it once again, when the above is resolved too.

packages/local-cli/generator/templates.js
```diff
@@ -25,21 +25,20 @@ function createProjectFromTemplate(
   template,
   yarnVersion
 ) {
-  // Expand the basic 'HelloWorld' template
-  copyProjectTemplateAndReplace(
-    path.resolve(__dirname, '../templates/HelloWorld'),
-    destPath,
-    newProjectName
+  const templatePath = path.resolve(
+    process.cwd(),
```
fson: I think this needs to be `destPath` (the project directory, passed from `init()`) instead of the current working directory (which could be any dir where the command to create a project was run).

grabbou: actually we could just use `require.resolve('react-native/template')` instead, since we are in context of `node_modules`, right?

Note that `destPath` is still passed on line 32. The diff might be confusing, but we are only changing the "path.resolve(__dirname, '../templates/HelloWorld')," part.

fson: Yeah, `require.resolve` should work I think.

PR #85, opened 2019-01-12
build: deprecate link all behavior

Esemesek:
Deprecate `react-native link` without specifying `packageName`.

Tests:
1. Init new project with `react-native init $project_name`
2. Install a couple of native dependencies that need to be linked
3. Invoke `link` command
```
node path/to/cli/packages/local-cli/cli.js link
```

screenshot 2019-01-12 at 13 26 24

Modules are linked correctly but we get deprecation notice.

### DIFF
--- a/packages/local-cli/link/linkAll.js
+++ b/packages/local-cli/link/linkAll.js
@@ -39,13 +39,12 @@ function linkAll(
   );
 
   const tasks = flatten(
-    depenendenciesConfig
-      .map(config => [
-        () => promisify(config.commands.prelink || commandStub),
-        () => linkDependency(platforms, project, config),
-        () => promisify(config.commands.postlink || commandStub),
-        () => linkAssets(platforms, project, assets)
-      ])
+    depenendenciesConfig.map(config => [
+      () => promisify(config.commands.prelink || commandStub),
+      () => linkDependency(platforms, project, config),
+      () => promisify(config.commands.postlink || commandStub),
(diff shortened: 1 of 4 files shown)


packages/local-cli/link/linkAll.js
```diff
@@ -0,0 +1,53 @@
...
+  project: ProjectConfigT
+) {
+  log.warn(
+    'Linking modules without specifying package name is deprecated and will be removed in next release'
+  );
+  const dependencies = getProjectDependencies(context.root);
+  const depenendenciesConfig = dependencies.map(dependnecy =>
+    getDependencyConfig(context, platforms, dependnecy)
+  );
+  const assets = dedupeAssets(flatten(depenendenciesConfig.map(d => d.assets)));
```
grabbou: One thing that was there was "project.assets.concat". 

We should also link the project assets. That's the only reason we have a deprecation notice - to make sure those that rely on this behaviour are aware that this is going away.

I wonder whether we should include a notice: 

> "Use react-native link xyz" instead.

and

> If you are using "react-native link" to link assets from your project, let us know here: goo.gl/...

Where goo.gl/... is a link to an issue in this repository. I would like to understand how people are using this feature (and why) and be able to offer them a reasonable workaround.

grabbou:
BTW - we really need a proper logger!

grabbou:
Fixed one thing and added URL to an issue where users will be able to let us know about the deprecation.

PR #94, opened 2019-01-16
fix: add newline in package.json

cpojer:
The initial version of the package.json file does not contain a newline at the end. This means that the first time a user modifies the file using either an IDE that inserts trailing newlines or using Yarn, the last line will be changed and visible in version control. This is not a big deal, but it is annoying and shows a lack of polish.

I tested this by manually patching this file in a current version of React Native because there is currently no way to test the master of RN and RN cli together.


thymikee:
Mind calling `eslint --fix`?

cpojer:
Done! Sorry about that.

PR #100, opened 2019-01-18
build: remove dependencies command

### LINKED ISSUE #92: Use `dependencies` command from Metro
Now, that our PR is merged, we could easily use `dependencies` command from Metro, instead of the code we have here.

Esemesek:
Remove deprecated `dependencies` command, it was moved to metro as `get-dependencies`.
Fixes #92


grabbou:
Well, technically speaking, the "deprecation message" is not live yet - it should go out in 0.59 and we should remove it in 0.60.

PR #101, opened 2019-01-18
feat: remove `link` without package name; add `link-assets` commands

### LINKED ISSUE #48: Make package name required for "link"

Esemesek:
Removes deprecated behavior of `react-native link` without providing `packageName`.
Fixes #48 

Update:
Added `link-assets` command to support linking project assets.


thymikee:
@Esemesek mind rebasing?

grabbou:
Let's rebase and get that merged. I would like to note that there's few people that use `link` to link assets and it would be great to provide them a workaround when we ship this.

thymikee:
Oh, that's a valid point, we need to introduce `link-assets` (or so) before we can ship this

packages/cli/src/commands/link/__tests__/link-test.js
```diff
@@ -20,9 +20,9 @@ describe('link', () => {
     jest.resetModules();
   });
 
-  it('should reject when run in a folder without package.json', done => {
+  it('should reject when run in a folder without package.json', () => {
```
thymikee:
```suggestion
  it('should throw when run in a folder without package.json', () => {
```

packages/cli/src/commands/link/link.js
```diff
@@ -7,64 +7,52 @@
...
+    throw new ReactNativeNotFound();
   }
   const hasProjectConfig = Object.keys(platforms).reduce(
     (acc, key) => acc || key in project,
     false,
   );
   if (!hasProjectConfig && findReactNativeScripts()) {
     throw new Error(
-      '`react-native link [package]` can not be used in Create React Native App projects. ' +
+      '`react-native link <package>` can not be used in Create React Native App projects. ' +
```
thymikee: CRNA is deprecated, let's remove this check

packages/cli/src/commands/link/linkAssets.js
```diff
@@ -32,4 +38,33 @@ const linkAssets = (
...
+  ctx: ContextT,
+  opts: LinkFlagsType,
+) {
+  try {
+    const {platforms, project} = getPlatformsAndProject(ctx, opts);
+    const projectAssets = getAssets(ctx.root);
+
+    return linkAssets(platforms, project, projectAssets);
+  } catch (err) {
+    throw new ReactNativeNotFound();
```
thymikee: this is quite brutal and eats all errors. The `ReactNativeNotFound` check should live inside `linkAssets`, same as in `link`.

packages/cli/src/commands/link/types.flow.js
```diff
@@ -0,0 +1,5 @@
+// @flow
+
+export type LinkFlagsType = {
```
thymikee: Can we rename it to just `LinkOptions`?

packages/cli/src/commands/link/linkAssets.js
Esemesek: True. Didn't notice that the `Promise.reject` call had `err` carried over.

Esemesek:
Closing in favor of auto-linking.

PR #104, opened 2019-01-21
feat: build before publishing

thymikee:
Fixes https://github.com/react-native-community/react-native-cli/issues/7

To make building easier I moved `local-cli` JS sources to `src/` directory and compile them to a sibling `build/` – thanks to it we can use the same paths to get resources like `package.json` or bat/sh files.

We need to think about setting up lerna or something similar to have one build process shared across all packages.


grabbou:
Now that we are moving files around, maybe it's a good idea to change the `local-cli` name of the folder to correspond to the real package name under which it's published?

cpojer:
This is so good.

PR #105, opened 2019-01-21
chore: followup after #104

thymikee:


packages/cli/src/eject/eject.js
```diff
@@ -91,7 +92,8 @@ function eject() {
       path.resolve(
         'node_modules',
         'react-native',
-        'local-cli',
+        '@react-native-community',
+        'cli',
```
thymikee: looks like we don't have a correct test for this scenario

grabbou: True. I don't even know when this command is being called.

packages/cli/src/generator/copyProjectTemplateAndReplace.js
```diff
@@ -17,7 +17,7 @@ const walk = require('../util/walk');
  * Util for creating a new React Native project.
  * Copy the project from a template and use the correct project name in
  * all files.
- * @param srcPath e.g. '/Users/martin/AwesomeApp/node_modules/react-native/local-cli/templates/HelloWorld'
+ * @param srcPath e.g. '/Users/martin/AwesomeApp/node_modules/react-native/templates/HelloWorld'
```
grabbou: chore: `react-native/template` instead of `templates/HelloWorld`.

thymikee: fixd

PR #109, opened 2019-01-21
chore: setup monorepo with Lerna

thymikee:
Let's take it easy and make @grabbou test it thoroughly.

This essentially moves building script to the root `scripts/build.js` script (taken from Jest, so expect see some nice output 🙂) based on a fact that source files are under `src/` directory.


lerna.json
```diff
@@ -0,0 +1,6 @@
+{
+  "lerna": "3.10.6",
+  "version": "independent",
```
thymikee: Currently packages are a bit disconnected (local cli at 1.1 and global at 2.0) so until we reach parity Lerna will ask you which version would you like to bump to

PR #116, opened 2019-01-23
Removes calls to process.cwd() in favor of using context.root

### LINKED ISSUE #75: Unify "process.cwd()" calls
This is easy one.

After creating a "config.root" value and explicitly defining "projectRoot" flag for every command, we should make sure that all the commands use it, instead of "process.cwd()". 

Otherwise, it may give unexpected errors during runtime.

How to go about implementing it?

1) Make sure that none of the commands define their own "--root" or something. If they do, remove it and replace occurrences of it with "ctx.root" (look at "link" as an example)

2) Make sure that no commands call "process.cwd()" themselves and replace as in the above.

kpsroka:
This probably removes all the ones that can be removed, and closes #75.


thymikee:
There are still couple of places that use `process.cwd`, would you mind checking that? Can be done as a separate PR as well.

kpsroka:
@thymikee I think that all of the remaining ones need to stay. They are either outside of the packages, or only used to print suggested command line calls.

PR #119, opened 2019-01-23
Update runAndroid.js

arnabsen:
Remove non-existent 'react-native android' and revise with messaging consistent with iOS per @grabbou 's suggestion


grabbou:
Thank you!

PR #120, opened 2019-01-24
feat: use logger to cliEntry and bin files

Esemesek:
Use logger to `cliEntry.js` and `bin.js` files.


packages/cli/src/bin.js
```diff
@@ -9,20 +9,19 @@
...
+      'Looks like you installed react-native globally, maybe you meant react-native-cli? ',
+      'To fix the issue, run: ',
+      chalk.white(
+        'npm uninstall -g react-native && npm install -g react-native-cli'
       ),
-      chalk.red('To fix the issue, run:'),
-      'npm uninstall -g react-native',
-      'npm install -g react-native-cli',
-    ].join('\n')
+    ].join('')
```
thymikee: wouldn't it be better to `join('\n')` in this case?

packages/cli/src/cliEntry.js
```diff
@@ -76,17 +74,15 @@ function printHelpInformation() {
...
 function printUnknownCommand(cmdName) {
-  console.log(
+  logger.error(
     [
-      '',
       cmdName
-        ? chalk.red(`  Unrecognized command '${cmdName}'`)
-        : chalk.red("  You didn't pass any command"),
-      `  Run ${chalk.cyan(
+        ? `  Unrecognized command '${cmdName}'`
```
thymikee:
```suggestion
        ? `  Unrecognized command "${cmdName}"`
```

PR #123, opened 2019-01-24
fix: ignore non-nativetargets during ios linking

dotconnor:
fixes issue #57


Kielan:
+1 really good job imho

packages/cli/src/link/ios/getTargets.js
```diff
@@ -15,25 +15,31 @@ module.exports = function getTargets(project) {
...
+        target: nativeTargetSection[key],
+        name: nativeTargetSection[key].productReference_comment,
+        isTVOS:
+          (buildConfiguration.buildSettings.SDKROOT &&
+            buildConfiguration.buildSettings.SDKROOT.indexOf('appletv') !==
+              -1) ||
+          false,
+      };
+    })
+    .filter(target => target !== undefined);
```
grabbou: I think this is a case where a simple `for/loop` with `continue` would be much more readable. Anyway, it's just a tiny nit pick.

packages/cli/src/link/ios/getTargets.js
```diff
@@ -15,25 +15,31 @@ module.exports = function getTargets(project) {
...
-          buildConfiguration.buildSettings.SDKROOT.indexOf('appletv') !== -1) ||
-        false,
-    };
-  });
+  return targets
+    .map(target => {
+      const key = target.value;
+      // If not a valid target, skip.
+      if (!project.pbxNativeTargetSection()[key]) return undefined;
+      const configurationListId = project.pbxNativeTargetSection()[key]
```
grabbou: While we are doing this, let's make it: `nativeTargetSection[key]` since we already have native target section obtained on line 17.

packages/cli/src/link/ios/getTargets.js
```diff
@@ -15,25 +15,31 @@ module.exports = function getTargets(project) {
...
-        (buildConfiguration.buildSettings.SDKROOT &&
-          buildConfiguration.buildSettings.SDKROOT.indexOf('appletv') !== -1) ||
-        false,
-    };
-  });
+  return targets
+    .map(target => {
+      const key = target.value;
+      // If not a valid target, skip.
+      if (!project.pbxNativeTargetSection()[key]) return undefined;
```
grabbou: Let's flip `filter` and `map` order, e.g:
```js
targets.filter(target => nativeTargetSection[target.value] !== undefined).map(...)
```

grabbou: It would be also great to elaborate a bit more on the comment itself. For example, I would say: "Filter out targets that are ....". I am trying to understand what exactly is an invalid target (if it's in targets section and not in the native targets section - what does it mean).

dotconnor: I run into the problem whenever I have an aggregate target alongside the default ones created by the cli. From the looks of it, we only need to modify native targets when linking, so anything not in the Native Target section should be ignored.

grabbou: Oh, now I remember! Due to limitations of Xcode library, I had to enumerate native targets by getting an array of project targets first. But yeah, I wrongly assumed that all targets will be in that section, which is not the case.

dotconnor:
@grabbou 
I have an empty aggregate target that I can add to the test fixtures to ensure that they get removed if you want me to.

PR #128, opened 2019-01-25
fix: schema option on run-ios not installing and loading the right bundle

misoguy:
Summary:
---------
This PR adds `schema` folder to the ios build path
to resolve issue https://github.com/facebook/react-native/issues/20497

Test Plan:
----------
1. Setup a different scheme with different environment variables for the same app. ex) exampleApp, exampleAppDev
2. run `react-native run-ios --scheme exampleApp`
3. should install and run exampleApp with `a` variable
4. run `react-native run-ios --scheme exampleAppDev`
5. should install and run exampleAppDev with `b` variable

### DIFF
--- a/packages/cli/src/runIOS/runIOS.js
+++ b/packages/cli/src/runIOS/runIOS.js
@@ -17,7 +17,7 @@ const findReactNativeScripts = require('../util/findReactNativeScripts');
 const parseIOSDevicesList = require('./parseIOSDevicesList');
 const findMatchingSimulator = require('./findMatchingSimulator');
 
-const getBuildPath = function(configuration = 'Debug', appName, isDevice) {
+const getBuildPath = function(configuration = 'Debug', appName, isDevice, scheme) {
   let device;
 
   if (isDevice) {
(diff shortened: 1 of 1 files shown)


packages/cli/src/runIOS/runIOS.js
```diff
@@ -28,7 +28,7 @@ const getBuildPath = function(configuration = 'Debug', appName, isDevice) {
     device = 'iphonesimulator';
   }
 
-  return `build/Build/Products/${configuration}-${device}/${appName}.app`;
+  return `build/${scheme}/Build/Products/${configuration}-${device}/${appName}.app`;
```
thymikee: this will fail with undefined scheme, right? can we guard against it?

packages/cli/src/runIOS/runIOS.js
```diff
@@ -282,7 +282,7 @@ function buildProject(
       '-destination',
       `id=${udid}`,
       '-derivedDataPath',
-      'build',
+      `build/${scheme}`,
```
thymikee: same stuff here

misoguy:
@thymikee I could have a default value for the `scheme` but I think the `scheme` is always present with a default as [this line](https://github.com/react-native-community/react-native-cli/pull/128/files#diff-4352e7b05219002fa12b950df8025bbdR195) is using it as a backup for `appName`. Let me know your thoughts :)

packages/cli/src/runIOS/runIOS.js
grabbou: Defined here: https://github.com/react-native-community/react-native-cli/blob/7d7238a7d53b42fe3b92517a0f89b1f038a16f28/packages/cli/src/runIOS/runIOS.js#L69

grabbou:
Looks good. Tested locally. Shipping. Thank you <3

PR #137, opened 2019-01-31
Fix: problem on run-android with deviceId

### LINKED ISSUE #132: --deviceId doesn't work with run-android
## Environment

```
    System:
      OS: macOS Sierra 10.12.1
      CPU: (8) x64 Intel(R) Core(TM) i7-4770HQ CPU @ 2.20GHz
      Memory: 893.90 MB / 16.00 GB
      Shell: 3.2.57 - /bin/bash
    Binaries:
      Yarn: yarn install v0.23.4
[1/4] Resolving packages...
[2/4] Fetching packages...
[3/4] Linking dependencies...
success Saved lockfile.
Done in 13.82s. - /usr/local/bin/yarn
      npm: 6.4.1 - ~/.nvm/versions/node/v10.13.0/bin/npm
    SDKs:
      iOS SDK:
        Platforms: iOS 10.3, macOS 10.12, tvOS 10.2, watchOS 3.2
      Android SDK:
        API Levels: 25, 27
        Build Tools: 19.1.0, 21.1.2, 23.0.1, 23.0.2, 23.0.3, 24.0.1, 24.0.3, 25.0.0, 25.0.1, 25.0.2, 25.0.3, 26.0.2, 27.0.3
        System Images: a...pis | Google APIs ARM EABI v7a Syste..., a...google_apis | Google APIs Intel x86 Atom Sys...
    IDEs:
      Android Studio: 3.0 AI-171.4408382
      Xcode: 8.3.2/8E2002 - /usr/bin/xcodebuild
    npmPackages:
      react: 16.6.1 => 16.6.1 
      react-native: 0.57.5 => 0.57.5 
    npmGlobalPackages:
      react-native-cli: 2.0.1
```

## Description

react-native run-android --deviceId=xxxxx will cause this issue:
```
Failed to install app/build/outputs/apk/app-debug.apk: Failure [INSTALL_FAILED_ALREADY_EXISTS: Attempt to re-install com.xxx.xxx without first uninstalling.]
Command failed: /Users/ysyyork/Library/Android/sdk/platform-tools/adb -s CBI9SH008I install app/build/outputs/apk/app-debug.apk
Could not install the app on the device, read the error above for details.
```
This is because the apk file is actually under `app/build/outputs/apk/debug/app-debug.apk`. The path is missing variant folder. 

## Reproducible Demo

Any repo run with --deviceId option set.

ifsnow:
Fixes #132

- Works well with `enableSeparateBuildPerCPUArchitecture`, `universalApk` options.
- Even if it is already installed, it works normally. (Added -r option to `adb install`)

I hope this helps.

### DIFF
--- a/packages/cli/src/runAndroid/adb.js
+++ b/packages/cli/src/runAndroid/adb.js
@@ -8,7 +8,7 @@
  * @flow strict
  */
 
-const { execSync } = require('child_process');
+const { execSync, execFileSync } = require('child_process');
 
 /**
  * Parses the output of the 'adb devices' command
@@ -34,10 +34,36 @@ function parseDevicesResult(result: string): Array<string> {
 /**
  * Executes the commands needed to get a list of devices from ADB
  */
-function getDevices(): Array<string> {
+function getDevices(adbPath: string): Array<string> {
   try {
-    const devicesResult = execSync('adb devices');
+    const devicesResult = execSync(`${adbPath} devices`);
     return parseDevicesResult(devicesResult.toString());
+  } catch (e) {
+    return [];
+  }
+}
+
+/**
+ * Gets available CPUs of devices from ADB
+ */
+function getAvailableCPUs(adbPath: string, device: string): Array<string> {
+  try {
+    const baseArgs = ['-s', device, 'shell', 'getprop'];
+
+    let cpus = execFileSync(
+      adbPath,
+      baseArgs.concat(['ro.product.cpu.abilist'])
+    ).toString();
+
+    // pre-Lollipop
+    if (!cpus || cpus.trim().length === 0) {
+      cpus = execFileSync(
+        adbPath,
+        baseArgs.concat(['ro.product.cpu.abi'])
+      ).toString();
+    }
+
+    return (cpus || '').trim().split(',');
   } catch (e) {
     return [];
   }
@@ -46,4 +72,5 @@ function getDevices(): Array<string> {
 module.exports = {
   parseDevicesResult,
   getDevices,
+  getAvailableCPUs,
 };
--- a/packages/cli/src/runAndroid/adb.js
+++ b/packages/cli/src/runAndroid/adb.js
@@ -9,7 +9,6 @@
  */
 
 import { execSync, execFileSync } from 'child_process';
-import logger from '../util/logger';
 
 /**
  * Parses the output of the 'adb devices' command
@@ -40,7 +39,6 @@ function getDevices(adbPath: string): Array<string> {
     const devicesResult = execSync(`${adbPath} devices`);
     return parseDevicesResult(devicesResult.toString());
   } catch (e) {
-    logger.error(e.message);
     return [];
   }
 }
@@ -67,7 +65,6 @@ function getAvailableCPUs(adbPath: string, device: string): Array<string> {
 
     return (cpus || '').trim().split(',');
   } catch (e) {
-    logger.error(e.message);
     return [];
   }
 }
--- a/packages/cli/src/runAndroid/adb.js
+++ b/packages/cli/src/runAndroid/adb.js
@@ -5,10 +5,11 @@
  * LICENSE file in the root directory of this source tree.
  *
  * @format
- * @flow strict
+ * @flow
  */
 
 const { execSync, execFileSync } = require('child_process');
+const logger = require('../util/logger');
 
 /**
  * Parses the output of the 'adb devices' command
@@ -39,6 +40,7 @@ function getDevices(adbPath: string): Array<string> {
     const devicesResult = execSync(`${adbPath} devices`);
     return parseDevicesResult(devicesResult.toString());
   } catch (e) {
+    logger.error(e.message);
     return [];
   }
 }
@@ -65,6 +67,7 @@ function getAvailableCPUs(adbPath: string, device: string): Array<string> {
 
     return (cpus || '').trim().split(',');
   } catch (e) {
+    logger.error(e.message);
     return [];
   }
 }
--- a/packages/cli/src/runAndroid/runAndroid.js
+++ b/packages/cli/src/runAndroid/runAndroid.js
@@ -130,7 +130,7 @@ function runOnSpecificDevice(
   packageName,
   adbPath
 ) {
-  const devices = adb.getDevices();
+  const devices = adb.getDevices(adbPath);
   if (devices && devices.length > 0) {
     if (devices.indexOf(args.deviceId) !== -1) {
       buildApk(gradlew);
@@ -166,16 +166,21 @@ function buildApk(gradlew) {
   }
 }
 
-function tryInstallAppOnDevice(args, device) {
+function tryInstallAppOnDevice(args, adbPath, device) {
   try {
     // "app" is usually the default value for Android apps with only 1 app
     const appFolder = args.appFolder || 'app';
     const variant = (args.variant || 'debug').toLowerCase();
     const buildDirectory = `${appFolder}/build/outputs/apk/${variant}`;
-    const apkFile = getInstallApkName(appFolder, device, buildDirectory);
-    const pathToApk = `${buildDirectory}/${apkFile}`;
+    const apkFile = getInstallApkName(
+      appFolder,
+      adbPath,
+      variant,
+      device,
+      buildDirectory
+    );
 
-    const adbPath = getAdbPath();
+    const pathToApk = `${buildDirectory}/${apkFile}`;
     const adbArgs = ['-s', device, 'install', pathToApk];
     console.log(
       chalk.bold(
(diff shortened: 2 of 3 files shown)


thymikee:
Thanks @ifsnow, appreciated! Would you mind fixing lint/flow?

ifsnow:
@thymikee Sure! All modified.

thymikee:
@ferrannp can you have a look?

packages/cli/src/runAndroid/runAndroid.js
```diff
@@ -191,6 +195,34 @@ function tryInstallAppOnDevice(args, device) {
   }
 }
 
+function getInstallApkName(appFolder, device, buildDirectory) {
+  const gradleFile = fs.readFileSync(`${appFolder}/build.gradle`, 'utf8');
+
+  const enableSeparateBuild = gradleFile.match(
+    /def enableSeparateBuildPerCPUArchitecture.*=.*(true|false)/i
```
ferrannp: So this is the default variable name `enableSeparateBuildPerCPUArchitecture` from RN template. I think we should check for:

```
 abi {
   enable true|false
 }
```

The same way you are looking for `universalApk`.

packages/cli/src/runAndroid/runAndroid.js
```diff
@@ -170,7 +170,11 @@ function tryInstallAppOnDevice(args, device) {
   try {
     // "app" is usually the default value for Android apps with only 1 app
     const appFolder = args.appFolder || 'app';
-    const pathToApk = `${appFolder}/build/outputs/apk/${appFolder}-debug.apk`;
+    const variant = (args.variant || 'debug').toLowerCase();
+    const buildDirectory = `${appFolder}/build/outputs/apk/${variant}`;
```
Esemesek: I think it would be safer if you used `path.join` everywhere you used `/`, because it uses platform specific delimiter.

thymikee: Gradle respects posix paths only

Esemesek: What about `fs` calls? It probably won't work on Windows. https://github.com/react-native-community/react-native-cli/blob/073389f5129a9b889a9db90b82e4278c4e847407/packages/cli/src/runAndroid/runAndroid.js#L199

thymikee: FS access needs to be normalized for Windows, yea. It's just paths that are injected into Gradle need to be posix (sorry for the confusion if that's not applicable here, I didn't review this PR yet :D )

packages/cli/src/runAndroid/runAndroid.js
ifsnow: @ferrannp Good point. I think the probability is low, but it is possible to set the value immediately without `enableSeparateBuildPerCPUArchitecture`. I'm thinking of a different approach.

packages/cli/src/runAndroid/runAndroid.js
ifsnow: @Esemesek @thymikee Thanks for reviews. I used the [same method used in the existing source.](https://github.com/react-native-community/react-native-cli/blob/master/packages/cli/src/runAndroid/runAndroid.js#L93) I will try to change it in a better way.

grabbou: I don't think `fs` needs paths to be normalised, I think it does it by default. It's `path` that needs it.

Context: https://nodejs.org/api/path.html#path_windows_vs_posix

packages/cli/src/runAndroid/runAndroid.js
```diff
@@ -191,6 +195,34 @@ function tryInstallAppOnDevice(args, device) {
...
+
+  const deviceApkName = `${appFolder}-armeabi-v7a-debug.apk`;
+  if (fs.existsSync(`${buildDirectory}/${deviceApkName}`)) {
+    return deviceApkName;
+  }
+
+  const universalApk = gradleFile.match(/universalApk.*(true|false)/i);
+  if (universalApk && universalApk[1] === 'true') {
+    return `${appFolder}-universal-debug.apk`;
+  }
```
grabbou: While I appreciate this, I wonder whether there's any better way to find an `apk` file (by talking to Gradle?)

ifsnow: @grabbou Gradle generates an `output.json` file after building. It has information about the built apk files. I think I can use this information. I will share a new version soon.

grabbou: Great! Not saying this code is bad whatsoever - I actually find it quite impressive and have to say - kudos for figuring this out! Just thinking from the long-term perspective, it's worth reinvestigating the entire approach to paths we use in this command!

Thank you!

packages/cli/src/runAndroid/runAndroid.js
ifsnow: @grabbou You are right. I've confirmed that it works fine on Windows. Thanks for the advice.

ifsnow:
@grabbou @thymikee @ferrannp @Esemesek  I think there was some problems with how to use `output.json` mentioned above. For example, some users delete this file. so, It seems most reasonable to find the appropriate apk file for the target device's cpu via `adb`.

Please review my new approach.

packages/cli/src/runAndroid/adb.js
```diff
@@ -34,16 +34,43 @@ function parseDevicesResult(result: string): Array<string> {
...
+    // pre-Lollipop
+    if (!cpus || cpus.trim().length === 0) {
+      cpus = execFileSync(
+        adbPath,
+        baseArgs.concat(['ro.product.cpu.abi'])
+      ).toString();
+    }
+
+    return (cpus || '').trim().split(',');
+  } catch (e) {
```
Krizzu: What about logging what happened?

packages/cli/src/runAndroid/runAndroid.js
```diff
@@ -191,6 +200,32 @@ function tryInstallAppOnDevice(args, device) {
...
+    }
+  }
+
+  // check if there is a default file like app-debug.apk
+  const apkName = `${appFolder}-${variant}.apk`;
+  if (fs.existsSync(`${buildDirectory}/${apkName}`)) {
+    return apkName;
+  }
+
+  throw new Error('Not found the correct install APK file!');
```
Krizzu: What about adding a "last resort" function to find apk by using unix `find` (or windows - [dir](https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/dir)) to search `appFolder` for all `*.apk` existance?

packages/cli/src/runAndroid/runAndroid.js
ifsnow: @Krizzu I know that. So I changed to use another way.

packages/cli/src/runAndroid/adb.js
thymikee: I think it won't hurt? Feel free to use `logger.error()` instead of `console.log`, with a message that there was an error but we continue with some defaults.

packages/cli/src/runAndroid/runAndroid.js
ifsnow: @Krizzu I don't think I need the logic you said. I'm sure I checked all available files in my previous logic. Please let me know if there is an exception case.

packages/cli/src/runAndroid/adb.js
ifsnow: @thymikee `logger.error` was added. Thank you for your review.

packages/cli/src/runAndroid/runAndroid.js
Krizzu: Just thought about product flavors, because they change output directories when used

packages/cli/src/runAndroid/runAndroid.js
Krizzu: Right, just realized when it was posted :)

packages/cli/src/runAndroid/runAndroid.js
ifsnow: @Krizzu `flavor` has been deprecated. Use `variant` instead. This option works properly with the source I worked on.

Krizzu: @ifsnow I'm talking about `product flavor` of Android plugin - if you have those in your build, you'll get different output dirs for your APKs. It's not the case in default RN project tho, so I think we can pass for now.

packages/cli/src/runAndroid/runAndroid.js
thymikee: @Krizzu @ifsnow this CLI is meant to be used with RN 0.59+ which uses 3.3.0 in the default template (so recommended). Even RN 0.57.0 uses 3.1.7. Shouldn't we be safe to assume folks use 3.0.0+ already? I'd go with that, as using `output.json` seems pretty good for this usecase

ifsnow: @thymikee @Krizzu I have a different idea. I think using adb is more efficient and reasonable than using output.json. That’s because output.json does not have any information about the target device.

packages/cli/src/runAndroid/adb.js
```diff
@@ -34,18 +35,45 @@ function parseDevicesResult(result: string): Array<string> {
...
+    logger.error(e.message);
```
grabbou: We shouldn't capture errors here, log and return an empty array. Let's let this error go up forward - there's a global try/catch that will print the error. 

You could throw a new error here, if you want to make a better message.

packages/cli/src/runAndroid/runAndroid.js
```diff
@@ -155,12 +155,21 @@ function buildApk(gradlew) {
...
 }
 
-function tryInstallAppOnDevice(args, device) {
+function tryInstallAppOnDevice(args, adbPath, device) {
   try {
     // "app" is usually the default value for Android apps with only 1 app
     const appFolder = args.appFolder || 'app';
-    const pathToApk = `${appFolder}/build/outputs/apk/${appFolder}-debug.apk`;
-    const adbPath = getAdbPath();
+    const variant = (args.variant || 'debug').toLowerCase();
```
grabbou: I think we can define "default: debug" at the bottom, where we have CLI arguments defined. What do you think?

ifsnow: @grabbou I think the way you said is good.

packages/cli/src/runAndroid/adb.js
ifsnow: @grabbou I don't think we need to log errors here. There is no meaningful information.

ifsnow:
@grabbou I'm done. Please let me know if you have any reviews.

packages/cli/src/runAndroid/adb.js
grabbou: Then, I think it's a question whether returning an empty array in case of an error and swallowing it is a good design decision or not.

I would assume we throw an error - is an empty array of CPUs a "good" return value? When this can happen?

thymikee:
@Krizzu can you give it one more look?

packages/cli/src/runAndroid/adb.js
ifsnow: @grabbou I agree with your idea. I’m sure that the probability of this error you said is very low. However, it returns an empty array for defense logic in case of an unknown situation.

grabbou: > it returns an empty array for defense logic

I am just worried that returning an empty array may provide confusing and hard to debug results when this error actually happens.

My question is:
- is it possible for an app to have no CPU architectures available? 
- is the outer code (in our case - runAndroid) able to handle this case? 

We should focus on letting the user know something bad happened so that they can debug it.

That said, I think it would be better to just throw an error here, e.g. `throw new Error("There was an error getting an array of XXXX, consider doing Y")`

ifsnow: @grabbou 
As far as I know, using `adb` can get the CPU architectures on all android devices. If an error occurs, it will be the part of running that. It may be an unknown issue. In this case, it is a matter of choosing whether to exit the program without trying to install the APK.

I chose to try to install even if I didn't know the CPU architectures.

- Even if the CPU array is empty, try to install the APK in two cases. [[code](https://github.com/ifsnow/react-native-cli/blob/fix/run-android-with-deviceId/packages/cli/src/runAndroid/runAndroid.js#L199)]
  - case 1) app-universal-debug.apk
  - case 2) app-debug.apk (`enableSeparateBuildPerCPUArchitecture = false`)
- If an installable APK is not found, throw an error. [[code](https://github.com/ifsnow/react-native-cli/blob/fix/run-android-with-deviceId/packages/cli/src/runAndroid/runAndroid.js#L212)]

It is very unlikely that this step will happen, so you don't have to worry about it. :)

grabbou: Okay, thank you for this answer and explaining it. Makes sense :)

PR #139, opened 2019-01-31
fix: improve overall run-ios experience and get rid of CFBundleIdentifier

### LINKED ISSUE #107: Get rid of "CFBundleIdentifier" error
This is super old issue that happens when you run "run-ios" (usually for the first time on a newly inited project). It's one of the pinned issues in React Native and one of the most popular errors people are struggling with.

If this is a "wont fix", we should consider changing the logic of the command itself to avoid this mistake. 

I am opening this issue as a memo for the future.

Related: https://github.com/facebook/react-native/issues/21303

Stack trace:
```
Error: Command failed: /usr/libexec/PlistBuddy -c Print:CFBundleIdentifier build/Build/Products/Debug-iphonesimulator/RNTesterPods.app/Info.plist
Print: Entry, ":CFBundleIdentifier", Does Not Exist

    at checkExecSyncError (child_process.js:611:11)
    at Object.execFileSync (child_process.js:629:13)
    at Promise.then (/Users/grabbou/Repositories/react-native/node_modules/@react-native-community/cli/build/runIOS/runIOS.js:183:36)
```

grabbou:
Fixes #107 

Summary:
---------

Bugs fixed:
- "CFBundleIdentifier" printed every time "run-ios" fails, shadowing the real cause for the problem
- "run-ios" running a previous/cached version of an app (usually when there was an error during build step that was ignored)

Changes made:
- Decided to turn Flow and ESLint and apply all the changes as I am working on it. I know the diff might be strange at first sight, but if you look closely, it's just cosmetics to satisfy various rules
- I also extracted some blocks to separate utilities for better readability
- I also removed a no longer needed check for CRNA which is deprecated (and IMO, is not needed)
- Converted few functions to "async" form for better readability
- Started capturing the "stderr" of "xcodebuild" and started relying on its error code in order to avoid going to next step when it fails
- Added a new type of error to differentiate from other errors and also be able to capture additional data. Logic and styling subject to change, encapsulated inside Logger.

Test Plan:
----------

Here's how the error messages look before/now:

While try to build a project with a wrong "schema" :

Before:
screenshot 2019-01-31 at 18 07 43

After:
screenshot 2019-01-31 at 18 05 23

While building a project that has a typo in "AppDelegate.m":

Before:
screenshot 2019-01-31 at 18 08 19

After:
screenshot 2019-01-31 at 18 06 36

You can try it out by running `run-ios` command on a broken project (typo, wrong configuration) or by passing a non-existing `schema`

### DIFF
--- a/packages/cli/src/runIOS/runIOS.js
+++ b/packages/cli/src/runIOS/runIOS.js
@@ -17,7 +17,7 @@ const path = require('path');
 const findXcodeProject = require('./findXcodeProject');
 const parseIOSDevicesList = require('./parseIOSDevicesList');
 const findMatchingSimulator = require('./findMatchingSimulator');
-const { ProcessError } = require('../util/logger');
+const { ProcessError } = require('../util/errors');
 
 type FlagsT = {
   simulator: string,
--- a/packages/cli/src/runIOS/runIOS.js
+++ b/packages/cli/src/runIOS/runIOS.js
@@ -319,8 +319,10 @@ function buildProject(
             [
               `Failed to build iOS project.`,
               `We ran "xcodebuild" command but it exited with error code ${code}.`,
-              `Consider building your app with XCode app to further debug this issue.`,
-            ].join(''),
+              `To debug build logs further, consider building your app with Xcode.app, by opening ${
+                xcodeProject.name
+              }`,
+            ].join(' '),
             errorOutput
           )
         );
--- a/packages/cli/src/runIOS/runIOS.js
+++ b/packages/cli/src/runIOS/runIOS.js
@@ -21,7 +21,7 @@ const { ProcessError } = require('../util/logger');
 
 type FlagsT = {
   simulator: string,
-  configuration: ?string,
+  configuration: string,
   scheme: ?string,
   projectPath: string,
   device: ?string,
@@ -80,7 +80,7 @@ function runIOS(_: Array<string>, ctx: ContextT, args: FlagsT) {
       );
     }
     if (devices && devices.length > 0) {
-      // $FlowIssue: `args.device` is never null here
+      // $FlowIssue: args.device is defined in this context
       console.log(`Could not find device with the name: "${args.device}".`);
       console.log('Choose one of the following:');
       printFoundDevices(devices);
@@ -88,13 +88,19 @@ function runIOS(_: Array<string>, ctx: ContextT, args: FlagsT) {
       console.log('No iOS devices connected.');
     }
   } else if (args.udid) {
+    // $FlowIssue: args.udid is defined in this context
     return runOnDeviceByUdid(args, scheme, xcodeProject, devices);
   }
 
   return runOnSimulator(xcodeProject, args, scheme);
 }
 
-function runOnDeviceByUdid(args, scheme, xcodeProject, devices) {
+function runOnDeviceByUdid(
+  args: FlagsT & { udid: string },
+  scheme,
+  xcodeProject,
+  devices
+) {
   const selectedDevice = matchingDeviceByUdid(devices, args.udid);
 
   if (selectedDevice) {
@@ -111,6 +117,7 @@ function runOnDeviceByUdid(args, scheme, xcodeProject, devices) {
   }
 
   if (devices && devices.length > 0) {
+    // $FlowIssue: args.udid is defined in this context
     console.log(`Could not find device with the udid: "${args.udid}".`);
     console.log('Choose one of the following:');
     printFoundDevices(devices);
@@ -257,7 +264,7 @@ function buildProject(
   xcodeProject,
   udid,
   scheme,
-  configuration = 'Debug',
+  configuration,
   launchPackager = false,
   verbose,
   port
@@ -339,7 +346,7 @@ function bootSimulator(selectedSimulator) {
   }
 }
 
-function getBuildPath(configuration = 'Debug', appName, isDevice, scheme) {
+function getBuildPath(configuration, appName, isDevice, scheme) {
   let device;
 
   if (isDevice) {
@@ -456,6 +463,7 @@ module.exports = {
     {
       command: '--configuration [string]',
       description: 'Explicitly set the scheme configuration to use',
+      default: 'Debug',
     },
     {
       command: '--scheme [string]',
--- a/packages/cli/src/runIOS/runIOS.js
+++ b/packages/cli/src/runIOS/runIOS.js
@@ -4,62 +4,47 @@
  * This source code is licensed under the MIT license found in the
  * LICENSE file in the root directory of this source tree.
  *
+ * @flow
  * @format
  */
 
-/* eslint-disable */
+import type { ContextT } from '../core/types.flow';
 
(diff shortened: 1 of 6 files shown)


packages/cli/src/runIOS/runIOS.js
```diff
@@ -308,28 +307,77 @@ function buildProject(
...
-        productNameMatch.length > 1
-      ) {
-        return resolve(productNameMatch[1]); // 0 is the full match, 1 is the app name
+      if (code !== 0) {
+        reject(
+          new ProcessError(
+            [
+              `Failed to build iOS project.`,
+              `We ran "xcodebuild" command but it exited with error code ${code}.`,
+              `Consider building your app with XCode app to further debug this issue.`,
```
eliperkins: s/XCode/Xcode

eliperkins: What do you think about phrasing this as: `To debug build logs furher, consider building your app with Xcode.app, by opening ${xcodeProject}`?

grabbou: Love that! I was struggling with error messages in particular as I always have a hard time phrasing it in a super friendly way! I'll definitely use that one!

elucaswork:
Nice work! 👏  

> Started capturing the "stderr" of "xcodebuild" and started relying on its error code in order to avoid going to next step when it fails

Cool, so this will prevent us to have a `"main.jsbundle does not exist."` error, right? 🙌(I've suffered with it this week - https://stackoverflow.com/a/54464135/1394414)

packages/cli/src/util/logger.js
```diff
@@ -23,13 +23,32 @@ const error = (...messages: Array<string>) => {
...
+  if (err instanceof ProcessError) {
+    console.error(`${chalk.grey(err.processError)}`);
+  }
+};
+
 const debug = (...messages: Array<string>) => {
   console.log(`${chalk.black.bgWhite(' DEBUG ')} ${joinMessages(messages)}`);
 };
 
+class ProcessError extends Error {
```
thymikee: This doesn't belong to logger. Can you move it to some kind of `errorHelpers`?

packages/cli/src/util/logger.js
```diff
@@ -23,13 +23,32 @@ const error = (...messages: Array<string>) => {
...
+
+  processError: string;
+}
+
 module.exports = {
   info,
   warn,
   error,
   debug,
+  logError,
```
thymikee: maybe `logProcessError`? It's very scoped to `ProcessError`

packages/cli/src/runIOS/runIOS.js
```diff
@@ -4,79 +4,68 @@
...
+type FlagsT = {
+  simulator: string,
+  configuration: string,
+  scheme: ?string,
+  projectPath: string,
+  device: ?string,
+  udid: ?string,
+  packager: boolean,
+  verbose: boolean,
+  port: number,
```
thymikee: This type seems wrong all the way. These are arguments, so they're all optional, right? Shouldn't it be typed like this?
```jsx
type FlagsT = {|
  simulator?: string,
  configuration?: string,
  scheme?: string,
  projectPath?: string,
  device?: string,
  udid?: string,
  packager?: boolean,
  verbose?: boolean,
  port?: number,
|};
```

packages/cli/src/runIOS/runIOS.js
```diff
@@ -91,23 +80,31 @@ function runIOS(argv, config, args) {
...
-  } else {
-    return runOnSimulator(xcodeProject, args, scheme);
   }
+
+  return runOnSimulator(xcodeProject, args, scheme);
 }
 
-function runOnDeviceByUdid(args, scheme, xcodeProject, devices) {
+function runOnDeviceByUdid(
+  args: FlagsT & { udid: string },
```
thymikee: see above comment

packages/cli/src/runIOS/runIOS.js
```diff
@@ -308,28 +307,79 @@ function buildProject(
...
+              `Failed to build iOS project.`,
+              `We ran "xcodebuild" command but it exited with error code ${code}.`,
+              `To debug build logs further, consider building your app with Xcode.app, by opening ${
+                xcodeProject.name
+              }`,
+            ].join(' '),
+            errorOutput
+          )
+        );
+        return;
```
thymikee: is this return necessary after we `reject`ed?

packages/cli/src/util/logger.js
grabbou: Yup, good idea. I'd suggest "errors" for now.

packages/cli/src/util/logger.js
grabbou: It is scoped for now, but the intent is to log all other kinds of errors too. We use it in the main cli interface where we catch the error from the command and don't know it's type yet.

packages/cli/src/runIOS/runIOS.js
grabbou: It seems, but if you take into consideration default values set by the CLI (see the object at the end where options are defined), it looks good. In this context, e.g. `verbose` is always false, unless set. Same for `port`. We do the same in other commands.

We can add a comment if that's confusing.

packages/cli/src/runIOS/runIOS.js
grabbou: It's not - just my personal preference of early returns over /else clause (visual appearance).

grabbou: I think I'll update it to if/else here 👌

packages/cli/src/util/logger.js
grabbou: See: https://github.com/react-native-community/react-native-cli/pull/139/files#diff-97059d274ee5d7bbcc5730ddee2a98daR30

packages/cli/src/runIOS/runIOS.js
thymikee: I mean, if an option is present in the object, why would it be falsy? Maybe that's something coming from commander, I didn't check locally.

thymikee: Just checked, this is what commander returns, so feel free to discard my comment

grabbou:
Check @thymikee if you are happy with the last commit and if yes, let's merge it :fire:

packages/cli/src/util/errors.js
```diff
@@ -0,0 +1,15 @@
+/**
+ * @flow
+ */
+const chalk = require('chalk');
+
+class ProcessError extends Error {
+  constructor(msg: string, processError: string) {
+    super(`${chalk.red(msg)}\n\n${chalk.gray(processError)}`);
```
thymikee: Looks silly, but I'm cool with that :)

grabbou: I would say it's a "temporary" solution when we figure out proper logging. I think it's a good starting point.

owcc:
I follow the issue " ":CFBundleIdentifier", Does Not Exist"to here, and this issue is still exist🙁

Issue #143, opened 2019-02-01
run-android|run-ios do not support RNTester out of the box


grabbou:
Instructions to run RNTester recommend running Gradle / opening Xcode in order to run the app. We should add support for using the CLI, which will make it faster and easier to run RNTester and check when releasing new versions.

The biggest issue is in the detection mechanism that are inside `run-android` and `run-ios` (that are quite dumb).

grabbou:
While doing it, rename `--root` in `run-android` to be less confusing (ideally, make it like the `run-ios`)

grabbou:
iOS part is being worked on in #287

grabbou:
@dulmandakh would you be able to explain why some Android projects have `android/gradlew` and others have `android/app/gradle` (RNTester)? Trying to handle both cases when doing `run-android` and I am a bit lost here.

thymikee:
AFAIK this is done.

PR #144, opened 2019-02-01
feat: allow setting different React Native path

grabbou:
Summary:
---------

This PR fixes `require.resolve` logic for looking up React Native location. It will now resolve React Native from the current working directory, instead of __dirname. Thanks to this subtle change, we don't need `react-native` to be a dev dependency of CLI. We will now load React Native from the "node_modules" of a project we are testing CLI on.

It also allows to overwrite path to `react-native` when require.resolve does not provide good results. For example, we can pass a location to React Native repository (when we have it cloned) to run CLI on master of React Native (case: checking compatibility without releasing a new version) or to run RNTester example app (where there is no React Native package in node_modules).

Partially addresses #143. I will send another follow-up fixing `run-android` and `run-ios` behaviour later.

Test Plan:
----------

When testing CLI locally, go to a project with React Native and run CLI: `node: ../<path_to_cli>/packages/cli/build/index.js`. It will load React Native from your project "node_module"s, instead of CLI.

### DIFF
--- a/packages/cli/src/cliEntry.js
+++ b/packages/cli/src/cliEntry.js
@@ -117,11 +117,20 @@ const addCommand = (command: CommandT, ctx: ContextT) => {
     )
   );
 
-  // This is needed to avoid `unknown option` error by Commander.js
-  cmd.option('--projectRoot [string]', 'Path to the root of the project');
+  cmd
+    .option('--projectRoot [string]', 'Path to the root of the project')
+    .option('--reactNativePath [string]', 'Path to React Native');
 };
 
 async function run() {
+  try {
+    await setupAndRun();
+  } catch (e) {
+    handleError(e);
+  }
+}
+
+async function setupAndRun() {
   const setupEnvScript = /^win/.test(process.platform)
     ? path.join('..', 'setup_env.bat')
     : path.join('..', 'setup_env.sh');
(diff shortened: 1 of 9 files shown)


grabbou:
When this gets merged, we can remove "metro.config.js" we added to React Native a while ago to overcome this issue.

packages/cli/src/init/init.js
```diff
@@ -50,6 +50,7 @@ function init(projectDir, argsOrName) {
  * @param options Command line arguments parsed by minimist.
  */
 function generateProject(destinationRoot, newProjectName, options) {
+  // eslint-disable-next-line
   const reactNativePackageJson = require('react-native/package.json');
```
grabbou: This is not run by our CLI, but by `global` CLI and we don't have context here. Ideally, this should be moved out to a sep. location to indicate it's not a command.

packages/cli/src/cliEntry.js
```diff
@@ -140,8 +149,26 @@ async function run() {
...
+        try {
+          return path.dirname(
+            // $FlowIssue: Wrong `require.resolve` type definition
+            require.resolve('react-native/package.json', {
+              paths: [root],
+            })
+          );
+        } catch (_ignored) {
+          throw new Error(
+            'Unable to find React Native files. Have you installed project dependencies?'
```
grabbou: I am wondering what is the best description of this error. Usually it means there's no `react-native` in `node_modules` (e.g. they are empty, deleted or an error) or we are running in an environment where `react-native` cannot be found (it's not in the regular require.resolve hierarchy)

packages/cli/src/init/init.js
```diff
@@ -50,6 +50,7 @@ function init(projectDir, argsOrName) {
  * @param options Command line arguments parsed by minimist.
  */
 function generateProject(destinationRoot, newProjectName, options) {
+  // eslint-disable-next-line
```
thymikee: can you explicitly say what rule is disabled?

packages/cli/src/cliEntry.js
thymikee:
```suggestion
            'Unable to find React Native files. Make sure "react-native" module is installed in your project dependencies.'
```

thymikee:
🎉

Issue #114, opened 2019-01-23
Convert requires to imports


thymikee:
Thanks to using `import`s instead of `require`s we're able to apply import transform to inline requires for faster startups :) 

Basically like this:

```diff
-const path = require('path');
+import path from 'path';
```

Trancever:
@thymikee I can try to do it. Assign me if possible.

sidferreira:
@Trancever I worked this morning on this but I haven't seen that you wanted to help. 

If you haven't started yet, you can take a look at https://github.com/sidferreira/react-native-cli/tree/refactor/replacesRequiresByImports and help me to check if there's anything missing...

Trancever:
@sidferreira Hi, I finished refactoring the codebase but I am currently struggling with making test passing again. I'll dm you tomorrow so we can collaborate on this.

Trancever:
@thymikee I believe this can be closed. We can open another issue for replacing module.exports with ES module exports.

PR #145, opened 2019-02-04
[WIP] Change require to imports 

Trancever:
Summary:
---------

Fix for #114 

Test Plan:
----------

This is still work in progress since many tests are failing.


packages/cli/src/bundle/assetPathUtils.js
```diff
@@ -81,7 +81,7 @@ function getBasePath(asset: PackagerAsset) {
   return basePath;
 }
 
-module.exports = {
+export default {
   getAndroidAssetSuffix,
```
thymikee: later on we can think about using named exports instead of this

scripts/watch.js
```diff
@@ -9,11 +9,11 @@
...
-const fs = require('fs');
-const { execSync } = require('child_process');
-const path = require('path');
-const chalk = require('chalk');
-const { getPackages } = require('./helpers');
+import fs from 'fs';
+import childProcess from 'child_process';
+import path from 'path';
+import chalk from 'chalk';
+import helpers from './helpers';
```
thymikee: we run scripts through regular node, without ESM loader, so please revert that as it won't work now

packages/global-cli/index.js
```diff
@@ -36,13 +36,12 @@
...
-const execSync = require('child_process').execSync;
-const chalk = require('chalk');
-const prompt = require('prompt');
-const semver = require('semver');
+import fs from 'fs';
+import path from 'path';
+import child_process from 'child_process';
+import chalk from 'chalk';
+import prompt from 'prompt';
+import semver from 'semver';
```
thymikee: `global-cli` is not compiled with Babel, please revert changes here for now.

packages/cli/src/index.js
```diff
@@ -8,12 +8,15 @@
...
+gracefulFs.gracefulify(fs);
+
+const cli = require('./cliEntry').default;
 
 if (require.main === module) {
   cli.run();
 }
 
-module.exports = cli;
+export default cli;
```
thymikee: please revert this export change as it changes our public API for those using `require` syntax.
Instead of doing:
```js
const rncli = require('@react-native-community/cli');
```
They would have to do:
```js
const rncli = require('@react-native-community/cli').default;
```
because of Babel interop. The API however won't change for folks using `import` syntax, because it has this check built-in (both in Babel and TS afaik)

Trancever:
Thanks @thymikee for review. Will try to fix those issues as soon as possible

sidferreira:
@Trancever Let's compare our branches. My tests are all passing...

packages/global-cli/index.js
grabbou: And it's going to be deprecated anyway, so I'll just skip it.

Trancever:
@sidferreira Nice! Can you open PR with your changes? I think we should proceed with your version if you have it already working.

sidferreira:
Added as #150

PR #148, opened 2019-02-04
Feature: adds prettier command

sidferreira:
That's a humble suggestion.
As I worked, I missed a lot this command...

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,6 +11,7 @@
     "lint": "eslint . --cache --report-unused-disable-directives",
     "flow-check": "flow check",
     "postinstall": "yarn build",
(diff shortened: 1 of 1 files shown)


thymikee:
Thanks for the PR.
However that's not necessary, since you can run `yarn lint --fix` to fix all Prettier errors, they're autofixable :)

thymikee:
Feel free to add that to contributing guide, that may be a good idea :)

sidferreira:
@thymikee I really added eslint and prettier to my workflow. But I keep them separated as sometimes I know there are errors and I just want to fix the code :)

thymikee:
In this project prettier and eslint are integrated and they shouldn't collide.

PR #150, opened 2019-02-04
[WIP] Change require to imports (alt)

sidferreira:


thymikee:
Mind fixing conflicts and address feedback I left on #145?

packages/cli/src/index.js
```diff
@@ -17,4 +17,4 @@ if (require.main === module) {
   cli.run();
 }
 
-module.exports = cli;
```
thymikee: this was correct

sidferreira: hahaha sorry

thymikee:
@Trancever mind checking that?

packages/cli/src/index.js
```diff
@@ -8,9 +8,10 @@
  */
 
 // gracefulify() has to be called before anything else runs
-require('graceful-fs').gracefulify(require('fs'));
+// eslint-disable-next-line no-unused-vars
+import gracefulFs from './util/gracefulFs';
```
thymikee: I don't see a reason to extract this to a util, and in this form you import with side-effects. Please inline the implementation here

packages/cli/src/server/util/messageSocket.js
```diff
@@ -232,7 +232,4 @@ function attachToServer(server, path) {
   };
 }
 
-module.exports = {
-  attachToServer,
-  parseMessage,
-};
+export default { attachToServer, parseMessage };
```
thymikee: while we're at, can you make these named exports instead of default?

packages/global-cli/index.js
```diff
@@ -57,7 +57,7 @@ const semver = require('semver');
  *     - "/Users/home/react-native/react-native-0.22.0.tgz" - for package prepared with `npm pack`, useful for e2e tests
  */
 
-const options = require('minimist')(process.argv.slice(2));
+import options from 'minimist')(process.argv.slice(2);
```
thymikee: Please revert

packages/cli/src/server/util/messageSocket.js
sidferreira: @thymikee Yes I can, but when I was working at it, I noticed:
```
  const wsProxy = webSocketProxy.attachToServer(
    serverInstance,
    '/debugger-proxy'
  );
  const ms = messageSocket.attachToServer(serverInstance, '/message');
```
( runServer.js )

so we can end up having 2 packages with `attachToServer`

I kept them as `default` to avoid a collision. But I can name it `attachToMessageSocket`?

packages/global-cli/index.js
sidferreira: @thymikee Isn't it better to

```
import minimist from 'minimist'
const options = minimist(process.argv.slice(2));
```

packages/cli/src/index.js
sidferreira: @thymikee the problem here is the 
```
import cli from './cliEntry';
```
cause we need to run `gracefulFs.gracefulify(realFs);` before it

sidferreira: @thymikee keep extracted?

packages/global-cli/index.js
sidferreira: fixed

packages/global-cli/index.js
```diff
@@ -57,7 +57,7 @@ const semver = require('semver');
  *     - "/Users/home/react-native/react-native-0.22.0.tgz" - for package prepared with `npm pack`, useful for e2e tests
  */
 
-const options = require('minimist')(process.argv.slice(2));
+const options = require('minimist')(process.argv.slice(2);
```
thymikee: Missing closing paren, there should be no changes to this file

sidferreira:
@thymikee I hope to help even more in the future 👍

packages/cli/src/server/util/messageSocket.js
grabbou: Would be great to open an issue tracking this. I would like to move to refine exports at some point to embrace this module system even further.

thymikee: We could leverage @Trancever work on #145 for that 🙂

grabbou:
@sidferreira amazing work, really impressive! Looks like `module.exports` are still there - is the intention to migrate to `export` syntax at some point, presumably in next PR?

sidferreira:
@grabbou as it isn't merged yet, I can take a look on it _now_

grabbou:
No worries, we can do it after we merge this one. Shall we land it?

thymikee:
If you don't have anything to add then feel free to press the button!

grabbou:
Had to force push as I forgot to reword the PR title.

PR #151, opened 2019-02-05
chore: use ESM exports where appropriate

sidferreira:
This is an extra for #150 
Fixes exports and tweaks some tests.

The `link`, `linkAssets`, and `linkDependency` where trickier as hell, but we have all tests working.

I suggest take it easy in this review as I'm worried in a "false positive" in tests as there was a LOT of changes.

We need to improve the code coverage for changes like this have smaller chances of failing...

### DIFF
--- a/packages/cli/src/cliEntry.js
+++ b/packages/cli/src/cliEntry.js
@@ -191,7 +191,7 @@ async function setupAndRun() {
   }
 }
 
-module.exports = {
+export default {
   run,
   init,
 };
--- a/packages/cli/src/cliEntry.js
+++ b/packages/cli/src/cliEntry.js
@@ -18,7 +18,7 @@ import getCommands from './core/getCommands';
 import getLegacyConfig from './core/getLegacyConfig';
 import init from './init/init';
 import assertRequiredOptions from './util/assertRequiredOptions';
-import logger from './util/logger';
+import { error } from './util/logger';
 import pkg from '../package.json';
 
 commander.version(pkg.version);
@@ -26,7 +26,7 @@ commander.version(pkg.version);
 const defaultOptParser = val => val;
 
 const handleError = err => {
-  logger.error(err.message);
+  error(err.message);
   process.exit(1);
 };
 
@@ -70,7 +70,7 @@ function printHelpInformation() {
 }
 
 function printUnknownCommand(cmdName) {
-  logger.error(
+  error(
     [
       cmdName
         ? `Unrecognized command "${cmdName}".`
@@ -195,3 +195,5 @@ export default {
   run,
   init,
 };
+
+export { run, init };
--- a/packages/cli/src/cliEntry.js
+++ b/packages/cli/src/cliEntry.js
@@ -18,7 +18,7 @@ import getCommands from './core/getCommands';
 import getLegacyConfig from './core/getLegacyConfig';
 import init from './init/init';
 import assertRequiredOptions from './util/assertRequiredOptions';
-import { error } from './util/logger';
+import logger from './util/logger';
 import pkg from '../package.json';
 
 commander.version(pkg.version);
@@ -26,7 +26,7 @@ commander.version(pkg.version);
 const defaultOptParser = val => val;
 
 const handleError = err => {
-  error(err.message);
+  logger.error(err.message);
   process.exit(1);
 };
 
@@ -70,7 +70,7 @@ function printHelpInformation() {
 }
 
 function printUnknownCommand(cmdName) {
-  error(
+  logger.error(
     [
       cmdName
         ? `Unrecognized command "${cmdName}".`
@@ -196,4 +196,4 @@ export default {
   init,
 };
 
-export { run, init };
+// export { run, init };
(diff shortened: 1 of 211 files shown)


sidferreira:
@grabbou

packages/cli/src/core/getCommands.js
```diff
@@ -3,31 +3,50 @@
...
+import logAndroid from '../logAndroid/logAndroid';
+import logIOS from '../logIOS/logIOS';
+import dependencies from '../dependencies/dependencies';
+import info from '../info/info';
+
 /**
  * List of built-in commands
  */
+
 const loadLocalCommands = (): Array<LocalCommandT> => [
```
grabbou: Doesn't need to be a function anymore. We did it for lazy require call.

packages/cli/src/core/__tests__/findPlugins-test.js
```diff
@@ -8,8 +8,9 @@
  * @emails oncall+javascript_foundation
  */
 
+import findPlugins from '../findPlugins';
+
 const path = require('path');
```
grabbou: Any reason `path` is required, not imported? Same in other files, where we use `fs`. Is that for mocking functionality?

sidferreira: There are places that there's a mock before the require. In this case, I believe the import may fail (had this issue in the past). I'll check anyway.

thymikee: `jest.mock` and `jest.unmock` are hoisted above imports with `babel-jest` by default by Jest so that's not an issue

sidferreira: But in some places it caused errors. For example in `getPlatforms` we have:

```
const builtInPlatforms = {
  ios: require('./ios'),
  android: require('./android'),
};
```

If I change it to imports, many errors pop.
Of course, I can give extra effort in order to filter _all_ this cases and try to fix the tests to work as expected.

packages/cli/src/core/getCommands.js
sidferreira: fixed

packages/cli/src/core/__tests__/findPlugins-test.js
thymikee: Nah, that's not necessary. `require`s are OK in tests when they make testing easier (e.g. when you want to reset modules in between tests in a describe block).

sidferreira: In this case I believe we're done

packages/cli/src/cliEntry.js
```diff
@@ -191,7 +191,7 @@ async function setupAndRun() {
   }
 }
 
-module.exports = {
+export default {
```
grabbou: I wonder whether this shouldn't be `export run` and `export init`

packages/cli/src/core/__fixtures__/projects.js
```diff
@@ -34,4 +34,4 @@ const withPods = {
   ios: ios.pod,
 };
 
-module.exports = { flat, nested, withExamples, withPods };
+export default { flat, nested, withExamples, withPods };
```
grabbou: Shouldn't this be a series of named exports instead of an object?

packages/cli/src/runAndroid/adb.js
```diff
@@ -43,7 +43,7 @@ function getDevices(): Array<string> {
   }
 }
 
-module.exports = {
+export default {
   parseDevicesResult,
```
grabbou: Shouldn't this be a set of named exports?

packages/cli/src/util/logger.js
```diff
@@ -29,7 +29,7 @@ const debug = (...messages: Array<string>) => {
   console.log(`${chalk.black.bgWhite(' DEBUG ')} ${joinMessages(messages)}`);
 };
 
-module.exports = {
+export default {
```
grabbou: Same here

packages/cli/src/util/yarn.js
```diff
@@ -52,7 +52,7 @@ function isGlobalCliUsingYarn(projectDir) {
   return fs.existsSync(path.join(projectDir, 'yarn.lock'));
 }
 
-module.exports = {
+export default {
```
grabbou: Same here

packages/cli/src/util/PackageManager.js
```diff
@@ -71,7 +71,7 @@ function remove(packageName, projectDir) {
   );
 }
 
-module.exports = {
+export default {
```
grabbou: Same here

sidferreira:
@grabbou I missed those indeed. Will check and ping back

grabbou:
Awesome @sidferreira! Thank you

packages/cli/src/index.js
```diff
@@ -17,4 +17,4 @@ if (require.main === module) {
   cli.run();
 }
 
-module.exports = cli;
+export default cli;
```
thymikee: please revert that, this is public API

sidferreira:
@grabbou the changes you requested ended being larger than expected. I reverted until we decide if we do the extra work now or wait for later.

grabbou:
@sidferreira sounds good. Did you update the public API?

thymikee:
Thanks @sidferreira 👍

PR #152, opened 2019-02-05
feat(perf): use lazy module evaluation

thymikee:
Summary:
---------

After #150 (thanks @sidferreira and @Trancever) we're finally able to use lazy module evaluation for additional perf! 

Before:
```
0.70 real         0.64 user         0.12 sys
```
After:
```
0.19 real         0.16 user         0.03 sys
```

Which gives us roughly 0.5s (~72%) improvement in startup time. With 0.2s startup time most commands feel instant now.

Test Plan:
----------

No new tests added.

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -10,5 +10,11 @@ module.exports = {
     ],
     require.resolve('@babel/preset-flow'),
   ],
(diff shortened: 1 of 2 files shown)


grabbou:
This is so good! We got from around 2.5s to 0.5s over two weeks. Can't wait to ship it to React Native developers!

PR #154, opened 2019-02-06
Esemesek/fix init pgkes

Esemesek:
Summary:
---------
- [x] - Added `PackageManager` to use throughout the application to install packages. Replaced usage in `install` and `uninstall` commands, but we should probably handle installing packages the same way everywhere
- [x] - Removed option to not include `jest` in deps. We will have upgraded template support in the future. I think we don't need this additional complexity here, and I don't know anyone who wouldn't install `jest` in a fresh project.

Test Plan:
----------

Added unit tests and tested in an existing project.

### DIFF
--- a/packages/cli/src/init/init.js
+++ b/packages/cli/src/init/init.js
@@ -7,7 +7,8 @@
  * @format
  */
 
-const { execSync } = require('child_process');
+import PackageManager from '../util/PackageManager';
+
 const fs = require('fs');
 const minimist = require('minimist');
 const path = require('path');
@@ -72,6 +73,11 @@ function generateProject(destinationRoot, newProjectName, options) {
     yarn.getYarnVersionIfAvailable() &&
     yarn.isGlobalCliUsingYarn(destinationRoot);
 
+  const packageManager = new PackageManager({
+    projectDir: destinationRoot,
+    forceNpm: options.npm,
+  });
+
   createProjectFromTemplate(
     destinationRoot,
     newProjectName,
@@ -79,28 +85,24 @@ function generateProject(destinationRoot, newProjectName, options) {
     yarnVersion
   );
 
-  if (yarnVersion) {
-    logger.info('Adding React...');
-    execSync(`yarn add react@${reactVersion}`, { stdio: 'inherit' });
-  } else {
-    logger.info('Installing React...');
-    execSync(`npm install react@${reactVersion} --save --save-exact`, {
-      stdio: 'inherit',
-    });
-  }
-  if (!options['skip-jest']) {
-    const jestDeps = `jest babel-jest metro-react-native-babel-preset react-test-renderer@${reactVersion}`;
-    if (yarnVersion) {
-      logger.info('Adding Jest...');
-      execSync(`yarn add ${jestDeps} --dev --exact`, { stdio: 'inherit' });
-    } else {
-      logger.info('Installing Jest...');
-      execSync(`npm install ${jestDeps} --save-dev --save-exact`, {
-        stdio: 'inherit',
-      });
-    }
-    addJestToPackageJson(destinationRoot);
-  }
+  const DEPENDENCIES = [`react@${reactVersion}`];
+
+  const DEV_DEPENDENCIES = [
+    '@babel/core',
+    '@babel/runtime',
+    'jest',
+    'babel-jest',
+    'metro-react-native-babel-preset',
+    `react-test-renderer@${reactVersion}`,
+  ];
+
+  logger.info('Adding required dependencies');
+  packageManager.install(DEPENDENCIES);
+
+  logger.info('Adding required dev dependencies');
+  packageManager.installDev(DEV_DEPENDENCIES);
+
+  addJestToPackageJson(destinationRoot);
   printRunInstructions(destinationRoot, newProjectName);
 }
 
--- a/packages/cli/src/init/init.js
+++ b/packages/cli/src/init/init.js
@@ -68,6 +68,7 @@ function generateProject(destinationRoot, newProjectName, options) {
     return;
   }
 
+  // TODO: Use `PackageManager` in generator to remove `yarn` calls.
   const yarnVersion =
     !options.npm &&
     yarn.getYarnVersionIfAvailable() &&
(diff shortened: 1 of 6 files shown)


packages/cli/src/init/init.js
```diff
@@ -73,35 +73,36 @@ function generateProject(destinationRoot, newProjectName, options) {
     yarn.getYarnVersionIfAvailable() &&
     yarn.isGlobalCliUsingYarn(destinationRoot);
```
thymikee: should this be handled by PackageManager as well?

packages/cli/src/util/PackageManager.js
```diff
@@ -1,77 +1,56 @@
...
- * @param  {string} projectDir  Root directory of the project
- * @return {Object}             spawnSync's result object
- */
-function remove(packageName, projectDir) {
-  return callYarnOrNpm(
-    `yarn remove ${packageName}`,
-    `npm uninstall --save ${packageName}`,
-    projectDir
-  );
+  uninstall(packageName: string) {
```
thymikee: this should accept an array for symmetry with `install`

packages/cli/src/init/init.js
Esemesek: IMO yes. We should have one unit in the system that decides how to install packages and it should work the same way in every scenario. Usage is quite cool and if there will be another package manager in the future, we will just add it here :P

packages/cli/src/util/PackageManager.js
Esemesek: True, I'll change it.

packages/cli/src/util/__tests__/PackageManager-test.js
```diff
@@ -0,0 +1,114 @@
...
+describe('yarn', () => {
+  beforeEach(() => {
+    jest
+      .spyOn(yarn, 'getYarnVersionIfAvailable')
+      .mockImplementation(() => true);
+    jest.spyOn(yarn, 'isGlobalCliUsingYarn').mockImplementation(() => true);
+  });
+
+  it('should install', () => {
+    const uut = new PackageManager({ projectDir: PROJECT_DIR });
```
thymikee: `uut`?

Esemesek: Unit Under Test. Can change that if it's confusing :P

packages/cli/src/init/init.js
grabbou: Why it's not handled by PackageManager then? To be honest, I feel like it could be a great candidate for a separate package.

packages/cli/src/init/init.js
```diff
@@ -73,35 +73,36 @@ function generateProject(destinationRoot, newProjectName, options) {
...
-      logger.info('Installing Jest...');
-      execSync(`npm install ${jestDeps} --save-dev --save-exact`, {
-        stdio: 'inherit',
-      });
-    }
-    addJestToPackageJson(destinationRoot);
-  }
+  const DEPENDENCIES = [`react@${reactVersion}`];
+
+  const DEV_DEPENDENCIES = [
```
grabbou: Do we need these to be defined as separate variables? If yes, then definitely it doesn't need to be uppercase.

packages/cli/src/util/PackageManager.js
```diff
@@ -1,77 +1,56 @@
...
   }
 
-  const args = command.split(' ');
-  const cmd = args.shift();
+  executeCommand(command: string) {
+    return execSync(command, { stdio: 'inherit' });
+  }
 
-  const res = spawnSync(cmd, args, spawnOpts);
+  shouldCallYarn() {
```
grabbou: How about we make this one private? (closure function)

packages/cli/src/init/init.js
Esemesek: We need to put in separate variables because they are installed differently. ALL_UPPER_CASE is a popular naming convention to mark constants.

grabbou: I mean why we don't use them inline, e.g.
`install([]` and `installDev([])`?

They are not really "constants" - at least `dependencies` is dynamic and depends on a `reactVersion`. You know, that way, every `const` variable should be UPPER_CASE :D

packages/cli/src/init/init.js
Esemesek: It is used here to get `yarnVersion` and pass it to a function that actually initializes project (so much spaghetti code there). I wanted to change this as well, but it would require me to do a lot more changes, making this PR too big. I can create follow-up PR to use `PackageManager` everywhere.

packages/cli/src/init/init.js
Esemesek: Constant is not Javascript's `const`. But you are right about `dependencies` - it is dynamic.

grabbou:
Sounds good!

packages/cli/src/init/init.js
grabbou: I know it's not - that's why I don't think we should uppercase variable declarations unless they are at the module level.

packages/cli/src/install/install.js
```diff
@@ -19,13 +19,13 @@ const spawnOpts = {
 function install(args, ctx) {
   const name = args[0];
 
-  let res = new PackageManager({ projectDir: ctx.root }).install([name]);
-
-  if (res.status) {
-    process.exit(res.status);
+  try {
+    new PackageManager({ projectDir: ctx.root }).install([name]);
```
grabbou: No need to do it. All you need to do is to make `install` an `async function` and `try/catch` we have in `cliEntry.js` will log the error automatically.

packages/cli/src/install/install.js
```diff
@@ -19,13 +19,13 @@ const spawnOpts = {
...
+  } catch (e) {
+    process.exit(e.status);
   }
 
-  res = spawnSync('react-native', ['link', name], spawnOpts);
+  const res = spawnSync('react-native', ['link', name], spawnOpts);
```
grabbou: Note for the future: We should do `return link(...)` here instead of spawning CLI - that way, we don't have two error handlers in case failure happens.

PR #157, opened 2019-02-08
Support for template post install script

emin93:
Summary:
---------

Thank you all for the work you've put into this CLI, I really appreciate it! 👍

With this PR I want to add support for template post install scripts. Those scripts come in handy when there's some extra work needed to set up the new project.

For example in the [TypeScript template](https://github.com/emin93/react-native-template-typescript) some extensions to the Jest configuration in the `package.json` have to be made. Instead of a manual process, a post install script could take care of it.

Test Plan:
----------

Went with the guide in the `CONTRIBUTION.md` document, installed Verdaccio, released a new version of React Native to it and used it to create a new project with a template that doesn't have a post install script and one that has it, to check if both scenarios will work.

### DIFF
--- a/packages/cli/src/generator/templates.js
+++ b/packages/cli/src/generator/templates.js
@@ -95,7 +95,7 @@ function createFromRemoteTemplate(
     });
     installTemplateDependencies(templatePath, yarnVersion);
     installTemplateDevDependencies(templatePath, yarnVersion);
-    executePostInstallScript(templatePath);
+    executeTemplatePostInstallScript(templatePath);
   } finally {
     // Clean up the temp files
     try {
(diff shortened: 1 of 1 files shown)


grabbou:
Hey, thanks for submitting this PR.

Looks interesting as a workaround to solving some of the issues we have faced with current implementation of generators, such as removing extraneous fonts.

@Esemesek and @thymikee - let me know what you think.

thymikee:
We could merge it as a temporary solution, but we're going to change templates to just run `yarn install`, so postinstall scripts and alike will _just work_. So I lean towards closing this and move to generic solution that @Esemesek is working on.

grabbou:
I also think we need to whitelist that `rn-post-install.js` file here: https://github.com/react-native-community/react-native-cli/blob/c9c3ca3b7c5e6d42435d10eb8a140e371d280e1c/packages/cli/src/generator/templates.js#L92. Otherwise, it will get added onto user space, no?

emin93:
@grabbou Good point, I added an entry to the `ignorePaths`.
@thymikee That would definitely be the cleaner solution 👍.

I'm closing this PR for now.

PR #160, opened 2019-02-11
feat: show help when no arguments are provided

### LINKED ISSUE #158: Show help by default
## Describe the Feature
Right now, when typing `react-native` without any arguments, it will print this:
```
You didn't pass any command
  Run react-native --help to see list of all available commands
```

Can we remove this and print the help screen? I think that's clearer.

lucasbento:
Summary:
---------

This PR adds the default behaviour of calling the CLI without any args to be the `--help` command.

Fixes #158.

Test Plan:
----------

Run the following:

```bash
node path/to/the/react-native-cli/packages/cli/build/index.js
node path/to/the/react-native-cli/packages/global-cli/index.js
```

### DIFF
--- a/packages/cli/src/cliEntry.js
+++ b/packages/cli/src/cliEntry.js
@@ -70,16 +70,15 @@ function printHelpInformation() {
 }
 
 function printUnknownCommand(cmdName) {
-  logger.error(
-    [
-      cmdName
-        ? `Unrecognized command "${cmdName}".`
-        : "You didn't pass any command.",
-      `Run ${chalk.white(
-        '"react-native --help"'
-      )} to see list of all available commands.`,
-    ].join(' ')
-  );
+  cmdName
+    ? logger.error(
+        [
+          `Run ${chalk.white(
+            '"react-native --help"'
+          )} to see list of all available commands.`,
+        ].join(' ')
+      )
+    : commander.outputHelp();
 }
 
 const addCommand = (command: CommandT, ctx: ContextT) => {
--- a/packages/cli/src/cliEntry.js
+++ b/packages/cli/src/cliEntry.js
@@ -72,11 +72,10 @@ function printHelpInformation() {
 function printUnknownCommand(cmdName) {
   cmdName
     ? logger.error(
-        [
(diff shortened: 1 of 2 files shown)


packages/cli/src/cliEntry.js
```diff
@@ -70,16 +70,15 @@ function printHelpInformation() {
...
-        : "You didn't pass any command.",
-      `Run ${chalk.white(
-        '"react-native --help"'
-      )} to see list of all available commands.`,
-    ].join(' ')
-  );
+  cmdName
+    ? logger.error(
+        [
+          `Run ${chalk.white(
```
thymikee: You missed the unrecognized command message, can you add it back?

packages/cli/src/cliEntry.js
```diff
@@ -70,16 +70,15 @@ function printHelpInformation() {
...
+            '"react-native --help"'
+          )} to see list of all available commands.`,
+        ].join(' ')
```
thymikee: Logger.error will join arguments with a space so no need for that now (array with join is a legacy behavior we changed revently)

lucasbento: Removed, thanks for pointing out!

packages/cli/src/cliEntry.js
lucasbento: Added back, thanks for noticing this!

I also found an issue with this part, if you do `cli --unknown-command` it will be parsed by commander but `cli unknown-command` will be handled by this function which will lead to two different error messages.

thymikee: >I also found an issue with this part, if you do `cli --unknown-command` it will be parsed by commander but `cli unknown-command` will be handled by this function which will lead to two different error messages.

We're going to rewrite `commander` to `yargs` soon, so let's not bother this inconsistency for now :)

thymikee:
I made some visual adjustments locally. What do you think?

screenshot 2019-02-11 at 11 47 46

lucasbento:
Cool, thanks @thymikee & @Esemesek!

PR #162, opened 2019-02-13
fix: improve cli interface

grabbou:
Summary:
---------

This fixes few of the issues that I've encountered working in a React Native repository.

When you pass a wrong command or provide no command at all (say, you want to print `--help`) and your CLI works by setting `--projectRoot` or `--reactNativePath` (in React Native repository, this is a must), you end up with a "Unrecognised option" error.

This PR changes that, so that it works and is accepted.

It also fixes a missing path to a inited application that is a regression introduced in https://github.com/react-native-community/react-native-cli/commit/72c3dd4282996891442c3ec779e99d989a74c440#diff-d9b7abe65a1a1a0e0b19ec93292008eaR16. We wrongly assumed `process.cwd()` remains the directory the command has been run from. Turns out, it's modified by the init script.

### DIFF
--- a/packages/cli/src/cliEntry.js
+++ b/packages/cli/src/cliEntry.js
@@ -118,6 +118,11 @@ const addCommand = (command: CommandT, ctx: ContextT) => {
       typeof opt.default === 'function' ? opt.default(ctx) : opt.default
     )
   );
+
+  // Redefined here to appear in the `--help` section
+  cmd
+    .option('--projectRoot [string]', 'Path to the root of the project')
+    .option('--reactNativePath [string]', 'Path to React Native');
 };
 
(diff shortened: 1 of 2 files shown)


packages/cli/src/cliEntry.js
```diff
@@ -177,17 +181,14 @@ async function setupAndRun() {
...
-  if (!isValidCommand) {
-    printUnknownCommand(process.argv[2]);
-    return;
+  if (!command) {
+    commander.help();
   }
 
-  if (!commander.args.length) {
-    commander.help();
+  if (typeof command === 'string') {
```
thymikee: this could be `else` clause to be more explicit :)

PR #166, opened 2019-02-13
fix: update react-native-cli README

thymikee:
Summary:
---------

Current README of `react-native-cli` is outdated and totally not user friendly. We already have a contributing guide on how to run it locally, so there's no point in keeping this one. I took this opportunity to write a proper README.

cc @kelset

### DIFF
--- a/packages/global-cli/README.md
+++ b/packages/global-cli/README.md
@@ -24,7 +24,7 @@ It will install `react-native`, `react`, `jest` and a bunch of other necessary p
 
 After it's finished, your AwesomeProject should be ready to use. From this point, you should use your local `react-native` binary to run the proper [React Native CLI](../cli):
 
-```
+```sh
 yarn react-native link native-dep
 ```
 
@@ -36,7 +36,7 @@ Use a custom template.
 
 Example: this will install init your AwesomeProject using template called `react-native-template-samplename` from npm:
 
-```
+```sh
 react-native init AwesomeProject --template samplename
 ```
 
@@ -48,8 +48,14 @@ Use a custom version. By default `react-native init` will use the latest stable.
 
 Example: this will install init your AwesomeProject using version `0.57.0`:
 
+```sh
+react-native init AwesomeProject --version 0.57.0
 ```
-react-native init AwesomeProject --template 0.57.0
+
+You can also install a specific tag, like `next`, using:
+
+```sh
+react-native init AwesomeProject --version react-native@next
 ```
 
 ## Future work
--- a/packages/global-cli/README.md
+++ b/packages/global-cli/README.md
@@ -56,4 +56,4 @@ react-native init AwesomeProject --template 0.57.0
 
 We understand this is counter-intuitive to have two packages for interacting with React Native and it makes the first experience with the framework a bit confusing.
 
-That's why, as a part of [Lean Core](https://github.com/facebook/react-native/issues/23313) initiative, there's an ongoing effort to remove this module so we can use just `react-native` as the only package necessary to install and run React Native commands.
+That's why, as a part of [Lean Core](https://github.com/facebook/react-native/issues/23313) initiative, there's an [ongoing effort](https://github.com/react-native-community/react-native-cli/issues/76) to remove this module so we can use just `react-native` as the only package necessary to install and run React Native commands.
--- a/packages/global-cli/README.md
+++ b/packages/global-cli/README.md
@@ -2,22 +2,54 @@
 
 `react-native-cli` is a command line tool for initializing React Native projects.
 
-## Usage
+## Installation
 
-We recommend using `react-native-cli` with `npx` so you don't need to pollute your global packages scope.
+Install it as a global module:
+
+```sh
+yarn global add react-native-cli
+```
+
+It will create a global `react-native` binary (even though the package name is `react-native-cli`).
+
+## Usage
 
 To create a new React Native Project called "AwesomeProject" you can run:
 
 ```sh
-npx react-native-cli init AwesomeProject
+react-native init AwesomeProject
 ```
 
-It will install `react-native`, `react`, `jest` and a bunch of other necessary packages from the default template.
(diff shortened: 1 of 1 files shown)


packages/global-cli/README.md
```diff
@@ -1,126 +1,27 @@
...
+## React Native Global CLI
 
-React Native is distributed as two npm packages, `react-native-cli` and `react-native`. The first one is a lightweight package that should be installed globally (`npm install -g react-native-cli`), while the second one contains the actual React Native framework code and is installed locally into your project when you run `react-native init`.
+`react-native-cli` is a command line tool for initializing React Native projects.
 
-Because `react-native init` calls `npm install react-native`, simply linking your local github clone into npm is not enough to test local changes.
+## Usage
 
-### Introducing Sinopia
+We recommend using `react-native-cli` with `npx` so you don't need to pollute your global packages scope.
```
kelset: TBH I’m personally again `npx` because of security reasons, maybe write the command to install globally too?

packages/global-cli/README.md
```diff
@@ -1,126 +1,27 @@
...
-[Sinopia] is an npm registry that runs on your local machine and allows you to publish packages to it. Everything else is proxied from `npmjs.com`. We'll set up sinopia for React Native CLI development. First, install it with:
+To create a new React Native Project called "AwesomeProject" you can run:
 
-    $ npm install -g sinopia
+```sh
+npx react-native-cli init AwesomeProject
+```
 
-Now you can run sinopia by simply doing:
+It will install `react-native`, `react`, `jest` and a bunch of other necessary packages from the default template.
```
kelset: Will the template option still be available? Should it be documented here?

packages/global-cli/README.md
```diff
@@ -1,126 +1,27 @@
...
+yarn react-native link native-dep
+```
 
-    storage: ./storage
+## Future work
 
-    auth:
-      htpasswd:
-        file: ./htpasswd
+We understand this is counter-intuitive to have two packages for interacting with React Native and it makes the first experience with the framework a bit confusing.
```
kelset: Oh wow I didn’t realize that we need both `react-native-cli` and `yarn react-native` 😰

Should we also mention somewhere that folsk should uninstall the previous version of the CLI? Maybe at the start where you mention how to install?

packages/global-cli/README.md
thymikee: What kind of security reasons? I don't think I'm aware :o

kelset: Well basically npx (AFAIK) is able to run *anything* (= a pack from the web that contains whatever) on your machine as basically an admin... which is a big no no in my books ;)

thymikee: Oh, I guess that's when you use global node version. Will adjust that.

packages/global-cli/README.md
thymikee: Haha yea, it's pretty bad. This is not necessary now that we advise to use global package. If someone is familiar with `npx` they can use it on their own anyway :)

packages/global-cli/README.md
thymikee: Added `--template` and `--version`. Any other options I'm not aware of?

packages/global-cli/README.md
```diff
@@ -1,126 +1,59 @@
...
-If you don't want to install Sinopia you could still test changes done on the cli by creating a sample project and installing your checkout of `react-native` on that project instead of downloading it from npm. The simplest way to do this is by:
-
-    $ npm init AwesomeProject
-    $ cd AwesomeProject
-    $ npm install $REACT_NATIVE_GITHUB
-
-Note that `REACT_NATIVE_GITHUB` should point to the directory where you have a checkout.
-
-Also, if the changes you're making get triggered when running `react-native init AwesomeProject` you will want to tweak the global installed `react-native-cli` library to install the local checkout instead of downloading the module from npm. To do so just change this [line](https://github.com/facebook/react-native/blob/master/react-native-cli/index.js#L191) and refer the local checkout instead.
+That's why, as a part of [Lean Core](https://github.com/facebook/react-native/issues/23313) initiative, there's an ongoing effort to remove this module so we can use just `react-native` as the only package necessary to install and run React Native commands.
```
grabbou: We could reference the issue where the discussion happens

thymikee: linked

packages/global-cli/README.md
```diff
@@ -1,126 +1,65 @@
...
+```sh
+react-native init AwesomeProject --version 0.57.0
+```
 
-### Running the local CLI
+You can also install a specific tag, like `next`, using:
 
-Now that the packages are installed in sinopia, you can install the new `react-native-cli` package globally and when you use `react-native init`, it will install the new `react-native` package as well:
+```sh
+react-native init AwesomeProject --version react-native@next
```
thymikee: @grabbou mind reviewing this section? It's not clear to me when users should use `react-native@0.57.0` and just `0.57.0`

packages/global-cli/README.md
kelset: don't remember tbh - but we can start with these 👍

packages/global-cli/README.md
kelset: 👍

packages/global-cli/README.md
```diff
@@ -1,126 +1,65 @@
...
-    auth:
-      htpasswd:
-        file: ./htpasswd
+After it's finished, your AwesomeProject should be ready to use. From this point, you should use your local `react-native` binary to run the proper [React Native CLI](../cli):
 
-    uplinks:
-      npmjs:
-        url: https://registry.npmjs.org/
+```sh
+yarn react-native link native-dep
```
kelset: does the `yarn` command still needed to be used here, or was it because you used `npx` earlier in the section?

thymikee: I stand that we should encourage using local `react-native` installation (with the real CLI, not a proxy like `react-native-cli`) to always use up-to-date CLI, instead of messing with globals.

PR #171, opened 2019-02-15
fix: printing error and unknown command messages when unnecessary

grabbou:
Summary:
---------

The logic for printing an unknown command was always a hack and the one for printing a help when no command provided never worked.

I did a research and found out there's a recommended (Commander way) of doing this.

It also fixes a regression I've noticed in https://github.com/react-native-community/react-native-releases/issues/79#issuecomment-463933483

Test Plan:
----------

Run w/o command, run with wrong command.


grabbou:
Gotta do it soon!

PR #174, opened 2019-02-15
feat: better logging for install/uninstall; less spawning; fix unlinking

thymikee:
Summary:
---------

- [x] use link/unlink directly from source instead of spawning
- [x] fix unlinking failing to get dependencies
- [x] improve logging

Test Plan:
----------

screenshot 2019-02-15 at 23 40 29
screenshot 2019-02-15 at 23 40 50

### DIFF
--- a/packages/cli/src/link/unlink.js
+++ b/packages/cli/src/link/unlink.js
@@ -9,16 +9,14 @@
 
 import { flatten, isEmpty, difference } from 'lodash';
 import type { ContextT } from '../core/types.flow';
-
-import log from '../util/logger';
+import logger from '../util/logger';
 import getProjectConfig from './getProjectConfig';
 import getDependencyConfig from './getDependencyConfig';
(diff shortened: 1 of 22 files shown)


packages/cli/src/link/unlink.js
```diff
@@ -85,13 +87,13 @@ function unlink(args: Array<string>, ctx: ContextT) {
...
   } catch (err) {
-    log.error(
+    logger.error(
       "No package.json found. Are you sure it's a React Native project?"
     );
     return Promise.reject(err);
   }
 
-  const allDependencies = getProjectDependencies().map(dependency =>
+  const allDependencies = getProjectDependencies(ctx.root).map(dependency =>
```
thymikee: this is a bugfix

Esemesek: Good to know C:

thymikee:
You think? I think it's nice, I didn't even know about it because the commands are not documented

PR #182, opened 2019-02-19
fix: setup_env script failure

### LINKED ISSUE #164: setup_env.sh ulimit script not actually calling
My issue is specifically regarding Environment config and react-native

I am testing multiple environments, including users without admin access, and I initially suspected reviewing the code that the react-native-cli wouldn't even run without admin privileges, however the app did run at least up until 0.58.* and now on react-native link command throws 

An error was encountered processing the command (domain=IXUserPresentableErrorDomain, code=1):
....
Failed to chmod x2

Underlying error (domain=MIInstallerErrorDomain, code=4):

Failed to chmod

Then fails launching the app.

## Reproducible Demo

I am building to provide something, however users should be able to test with their own mac users and see similar failing behavior; and maybe the error should really be explicitly where admin access is needed or what is causing this error if we can simply fix it. The idea should be I think that root but non-admin users should be able to develop with react native.

the setup_env.sh script does not seem to be calling.

thymikee:
Summary:
---------

Likely fixes #164 by wrapping the `spanwSync` call in a try/catch.

cc @Kielan

Test Plan:
----------

It works™

### DIFF
--- a/packages/cli/src/cliEntry.js
+++ b/packages/cli/src/cliEntry.js
@@ -147,7 +147,7 @@ async function setupAndRun() {
       childProcess.execFileSync(absolutePath, { stdio: 'pipe' });
     } catch (error) {
       logger.warn(
-        `Failed to run environment setup script "${scriptName}".\n\n${chalk.red(
+        `Failed to run environment setup script "${scriptName}"\n\n${chalk.red(
           error
         )}`
       );
--- a/packages/cli/src/cliEntry.js
+++ b/packages/cli/src/cliEntry.js
@@ -139,15 +139,20 @@ async function run() {
 
 async function setupAndRun() {
   // We only have a setup script for UNIX envs currently
-  if (/^win/.test(process.platform) === false) {
+  if (process.platform !== 'win32') {
     const scriptName = 'setup_env.sh';
     const absolutePath = path.join(__dirname, '..', scriptName);
 
     try {
-      childProcess.execFileSync(absolutePath);
+      childProcess.execFileSync(absolutePath, { stdio: 'pipe' });
     } catch (error) {
       logger.warn(
-        `Failed to run environment setup script "${scriptName}". This script can continue to run if your local environment matches what React Native expects. If it does fail, check out "${absolutePath}" and adjust your environment to match it.\n${error}`
+        `Failed to run environment setup script "${scriptName}".\n\n${chalk.red(
+          error
+        )}`
+      );
+      logger.info(
+        `React Native CLI will continue to run if your local environment matches what React Native expects. If it does fail, check out "${absolutePath}" and adjust your environment to match it.`
       );
     }
   }
--- a/packages/cli/src/cliEntry.js
+++ b/packages/cli/src/cliEntry.js
@@ -138,16 +138,18 @@ async function run() {
 }
 
 async function setupAndRun() {
-  const setupEnvScript = /^win/.test(process.platform)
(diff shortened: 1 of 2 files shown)


packages/cli/src/cliEntry.js
```diff
@@ -142,7 +142,13 @@ async function setupAndRun() {
     ? path.join('..', 'setup_env.bat')
     : path.join('..', 'setup_env.sh');
 
-  childProcess.execFileSync(path.join(__dirname, setupEnvScript));
+  try {
+    childProcess.execFileSync(path.join(__dirname, setupEnvScript));
+  } catch (error) {
+    logger.warn(
+      `Failed to run environment setup script, but we can continue without it. You may want to apply the steps manually.\n${error}`
```
cpojer: This isn't super descriptive and doesn't tell people what to do to fix it either. What about something like:
```
Failed to run environment setup script (`setup_env.sh|bat`). This script can continue to run if your local environment matches what React Native expects. If it does fail, check out `path/to/setup_env` and adjust your environment to match it.
```

?

thymikee: Oh yea, way better wording. Btw while doing that I noticed that the windows setup_env script is (and has always been) a file with only comments. What's the idea behind it?

cpojer: It may have just been a placeholder. Feel free to kill it and ignore this codepath on Windows altogether.

thymikee: Adjusted the wording and removed Windows script ✨

packages/cli/src/cliEntry.js
```diff
@@ -138,11 +138,19 @@ async function run() {
 }
 
 async function setupAndRun() {
-  const setupEnvScript = /^win/.test(process.platform)
-    ? path.join('..', 'setup_env.bat')
-    : path.join('..', 'setup_env.sh');
-
-  childProcess.execFileSync(path.join(__dirname, setupEnvScript));
+  // We only have a setup script for UNIX envs currently
+  if (/^win/.test(process.platform) === false) {
```
cpojer: This seems like a shady test. Is there something that is not regex that would be more reliable?

packages/cli/src/cliEntry.js
```diff
@@ -138,11 +138,19 @@ async function run() {
...
+    const scriptName = 'setup_env.sh';
+    const absolutePath = path.join(__dirname, '..', scriptName);
+
+    try {
+      childProcess.execFileSync(absolutePath);
+    } catch (error) {
+      logger.warn(
+        `Failed to run environment setup script "${scriptName}". This script can continue to run if your local environment matches what React Native expects. If it does fail, check out "${absolutePath}" and adjust your environment to match it.\n${error}`
```
cpojer: Actually it seems like `This script can continue to run` may be confusing (what is "script"). Could you change it to `The react-native CLI continues to run` or something?

packages/cli/src/cliEntry.js
cpojer: Loads better! With your change, if there is an OS variant of linux coming around that they call `linuxIsNotWin`, we'd be screwed.

packages/cli/src/cliEntry.js
thymikee: Adjusted the wording and added red color to original error:

screenshot 2019-02-20 at 09 02 43

cpojer: Awesome!

PR #185, opened 2019-02-20
Allow absolute path for templates

### LINKED ISSUE #184: Template arg doesn't pick up absolute path
## Environment

N/A

## Description

Trying to point template to an absolute path fails, even though this is a valid way to use `yarn`.
```
react-native init EndToEndTest --template file:/User/ericlewis/work/react-native
```

Offending code: https://github.com/react-native-community/react-native-cli/blob/72c3dd4282996891442c3ec779e99d989a74c440/packages/cli/src/generator/templates.js#L63

## Reproducible Demo

N/A

ericlewis:
Summary:
---------

Fix #184 

Test Plan:
----------

The test case is now correct.

### DIFF
--- a/packages/cli/src/generator/templates.js
+++ b/packages/cli/src/generator/templates.js
@@ -60,8 +60,8 @@ function createFromRemoteTemplate(
 ) {
   let installPackage;
   let templateName;
-  if (template.includes('://')) {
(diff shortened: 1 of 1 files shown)


cpojer:
Thanks!

PR #191, opened 2019-02-27
feat: Allow projectRoot and reactNativePath to be read from metro config

rajivshah3:
Summary:
---------

Currently, there's no way to specify `projectRoot` or `reactNativePath` in a config file (as shown in the example in https://github.com/react-native-community/react-native-cli/pull/167). This means that these parameters have to be specified as arguments in every `react-native` command, which poses a problem for repos with a nonstandard `reactNativePath`. As a result, these repos (namely `react-native` itself) cannot upgrade to later versions of `@react-native-community/cli` without updating all of their test scripts (see https://github.com/facebook/react-native/pull/23627, where the tests are failing because the scripts haven't been updated). An easier way to do this would be to simply specify `projectRoot` and `reactNativePath` in the CLI/Metro config (see [example](https://github.com/facebook/react-native/pull/23627/commits/6f731fe7d247cb014c4b827edfd23c7c6fb670ef))

Test Plan:
----------

Running `node ../react-native-cli/packages/cli/build/index.js` in facebook/react-native@6f731fe no longer throws an error

### DIFF
--- a/packages/cli/src/cliEntry.js
+++ b/packages/cli/src/cliEntry.js
@@ -12,6 +12,7 @@ import childProcess from 'child_process';
 import commander from 'commander';
 import minimist from 'minimist';
 import path from 'path';
+import { loadConfig } from 'metro';
 import type { CommandT, ContextT } from './core/types.flow';
 import getCommands from './core/getCommands';
 import getLegacyConfig from './core/getLegacyConfig';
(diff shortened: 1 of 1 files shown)


thymikee:
These are not valid Metro configs and we shouldn't mix them. 
@grabbou I think it's the right time to think about config for the the CLI itself, separate from Metro.

Something like:
```js
// rncli.config.js
module.exports = {
  projectRoot: '/foo',
  bundlerOptions: {} // likely not necessary, metro.config.js / webpack.config.js should handle bundler options
}
```

We could use https://github.com/davidtheclark/cosmiconfig for resolving/loading config easily.

rajivshah3:
> These are not valid Metro configs and we shouldn't mix them.

I agree. I was basically trying to avoid making any significant changes, but if you and @grabbou would prefer to add a more "official" config file (like the example you posted) then I think that would be a good idea too.

grabbou:
Yeah, part of the reason we moved files around was to decouple CLI from Metro. Historically, Metro configuration was shared with CLI one (in fact, there was nothing CLI was reading from that config, so it was just misleading). 

I think `cosmiconfig` is a great and standard solution.

grabbou:
Thank you for doing investigation on this issue @rajivshah3. I have already answered in the React Native PR with some hints how to get this fixed for time being.

Do you feel like having some time in the nearest future to work on the configuration? I would be more than happy to assist when needed.

grabbou:
Going to close this PR in the meantime.

rajivshah3:
> Do you feel like having some time in the nearest future to work on the configuration? I would be more than happy to assist when needed.

Sure!

PR #200, opened 2019-03-04
chore: bump flow to 0.94

thymikee:
Summary:
---------

It's usually better to be on the latest available :D 

Test Plan:
----------

Green CI

### DIFF
--- a/packages/cli/src/core/types.flow.js
+++ b/packages/cli/src/core/types.flow.js
@@ -18,7 +18,7 @@ export type LocalCommandT = {
     command: string,
     description?: string,
     parse?: (val: string) => any,
-    default?: ((ctx: ContextT) => mixed) | mixed,
+    default?: string | boolean | number,
   }>,
   examples?: Array<{
     desc: string,
--- a/packages/cli/src/util/loadMetroConfig.js
+++ b/packages/cli/src/util/loadMetroConfig.js
@@ -2,7 +2,6 @@
  * Configuration file of Metro.
  * @flow
  */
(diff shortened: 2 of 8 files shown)


packages/cli/src/core/types.flow.js
```diff
@@ -18,7 +18,7 @@ export type LocalCommandT = {
     command: string,
     description?: string,
     parse?: (val: string) => any,
-    default?: ((ctx: ContextT) => mixed) | mixed,
```
thymikee: turned out the only `default` values we use are `string | boolean | number `

packages/cli/src/util/loadMetroConfig.js
```diff
@@ -90,7 +89,7 @@ export default async function load(
   ctx: ContextT,
   // $FlowFixMe - troubles with empty object being inexact
   options?: ConfigOptionsT = {}
-): Promise<ConfigT> {
+) {
```
Esemesek: Why you removed type signature here?

packages/cli/src/core/types.flow.js
Esemesek: Sweet 🍰

packages/cli/src/util/loadMetroConfig.js
thymikee: because Flow sees `ConfigT` as `any`, for some reason it's unable to get the types from `metro-config` even though they have `.js.flow` files. Didn't investigate though

PR #201, opened 2019-03-04
chore: simplify cli structure and common tools

thymikee:
Summary:
---------

Cleaning up the `packages/cli` structure. This might be pretty disruptive to already open PRs :D.

- Renamed `core` -> `tools`
- Moved `util` to `tools`
- Flatten-out `server/util`
- Create `commands` dir and put all commands there

```
packages/cli/src
├── bin.js
├── cliEntry.js
├── commands
│   ├── bundle
│   ├── dependencies
│   ├── eject
│   ├── index.js
│   ├── info
│   ├── init
│   ├── install
│   ├── library
│   ├── link
│   ├── logAndroid
│   ├── logIOS
│   ├── runAndroid
│   ├── runIOS
│   ├── server
│   └── upgrade
├── index.js
└── tools
    ├── ...
``` 

Fixes https://github.com/react-native-community/react-native-cli/issues/49

Test Plan:
----------

CI green


grabbou:
Thanks @thymikee for taking up this PR. If tests are green, I believe it's good to go. 

Might require some work on React Native side that uses private API to make sure the paths are correct. If there's work required on React Native side, we need to make this a breaking change.

thymikee:
Can you check that?

thymikee:
Also, no RN version lower than 0.59 installs this CLI, so we could only break RC (which is not good, but you know, it's RC)

cpojer:
Breaking RC should be fine. At FB, nothing depends on this code, so you can do whatever you want as long as it works with RN.

thymikee:
So it turns out that `hasteImpl` in RN uses it:
https://github.com/facebook/react-native/blob/9cb4d3f2c1e133d40d5c5d44583dc45243e033bd/jest/hasteImpl.js#L28-L29

We could apply this diff to make it backcompat:

```diff
--- a/packages/cli/src/cliEntry.js
+++ b/packages/cli/src/cliEntry.js
@@ -17,6 +17,7 @@ import getLegacyConfig from './tools/getLegacyConfig';
 import { getCommands } from './commands';
 import init from './commands/init/init';
 import assertRequiredOptions from './tools/assertRequiredOptions';
+import findPlugins from './tools/findPlugins';
 import logger from './tools/logger';
 import pkg from '../package.json';

@@ -208,4 +209,6 @@ export default {
   init,
 };

+export { findPlugins };
+
 // export { run, init };
diff --git a/packages/cli/src/core/findPlugins.js b/packages/cli/src/core/findPlugins.js
new file mode 100644
index 0000000..d1ac206
--- /dev/null
+++ b/packages/cli/src/core/findPlugins.js
@@ -0,0 +1,5 @@
+/**
+ * TODO: This file exists only because RN 0.59.0-rc depends on importing private
+ * methods. Remove it once it's updated with public API.
+ */
+module.exports = require('../tools/findPlugins');
```

grabbou:
Double checked and we should be good to make a breaking change now.

The CLI is only consumed by 0.59-stable release candidates. I am about to make a new one with few of the cherry-picks, so we can technically do the breaking change (we need to bump major to avoid older RCs installing this one).

PR #209, opened 2019-03-07
chore: simplify commander/minimist usage

thymikee:
Summary:
---------

I was playing a bit with `commander` and took the opportunity to remove some code.
If we ever move to `yargs` (I don't see this as beneficial, as we'd need to do the same hacks with `version`), it should be easier now. 

Test Plan:
----------

Test it carefully, I might have broken something (although from testing locally on a bunch of commands it seemed ok)


packages/cli/src/cliEntry.js
```diff
@@ -93,26 +91,18 @@ const addCommand = (command: CommandT, ctx: ContextT) => {
...
-    })
+    .command(command.name, undefined, {noHelp: !command.description})
     .description(command.description)
     .action(function handleAction(...args) {
       const passedOptions = this.opts();
       const argv: Array<string> = Array.from(args).slice(0, -1);
 
       Promise.resolve()
-        .then(() => {
-          assertRequiredOptions(options, passedOptions);
```
thymikee: `commander` validation actually works and commands fail when required arguments are not passed, so this fn was never called?

packages/cli/src/cliEntry.js
```diff
@@ -93,26 +91,18 @@ const addCommand = (command: CommandT, ctx: ContextT) => {
...
-          return command.func(argv, ctx, passedOptions);
-        })
+        .then(() => command.func(argv, ctx, passedOptions))
         .catch(handleError);
     });
 
-  cmd.helpInformation = printHelpInformation.bind(cmd);
-  cmd.examples = command.examples;
-  // $FlowFixMe: This is either null or not
-  cmd.pkg = command.pkg;
```
thymikee: seems like it's not used anywhere, @grabbou ideas?

grabbou: `pkg` is still used on line 48. You may consider passing it to `bind` too

packages/cli/src/cliEntry.js
```diff
@@ -122,11 +112,6 @@ const addCommand = (command: CommandT, ctx: ContextT) => {
       opt.default,
     ),
   );
-
-  // Redefined here to appear in the `--help` section
-  cmd
-    .option('--projectRoot [string]', 'Path to the root of the project')
-    .option('--reactNativePath [string]', 'Path to React Native');
```
grabbou: Once removed, they will not appear in the `--help` for a command, right?

packages/cli/src/cliEntry.js
grabbou: This one checks for required options, see the linked issue in `assertRequiredOptions.js` file. It allows to require options, when the: "required: true" is set.

packages/cli/src/cliEntry.js
grabbou: I guess we can tweak `printHelpInformation` a bit to render global options too.

packages/cli/src/cliEntry.js
grabbou: I don't think we use `required: true` in any of the options for built-in commands, but it's a breaking change (potentially) for other users that provide 3rd party commands. Would consider leaving it as it's a nice feature - we can always work upstream.

thymikee: Oh didn't realise that :<

thymikee: You definitely need to document stuff like that 😛

grabbou: It's all documented in the `assertRequiredOptions` file description! I wouldn't remember that myself otherwise

thymikee: I mean documenting plugins to CLI and how they work

packages/cli/src/cliEntry.js
thymikee: Looks like they show up just right:

Screenshot 2019-03-08 at 13 54 48

thymikee: Ok, it's not right, I've added extra `.parse()` call and it skips commands. Gonna work on it

grabbou: Yeah, but this was to handle `react-native start --help` (help for a command) where they don't show up anymore.

Screenshot 2019-03-11 at 10 41 07

thymikee: Ok, just removed this extra call and help is complete:

Screenshot 2019-03-11 at 10 43 30

thymikee: ugh

packages/cli/src/cliEntry.js
grabbou: Yeah, CLI was historically missing some documentation, but it's finally improving!

packages/cli/src/cliEntry.js
thymikee: brought it back with (imho) more appropriate comment

packages/cli/src/cliEntry.js
```diff
@@ -120,6 +120,15 @@ const addCommand = (command: CommandT, ctx: ContextT) => {
       opt.default,
     ),
   );
+
+  /**
+   * We want all commands (like "start", "link") to accept these flags, so we
+   * add them here instead of in every single command. This way they'll be
```
grabbou: Hmm... "we add them here instead of in every single command". In this context, "here" is "in every single command" actually.

cpojer:
Nice work! :)

PR #217, opened 2019-03-11
chore: remove some unused files

thymikee:
Summary:
---------

Seems like we have a bunch of unused files, so I removed them.
I also removed `fs-extra` (although we still have it in deps because of Metro 0.51) because it's not necessary with Node 8 anymore and we were on v1.0 (current is 7.0 :D )

Test Plan:
----------

Green CI


packages/cli/src/commands/server/jsPackagerClient.js
```diff
@@ -1,121 +0,0 @@
-/**
```
thymikee: I also checked Metro for any calls to this file, but there's none, so removed.

PR #221, opened 2019-03-11
Fix Android unlink regression

matt-block:
Summary:
---------
This PR fixes a regression introduced in https://github.com/react-native-community/react-native-cli/commit/c6c34a2a6e0d5898a62bc10b7c2936cadba3e071 where only `implementation` type dependencies would be unlinked correctly. Previously linked `compile` dependencies would not be fully unlinked. This issue has been discussed also in #43, #81 and [here](https://github.com/facebook/react-native/pull/21043).

This PR augments the Android `makeBuildPatch` to accept the path of the Gradle build configuration as a second parameter so that the correct patch string can be detected instead of returning the hardcoded `implementation` variant.

The path to the Gradle file is optional and if none is provided, `makeBuildPatch` will return the default  `implementation` patch string as it has always been since the introduction of the regression. This behavior is preferred since this feature is only needed when unlinking and other files like `registerNativeModule` used by `link` don't need to be edited.

An additional test is added to verify the detection of the patch string used in the Gradle build file.

Test Plan:
----------
Try unlinking any `compile`, `implementation` or `api` dependency.

CC @grabbou

### DIFF
--- a/packages/cli/src/commands/link/android/patches/makeBuildPatch.js
+++ b/packages/cli/src/commands/link/android/patches/makeBuildPatch.js
@@ -10,10 +10,12 @@
 import fs from 'fs';
 import normalizeProjectName from './normalizeProjectName';
 
+const depTypes = ['compile', 'api', 'implementation'];
+
 export default function makeBuildPatch(name, buildGradlePath) {
   const normalizedProjectName = normalizeProjectName(name);
   const installPattern = new RegExp(
-    `(implementation|api|compile)\\w*\\s*\\(*project\\s*\\(['"]:${normalizedProjectName}['"]\\)`,
+    buildDepRegExp(normalizedProjectName, ...depTypes),
   );
 
   return {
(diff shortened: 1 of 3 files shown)


packages/cli/src/commands/link/android/patches/makeBuildPatch.js
```diff
@@ -18,6 +19,28 @@ export default function makeBuildPatch(name) {
...
+  if (!buildGradlePath) {
+    return defaultPatchString;
+  }
+
+  const buildGradle = fs.readFileSync(buildGradlePath);
+
+  const depTypes = ['compile', 'api', 'implementation'];
+  for (const type of depTypes) {
+    const depPattern = new RegExp(
+      `(${type})\\w*\\s*\\(*project\\(['"]:${normalizedProjectName}['"]\\)`,
```
thymikee: Should we care about a whitespace between `project` and `(:lib-name)`? It's our CLI that generates it, so that shouldn't be an issue unless somebody modifies gradle config, maybe with a formatter?

matt-block: True, we should account for possible white spaces in that spot, just in case.

packages/cli/src/commands/link/android/patches/makeBuildPatch.js
```diff
@@ -7,17 +7,40 @@
...
 
+import fs from 'fs';
 import normalizeProjectName from './normalizeProjectName';
 
-export default function makeBuildPatch(name) {
+export default function makeBuildPatch(name, buildGradlePath) {
   const normalizedProjectName = normalizeProjectName(name);
   const installPattern = new RegExp(
-    `(implementation|api|compile)\\w*\\s*\\(*project\\(['"]:${normalizedProjectName}['"]\\)`,
+    `(implementation|api|compile)\\w*\\s*\\(*project\\s*\\(['"]:${normalizedProjectName}['"]\\)`,
```
grabbou: How about we extract https://github.com/react-native-community/react-native-cli/pull/221/files#diff-7397ff56d9926b7ce487df7df7efbf8dR37 to a separate function and make `installPattern` based on it? They seem to be the same and I am afraid of having them defined in two places.

matt-block: Right, better to have a single source of truth. The same goes for the dependency types.

PR #222, opened 2019-03-11
chore: bump metro to 0.53

### LINKED ISSUE #210: Bump metro dependencies
Would it be possible to bump the version of metro dependencies in next release ?

We're currently evaluating upgrade to RN 59 (using RC 3 as of now) for one of our application, and we faced a new error during bundling. After taking a closer look we realized that we were running into a metro bug, related to assets path resolution, that was fixed in Metro 0.52.0 (https://github.com/facebook/metro/commit/ce6824ced730d4aa6c4b63c007ed915e1180c258).
Manually updated cli metro dependencies locally to give it a try, and bundling issue was gone.

Thanks.

belemaire:
Summary:
---------

Bump metro dependencies from `0.51` to `0.53`, mostly to propagate some recent bug fixes.
This version bump does not seem to have any impact on the CLI.

In terms of metro breaking changes, in [0.52.0](https://github.com/facebook/metro/releases/tag/v0.52.0), the `dependencies` command was renamed to `get-dependencies`. This change already seems to be accounted for, in PR https://github.com/react-native-community/react-native-cli/pull/100.
In [0.53.0](https://github.com/facebook/metro/releases/tag/v0.54.0), the only breaking change is related to the use of [custom (non production) `BABEL_ENV`](https://github.com/facebook/metro/pull/364). This is unusual, so it shouldn't impact most users, and workaround is documented in the PR in case some users are facing the problem.

Test Plan:
----------

- flow/eslint/test ✅ 
- Used verdaccio to publish custom new patch version of the CLI, including bump to metro 53 and used react-native@0.59.0-rc.3 (which resolved the CLI to my custom version). Ran `react-native init` followed by `run-ios` and `run-android` with success (no issue with bundler but for a bug identified in current CLI, fixed by https://github.com/react-native-community/react-native-cli/pull/220)

Close #210


thymikee:
As you can see #100 is marked `onhold` for 2.0 release, because it's a breaking change (even though RN stable is not yet using the new CLI). 

If we don't want to make this PR breaking, can you proxy `dependencies` to `get-dependencies`? 

What do you think @grabbou, should we cut 2.0 already or break 0.59 RCs (and maybe some other consumers?) just a bit more?

grabbou:
I think we can just follow `semver` and release a major version (to indicate two breaking changes) and not be afraid of the big number we might get into eventually.

grabbou:
Thanks for sending this over! React Native still uses `metro-react-native-babel-transformer` that is `0.52.0`, but I don't think we need to keep the versions in sync.

(I've checked that package source code and it doesn't have a peer dependency nor dependency on any of the Metro packages that we use here).

CC: @cpojer - shall we keep Metro version here in sync with https://github.com/facebook/react-native/blob/master/package.json#L95-L96 or we are good to just keep updating it? (since you follow master anyway)

cpojer:
Yes we should keep it in sync in both places.

grabbou:
Then we need to figure out an update strategy to avoid issues in the future. Right now, any major/minor update to the CLI will be automatically pulled by all React Native installations that use it (from 0.59 onwards).

Shall we scope React Native CLI version for every React Native release to patches only?

grabbou:
Ideally, Metro would be a React Native `dependency` so it could be dynamically installed based on the version needed.

@cpojer - checking the packages that React Native depends on - they don't seem to have any Metro related dependencies. Do they really depend on other Metro packages?

cpojer:
@grabbou the CLI is responsible for all the interactions with Metro so when we extracted the CLI, we were able to remove the direct dependency between RN and Metro.

For the metro-react-native-babel-transformer, maybe we can set it to `*` and set it to the latest that the CLI supports upon RN releases.

grabbou:
That would make sense and let us ship updates to React Native a bit easier and faster. Is this something we can try doing for 0.59 / 0.60 release?

belemaire:
Thanks for the update.

We're good on our end with RN 0.59 that was just released with metro 0.51 (via current 1.x line of he CLI). As part of our automation, we're manually patching metro with the fix we need from metro 0.52, so that we have a workaround for now (for #210).

@thymikee do you still want to proxy `dependencies` command to metro `get-dependencies` command ? I'm a bit confused here because PR #100 completely removes the `dependencies` command from the CLI and is planned to be merged for 2.0 along with my PR, and it seems like there was no plan to replace it with a proxy (as far as I understand the PR comments).

thymikee:
To be frank, I recently noticed that `react-native dependencies` is actually broken on CLI 1.x. Not sure how many projects use that, but I assume not too many.

By proxying (and actually fixing) `dependencies` to `get-dependencies` we could squeeze in Metro 0.53 to 1.x without a breaking change, which would be nice, isn't it?

belemaire:
But per comments from @grabbou and @cpojer it does not seem possible to bump metro version in CLI 1.x because of the `metro-react-native-babel-transformer` dependency in RN 0.59 that is still set to `0.51.0` and metro dependencies need to match (do they really ? the CLI is currently using version range `^0.51.0` which resolves to `0.51.1` whereas RN used fixed version `0.51.0` so they are not at this point using the exact same versions (same minor only).
So if the version mismatch restriction still applies, there is no way to squeeze metro 0.53 to 1.x anyway.

thymikee:
Oh, I thought they're updated already. Then don't bother and just wait for us to merge it in a couple of days :). Sorry for misunderstanding!

cpojer:
Could we change this PR to update Metro to 0.53.1?

It seems like we can merge this afterwards, right @thymikee?

thymikee:
Yup, we're gonna need to release 2.0 from that point.

cpojer:
I changed it to 0.53.1, merging now :)

Vyazovoy:
Hi, @cpojer ! Is it possible to update Metro to `0.54.1` in `1.x` That Metro version has a fix for custom BABEL_ENV

thymikee:
@Vyazovoy I just merged a PR with that, will be available in RN 0.60.0-rc.1 :)

thymikee:
But I don't think we can put this in 1.x because there are breaking changes

PR #229, opened 2019-03-13
Add `eslint` to react native project template.

michalchudziak:
Summary:
---------
The goal of this PR is to enable `eslint` checks in the projects generated by `react-native init` command. I added `eslint` and `@react-native-community/eslint-config` as dependencies. Also there is a new script in generated `package.json` called `lint`.

PR to update template in `react-native` repo: https://github.com/facebook/react-native/pull/23901

Test Plan:
----------
Follow setup from https://github.com/react-native-community/react-native-cli/blob/master/CONTRIBUTING.md
```bash
react-native init TestProject --template eslint-integration
cd TestProject && yarn lint
```

### DIFF
--- a/packages/cli/src/tools/generator/copyProjectTemplateAndReplace.js
+++ b/packages/cli/src/tools/generator/copyProjectTemplateAndReplace.js
@@ -127,7 +127,7 @@ function translateFilePath(filePath) {
     .replace('_gitignore', '.gitignore')
     .replace('_gitattributes', '.gitattributes')
     .replace('_babelrc', '.babelrc')
-    .replace('_eslintrc', '.eslintrc')
+    .replace('_eslintrc.js', '.eslintrc.js')
     .replace('_flowconfig', '.flowconfig')
     .replace('_buckconfig', '.buckconfig')
     .replace('_watchmanconfig', '.watchmanconfig');
--- a/packages/cli/src/tools/generator/copyProjectTemplateAndReplace.js
+++ b/packages/cli/src/tools/generator/copyProjectTemplateAndReplace.js
@@ -127,6 +127,7 @@ function translateFilePath(filePath) {
     .replace('_gitignore', '.gitignore')
     .replace('_gitattributes', '.gitattributes')
     .replace('_babelrc', '.babelrc')
+    .replace('_eslintrc', '.eslintrc')
     .replace('_flowconfig', '.flowconfig')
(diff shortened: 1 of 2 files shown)


thymikee:
Also this needs to wait for facebook/react-native#23901 being merged and released, otherwise we'll break `init`. Marking as "on hold" for now.

grabbou:
See this: https://github.com/reasonml-community/bs-react-native/blob/master/generator/dependencies.json for an example on how to use it in production.

That said, please send a PR against React Native where this template is located.

michalchudziak:
@grabbou How about the changes that replace the underscore in `_eslintconfig` file with dot? I guess there is no option to handle that in `react-native` repo. Same story for the `lint` script.

michalchudziak:
Also, don't you think that keeping the list of dependencies in one place is better than spreading it between repos? Why would we keep `jest` in CLI and `eslint` outside?

thymikee:
@michalchudziak you only have a half of the story, sorry for that :D @Esemesek is rewriting template handling and soon it will look like this: https://github.com/Esemesek/react-native-new-template

The idea is to have a `template.config.js` where you define stuff like project name, template path, maybe after-init scripts and more to come if necessary. Thanks to that the template itself would be entirely copied over so you could define anything you like in your `package.json` or dotfiles files and not be constrained on CLI inability to foresee the future for you.

We want to have this for 0.60, hopefully we can make it in time :D

michalchudziak:
@thymikee Should I wait with implementing changes to the template until this lands?

thymikee:
I think that makes sense. Feel free to help @Esemesek land this :)

grabbou:
@michalchudziak you are right, we need to handle _eslintconfig files!

packages/cli/src/tools/generator/copyProjectTemplateAndReplace.js
```diff
@@ -127,6 +127,7 @@ function translateFilePath(filePath) {
     .replace('_gitignore', '.gitignore')
     .replace('_gitattributes', '.gitattributes')
     .replace('_babelrc', '.babelrc')
+    .replace('_eslintrc', '.eslintrc')
```
grabbou: This is I guess the only thing we need to add as a part of this PR

thymikee:
This gonna be released as part of 2.x, as it's not really backwards compatible, isn't it?

PR #231, opened 2019-03-13
fix: linking assets on Android

thymikee:
Summary:
---------

Fixes https://github.com/react-native-community/react-native-cli/issues/228 introduced by removing `fs-extra`. Brought it back and left a todo when we move to Node 10.

Test Plan:
----------

Green CI (need to covert this in e2e tests once we set them up)


thymikee:
@tallpants could you verify if this works for you?

Issue #96, opened 2019-01-16
Add verbose logging to all commands


Esemesek:
Add `--verbose` flag to enable `DEBUG` log level.
Logger unification and implementation is tracked in #95

matt-block:
Is anyone working on this? If not, I'd like to give it a try 😄

thymikee:
@matt-block afaik nobody works on this currently. Feel free to grab it! :)

lucasbento:
Can we close this? it was done in #239 I believe.

matt-block:
It's been implemented but many commands of the CLI still don't have any debug level log messages.

Maybe we should track those as a list in a dedicated issue and close this 🤔

lucasbento:
@matt-block I second this :)

thymikee:
Who's willing to make a list of commands that have `logger.debug` calls and ones that don't? :)

matei-radu:
@thymikee I can do that. Should I create a new issue for this?

thymikee:
Nah, let's add this list here and I can keep it updated :)

theweavrs:
@thymikee what's the status of this?

thymikee:
Debug logging is still not present in every command. We need to compile a list and let people work on it

maciejsimka:
OK, so here's the list of commands, with the ones having `logger.debug` calls checked.

- [ ] `start`
- [ ] `bundle`
- [ ] `ram-bundle`
- [x] `link`
- [x] `unlink`
- [ ] `install`
- [ ] `uninstall`
- [x] `upgrade`
- [ ] `info`
- [ ] `config`
- [x] `init`
- [ ] `log-ios`
- [x] `run-ios`
- [ ] `log-android`
- [x] `run-android`

Esemesek:
In addition to @maciejsimka list:
- `log-android` and `log-ios` are using `logkitty` to log device logs. We don't need verbose logging there, because they are verbose by default.
- `info` and `config` are very basic and verbose logging there isn't necessary
- `install` and `uninstall` have verbose logging when `linking` happens - I think we don't need more logging there

Naturalclar:
are there any additional logging required for this issue?
the unchecked options are `bundle` and `ram-bundle`, which both leads to `buildBundle` which has loggings.

Esemesek:
@Naturalclar it seems like this issue should be closed. C:

PR #239, opened 2019-03-15
Add verbose logging flag, debug logging for (un)link commands

matt-block:
Summary:
---------
This PR addresses #96 and:

* Adds the `--verbose` flag to enable `DEBUG` logs
* Adds `setVerbose(boolean)` to `tools/logger.js` so that it can enable/disable `DEBUG` logs
* Adds debug logs for the (un)link commands
* Adds Flow typings to `commands/link/android/unlinkAssets.js`

I haven't added `DEBUG` logs to other parts of the CLI because:
* I'm not that familiar with the other CLI commands
* I'd like some feedback on what I've done so far before going on

Also, would it be better to add debug logs in chunks, say one PR per command? I suspect this PR would become too big otherwise.

Test Plan:
----------
These additions do not change any implementations but I ran the tests anyway, all green.
I also ran it against two existing projects and a freshly created one: no changes in behavior.

Screenshots:
----------
How the new flag is presented when using `--help`:
Screenshot 2019-03-15 at 22 07 26

How `link packageName --verbose` looks like:
Screenshot 2019-03-15 at 22 09 08

How `link packageName --verbose` looks like (module with assets):
Screenshot 2019-03-15 at 22 33 28

How `unlink packageName --verbose` looks like:
Screenshot 2019-03-15 at 22 09 34
Screenshot 2019-03-15 at 22 13 58

### DIFF
--- a/packages/cli/src/commands/link/android/patches/applyPatch.js
+++ b/packages/cli/src/commands/link/android/patches/applyPatch.js
@@ -8,8 +8,13 @@
  */
 
 import fs from 'fs';
+import logger from '../../../../tools/logger';
 
 export default function applyPatch(file, patch) {
+  if (file) {
+    logger.debug(`Patching ${file}`);
+  }
+
   fs.writeFileSync(
     file,
     fs
--- a/packages/cli/src/cliEntry.js
+++ b/packages/cli/src/cliEntry.js
@@ -24,7 +24,8 @@ import pkgJson from '../package.json';
 commander
   .option('--version', 'Print CLI version')
   .option('--projectRoot [string]', 'Path to the root of the project')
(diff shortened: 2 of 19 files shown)


packages/cli/src/commands/link/android/patches/applyPatch.js
```diff
@@ -8,8 +8,13 @@
  */
 
 import fs from 'fs';
+import logger from '../../../../tools/logger';
```
Esemesek: This is getting really ugly. @thymikee Do you think we should introduce `absolute` paths to the project?

thymikee: Sure

PR #240, opened 2019-03-15
Throw an error if an invalid platform is passed to the bundler

empyrical:
Summary:
---------

This pull request adds a new error message to the `react-native bundle` command that will throw an error if an invalid platform is passed to the `--platform` command, instead of failing with cryptic Haste errors.

The error message looks like this:

```
> react-native bundle --entry-file index.js --platform foo --bundle-output test.jsbundle
Invalid platform (foo) selected.
Available platforms are:
ios, android, native
If you are trying to bundle for an out-of-tree platform, it may not be installed.
error Invalid platform selected.
```

Test Plan:
----------

If you provide a platform that is not installed to `react-native bundle`, it will throw this error. If you provide a valid platform, it will continue bundling without issue.

### DIFF
--- a/packages/cli/src/commands/bundle/buildBundle.js
+++ b/packages/cli/src/commands/bundle/buildBundle.js
@@ -32,7 +32,9 @@ async function buildBundle(
     console.log(
       chalk.red(
         [
-          `Invalid platform (${args.platform}) selected.`,
+          `Invalid platform ${
+            args.platform ? `(${args.platform}) ` : ''
+          }selected.`,
           'Available platforms are:',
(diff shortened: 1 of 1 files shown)


empyrical:
No problem! It looks good 😄

grabbou:
Great work! Thanks for submitting it!

I think we should replace "out-of-tree" platform with something a bit better. There was a discussion regarding that in the past.

CC: @cpojer

cpojer:
Haven't figured out a better naming yet, so this is fine for now :)

PR #241, opened 2019-03-16
feat: new `init` command, leave legacy

Esemesek:
Summary:
---------

New `init` command will exist in `@react-native-community/cli` and will deprecate current initialization flow.

### What has been done:
- [x] - Added `init` command to `cli` (the previous implementation was partially existing in `global cli`)
- [x] - Created a more friendly interface to create custom templates for initialization
- [x] - Redesigned template to match custom template interface (https://github.com/Esemesek/react-native-new-template)
- [x] - Added support to pass local repositories to `--version` and `--template` options through `file:`

### New template
Structure:
```
.
├── package.json
├── template
│   ├── App.js
│   ├── __tests__
│   ├── android
│   ├── babel.config.js
│   ├── index.js
│   ├── ios
│   ├── metro.config.js
│   ├── package.json
│   └── yarn.lock
└── template.config.js
```

* Root level `package.json` defines package
* `template.config.js` is a configuration file. Not a lot of properties we can configure for now, but can be extended in the future. (https://github.com/Esemesek/react-native-new-template/blob/master/template.config.js - supported properties)
* Directory with `template` which will be copied and processed during initialization process. In this structure it is called `template`, but it can be configured.

### Initalization

You can try out this PR, by cloning repository, checkouting to `Esemesek/init` branch, building `cli` and then running command:
```sh
node react-native-cli/packages/cli/buld init {projectName} --template react-native-new-template
```

To test intialization process without using custom template, `react-native` package you are using should contain update template. You can either create new version of `react-native` and publish it to local npm repository or use `--version file:{/absolute/path/to/update/react-native}`

### Example usage

Init with custom version:
```
npx react-native init ProjectName --version 0.60.0
```

Init with custom local version:
```
npx react-native init ProjectName --version file:///Users/username/local/version/of/react-native
```

Init with custom template:
```
npx react-native init ProjectName --template custom-template
```

Inite with custom local template:
```
npx react-native init ProjectName --template file:///Users/username/local/custom/template
```

> Note that `--template` needs to be valid absolute URL to react native or custom template 

Initialization could also be done by first installing `react-native` and then running init.
```sh
mkdir SomeDir && cd SomeDir
yarn add react-native
react-native init ProjectName
```

Hopefully we will be able to support yarn flows in the future: https://github.com/yarnpkg/rfcs/pull/113

### TODO:
- [x] - Add better logging
- [x] - Add more unit tests
- [x] - Update template in `react-native` repository (https://github.com/facebook/react-native/pull/24138)
- [ ] - Update documentation

Test Plan:
----------

- [x] - Tested manually
- [x] - Unit tests
- [x] - E2E tests - probably most important one


packages/cli/src/cliEntry.js
```diff
@@ -167,9 +167,10 @@ async function setupAndRun() {
             }),
           );
         } catch (_ignored) {
-          throw new Error(
-            'Unable to find React Native files. Make sure "react-native" module is installed in your project dependencies.',
-          );
+          // Do not check for it in `init` command
```
Esemesek: We don't want to check for the existence of `react-native` during `initialization`. Need to refactor that logic.

packages/cli/src/commands/init/protocols.js
```diff
@@ -0,0 +1,37 @@
+// @flow
+import path from 'path';
+
+const FILE_PROTOCOL = /file:/;
```
thymikee: so we use `file:`, shouldn't it be `file://`?

packages/cli/src/commands/init/template.js
```diff
@@ -0,0 +1,38 @@
...
+export function fetchTemplate(templateName: string) {
+  new PackageManager({}).install([templateName]);
+}
+
+export function getTemplateConfig(templateName: string): TemplateConfig {
+  return require(path.join(
+    process.cwd(),
+    'node_modules',
+    templateName,
+    'template.config',
```
thymikee: process.cwd is not necessary, use:
```js
  require(path.resolve(
    'node_modules',
    templateName,
    'template.config.js',
  ))
```

packages/cli/src/tools/PackageManager.js
```diff
@@ -21,11 +21,7 @@ export default class PackageManager {
   }
 
   shouldCallYarn() {
-    return (
-      !this.options.forceNpm &&
-      yarn.getYarnVersionIfAvailable() &&
-      yarn.isGlobalCliUsingYarn(this.options.projectDir)
```
thymikee: why this was removed? `projectDir` is used by other commands and, while with `init` we have not reliable way to say which package manager to use by default, other commands have this data (`yarn.lock` exists or not).

packages/cli/src/commands/init/template.js
```diff
@@ -0,0 +1,38 @@
...
+import PackageManager from '../../tools/PackageManager';
+
+export type TemplateConfig = {
+  placeholderName: string,
+  templateDir: string,
+  postInitScript?: string,
+};
+
+export function fetchTemplate(templateName: string) {
+  new PackageManager({}).install([templateName]);
```
thymikee: consider passing `{silent: true}` to `install`

packages/cli/src/commands/init/protocols.js
Esemesek: I should probably use https://nodejs.org/api/url.html.

packages/cli/src/commands/init/editTemplate.js
```diff
@@ -0,0 +1,68 @@
...
+    .replace(new RegExp(templateName, 'g'), projectName)
+    .replace(
+      new RegExp(templateName.toLowerCase(), 'g'),
+      projectName.toLowerCase(),
+    );
+
+  fs.writeFileSync(filePath, content, 'utf8');
+}
+
+const BINARY_EXT = ['.png', '.jar'];
```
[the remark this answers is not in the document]
Esemesek: It was carried over from the previous implementation, but I agree and will change this.

packages/cli/src/commands/init/__tests__/protocols.test.js
```diff
@@ -0,0 +1,31 @@
...
+import {supportProtocols} from '../protocols';
+
+const VERSION = '0.58.0';
+const RN_WITH_VERSION = 'react-native@0.58.0';
+const ABS_RN_PATH = '/path/to/react-native';
+const REL_RN_PATH = '../path/to/react-native';
+const CWD = '/User/name/projects';
+
+test('should support file protocol with absolute path', () => {
+  expect(supportProtocols(`file:${ABS_RN_PATH}`)).toEqual(ABS_RN_PATH);
```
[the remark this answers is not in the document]
Esemesek: That's why this PR is in draft C:

packages/cli/src/commands/init/printRunInstructions.js
```diff
@@ -0,0 +1,40 @@
...
+    projectName,
+  )}.xcodeproj`;
+  const relativeXcodeProjectPath = path.relative(
+    process.cwd(),
+    xcodeProjectPath,
+  );
+
+  logger.log(`
+  ${chalk.cyan(`Run instructions for ${chalk.bold('iOS')}`)}:
+    • cd ${absoluteProjectDir} && react-native run-ios
```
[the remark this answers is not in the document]
Esemesek: It was also part of previous implementation. Looked briefly into CRA, and they also tell to `cd` into directory.
Screenshot 2019-03-16 at 15 32 38

packages/cli/src/tools/PackageManager.js
Esemesek: That's true. Will change that.

packages/cli/src/commands/init/protocols.js
Esemesek: Since API is supported from `v10` of node, I will use https://www.npmjs.com/package/whatwg-url

packages/cli/src/commands/init/editTemplate.js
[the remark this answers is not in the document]
Esemesek: Turns out that `isDirectory` check is enough.

packages/cli/src/commands/init/init.js
```diff
@@ -79,13 +80,19 @@ export default function initialize(
      * Commander is stripping `version` from options automatically.
      * We have to use `minimist` to take that directly from `process.argv`
      */
-    const version: string = minimist(process.argv).version || 'latest';
+    const version: string = minimist(process.argv).version;
+
+    if (version && !semver.satisfies(version, '0.60.0')) {
```
thymikee: this should also check if default template from `react-native/templates` was used

Esemesek: Maybe we should just move this check to `createFromReactNativeTemplate`?

thymikee: sure

packages/cli/src/commands/init/init.js
```diff
@@ -1,101 +1,102 @@
...
+  const {packageDir, packageName} = supportProtocols(templateName);
+
+  installTemplatePackage(packageDir);
+  const templateConfig = getTemplateConfig(packageName);
+  copyTemplate(packageName, templateConfig.templateDir);
+  changePlaceholderInTemplate(packageName, templateConfig.placeholderName);
 
-  const newProjectName = args[0];
-  const options = minimist(args);
+  new PackageManager({}).installAll();
```
Esemesek: We should pass options to make `forceNpm` option work.

packages/cli/src/tools/PackageManager.js
Esemesek: Depending on https://github.com/react-native-community/react-native-cli/pull/243

thymikee:
cc @orta

packages/cli/src/commands/init/__tests__/editTemplate.test.js
```diff
@@ -0,0 +1,74 @@
...
+    'utf8',
+  );
+  const newCFile = fs.readFileSync(
+    path.resolve(testPath, 'ios', PROJECT_NAME, 'AppDelegate.m'),
+    'utf8',
+  );
+
+  expect(snapshotDiff(oldCFile, newCFile)).toMatchSnapshot();
+  expect(snapshotDiff(oldJavaFile, newJavaFile)).toMatchSnapshot();
+  expect(snapshotDiff(fixtureTree, transformedTree)).toMatchSnapshot();
```
thymikee: can we use `{contextLines: 1}` for options? We don't really need too much context in these snapshots :)

Esemesek: Sure.

thymikee:
FYI looks like this is not gonna be shipped in 0.60 buy we can make it in 0.61, CLI v3.x

grabbou:
The 0.60 is delayed to wait for auto-linking feature, so I think we can ship both next week. I am going to give it a proper review next week on Monday.

Meanwhile, @Esemesek, would you set up a demo internally to present how this works in practice?

patrickkempff:
First of all, thank you all for the awesome improvements! Really looking forward to this!

I have a couple of questions about the new implementation:

Are custom templates based on the default template or should one provide a 'full' template? If so, how can template creators determine the compatible react version when using --version? Do we need to keep a list or are these versions provided by the rn cli or are the react & react-test-renderer & react-native version added to package.json by the cli?

Is there a way to expose the cli arguments to template.config.js/the post install script? We would love to provide different configurations for our template. For example --linter=standard, --linter=eslint or --kotlin / --swift.

Esemesek:
@patrickkempff 
1) You should provide a full template, meaning with ios, android directories, and package.json, yarn.lock/package-lock.json. All the versions should be specified in the package.json.

2) For now, there is no way to provide a version to custom templates - version is determined by the template. Right now I am working to change that, and if a user will pass a version that is incompatible with the template, we will print a warning.

3) I was also considering passing arguments to post install script. Might implement it in this PR - shouldn't be that complicated. We plan to support kotlin/swift in react-native template in the future.

cpojer:
Man this is so exciting. I definitely think this should go into 0.60 if we can make it happen. What is the remaining work necessary to ship this? How can I help?

grabbou:
Its waiting for my review - going to do it ASAP today so we can merge and
test!

packages/cli/src/cliEntry.js
```diff
@@ -167,22 +167,27 @@ async function setupAndRun() {
...
-        } catch (_ignored) {
-          throw new Error(
-            'Unable to find React Native files. Make sure "react-native" module is installed in your project dependencies.',
-          );
-        }
-      })();
+  let reactNativePath;
+
+  // Do not check for it in `init` command
+  if (!process.argv.includes('init')) {
```
grabbou: Not needed anymore after I ship my PR, so we can remove it later.

packages/cli/src/commands/init/errors/DirectoryAlreadyExistsError.js
```diff
@@ -0,0 +1,12 @@
+// @flow
+import path from 'path';
+
+export default class DirectoryAlreadyExistsError extends Error {
+  constructor(directory: string) {
```
grabbou: Shall we use this approach throughout other parts of the codebase?

packages/cli/src/commands/init/printRunInstructions.js
```diff
@@ -0,0 +1,40 @@
...
+    - or -
+    • Open ${relativeXcodeProjectPath} in Xcode
+    • Hit the Run button
+
+  ${chalk.green(`Run instructions for ${chalk.bold('Android')}`)}:
+    • Have an Android emulator running (quickest way to get started), or a device connected.
```
grabbou: Shall we use `dedent` here to make sure white spaces are not preserved?

.gitignore
```diff
@@ -6,3 +6,4 @@ package-lock.json
 *.log
 build/
 .eslintcache
+!packages/cli/src/commands/init/__fixtures__/editTemplate/node_modules
```
grabbou: Why?

packages/cli/src/commands/init/init.js
```diff
@@ -1,100 +1,115 @@
...
+import printRunInstructions from './printRunInstructions';
 import logger from '../../tools/logger';
+import {
+  installTemplatePackage,
+  getTemplateConfig,
+  copyTemplate,
+  executePostInitScript,
+} from './template';
+import {changePlaceholderInTemplate} from './editTemplate';
+import * as PackageManager from '../../tools/PackageManager';
```
grabbou: Do we use `PackageManager` anywhere outside of this command? Maybe we can move it here. Just a "misc", but I am trying to remove most of the "tools" as they are not useful except for the really particular use-case a command might be serving.

grabbou: Also, should be probably lower-case since it's not a "class"?

grabbou: Let's do it in a separate PR

packages/cli/src/commands/init/errors/DirectoryAlreadyExistsError.js
Esemesek: I like having one error per file. It is just my preference carried over from `Java`. Also having:
```jsx
throw new DirectoryAlreadyExistError();
```
seems more readable than
```jsx
throw new Error('This is the description why this directory should not be present');
```

packages/cli/src/commands/init/init.js
thymikee: We use it in a bunch of other placed, e.g. install/uninstall commands (feel free to grep for this)

thymikee:
@grabbou CRNA is archived: https://github.com/react-community/create-react-native-app and they suggest using Expo. We can get rid of eject straight away ;). Wanna send a PR?

Esemesek:
@thymikee I'll do follow up PR with eject removed.

thymikee:
Or we should break `react-native-cli` in RN 0.60 and let use `npx react-native init` for now? @cpojer @hramos can you weigh in?

cpojer:
@thymikee let's not break `react-native init` because everyone is used to it, or at least let's not break it as quickly.

e2e/__tests__/init.test.js
```diff
@@ -0,0 +1,61 @@
...
+  ]);
+  expect(stdout).toContain('Initializing new project from extrenal template');
+  expect(stdout).toContain('Run instructions');
+
+  // make sure we don't leave garbage
+  expect(fs.readdirSync(DIR)).toEqual(['TestInit']);
+  expect(fs.readdirSync(path.join(DIR, 'TestInit'))).toEqual(templateFiles);
+
+  const pkgJson = require(path.join(DIR, 'TestInit', 'package.json'));
+  expect(pkgJson).toMatchSnapshot(
```
thymikee: `package.json` will be updated with every release, so went with using snapshot for that.

patrickkempff:
This is great! I have a question:

How does the template argument resolve the actual template package? Is it still be appending the value --template to `react-native-template-`? Or should I pass the full npm package name, react-native-template-typescript? Personally I like the first one better because it reduces typing and looks cleaner, but it is less explicit.

I looked at the code but it wasn't really clear to me, but i am on mobile.

thymikee:
@patrickkempff that's a good point, I'll add a test for that

packages/cli/src/commands/init/init.js
```diff
@@ -1,100 +1,130 @@
...
   }
+  return name;
+}
+
+async function createFromExternalTemplate(
+  projectName: string,
+  templateName: string,
+  npm?: boolean,
+) {
+  logger.info('Initializing new project from extrenal template');
```
cpojer: Typo: `extrenal`.

packages/cli/src/commands/init/errors/DirectoryAlreadyExistsError.js
```diff
@@ -0,0 +1,12 @@
...
+    super(
+      `Cannot initialize new project because directory ".${
+        path.sep
```
cpojer: It looks weird to have this as a local reference including `./`. I would just remove that tbh.

packages/cli/src/commands/init/templateName.js
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * `init` may be invoked with a shorthand like `--template typescript`
+ * which should resolve to `react-native-template-typescript` package.
+ * To support that, we query npm registry if a package like this exists, if not
+ * we return the original name without a change.
+ */
+export async function tryTemplateShorthand(maybeTemplateShorthand: string) {
+  try {
+    const reactNativeTemplatePackage = `react-native-template-${maybeTemplateShorthand}`;
+    const response = await fetch(
```
cpojer: I don't think it's good to make an HTTP request to the registry with the local string here. Can we first check if it is a local path (has path.separator in it) or whether it has `file://` prefixed?

cpojer:
I'm trying to give this a spin and create a new project from a local checkout of react native (master) and it does not seem to work for me.

At first I tried: `node react-native-cli/packages/cli/build init NewTestApp --template file:///Users/cpojer/Projects/react-native/template --version file:///Users/cpojer/Projects/react-native` but that always failed. It seems to me that the `--template` option expects a path to React Native, not the template folder, is that right?

Then I tried: `node react-native-cli/packages/cli/build init NewTestApp --template file:///Users/cpojer/Projects/react-native` which works and as far as I can tell did a local checkout of the template in that version of the repo. However, after copying over the package.json and installing react-native proper, it would overwrite the local version of RN with `0.59.3`, the latest version of RN.

Ideally the way it would work is:
* Use a local template
* Put the local version of RN into the dependencies
* Be able to build a local version of the template with a local version of RN

The intention of that PR seems to enable that functionality but it doesn't seem to be quite there yet.

My questions are:
* Is `--template` supposed to be used with an RN fork that has a `template` folder or is it supposed to work with the template folder itself?
* How does `--version` play into this?
* I think `react-native` in the template should be either removed and added by the CLI, at least we shouldn't hardcode the version. It seems like an easy thing to miss during our release process (see https://github.com/facebook/react-native/blob/master/template/package.json#L11 ). Similarly, I think it would be best to take the version of React from the package.json of React Native, just like we do now. I don't really see the value of hardcoding these tbh. Maybe I'm missing something?
* Even if we copy the template or local version of RN into the node_modules folder, if we do not explicitly add it to the package.json, running `yarn` will wipe it away.

Esemesek:
> Is --template supposed to be used with an RN fork that has a template folder or is it supposed to work with the template folder itself?

I've implemented `file` protocol to work with `--version`, but it seems like my change was reverted. We still can use `--template` to init with local versions, and that would be preferred way of doing things. `node react-native-cli/packages/cli/build init NewTestApp --template file:///Users/cpojer/Projects/react-native` is fine.

> How does --version play into this?

`--version` accepts only semantic versions for now. We fetch template from `react-native@version` and the process it. We decided that we don't want to override any of the dependencies inside templates, so template inside react-native should have correct version. There is no mixing `template` and `version`, we assume that react-native template has already correct dependency to itself.

> I think react-native in the template should be either removed and added by the CLI, at least we shouldn't hardcode the version. It seems like an easy thing to miss during our release process (see facebook/react-native:template/package.json@master#L11 ). Similarly, I think it would be best to take the version of React from the package.json of React Native, just like we do now. I don't really see the value of hardcoding these tbh. Maybe I'm missing something?

I agree. We shouldn't hardcode it. Maybe we can change version during the releasing?

> Even if we copy the template or local version of RN into the node_modules folder, if we do not explicitly add it to the package.json, running yarn will wipe it away.

That is why we assume that every template will have correct dependency to react-native, including react-native's template.

PR #242, opened 2019-03-16
test with pvinis repo with the two-dot diff

pvinis:
Summary:
---------

This is in preparation of using the new branch structure of the purge repo, using one branch per version.

GitHub does not support getting a patch file from two-dot diffs by appending `.diff` in the url, so I keep them in https://github.com/pvinis/rn-diff-purge/tree/master/diffs.

Test Plan:
----------

We just test it by upgrading a project.


thymikee:
@pvinis what's the status of adding `diffs/` dir to https://github.com/react-native-community/rn-diff-purge?

pvinis:
planned for today :)

I came back from holidays yesterday evening and I'm just leaving the office.

pvinis:
Done. We now have the old list in `old/master`, that has the tags like `version/x.y.z`, and the new table in `master`, with all the diffs in the `diffs` branch, and [purge-web](https://pvinis.github.io/purge-web) pointing to rnc.

I'll close this issue since everything is in sync now, and we store the diffs ourselves.

Issue #244, opened 2019-03-17
Do we need to add the purge repo as a remote?


pvinis:
## Describe the Feature

Maybe we can skip adding any extra remote repos to do an upgrade.
Let's investigate.

## Possible Implementations

We use the diff files from github, so maybe we don't need the remote anymore.

pvinis:
We can close this then :D

PR #245, opened 2019-03-17
refactor: don't add purge as remote when upgrading

pvinis:
Summary:
---------

I tested without adding any remotes when upgrading, and it seems to be working fine, nothing I saw was affected.

It is based on #242, and it addresses #244.

Test Plan:
----------

Try to upgrade a rn project.


thymikee:
Just tried this locally and looks like we don't need a remote for 3way merging, yay! (confirmed that it worked without a remote with --3way flag and failed without it to make sure falling back actually does something)

Can you rework this PR to only remove the remote, so without changes from #242?

pvinis:
I can rebase.

pvinis:
👍

thymikee:
Thank you!

pvinis:
I'll go fix the conflicts on the other PR.

PR #258, opened 2019-03-21
feat: autolinking for Android with Gradle

Salakar:
Provides the Android/Gradle functionality to go with https://github.com/react-native-community/react-native-cli/pull/256 & https://github.com/react-native-community/react-native-cli/pull/254

## CLI Changes

- Updated `react-native config` to include the android `sourceDir` for react native module packages

----

## Usage

These are the one-off changes required to enable this in the React Native init template (or for existing user projects).

### Modify `android/settings.gradle`

Include the following line in your Android projects `settings.gradle` file:

```groovy
apply from: file("../node_modules/@react-native-community/cli-platform-android/native_modules.gradle"); applyNativeModulesSettingsGradle(settings)
```

### Modify `android/app/build.gradle`

Add the following line at the bottom of your Android projects `android/app/build.gradle` file:

```groovy
apply from: file("../../node_modules/@react-native-community/cli-platform-android/native_modules.gradle"); applyNativeModulesAppBuildGradle(project)
```

### Modify `MainApplication.java`

Import the PackageList class;

```java
import com.facebook.react.PackageList;
```

Replace the `protected List<ReactPackage> getPackages()` method with the following:

```java
    @Override
    protected List<ReactPackage> getPackages() {
      @SuppressWarnings("UnnecessaryLocalVariable")
      List<ReactPackage> packages = new PackageList(this).getPackages();
      // additional non auto detected packages can still be added here:
      // packages.add(new SomeReactNativePackage());
      return packages;
    }
```

### Testing

I've tested the following scenarios locally:

- User has no React Native global CLI installed
  - Will warn but continue without issue
- User has no React Native packages installed yet
  - Continues as normal with no warning
- User has more than one React Native package installed
  - Correctly adds multiple packages without issue
- User has a package installed that does not support autodetection (config == null)
  - Prints a warning but continues gracefully
- User has custom/non-linked React Native packages
  - Can be sucessfully manually added via `packages.add(new SomeReactNativePackage());` as documented above in `MainApplication.java`

To test this in a new `react-native init` project locally:

- Clone the CLI project
- Run `yarn` in the CLI project
- Run `cd packages/cli && yarn && yarn link`  in the CLI project
  - (a little hacky as it's a mono-repo)
- Run `yarn link "@react-native-community/cli"` on your React Native project
- Make the required changes as above to your React Native project

----

### Codegen Output Example

An example of the generated `PackageList.java` class:

```java
package com.facebook.react;

import android.app.Application;
import android.content.Context;
import android.content.res.Resources;

import com.facebook.react.ReactPackage;
import com.facebook.react.shell.MainReactPackage;
import java.util.Arrays;
import java.util.List;

import com.test.BuildConfig;

// react-native-webview
import com.reactnativecommunity.webview.RNCWebViewPackage;

public class PackageList {
  private ReactNativeHost reactNativeHost;
  public PackageList(ReactNativeHost reactNativeHost) {
    this.reactNativeHost = reactNativeHost;
  }

  private ReactNativeHost getReactNativeHost() {
    return this.reactNativeHost;
  }

  private Resources getResources() {
    return this.getApplication().getResources();
  }

  private Application getApplication() {
    return this.reactNativeHost.getApplication();
  }

  private Context getApplicationContext() {
    return this.getApplication().getApplicationContext();
  }

  public List<ReactPackage> getPackages() {
    return Arrays.<ReactPackage>asList(
      new MainReactPackage(),
      new RNCWebViewPackage()
    );
  }
}
```


Salakar:
cc @dulmandakh & @bartolkaruza for review

grabbou:
Before reviewing, I'll implement JS changes you did here over in my PR, so you can rebase the branch. That way, it will merge cleanly. 

Going to attempt a review next week, on Monday.

grabbou:
Regarding: 

> I've tested the following scenarios locally:

Shall we also make sure that we print similar messages on iOS, e.g. when the package is not supported or global CLI wasn't recognised?

Salakar:
> Shall we also make sure that we print similar messages on iOS, e.g. when the package is not supported or global CLI wasn't recognised?

I think so - cc @alloy @orta - what are your thoughts on this?

alloy:
I definitely think we should try to be helpful, but I think we should do this when the user uses `react-native install [pkg]` by running the spec validation (`react-native validate [pkg]`?), as that’s the time we know their intent. Afterwards, during eg `pod install` I personally prefer only relevant output (unless perhaps `—verbose` is passed), i.e. not list all packages that are not included. I’ll need to read the PR more to make sure I fully understand the other warnings.

punksta:
This looks much better than javascript linking!
There are react-native packages with constructor accepts parameters. For example  `new FBSDKPackage(mCallbackManager)` from https://github.com/facebook/react-native-fbsdk#31-android. 

Will there be a way of excluding package from auto-linking? annotations may be used.

```
@Retention(RetentionPolicy.SOURCE)
@Target(ElementType.TYPE)
public @interface NoAutoLink {
   String message();
}

@NoAutoLink(message = "Link module manually, providing ReactApplication class instance")
class LibraryPackage {
    public LibraryPackage(ReactApplication application) {
     
    }
}
```

Salakar:
Hey @punksta, I was just discussing args support with @grabbou - for now, we'd look at only supporting arguments where they are set up already to be linked, for example  https://github.com/Microsoft/react-native-code-push/blob/master/package.json#L45 is one I'm looking at supporting as it works as is currently on the old way of linking. 

Whereas the FBSDK one you linked doesn't link the package with args by default - you have to manually link it, so probably not a scenario we'd support yet.

grabbou:
Ideally, `fbsdk` would provide a custom `packageInstance` property in their `package.json` so that we know it needs extra parameters.

Alternatively, maybe we can guess it and provide some stubs? 

> Will there be a way of excluding package from auto-linking? annotations may be used.

We are keeping JavaScript as a source of truth, so you will be able to modify settings for each package (in case it has broken settings or doesn't link properly, e.g. providing that custom `packageInstance` yourself to tell us to use it to link `fbsdk` or just disable linking on Android/iOS/both entirely). Details to follow in my PR #254

Salakar:
@grabbou I've pushed up changes to allow support for the various arguments that CodePush was using;

These were the specific commits:

**CLI**: https://github.com/react-native-community/react-native-cli/pull/258/commits/505e4d01a721fff404d37c838500bcf759cc44bf
**Add Build Config Arg support**: https://github.com/react-native-community/react-native-cli/pull/258/commits/64f3f5a3db137c5cee4bbe26b18b26a7c55a205e
**Template**: https://github.com/facebook/react-native/pull/24099/commits/a58d6e6f0733bb5f1a46f1d133e042ed2507d084

Let me know your thoughts.

gengjiawen:
Is it possible to add some test for this ?

thymikee:
@Salakar ping, the #254 is done, so you should be good to continue working on this :)

Salakar:
Updated to work with latest config changes + new platform-android package

cc @grabbou @thymikee

packages/platform-android/native_modules.gradle
```diff
@@ -0,0 +1,249 @@
...
+
+class ReactNativeModules {
+  private Logger logger
+  private Project project
+  private DefaultSettings defaultSettings
+  private ExtraPropertiesExtension extension
+  private ArrayList<HashMap<String, String>> reactNativeModules
+
+  private static String LOG_PREFIX = ":ReactNative:"
+  private static String REACT_NATIVE_CONFIG_CMD = "react-native config"
```
thymikee: This will execute config through global `react-native-cli`. Since we're aiming to get rid of the global package (or at least make users forget about it), please change it to a "require" from Node, like in CocoaPods implementation:
https://github.com/react-native-community/cli/blob/7d9f8cd96f8e5abc5b7c517d02fb166f9e51b877/packages/platform-ios/native_modules.rb#L3-L4

Salakar: Sure, give me a few minutes.

packages/platform-android/native_modules.gradle
```diff
@@ -0,0 +1,249 @@
+import groovy.json.JsonSlurper
```
thymikee: please add this file `"files"` array in `package.json` so it's published: https://github.com/react-native-community/cli/blob/7d9f8cd96f8e5abc5b7c517d02fb166f9e51b877/packages/platform-android/package.json#L12-L14

thymikee: We forgot about it for iOS 😄 https://github.com/react-native-community/cli/pull/331

Salakar: Done :D

packages/platform-android/native_modules.gradle
Salakar: Done 👍

Salakar:
@thymikee changes requested have been made, thanks for the review 🎉 

For testing: the only way I can see testing being added is if we set up an Android project and run a build and validate that it:

 1) builds the apk without error 
 2) the generated output at `generated/rncli/src/main/java/com/facebook/react/PackageList.java` contains the RN modules that should be available as part of the test (e.g. `yarn add react-native-webview`)

I think this should be based off the RN template though, init a fresh project as part of the test, add a couple of RN modules via npm/yarn, build android and validate output. 

Current problem is the template is not yet live with the tests so not sure how best to do this until then? Thoughts? Or shall we add test suites once the template is live on RN master? This also needs Windows testing - have tested locally though for Mac + Windows.

packages/platform-android/native_modules.gradle
```diff
@@ -54,7 +54,8 @@ class ReactNativeModules {
   private ArrayList<HashMap<String, String>> reactNativeModules
 
   private static String LOG_PREFIX = ":ReactNative:"
-  private static String REACT_NATIVE_CONFIG_CMD = "react-native config"
+  private static String REACT_NATIVE_CLI_BIN = "node_modules${File.separator}@react-native-community${File.separator}cli${File.separator}build${File.separator}index.js"
```
thymikee: RN can technically be in monorepo so we can't make assumptions about where node_modules with cli are. That's why in iOS implementation we spawn a Node process and use its `require.resolve` to output the resolved path to stdout

packages/platform-android/native_modules.gradle
```diff
@@ -0,0 +1,250 @@
...
+    try {
+      cmdProcess = Runtime.getRuntime().exec(REACT_NATIVE_CONFIG_CMD, null, getReactNativeProjectRoot())
+      cmdProcess.waitFor()
+    } catch (Exception exception) {
+      this.logger.warn("${LOG_PREFIX}${exception.message}")
+      this.logger.warn("${LOG_PREFIX}Automatic import of native modules failed. (UNKNOWN)")
+      return reactNativeModules
+    }
+
+    def reactNativeConfigOutput = cmdProcess.in.text
```
thymikee: does this take output from `stdout` only or `stderr` is there as well? `config` outputs warnings in `stderr` so we need to either only take stdout or filter out the warnings like iOS implementation does (because `execute_command` cannot into stdout only)

packages/platform-android/native_modules.gradle
Salakar: This is handled by the final arg here: https://github.com/react-native-community/cli/blob/7a7115f6e8dd10f9eb33eb9d1e8b4c085e08a67f/packages/platform-android/native_modules.gradle#L178 - the process is started in the project root, which can also be overridden here: https://github.com/react-native-community/cli/blob/7a7115f6e8dd10f9eb33eb9d1e8b4c085e08a67f/packages/platform-android/native_modules.gradle#L114

thymikee:
I'm cool with waiting for RN to publish the new template and using that as a source. Or you could publish a custom template (based on the PR you made to RN) and create a custom test in `e2e/` directory. See how the new template looks like: https://github.com/Esemesek/react-native-new-template/ (basically a `template.config.js` with a path to the template that will be straight copied over to your project)

packages/platform-android/native_modules.gradle
Salakar: `stdout` only - `.in` is short hand for `getInputStream()`, `getErrorStream()` is where `stderr` is piped to

thymikee: That's what I assumed, but wanted to be sure. Thanks! Too bad we don't have cross-platform API for that in Ruby and CocoaPods :< (or at least I don't know about them and quick google didn't reveal there's such; cc @orta)

packages/platform-android/native_modules.gradle
thymikee: Ok, so it won't be seamless but at least that's supported. We need to add documentation on how the autolinking is working on how to configure it.

grabbou: @Salakar, we already have `config.root` in our config outputted from `react-native config`. You could use that one.

Also, you can just do `console.log(require.resolve('react-native'))` script and exec `node` to get the exact location. I think this is what we use on iOS.

packages/platform-android/native_modules.gradle
```diff
@@ -0,0 +1,250 @@
...
+      packageImports = "import ${applicationId}.BuildConfig;\n\n"
+      packageImports = packageImports + packages.collect {
+        "// ${it.name}\n${it.packageImportPath}"
+      }.join(';\n')
+      packageClassInstances = ",\n      " + packages.collect { it.packageInstance }.join(',')
+    }
+
+    String generatedFileContents = generatedFileContentsTemplate
+      .replace("{{ packageImports }}", packageImports)
+      .replace("{{ packageClassInstances }}", packageClassInstances)
```
grabbou: I am wondering if there's any way we can wrap an instance of `ReactPackage` here to `LazyReactPackage`. By default, `packageClassInstance` is e.g. `new CodePushPackage()` and in order to convert it to a LazyPackage, we would have to update that package itself.

Is there a way to wrap it or create our custom lazy package for each ReactPackage, that will lazily create it and return everything what's needed?

Salakar: I think `CodePushPackage` class itself would need to extend `LazyReactPackage` instead of `ReactPackage` for this to work, not something we can do? 🤷‍♂️ 

Forcing every package to be lazy will also cause issues for packages that are not inherently meant to be lazy, e.g. modules that have app initialization logic that must always be run?

packages/platform-android/native_modules.gradle
alloy: There’s definitely APIs for that and `execute_command` uses it under the hood, but it doesn’t currently expose a way to keep that output separate, which it probably _should_, but that would be an iteration.

thymikee:
let's give it a shot!

PR #259, opened 2019-03-22
fix: pass `--projectRoot` argument to Metro

oblador:
Despite https://github.com/react-native-community/react-native-cli/pull/234 the `--projectRoot` argument is ignored when running the `react-native start` command as reported in https://github.com/react-native-community/react-native-cli/issues/246. The root of the issue is probably that the value is called multiple things like `root` and `projectRoot`. This PR fixes the immediate issue, but seems to be ripe for a refactor as tracked in https://github.com/react-native-community/react-native-cli/issues/246.


grabbou:
Looks good!

thymikee:
Thank you @oblador 	:)

thymikee:
FYI released 1.5.0 with this fix :)

PR #264, opened 2019-03-23
chore: setup e2e tests

### LINKED ISSUE #134: Implement E2E tests
We should start implementing E2E tests (to test init and e.g. running an actual "start" command) to make sure many simple issues do not happen in production.

@thymikee said he would prepare a good infrastructure for us so that we can start working on them.

thymikee:
Summary:
---------

This diff introduces a working setup for e2e tests, happily ported from Jest codebase :). As an example I decided to add tests (and documentation, because why not) for `install` and `uninstall` commands. We'll add more testing in a followup.

Fixes #134 	

Test Plan:
----------

Added some, doh!

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -63,6 +63,3 @@ unclear-type
 unsafe-getters-setters
 untyped-import
 untyped-type-import
-
-[version]
-^0.94.0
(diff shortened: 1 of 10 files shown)


grabbou:
Thanks! Really excited to have better E2E tests now!

PR #265, opened 2019-03-23
feat: upgrade envinfo for better output of info command

thymikee:
Summary:
---------

Upgrading `evninfo` to v7. Looks like [in 6.0 they removed `clipboard` option](https://github.com/tabrindle/envinfo/releases/tag/6.0.0) because it was unreliable, so I removed it as well.

The biggest change I can see on my machine is actually printing Android SDKs, which is nice.

Apart from that I added `@react-native-community/cli` to the output, because it's pretty relevant now.

Not marking it as breaking change as I don't believe the `--clipboard` was used anywhere.

Test Plan:
----------

Added a test which needs to go to e2e, so I left some todos

### DIFF
--- a/packages/cli/src/commands/info/info.js
+++ b/packages/cli/src/commands/info/info.js
@@ -14,7 +14,7 @@ import type {ContextT} from '../../tools/types.flow';
 const info = async function getInfo(
   argv: Array<string>,
   ctx: ContextT,
-  options: {packages?: boolean | string},
+  options: {},
 ) {
   try {
     logger.info('Fetching system and libraries information...');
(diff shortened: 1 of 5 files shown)


packages/cli/src/commands/info/info.js
```diff
@@ -4,40 +4,36 @@
...
+    logger.info('Fetching system and libraries information...');
+    const output = await envinfo.run({
+      System: ['OS', 'CPU', 'Memory', 'Shell'],
+      Binaries: ['Node', 'Yarn', 'npm', 'Watchman'],
+      IDEs: ['Xcode', 'Android Studio'],
+      SDKs: ['iOS SDK', 'Android SDK'],
+      npmPackages:
+        (typeof options.packages === 'string' &&
+          !options.packages.includes('*')) ||
+        !options.packages
```
thymikee: I'm actually thinking about removing the `--packages` options, because why would anybody use that? `react-native info` is only helpful in bug reports and we don't use the option there.

Esemesek: I agree. This `--packages` option seems pretty useless.

thymikee: yolo, removed

termux-io:
https://github.com/?q=https%3A%2F%2Fgithub.com%2Ftermux-io%2FComputer-Backup%2Ftree%2F%2524

Issue #189, opened 2019-02-25
Show new version + upgrade help text in CLI


cpojer:
The CLI is a good place to notify people of new releases and give them more information about upgrades.

My proposal:
* Periodically check for new RN releases when invoking the RN CLI and surface that people can upgrade with `react-native upgrade`.
* When running `react-native upgrade`, show the breaking changes in red with a link to a blog post and the changelog:

```
React Native 0.60

Breaking Changes:
* Description of a breaking change and a link. 
* Description of a breaking change and a link.
* Description of a breaking change and a link.
For more information read:
* Blog post with link
* See the full changelog
```

I'm not entirely sure what the best way to implement is. I was thinking that release branches could have a file `Upgrade.txt` and the CLI could read that file straight from GitHub and display its contents.

What do you think?

grabbou:
> I'm not entirely sure what the best way to implement is. I was thinking that release branches could have a file `Upgrade.txt` and the CLI could read that file straight from GitHub and display its contents.

Since we need to check for a release anyway with a network request, we could read a section from the releases tab, e.g: https://github.com/facebook/react-native/releases/tag/v0.58.5

Additionally to what you've said, I'd leave an option to mute this banner on per-project for those who are intentionally using CLI with older versions.

cpojer:
Parsing the info from the release and tagging the CLI output in a special block seems like a great idea to me.

grabbou:
Sounds good. I'd love to pick this up after we cut 0.60-stable branch (around a week from now) and see what are the next steps. What do you think?

cpojer:
Yeah!

cpojer:
(You meant 0.59 stable I assume :) )

grabbou:
I'll try to see if I can make it land before we make 0.59 stable. If not, then it will be a 0.60.0 feature. Will do my best!

matt-block:
May I interest you in some GitHub APIs? Try this in your browser:

```
https://api.github.com/repos/facebook/react-native/releases/latest
```
Docs: https://developer.github.com/v3/repos/releases/#get-the-latest-release

Here we have all we need:
* `tag_name` to check if the release is newer
* `draft` and `prerelease` to filter out newer but undesired releases
* `body` which is the entire release markdown as a string that shouldn't be too hard to display.


### What about API rate limits?
> For unauthenticated requests, the rate limit allows for up to 60 requests per hour. Unauthenticated requests are associated with the originating IP address, and not the user making requests.

[...]

PR #268, opened 2019-03-25
feat: notify of new major RN version at most once a week

matei-radu:
Summary:
---------

This PR addresses the first part of #189: notifying the user when a newer version of React Native is available to upgrade to when using the CLI.

**Behaviour**
* The remote release check mechanism uses the [GitHub API](https://developer.github.com/v3/repos/releases/#get-the-latest-release).
* The mechanism is triggered for any CLI command.
* Nothing will be displayed if the project is already up-to-date.
* The release data returned from the API call is cached per-project. Caching is necessary to not hit the GitHub API rate limit.
* For the time being, there is no way to "mute" this behaviour if a project is intentionally using an older RN version because we have not discussed yet how the UX for that will be.

**Caveats**
The GitHub API we are using is specific to [Releases](https://help.github.com/en/articles/creating-releases), so we need to make sure that each new version of React Native has one. For example, 0.59.1 has been cut and published on NPM on March 14th but the release has been made only on the 25th. During that interval, the CLI would have not notified users of the newer version.

Test Plan:
----------

* Clone this branch
* Install dependencies with `yarn install`
* Test the behaviour in newer and older projects by calling `cli.js` directly from the cloned repository as described in `CONTRIBUTING.md`.

Screenshots:
---------
Linking a module in an older project:
Screenshot from 2019-03-25 19-22-14

Output when a project is up-to-date (showing debug messages as normally there would be no output):
Screenshot from 2019-03-25 19-35-42


packages/cli/src/cliEntry.js
```diff
@@ -210,6 +212,16 @@ async function setupAndRun() {
...
   logger.setVerbose(commander.verbose);
+
+  const {version: currentVersion} = require(path.join(
+    ctx.root,
+    'node_modules/react-native/package.json',
+  ));
+  const latestRelease = await getLatestRelease(currentVersion);
+
+  if (latestRelease) {
+    printNewRelease(latestRelease, currentVersion);
```
thymikee: this will print at any time, before, during and after running commands. Not cool. It will also print the warning with every command multiple times during the day (depending on how much you use the cli). I think we should show the prompt _at most_ once a day, ideally once per week. This should be a gentle reminder. It's very easy to annoy users with such messages, so we need to be cautious

packages/cli/src/tools/releaseChecker/getLatestRelease.js
```diff
@@ -0,0 +1,157 @@
...
+  eTag: string,
+  release: ?Release,
+  statusCode: number,
+};
+
+function getLatestRelease(eTag: ?string) {
+  return new Promise<Response>((resolve, reject) => {
+    const options = {
+      hostname: 'api.github.com',
+      path: '/repos/facebook/react-native/releases/latest',
```
thymikee: We cannot check latest from RN github for the same reasons you mentioned in OP. We can either:
1) check from npm (not ideal, because rn-diff-purge is usually updated with a delay)
2) check from rn-diff-purge – this way we can be sure that running `yarn react-native upgrade` will actually have something to fetch

packages/cli/src/tools/releaseChecker/printNewRelease.js
```diff
@@ -0,0 +1,23 @@
...
+import chalk from 'chalk';
+import type {Release} from './getLatestRelease';
+
+/**
+ * Notifies the user that a newer version of React Native is available.
+ */
+export default function(latestRelease: Release, currentVersion: string) {
+  console.log(
+    boxen(
+      'A newer release of React Native is available!\n\n' +
```
thymikee: I'm not a fan of `boxen`, it reminds me bad times when Bower was prompting me all the time to use new version. I can sometimes see npm doing so, but I use it so rarely that I don't really notice it. The memories still remain :D

Maybe we can just prompt a small info about new version + link to changelog? Similar how Yarn does it

cpojer: I tend to agree, actually. I kinda feel like only outdated and old packages use boxen.

patrickkempff:
@matt-block awesome feature! very nice 👍

I agree with @thymikee & @cpojer about the usage of boxer, my main concern is the size of the message. Personally I would prefer a more subtle notice.

matt-block:
Thank you all for the feedback! I'll make some changes over the week.

matt-block:
I've made some changes:

* Removed `boxen` and reverted to using `logger.info`.
* Upgrade notification is less invasive and uses message templates taken from `yarn`.
* Ensured that upgrade notifications are consistent: always shown before running commands.
* Upgrade availability is now checked against the `rn-diff-purge` repo to ensure that an upgrade is truly possible.
  * Note: the changelog URL will still point to the `facebook/react-native` repo.
* Upgrade notifications will be shown at most once a week instead of each time the CLI is used.

Example of notification when using `link`:
Screenshot from 2019-03-30 21-14-52

matthargett:
This is cool for the "major" version of react-native, like if I'm on 0.59.1 and 0.59.3 is available. Similarly, a notice when a react-native-cli is available that is compatible with my version of react-native, is also nice. What will get noisy really fast is if 0.60 comes out and I am unable to migrate for a long period of time (or maybe never) -- this is very common when RN updates xcode dependencies (by necessity due to Apple) and xcode drops device support that is still needed by a user.

So, besides having the default policy not be as aggressive, I think the "mute" use case needs to be implemented before the feature is integrated. If the app's package.json can specify the update notification policy ("minor", "patch", "mute"), with a default to "patch", that would be ideal.

matt-block:
@thymikee any thoughts on this "mute" feature? It could be a property in `package.json` 🤔

Also, sorry I haven't pinged you earlier on this PR but I've been super busy last month.

thymikee:
> any thoughts on this "mute" feature? It could be a property in package.json 🤔

Not really a fan of configuring this. We should be as non-intrusive as possible, so it doesn't bother anybody too much. 

Btw, I like how the upgrade message is displayed as regular `info`. It's visible, but also easy to filter out. Once a week seems good as well. I'll try to review the implementation shortly. In the meantime, mind resolving the conflicts?

matt-block:
@thymikee  rebased onto `master` and solved the conflicts.

packages/cli/src/tools/releaseChecker/getLatestRelease.js
```diff
@@ -0,0 +1,125 @@
...
+  // Should be returned only if something went wrong.
+  return '0.0.0';
+}
+
+type Headers = {
+  'User-Agent': string,
+  [header: string]: string,
+};
+
+function httpsGet(options: any) {
```
thymikee: We already have a helper for that: https://github.com/react-native-community/cli/blob/master/packages/cli/src/tools/fetch.js please use it :)

packages/cli/src/cliEntry.js
```diff
@@ -162,6 +164,27 @@ async function setupAndRun() {
 
   setProjectDir(ctx.root);
 
+  try {
+    // New version check must occur before `commander.parse` to ensure that
+    // the message of a new release happens before anything else.
```
thymikee: We don't need such guarantee. This check is low-priority and shouldn't block anything. Please print the message at the end of `setupAndRun` function (you can trigger the check in parallel earlier of course, just don't block it with `await`)

packages/cli/src/tools/releaseChecker/releaseCacheManager.js
```diff
@@ -0,0 +1,49 @@
...
+ */
+import path from 'path';
+import fs from 'fs';
+import logger from '../logger';
+
+type ReleaseCacheKey = 'eTag' | 'lastChecked' | 'latestVersion';
+
+function loadCache(): ?Object {
+  try {
+    const cacheRaw = fs.readFileSync(path.resolve(__dirname, '.cache'), 'utf8');
```
thymikee: We can't store cache like this in user directory because they'll need to git-ignore it. Please use `os.tmpdir()`

packages/cli/src/tools/releaseChecker/printNewRelease.js
```diff
@@ -0,0 +1,33 @@
...
+ */
+import chalk from 'chalk';
+import logger from '../logger';
+import type {Release} from './getLatestRelease';
+import cacheManager from './releaseCacheManager';
+
+/**
+ * Notifies the user that a newer version of React Native is available.
+ */
+export default function(latestRelease: Release, currentVersion: string) {
```
thymikee:
```suggestion
export default function printNewRelease(latestRelease: Release, currentVersion: string) {
```
this will likely need the `eslint --fix` run as well

packages/cli/src/tools/releaseChecker/getLatestRelease.js
```diff
@@ -0,0 +1,125 @@
...
+    ) {
+      return {
+        version: latestVersion,
+        changelogUrl: buildChangelogUrl(latestVersion),
+      };
+    }
+  } catch (e) {
+    logger.debug(
+      'Something went wrong with remote version checking, moving on',
+    );
```
thymikee: Print the error in debug
```suggestion
    );
    logger.debug(e);
```

packages/cli/src/cliEntry.js
```diff
@@ -162,6 +164,27 @@ async function setupAndRun() {
...
+    const {version: currentVersion} = require(path.join(
+      ctx.root,
+      'node_modules/react-native/package.json',
+    ));
```
thymikee: Please use `require.resolve` or `resolveNodeModuleDir` helper for that as in future it may happen that there's no `node_modules`, something like this should do:
```js
const {version: currentVersion} = require(
  path.join(
    resolveNodeModuleDir(ctx.root, 'react-native'), 
    'package.json'
  )
);
```

packages/cli/src/tools/releaseChecker/releaseCacheManager.js
matt-block: But shouldn't `path.resolve(__dirname, '.cache')` always resolve to `MyProject/node_modules/.../cli/build/tools/releaseChecker/.cache`? If so, the cache should already be ignored.

packages/cli/src/tools/releaseChecker/getLatestRelease.js
matt-block: That implementation is a bit too basic for my use case: it rejects for status codes `300` and above with only an error and nothing else. However, in my case we are also interested in `304`s and result headers, so it wouldn't really work without first augmenting that `fetch` implementation.

packages/cli/src/tools/releaseChecker/releaseCacheManager.js
thymikee: You're right, I somehow missed the `__dirname` there

packages/cli/src/tools/releaseChecker/getLatestRelease.js
thymikee: But, I'd be glad if you had ideas on how to unify both functions in a followup :)

matt-block: I'd be glad to. I should also point out that we are also using `node-fetch` [here](https://github.com/react-native-community/cli/blob/master/packages/tools/src/isPackagerRunning.js), adding up to three different used solutions 🙈.

thymikee: Oh, good catch! We should kill it. I didn't notice that we don't have `node-fetch` in our dependencies so it works because of luck. cc @lucasbento

packages/cli/src/cliEntry.js
matt-block: The problem is that without blocking we cannot guarantee __when__ the message is printed.

Depending on the complexity of the command and time to fetch data from GH, the new version message might be printed before, during or at the end of a command. This might pollute the command's own logging flow and the user might also lose track of the new release message since it could appear whenever.

thymikee: so call it after the `await setupAndRun();`, it's all async and you can be sure all the output is already there in the console then

matt-block:
Initially the `unit-tests` job failed ([Job 4393](https://circleci.com/gh/react-native-community/cli/4393)) but after a `commit --amend` without changing code and a `push --force` the second run passed ([Job 4404](https://circleci.com/gh/react-native-community/cli/4404)).

Is it normal for some tests to _occasionally_ fail without changes?

thymikee:
>Is it normal for some tests to occasionally fail without changes?

It sometimes happen for `config` tests, but we didn't figure out why yet.

thymikee:
btw this still hits the network every time. Can we make it so that it happens ~once a week?

packages/cli/src/cliEntry.js
thymikee: I was wrong about it, sent a commit with a fix: cd5fa9f

packages/cli/src/tools/releaseChecker/releaseCacheManager.js
patrickkempff: Quick question about the cache folder: I am not sure if I remember it correctly but when running `yarn install` (or yarn add / remove), yarn replaces the contents of the dependencies inside node_modules with its cached version. Does this mean the `.cache` is always cleared after an yarn install?

matt-block: Good question! I'm unaware of the behavior of yarn when it comes to already installed packages, but if you are correct then yes, the cache would be cleared. We should investigate how both Yarn and Npm actually behave.

matt-block: We could use `os.tmpdir()` to completely eliminate any doubts about this but we would need to create a cache file per project (maybe inferring the file name from the project name?).

patrickkempff: A quick test showed me that yarn replaces the node_modules when the node_modules is not up to date with the yarn.lock / package.json, this applies to yarn install / add / remove. 

I am not sure if this behaviour is officially documented. At least the `patch-package` mentioned it: https://www.npmjs.com/package/patch-package#why-use-postinstall-postinstall-with-yarn

I am not sure about npm.

patrickkempff: About `os.tmpdir()` on macOS, I think this folder gets cleared at least every reboot on macOs. You can compare the content of `ls $TMPDIR` with the content it after reboot.

Why not create a `.react-native-cli` folder in `os.homedir()` to store the .cache dir in? This way you have more control over how long you want to retain the cache

matt-block:
This is more of a note to myself for this weekend:
* Hit the network only if the cache is old (1 week).
* Move cache to user home dir, although each project will need to have its own unique file.

matt-block:
The `.react-native-cli/cache` dir is created in the home directory by the release cache manager if it doesn't exist already.

In it, a cache file for each project will be created bearing the name of their projects.

Screenshot 2019-05-25 at 10 21 45

matei-radu:
@thymikee

thymikee:
Please bear with me, I'll get to it! Looks promising so far, I'd like to play with it later today :)

packages/cli/src/tools/releaseChecker/printNewRelease.js
```diff
@@ -0,0 +1,26 @@
...
+ */
+export default function printNewRelease(
+  name: string,
+  latestRelease: Release,
+  currentVersion: string,
+) {
+  logger.info(
+    `React Native v${
+      latestRelease.version
+    } is now available (your project is running on v${currentVersion}).`,
```
thymikee: Adjusted the message a bit, looks like this now:

image

Hope you don't mind :)

thymikee: cc @cpojer is this ok, or do you have  better ideas?

matei-radu: Not at all, I like this version more!

thymikee:
Thank you so much @matei-radu! <3

matei-radu:
Thank you guys for the patience 😄

thymikee:
FYI just released `2.0.0-rc.0` with this included

PR #272, opened 2019-03-27
bug: apply upgrade patch in subdirectory if necessary

### LINKED ISSUE #271: Upgrading does not work inside a monorepo
I have a monorepo with this kind of structure :

```
root
  ├ .git
  ├ mobile
  |  ├ android
  |  ├ app
  |  ├ ios
  |  ├ ...
  ├ server
  └ web
```
`react-native upgrade` does nothing when I run it from my app's folder (`mobile`) because it tries to apply the patch at the root of the git repo (I guess). 

Anyways I managed to upgrade manually with `git --apply ... --directory="mobile"` but there have to be other people expecting this to work out of the box, given that it did with the legacy `react-native-git-upgrade`.

maxaggedon:
Summary:
---------

Resolves #271 

Test Plan:
----------

1. Init a react native app on an old version at the root of a repo
2. Upgrade to latest rn version and check if it worked as expected (to avoid regression)
3. Discard upgrade changes and move the app to a subfolder
4. Try to upgrade again

I did test that between 0.59.0 and 0.59.2

### DIFF
--- a/.prettierrc.js
+++ b/.prettierrc.js
@@ -1,0 +1,5 @@
+// added for Jest inline snapshots to not use default Prettier config
+module.exports = {
+  bracketSpacing: false,
+  trailingComma: "all"
+}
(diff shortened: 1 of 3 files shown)


maxaggedon:
Thanks for the merge !

thymikee:
Released 1.7.0 with this diff

PR #275, opened 2019-03-28
Refactor: Remove eject

Esemesek:
Summary:
---------

Since `create-react-native-app` is deprecated this functionality is no longer needed.


thymikee:
Can you also remove this? https://github.com/react-native-community/react-native-cli/blob/43e37dd946a9296231577923c46435703295f015/packages/cli/src/commands/link/link.js#L66-L74

Esemesek:
@thymikee Missed that one :P

ansarikhurshid786:
is there any other way to re-generate android and ios folder?

thymikee:
@ansarikhurshid786 run `npx react-native init YourProjectName` and copy over what's necessary.

zerobytes:
That is unacceptable. @thymikee, so you're saying that in order to to something on our project, we have to create another project only to copy and paste things? really?

There are several reasons why we, sometimes, need to delete android/ios folders, and generate them again. And that we did with eject. Now, removing eject, is not my concern, although there should be a function to cover the ability to generate android and ios folders.

E.g. 
I have to change the name of the project. Easy, just remove the folders, and recreate (such as eject would do). But now you're telling us to do manually what we did automatically because 'NO'

thymikee:
I'd be glad if you could create a new RFC issue with a command that would suit your needs. Eject was created for CRNA which is archived now, and we didn't know about such use cases. Glad to discuss further, but please make the RFC first :)

radicalswetank:
Is there any new way to re-generate the Android and ios folders apart from initializing a new project?

phantom-factotum:
I've been using expo eject

jhoanborges:
I just faced an issue. Seems like xcode was returning some errors because it can find the ios files. So i came to this post looking for ejetct... as the guy said, unacceptable!!!

image


image

PR #276, opened 2019-03-28
Better name selecting for upgrade command

Krizzu:
Summary:
---------

Small change regarding upgrade command.
Before, if `name` value in `package.json` has changed since you init'ed a project, `react-native upgrade` would fail. This is because directory paths in diffing were selected based on that `name` value.

Now, we check if names (from `package.json` and `app.json`) are different, and if so, use the one from `app.json` (to keep `name` value in `package.json` untouched).

Test Plan:
----------

Probably green CI should be sufficient.


Krizzu:
Closing in favor of #299

PR #279, opened 2019-04-01
Sort modules in debugger delta patcher to match source map order

motiz88:
Summary:
---------

Context: https://github.com/facebook/metro/issues/380, https://github.com/facebook/react-native/issues/23955

Fixes a discrepancy between the order of modules in the source map and in the actual bundle, caused by a change in Metro (see details in https://github.com/facebook/metro/issues/380#issuecomment-478633831). Consequently fixes React Native remote JS debugging.

Test Plan:
----------

1. Made the same change in my checkout of the repro code from https://github.com/facebook/metro/issues/380 and verified debugging works.
2. Added unit tests.

### DIFF
--- a/packages/cli/src/commands/server/debugger-ui/DeltaPatcher.js
+++ b/packages/cli/src/commands/server/debugger-ui/DeltaPatcher.js
@@ -119,7 +119,10 @@
     getAllModules() {
       return [].concat(
         [this._lastBundle.pre],
-        Array.from(this._lastBundle.modules.values()),
+        // Sort modules so they match the source map emitted by Metro >= 0.51.0
+        Array.from(this._lastBundle.modules.entries())
+          .sort(([id1], [id2]) => id1 - id2)
+          .map(([id, contents]) => contents),
         [this._lastBundle.post],
       );
     }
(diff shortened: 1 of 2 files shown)


thymikee:
Do you think it's worth to add a comment on why we sort (and when to stop, in case Metro fixes it)?

motiz88:
Yes, I was going to add a comment and it slipped my mind, thanks 😅

cpojer:
Thank you ❤️

thymikee:
Published a patch version 1.5.2 with this fix included.

hypest:
👋 friends, I'm the author of [the original ticket](https://github.com/facebook/react-native/issues/23955) and repro code repo. I wonder if using react-native-cli v1.5.2 (as per @thymikee 's comment above) is a good way to incorporate the fix right now, or should we normally wait to see its way to a future React Native release. I tried v1.5.2 out in the repro repo ([this branch](https://github.com/hypest/rn-059-stacktrace-mess-demo-app/compare/try/cli-1.5.2)) by force-resolving the package to v1.5.2 but it doesn't seem to solve the problem for me.

PR #281, opened 2019-04-02
feat: make legacy `link` use new configuration

grabbou:
Summary:
---------

This PR switches `link` to rely on the new configuration (from "ctx") that is also used by "react-native config" and soon, "auto linking".

Test Plan:
----------

You should be able to run link w/o any issues


thymikee:
Tests and linters are failing, mind fixing that?

grabbou:
It's WIP @thymikee :) I'll make sure it's awesome when it's ready.

packages/cli/src/commands/link/link.js
```diff
@@ -32,31 +32,11 @@ type FlagsType = {
...
-      'Available platforms: ' +
-        `${Object.getOwnPropertyNames(platforms)
-          .map(platform => getPlatformName(platform))
-          .join(', ')}`,
-    );
-    if (opts.platforms) {
-      platforms = pick(platforms, opts.platforms);
-    }
-    logger.debug(
-      'Targeted platforms: ' +
```
thymikee: can you bring back the debug logs?

packages/cli/src/cliEntry.js
```diff
@@ -170,6 +169,9 @@ async function setupAndRun() {
   logger.setVerbose(commander.verbose);
 }
 
+// @todo replace this
+const findPlugins = () => {};
```
thymikee: you'll need to migrate RN itself to not use it for 0.60 anyway: https://github.com/facebook/react-native/blob/a2285b17907cfd3d7cfd311e18027da800a48b2e/jest/hasteImpl.js#L27

so feel free to remove it

grabbou: I was thinking about it but didn't have time to look into how React Native is using it yet. It would be great to get rid of it entirely and maybe just read config (`loadConfig` sounds like a better public API than `findPlugins`)

packages/cli/src/commands/link/link.js
grabbou: Brought them back in a different form, you should be satisfied.

grabbou:
There's one `@todo` left in the `config/index.js`, but it doesn't affect the mergability of this PR.

packages/cli/src/commands/link/linkAll.js
```diff
@@ -1,64 +1,53 @@
...
 const dedupeAssets = (assets: Array<string>): Array<string> =>
   uniqBy(assets, asset => path.basename(asset));
 
-function linkAll(
-  context: ContextT,
-  platforms: PlatformsT,
-  project: ProjectConfigT,
-) {
-  logger.warn(
-    'Running `react-native link` without package name is deprecated and will be removed ' +
```
grabbou: We decided not to deprecate it, since whole `link` will become obsolete at some point.

packages/cli/src/tools/config/index.js
```diff
@@ -106,7 +106,7 @@ function loadConfig(projectRoot: string = process.cwd()): ConfigT {
           (project, platform) => {
             project[platform] = this.platforms[platform].projectConfig(
               projectRoot,
-              userConfig.project[platform],
+              userConfig.project[platform] || {},
```
thymikee: could you fix that better typings?

grabbou: We already have the good typings. Flow can't handle them. userConfig.project is a map, we just need to handle missing keys.

grabbou: iOS/Android is always defined because we use Joi schema validation and apply some defaults to not need to handle with that.

packages/cli/src/commands/link/__tests__/link-test.js
```diff
@@ -1,108 +1,93 @@
-/**
- * Copyright (c) Facebook, Inc. and its affiliates.
- *
- * This source code is licensed under the MIT license found in the
- * LICENSE file in the root directory of this source tree.
- *
- * @format
- * @emails oncall+javascript_foundation
- */
+import {func as link} from '../link';
```
thymikee: I guess it's too much of a hassle to add Flow here?

packages/cli/src/commands/link/link.js
```diff
@@ -32,60 +33,57 @@ type FlagsType = {
...
   ];
 
   return promiseWaterfall(tasks).catch(err => {
-    logger.error(
-      `Something went wrong while linking. Error: ${err.message} \n` +
-        'Please file an issue here: https://github.com/react-native-community/react-native-cli/issues',
-    );
-    throw err;
+    throw new Error(dedent`
+      Something went wrong while linking. Reason: ${err.message}
```
thymikee: can you include the `error.stack` as well?

packages/cli/src/commands/link/linkAll.js
```diff
@@ -1,64 +1,53 @@
...
+  ]);
 
   return promiseWaterfall(tasks).catch(err => {
-    logger.error(
-      `Something went wrong while linking. Error: ${err.message} \n` +
-        'Please file an issue here: https://github.com/react-native-community/react-native-cli/issues',
-    );
-    throw err;
+    throw new Error(dedent`
+      Something went wrong while linking. Reason: ${err.message}
```
thymikee: can you include the `error.stack` instead? This eats where the error comes from

packages/cli/src/commands/link/linkDependency.js
```diff
@@ -51,9 +51,10 @@ const linkDependency = async (
 
     linkConfig.register(
       dependency.name,
-      dependency.config[platform] || {},
+      // $FlowExpectedError: We already checked if dependency.platforms[platform] exists
+      dependency.platforms[platform],
       params,
-      // $FlowFixMe: We check if project[platform] exists on line 42
+      // $FlowFixMe: We already checked if project[platform] exists
```
thymikee: got rid of these fixmes by extracting variables and asserting on that (pushed in 00abd83)

packages/cli/src/tools/config/types.flow.js
```diff
@@ -2,75 +2,222 @@
...
+      },
       params: InquirerPromptT[],
     },
   },
-  platforms: PlatformsT,
+
+  // Map of available platforms (built-ins and dynamically loaded)
+  platforms: {
+    [name: string]: any,
+    ios: {
```
thymikee: simplified this a bit to avoid duplication and copy-paste errors (see 3d65cda)

grabbou: actually used to be like that, I removed it for better readability. But whatever makes sense :)

packages/cli/src/commands/link/__tests__/link-test.js
grabbou: Yeah, it complains about "config" missing some properties. Don't want to define all properties for the sake of the test.

packages/cli/src/commands/link/link.js
```diff
@@ -32,60 +33,57 @@ type FlagsType = {
...
 
-  logger.debug(`Package to link: ${rawPackageName}`);
-
   // Trim the version / tag out of the package name (eg. package@latest)
   const packageName = rawPackageName.replace(/^(.+?)(@.+?)$/gi, '$1');
 
-  const dependencyConfig = getDependencyConfig(ctx, platforms, packageName);
+  if (!Object.keys(ctx.dependencies).includes(packageName)) {
+    throw new Error(dedent`
+      Unknown dependency.
```
thymikee: @grabbou I get this error while linking `appcenter`. Is this because you only try to detect `react-native-` etc modules as deps? We should allow all of them I guess

grabbou: Yes. What is the command you are running? I haven't changed the discovery behaviour. I could look at all `package.json` dependencies, but that would be a change to `link` command too.

grabbou: This is the behaviour. We might consider changing that at some point, but it's out of scope for this PR.

grabbou:
Fixed a little inconsistency between `link` and `link [packageName]` where `link [packageName]` would work for any name whereas `link` would only take `react-native-*` packages to link.

From now on, `link` will check all dependencies, regardless of the name.

packages/cli/src/cliEntry.js
```diff
@@ -46,6 +46,7 @@ const handleError = err => {
         )}`,
       )}`,
     );
+    console.log(err.stack);
```
thymikee: remove

packages/cli/src/tools/config/index.js
```diff
@@ -111,25 +112,20 @@ function loadConfig(projectRoot: string = process.cwd()): ConfigT {
       get assets() {
         return findAssets(projectRoot, userConfig.assets);
       },
-      platforms: {
-        ios,
-        android,
-      },
+      platforms: {},
```
thymikee: oh yea!

packages/cli/src/tools/config/types.flow.js
```diff
@@ -79,9 +79,9 @@ export type ConfigT = {|
 
   // Object that contains configuration for a project (null, when platform not available)
   project: {
-    android: ?ProjectConfigAndroidT,
-    ios: ?ProjectConfigIOST,
-    [key: string]: any,
+    android?: ?ProjectConfigAndroidT,
+    ios?: ?ProjectConfigIOST,
```
thymikee: Shouldn't it be this now?
```jsx
android?: ProjectConfigAndroidT,
ios?: ProjectConfigIOST,
```

grabbou: `android` and `ios` can be missing (no `react-native` in package.json), still they can be `null` when there's `react-native` in a project but no Android or iOS files (say I am working on Android only project).

PR #283, opened 2019-04-02
feat: point to --verbose when command fails

thymikee:
Summary:
---------

Our command error handler eats error stack and leaves the user with only a message. Example: https://github.com/react-native-community/react-native-cli/issues/273

I don't think we need to put stacks in every error out there, but we should enable them in `--verbose` mode.

Test Plan:
----------

Screenshot 2019-04-02 at 12 49 06


grabbou:
So good! I actually just had this thought while we were chatting!

PR #285, opened 2019-04-03
fix: properly copy template binary files

cpojer:
Summary:
---------
`fs.copySync` does not properly copy binary files. I am using the copy method that we already have in the repo.

Test Plan:
----------
Without this change `react-native run-android` breaks with the following error:
Screen Shot 2019-04-02 at 16 58 14


grabbou:
Would be good to move this util out of that folder at some point now that we want to remove generator entirely (and yeoman)

packages/cli/src/tools/copyFiles.js
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+import path from 'path';
+import copyAndReplace from './copyAndReplace';
+import walk from './walk';
+
+/**
+ * Copy files (binary included) recursively.
+ */
+function copyFiles(srcPath: string, destPath: string) {
```
thymikee: Created a new utility to just copy files recursively, without replacing stuff around (which could theoretically break some 3rd templates in the future). It's based on `copyAndReplace` but with empty replacements, so binaries are still copied and I verified that it works with `run-android` (previously was breaking)

cpojer:
@thymikee thanks for fixing this up so nicely!

PR #287, opened 2019-04-03
feat: use config in `run-ios` to resolve project location

### LINKED ISSUE #12: `run-ios` should use config
Right now it has its own detection mechanism for `ios` files. Should use the config that already does that job.

grabbou:
Summary:
---------

Right now, iOS path is hardcoded to `ios`, which makes it impossible to run `RNTester` "out-of-the-box".

This is a slight change that updates the default value of the option.

Note: We should probably also move the "xcworkspace" information to the config instead of it being here, but this can be done in a follow-up.

Fixes #12 
Partially fixes (not close yet): #143 

Test Plan:
----------

Run `react-native run-ios` in `react-native` root. It should just work.


grabbou:
To be rebased and merged after #281 is merged.

thymikee:
mind rebasing to master (it produces awful merge conflicts, so probably easier to open it once again)

PR #290, opened 2019-04-04
feat: extract iOS and Android specific linking files out of the core

grabbou:
Summary:
---------

Historically, Android and iOS platform were hardcoded inside CLI and provided by default. Then, other packages could export additional platforms on top of it.

React Native as a platform was hardcoded in Metro configuration (default platforms, default extensions) and inside link. That means all new features were always developed with these platforms in mind, leaving others behind. We often used private APIs, which made things more complicated for 3rd party platform developers.

From this PR on, there's no longer concept of an out-of-tree platform inside of the CLI. In a follow up PR to React Native, I am going to configure iOS and Android platforms inside React Native, just like React Native Windows exposes Windows platform.

This will open room for improvements and should make it even easier for the CLI to welcome other platforms.


packages/cli/src/tools/config/index.js
```diff
@@ -22,8 +22,7 @@ import assign from '../assign';
 /**
  * Built-in platforms
  */
-import * as ios from '../ios';
-import * as android from '../android';
+import {ios, android} from '../../..';
```
grabbou: @todo let's remove it when we are done, update tests and then, ship a PR to React Native to turn this on.

PR #291, opened 2019-04-04
feat: morgan - skip successful requests if 'verbose' is set to false

vancehu:
Summary:
---------
Metro server can become too verbose if there are a lot of assets due to morgan verbosely logging every request. Trying to make use of `verbose` flag to allow users skipping successful requests if 'verbose' is set to false (default)

Test Plan:
----------
1. run `react-native start`
Expected Result: no console output on a successful request
2. run `react-native start --verbose`
Expected Result: console output for every request


thymikee:
Nice. Can you fix the CI failures?

vancehu:
@thymikee fixed!

packages/cli/src/commands/server/runServer.js
```diff
@@ -62,7 +62,12 @@ async function runServer(argv: Array<string>, ctx: ContextT, args: Args) {
     watchFolders: metroConfig.watchFolders,
   });
 
-  middlewareManager.getConnectInstance().use(morgan('combined'));
+  middlewareManager.getConnectInstance().use(
+    morgan('combined'),
+    !args.verbose && {
```
thymikee: this should be a second argument to `morgan`.
Also, I've spotted and issue with `--verbose` flag being omitted because we already declare it globally. Need to investigate before merging this, so please bare with us :)

vancehu: @thymikee updated. I'll run eslint locally to avoid simple errors like this in the future. thanks

thymikee: Please use `logger.isVerbose()` for now, instead of `args.verbose`

packages/cli/src/commands/server/runServer.js
```diff
@@ -19,6 +19,7 @@ import webSocketProxy from './webSocketProxy';
 import MiddlewareManager from './middleware/MiddlewareManager';
 
 import loadMetroConfig from '../../tools/loadMetroConfig';
+import logger from '../../../../tools/src/logger';
```
thymikee:
```suggestion
import {logger} from '@react-native-community/cli-tools';
```

vancehu: @thymikee did not get a chance fixing this yesterday. thanks for helping

thymikee:
Released in 1.8.0, thank you so much!

PR #292, opened 2019-04-05
feat: add React logo and spinners to make `init` UI nicer

### LINKED ISSUE #91: Improve `init` command
We should look at the `react-native init` behaviour and think about improving the way it works.

Here are some of the things that could be improved:
- Using custom templates. Right now, when you init a new project with a custom template, it gets applied on top of the default one. This makes it a bit confusing when applying e.g. a Typescript template, where you end up with Flow configuration (left-over from the default template).
- We could make it interactive

Please comment with other improvements/feedback and we will convert them into issues at some point.

### LINKED ISSUE #266: ASCII art for init command
Lets make the init command looking really awesome with some nice ascii art! I have spend some time trying to give the init command a cleaner look and posted a screenshot and video of the result here https://github.com/react-native-community/react-native-cli/issues/91#issuecomment-475914736. 

Based on @thymikee comment I have tried some different variations and sizes. I would love to discuss those.

There are different terminals, platforms and configurations out there. We should make sure that the ASCII image is recognizable a cross different platforms. We could learn from the yeoman implementation: https://github.com/yeoman/yosay

**Original:**

https://sendvid.com/embed/a9lrtvjj

54873024-0f7d9180-4dce-11e9-8eb0-ffe3cab33dc1

Variant 2
--------------
variant 2

Variant 3
--------------
variant 3

Variant 4
--------------
variant 5

Variant 5
--------------
variant 5

Variant 6
--------------
variant 6

Variant 7
--------------
variant 7

Variant 8
--------------
variant 8

Variant 9
--------------
variant-9

Although I tried to include multiple sizes, small that variant will render the logo unrecognizable. Variant 8 is the smallest I could get, but I can spend some time trying to create a smaller version.

What do you think? Please be critical!

Esemesek:
Summary:
---------

Adding spinners and react logo during initialization.

Kapture 2019-04-05 at 18 55 57

- [x] - Spinners not included when `--verbose` flag is provided
- [ ] - Spinners are frozen on `copying template` and `processing template` steps, because they are synchronous. Those steps are very fast and it is almost not visible.

Test Plan:
----------

- [x] - Manual testing
- [x] - Improve e2e

Closes #91 
Closes #266


packages/cli/src/commands/init/__tests__/template.test.js
```diff
@@ -68,21 +70,23 @@ test('copyTemplate', () => {
...
 });
 
-test('executePostInitScript', () => {
+test('executePostInitScript', async () => {
   const RESOLVED_PATH = '/some/path/script.js';
   const SCRIPT_PATH = './script.js';
 
+  const mockExec = jest.fn();
+
+  jest.spyOn(Util, 'promisify').mockImplementationOnce(() => mockExec);
```
thymikee: looks very fragile, can you rethink this spy?

packages/cli/src/commands/init/banner.js
```diff
@@ -0,0 +1,47 @@
...
+  '             ######    ###            ###    ######             ',
+  '                ##  ########################  ##                ',
+  '               ##         ###      ###         ##               ',
+  '               ##           ##    ##           ##               ',
+  '               ##             ####             ##               ',
+  '               ##             ####             ##               ',
+  '               ##          ###    ###          ##               ',
+  '                ###     ####        ####     ###                ',
+  '                  ######                ######                  ',
+  '                                                                ',
```
thymikee: we need to add a very random Guy Fieri at some point behind a flag  :trollface:

Esemesek: I'm down XD

packages/cli/src/commands/init/banner.js
```diff
@@ -0,0 +1,47 @@
...
+  '                                                                ',
+  '                                                                ',
+  '                                                                ',
+];
+
+const welcomeMessage = '                   Welcome to React Native!';
+const heyThere =
+  'Hey there! I am going to initialize a fresh react-native project: ';
```
thymikee: Maybe we don't need a "hey message", but rather a "Learn once, write everywhere" tagline? :)

packages/cli/src/commands/init/init.js
```diff
@@ -14,8 +14,10 @@ import {
   executePostInitScript,
 } from './template';
 import {changePlaceholderInTemplate} from './editTemplate';
-import * as PackageManager from '../../tools/PackageManager';
+import * as PackageManager from '../../tools/packageManager';
 import {processTemplateName} from './templateName';
+import banner from './banner';
+import getLoader from '../../tools/loader';
```
thymikee: can we stick to the module name? So `loader` instead of `getLoader`

packages/cli/src/commands/install/install.js
```diff
@@ -9,14 +9,14 @@
 
 import type {ContextT} from '../../tools/types.flow';
 import logger from '../../tools/logger';
-import * as PackageManager from '../../tools/PackageManager';
+import * as PackageManager from '../../tools/packageManager';
```
thymikee: you changed module name from `PackageManager` to `packageManager`, but the variable name stayed without changes. mind fixing?

packages/cli/src/commands/init/template.js
```diff
@@ -44,5 +49,5 @@ export function executePostInitScript(
 
   logger.debug(`Executing post init script located ${scriptPath}`);
 
-  execFileSync(scriptPath, {stdio: 'inherit'});
+  return promisify(execFile)(scriptPath, {stdio: 'inherit'});
```
thymikee: use `execa` instead of `child_process`, it accepts file as a first parameter and has promise interface

packages/cli/src/tools/PackageManager.js
```diff
@@ -1,16 +1,19 @@
 // @flow
-import {execSync} from 'child_process';
+import {exec} from 'child_process';
```
thymikee: use `execa`

packages/cli/src/commands/install/install.js
Esemesek: Variable name left uppercased because it is a grouping of all methods in `packageManager` through `* as` importing.

packages/cli/src/commands/init/template.js
Esemesek: I was thinking the same but didn't want to introduce the next dependency. I'll gladly use `execa` because testing will be much easier.

packages/cli/src/tools/PackageManager.js
Esemesek: Sure sure sure.

packages/cli/src/tools/loader.js
```diff
@@ -0,0 +1,14 @@
...
+import Ora from 'ora';
+import logger from './logger';
+
+class OraMock {
+  succeed() {}
+  start() {}
+}
+
+function getLoader(): typeof Ora {
+  return logger.isVerbose() ? OraMock : Ora;
```
thymikee: what's the idea behind disabling ora in verbose?

packages/cli/src/commands/init/template.js
thymikee: fortunately we already use it!

packages/cli/src/tools/loader.js
Esemesek: If we go for `verbose` logging, additional logs `break` spinners, by adding additional lines with logs. I thought this would be a nice addition for `ci` runs and `bug` reports.

packages/cli/src/commands/init/template.js
Esemesek: OH, SICK!

packages/cli/src/tools/__tests__/PackageManager-test.js
```diff
@@ -114,10 +113,30 @@ it('should use yarn if project is using yarn', () => {
...
+  jest.spyOn(logger, 'isVerbose').mockImplementation(() => false);
+
+  PackageManager.install(PACKAGES, {silent: true});
+
+  expect(execa).toHaveBeenCalledWith('yarn', ['add', ...PACKAGES], {
+    stdio: 'pipe',
+  });
+});
+
+it('should ignore silent when in verbose mode', () => {
```
thymikee: these tests are mostly duplicated, how about using `test.each`? :)

Esemesek: Sure :P

packages/cli/src/commands/init/init.js
Esemesek: I'll just replace default export. I think `getLoader` is good naming here. It shouldn't be confusing when I use `import { getLoader } from '../../tools/loader`;

e2e/__tests__/init.test.js
```diff
@@ -84,3 +84,17 @@ test('init --template file:/tmp/custom/template', () => {
...
   expect(stdout).toContain('Run instructions');
 });
+
+test('init --verbose', () => {
+  const {stdout} = run(DIR, [
+    'init',
+    '--template',
+    'react-native-new-template',
+    'TestInit',
+    '--verbose',
```
thymikee: I don't think we need a test with verbosity on, not a full one at least. Maybe you could use `runUntil` to pause execution after certain log output is there or create a unit test instead?

thymikee: The main problem I have is that running multiple `run`s in one test gets _very_ slow because there's no parallelization there.

packages/cli/src/commands/init/__tests__/template.test.js
```diff
@@ -16,13 +17,14 @@ afterEach(() => {
   jest.restoreAllMocks();
```
thymikee: this doesn't clean the `execa` because it's not a spy. Either move `execa` to spies or call `jest.clearAllMocks` additionally (same for other cases)

packages/cli/src/commands/init/banner.js
```diff
@@ -0,0 +1,48 @@
...
+  '               ##             ####             ##               ',
+  '               ##             ####             ##               ',
+  '               ##           ##    ##           ##               ',
+  '               ##         ###      ###         ##               ',
+  '                ##  ########################  ##                ',
+  '             ######    ###            ###    ######             ',
+  '         ###     ##    ##              ##    ##     ###         ',
+  '      ###         ## ###      ####      ### ##         ###      ',
+  '     ##           ####      ########      ####           ##     ',
+  '    ##             ###     ##########     ###             ##    ',
```
thymikee: btw, have you tried if it looks good with 2 spaces padding instead of 4? (similar for the top, I think 1 or 2 is enough)

packages/cli/src/commands/init/init.js
```diff
@@ -39,50 +41,98 @@ async function createFromExternalTemplate(
...
+
+    loader.succeed();
+    if (templateConfig.postInitScript) {
+      loader.start('Executing post init script');
+      // $FlowFixMe
+      await executePostInitScript(name, templateConfig.postInitScript);
+      loader.succeed();
+    }
+
+    loader.start('Installing all required dependencies');
```
thymikee: maybe `Installing template dependencies`?

e2e/__tests__/init.test.js
Esemesek: I don't think this test is necessary.

thymikee: kill it

packages/cli/src/commands/init/init.js
Esemesek: Both sound good :P

PR #295, opened 2019-04-05
feat: commands should be defined as object, unless defined via legacy config

grabbou:
Summary:
---------

This PR moves the logic for loading commands from CLI to the dependency authors. They should define a JavaScript configuration and return an array of commands (an object with required properties)  they want to load.

Previously, we would require them to define paths (relative to root) to a JavaScript file to require.

This is necessary to send a follow-up PR, which removes built-in platform commands to be bundled with React Native.

WIP: Need to check tests and update one or two snapshots to make sure we output Objects, instead of strings in a configuration.


packages/cli/src/tools/config/readConfigFromDisk.js
```diff
@@ -82,7 +98,9 @@ export function readLegacyDependencyConfigFromDisk(
       hooks: config.commands,
       params: config.params,
     },
-    commands: [].concat(config.plugin || []),
+    commands: []
+      .concat(config.plugin || [])
+      .map(p => loadProjectCommand(rootFolder, p)),
```
grabbou: The RNPM config is confusing. Can be `plugin: ['a.js']`, `plugin: 'a.js'` and contents of `a.js` can be either `module.exports = {}` or `module.exports = [{}]`. This code tries to handle that.

packages/cli/src/tools/config/readConfigFromDisk.js
```diff
@@ -82,7 +98,9 @@ export function readLegacyDependencyConfigFromDisk(
       hooks: config.commands,
       params: config.params,
     },
-    commands: [].concat(config.plugin || []),
+    commands: []
+      .concat(config.plugin || [])
```
thymikee: There's no need for the `[].concat`, can you clean it up while at it?

packages/cli/src/tools/config/readConfigFromDisk.js
thymikee: Hm, so the type is wrong, no? If I understand you correctly `p: Array<string> | string`, while `loadProjectCommand` assumes only `Array<string>`. I also don't see any handling for `module.exports = [{}]` scenario, am I missing something?

packages/cli/src/tools/config/readConfigFromDisk.js
grabbou: Why not? `config.plugin` can be either an array or a string, want to cast it to array

packages/cli/src/tools/config/readConfigFromDisk.js
grabbou: Yup. We use the `concat` on the line above to convert it to an array (concat takes an array or an item to append). Then, we do the same with concat(require) inside `loadProjectCommand`

packages/cli/src/tools/config/readConfigFromDisk.js
thymikee: Wrote that before seeing comment on this, feel free to ignore

thymikee:
Looks like one more test to fix and this should be good to go

grabbou:
Yeah, still working on it! Give me few minutes

thymikee:
Don't get me wrong, I'm not rushing this :D Keep working at any pace

grabbou:
It's all right. I like to move fast

packages/cli/src/tools/config/__tests__/__snapshots__/index-test.js.snap
```diff
@@ -1,5 +1,60 @@
...
+    "windows": Object {},
+  },
+}
+`;
+
+exports[`should load commands from "react-native-foo" and "react-native-bar" packages 1`] = `
+Array [
+  Object {
+    "name": "foo-command",
+  },
```
thymikee: Can we leave one command as is and add `func` to the other one? This way we can have some logic tested for commands that are valid and the ones that are not (we could also inspect stdout to check if there's a warning message)

packages/cli/src/tools/config/__tests__/index-test.js
```diff
@@ -150,27 +141,28 @@ test('should read `rnpm` config from a dependency and transform it to a new form
...
-        "commands": [
-          "./command-bar.js"
-        ]
-      }
+    'node_modules/react-native-bar/package.json': '{}',
+    'node_modules/react-native-bar/react-native.config.js': `module.exports = {
+      commands: [
+        {
+          name: 'bar-command',
+          func: () => console.log('foo')
```
thymikee:
```suggestion
          func: () => console.log('bar')
```

packages/cli/src/tools/config/__tests__/__snapshots__/index-test.js.snap
grabbou: I think it's a good idea, how about we do it in a follow-up PR where I do the "skip invalid packages" feature? That way, I can test that "package with invalid command is not available in the output" and have both capabilities tested at the same time (check resulting object and inspect stdout)

thymikee: Hm, lines https://github.com/react-native-community/cli/pull/295/files#diff-60a96ddc584a7d7c65942648e9e88c57R158 and https://github.com/react-native-community/cli/pull/295/files#diff-60a96ddc584a7d7c65942648e9e88c57R149 indicate that `func` should be there. Can you investigate?

grabbou: I run the test with `-u` flag. Maybe `jest` is not keeping functions in a snapshot?

thymikee: it should show a `[Function]` replacement, so rather nope

grabbou: Logging the config right before returning from the function:
```
[ { name: 'foo-command', func: [Function: func] },
        { name: 'bar-command', func: [Function: func] } ]
```
the properties are there.

grabbou: I need to run to a meeting now, would appreciate if you could investigate it further.

thymikee: I'd blame `removeString` then

grabbou: Could be

grabbou: I can actually remove it from here.

packages/cli/src/tools/config/index.js
```diff
@@ -125,7 +125,7 @@ function loadConfig(projectRoot: string = process.cwd()): ConfigT {
       },
     }: ConfigT),
   );
-
+  console.log(finalConfig.commands);
```
thymikee: remove

PR #296, opened 2019-04-06
chore: migrate `tools` to TypeScript

Esemesek:
Summary:
---------

- [x] - setup Typescript build process
- [x] - migrate `@react-native-community/cli-tools` to Typescript

Test Plan:
----------

Everything builds and tests are passing.


packages/tools/src/groupFilesByType.ts
```diff
@@ -4,22 +4,11 @@
  * This source code is licensed under the MIT license found in the
  * LICENSE file in the root directory of this source tree.
  *
- * @flow
  */
 
-import {groupBy} from 'lodash';
+import { groupBy } from 'lodash';
```
thymikee: fix spacing

packages/tools/src/groupFilesByType.ts
```diff
@@ -4,22 +4,11 @@
...
 import mime from 'mime';
 
-/**
- * Since there are no officially registered MIME types
- * for ttf/otf yet http://www.iana.org/assignments/media-types/media-types.xhtml,
- * we define two non-standard ones for the sake of parsing
- */
-mime.define({
-  'font/opentype': ['otf'],
-  'font/truetype': ['ttf'],
```
thymikee: why is this removed?

scripts/buildTs.js
```diff
@@ -0,0 +1,48 @@
...
+const {getPackages, adjustToTerminalWidth, OK} = require('./helpers');
+
+const packages = getPackages();
+
+const packagesWithTs = packages.filter(p =>
+  fs.existsSync(path.resolve(p, 'tsconfig.json')),
+);
+
+const args = [
+  '--max-old-space-size=4096',
```
thymikee: I think we have too small code base for this?

tsconfig.json
```diff
@@ -0,0 +1,30 @@
...
+    /* Module Resolution Options */
+    "moduleResolution": "node",
+    "esModuleInterop": true,
+    "resolveJsonModule": true
+  },
+  "exclude": [
+    ".yarn/releases/*",
+    "**/__tests__/**/*",
+    "**/build/**/*",
+    "**/build-es5/**/*"
```
thymikee: remove this

tsconfig.json
```diff
@@ -0,0 +1,30 @@
...
+    "noImplicitReturns": true,
+    "noFallthroughCasesInSwitch": true,
+
+    /* Module Resolution Options */
+    "moduleResolution": "node",
+    "esModuleInterop": true,
+    "resolveJsonModule": true
+  },
+  "exclude": [
+    ".yarn/releases/*",
```
thymikee: remove this

packages/tools/src/groupFilesByType.ts
Esemesek: I've updated version of `mimetype` which has those types buildin.

packages/tools/src/groupFilesByType.ts
Esemesek: I will need to configure linter with prettier for Typescript.

grabbou:
Awesome! A small step for a human, big step for CLI xD

lucasbento:
After this PR lands, can we start migrating the rest of the stuff away from Flow?

grabbou:
Yes, that's the idea!

.eslintrc.js
```diff
@@ -16,6 +16,8 @@ module.exports = {
       alias: {
         map: [['types', './types/index.js']],
       },
+      // Use <rootDir>/tsconfig.json for typescript resolution
+      typescript: {},
```
thymikee: we should add TS plugins for ESLint, similar to: https://github.com/callstack/eslint-config-callstack#typescript

packages/tools/src/isPackagerRunning.ts
```diff
@@ -16,18 +16,17 @@ import fetch from 'node-fetch';
...
-          body === 'packager-status:running' ? 'running' : 'unrecognized',
-        ),
-    () => 'not_running',
-  );
+  try {
+    const result = await fetch(`http://localhost:${packagerPort}/status`);
+    const body = await result.text();
+
+    return body === 'packager-status:running' ? 'running' : 'unrecognized';
+  } catch (e) {
```
thymikee:
```suggestion
  } catch (_error) {
```

.eslintrc.js
Esemesek: TS plugins are included in react-native eslint configuration. https://github.com/facebook/react-native/blob/master/packages/eslint-config-react-native-community/index.js#L48

termux-io:
https://patch-diff.githubusercontent.com/raw/vadimdemedes/ink/pull/9.patch

termux-io:
https://patch-diff.githubusercontent.com/raw/react-native-community/cli/pull/296.patch

.vscode/settings.json
```diff
@@ -0,0 +1,24 @@
...
+  "prettier.eslintIntegration": true,
+  "eslint.validate": [
+    "javascript",
+    "javascriptreact",
+    {
+      "language": "typescript",
+      "autoFix": true
+    },
+    {
+      "language": "typescriptreact",
```
grabbou: Should we commit `vscode`? I've never included it and kept the local file

package.json
```diff
@@ -4,14 +4,16 @@
...
+    "prebuild": "yarn build:ts",
     "build": "node ./scripts/build.js",
-    "build-clean": "rm -rf ./packages/*/build",
+    "build:ts": "node ./scripts/buildTs.js",
+    "build-clean": "rm -rf ./packages/*/build ./packages/*/tsconfig.tsbuildinfo",
     "watch": "node ./scripts/watch.js",
     "test": "jest",
     "test:ci:unit": "jest packages --ci --coverage",
     "test:ci:e2e": "jest e2e --ci -i",
+    "lint": "eslint --ext .js,.ts . --cache --report-unused-disable-directives",
```
grabbou: Do we need to specify extensions manually even if we use ESLint config?

.vscode/settings.json
thymikee: I've added it on purpose because of eslint integration issues that we fixed by adding these entries

package.json
Esemesek: For Typescript, yes.

packages/tools/src/groupFilesByType.ts
```diff
@@ -32,5 +21,5 @@ mime.define({
  * the returned object will be: {font: ['fonts/a.ttf'], image: ['images/b.jpg']}
  */
 export default function groupFilesByType(assets: Array<string>) {
-  return groupBy(assets, type => mime.lookup(type).split('/')[0]);
+  return groupBy(assets, type => (mime.getType(type) || '').split('/')[0]);
```
grabbou: When this can return `undefined`?

.vscode/settings.json
Esemesek: We should. It gives contributors recommendations for plugins, and provides better integration with vscode eslint plugin. We have to configure additional rule for plugin to check typescript files.

packages/tools/src/isPackagerRunning.ts
```diff
@@ -16,18 +16,17 @@ import fetch from 'node-fetch';
...
-    () => 'not_running',
-  );
+  try {
+    const result = await fetch(`http://localhost:${packagerPort}/status`);
+    const body = await result.text();
+
+    return body === 'packager-status:running' ? 'running' : 'unrecognized';
+  } catch (_error) {
+    return 'not_running';
+  }
```
grabbou: Feels like a good opportunity for enum here

packages/tools/src/groupFilesByType.ts
Esemesek: When you try ot get mime type which doesn't exist.

grabbou: Was it always like that, even with the previous version? Trying to get an idea whether we had a hidden `cannot read property split of undefined` error or the method changed.

Esemesek: Probably yes, because `mime` lib didn't have any flow typings. When upgraded `mime` lib and attaching typescript types, we caught this error.

packages/tools/src/isPackagerRunning.ts
Esemesek: Yes, but the point of this PR was to configure Typescript ecosystem and prepare for migration, not refactoring every LOC.

grabbou: Well, the PR says "migrate", not "prepare for migration", so that's the reason I am asking. I'd love us to migrate to Typescript fully and embrace its features where possible. 

Shall we open a follow-up issue to track the backlog of things like this, and possibly others that you may already have in mind?

Esemesek: Ye, sure.

termux-io:
https://weedmaps.com/order-details/confirmation/1947167

PR #297, opened 2019-04-06
fix: resolve relative paths for reactNativePath

grabbou:
Summary:
---------

Path to React Native can also be relative, in that case, we should always resolve so that it works.

Test Plan:
----------

Setting `reactNativePath` to `.` works in React Native source


thymikee:
Mind updating the snapshots?

PR #299, opened 2019-04-08
feat: infer package and project name for upgrade from config

thymikee:
Summary:
---------

We can infer package name (for Android) and project name (for iOS) from the config.
This should produce better results than current solution, because we don't need to make assumptions about project name to be synced with package.json.

Supersedes and closes https://github.com/react-native-community/react-native-cli/pull/276
Closes https://github.com/react-native-community/react-native-cli/issues/190

cc @Krizzu, mind testing this branch on your project?

Test Plan:
----------

Added a test

### DIFF
--- a/packages/cli/src/commands/upgrade/__tests__/upgrade.test.js
+++ b/packages/cli/src/commands/upgrade/__tests__/upgrade.test.js
@@ -172,7 +172,24 @@ success Upgraded React Native to v0.58.4 🎉. Now you can review and commit the
 
 test('fetches regular patch, adds remote, applies patch, installs deps, removes remote,', async () => {
   (fetch: any).mockImplementation(() => Promise.resolve(samplePatch));
-  await upgrade.func([newVersion], ctx, opts);
+  await upgrade.func(
+    [newVersion],
+    {
+      ...ctx,
+      project: {
+        ...ctx.project,
+        ios: {
+          ...ctx.project.ios,
+          projectName: 'TestApp.xcodeproj',
+        },
+        android: {
+          ...ctx.project.android,
+          packageName: 'com.testapp',
+        },
+      },
+    },
+    opts,
+  );
   expect(flushOutput()).toMatchInlineSnapshot(`
 "info Fetching diff between v0.57.8 and v0.58.4...
 [fs] write tmp-upgrade-rn.patch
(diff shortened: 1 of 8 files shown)


packages/cli/src/commands/upgrade/__tests__/upgrade.test.js
```diff
@@ -172,7 +172,24 @@ success Upgraded React Native to v0.58.4 🎉. Now you can review and commit the
 
 test('fetches regular patch, adds remote, applies patch, installs deps, removes remote,', async () => {
   (fetch: any).mockImplementation(() => Promise.resolve(samplePatch));
-  await upgrade.func([newVersion], ctx, opts);
+  await upgrade.func(
+    [newVersion],
+    {
+      ...ctx,
+      project: {
+        ...ctx.project,
```
grabbou: You could use `merge` util for deep merge. Doing it this way can execute lazy getters, which is not necessarily bad, but why not.

packages/cli/src/commands/upgrade/upgrade.js
```diff
@@ -60,9 +57,32 @@ const getPatch = async (currentVersion, newVersion, projectDir) => {
...
+  // rn-diff-purge only supports iOS and Android platforms
+  if (config.project.ios) {
+    patchWithRenamedProjects = patchWithRenamedProjects.replace(
+      new RegExp('RnDiffApp', 'g'),
+      // $FlowFixMe - poor typings of ProjectConfigIOST
+      config.project.ios.projectName.replace('.xcodeproj', ''),
+    );
+  }
+
+  if (config.project.android) {
```
grabbou: Maybe you would iterate on `config.platforms` and print a warning for every platform that is not android/ios? Just in case someone has `windows` integrated.

packages/cli/src/commands/upgrade/upgrade.js
```diff
@@ -60,9 +57,32 @@ const getPatch = async (currentVersion, newVersion, projectDir) => {
...
+    patchWithRenamedProjects = patchWithRenamedProjects
+      .replace(
+        new RegExp('com\\.rndiffapp', 'g'),
+        // $FlowFixMe - poor typings of ProjectConfigAndroidT
```
grabbou: Maybe you would add the typing for these? At lest for the ones you are using, we can extend the types partially.

packages/cli/src/commands/upgrade/upgrade.js
```diff
@@ -214,7 +234,7 @@ async function upgrade(argv: Array<string>, ctx: ContextT, args: FlagsT) {
     return;
   }
 
-  const patch = await getPatch(currentVersion, newVersion, projectDir);
+  const patch = await getPatch(currentVersion, newVersion, projectDir, ctx);
```
grabbou: I think `projectDir` is also on `ctx` as `root` so might be better to stick to that one?

packages/cli/src/commands/upgrade/upgrade.js
thymikee: I tried, but it seemed to result in quite a bunch of errors I didn't want to handle in this diff. Will try to address that later

packages/cli/src/commands/upgrade/upgrade.js
thymikee: Turned out `projectDir` was not used :D

PR #302, opened 2019-04-09
feat: extract commands out to platform packages

grabbou:
Summary:
---------

Follow up to #290, we also move out commands.

Side note: In the future, we might want to move `errors` to tools too in order to have better error reporting. Not important for now.


package.json
```diff
@@ -31,6 +31,7 @@
     "glob": "^7.1.3",
     "jest": "^24.6.0",
     "lerna": "^3.13.1",
+    "metro-memory-fs": "^0.53.1",
```
thymikee: Please move it to `devDependencies`, it's only used for testing, no need to feed it to the users

packages/cli/package.json
```diff
@@ -57,10 +52,7 @@
     "semver": "^5.0.3",
     "serve-static": "^1.13.1",
     "shell-quote": "1.6.1",
-    "slash": "^2.0.0",
-    "ws": "^1.1.0",
-    "xcode": "^2.0.0",
-    "xmldoc": "^0.4.0"
+    "ws": "^1.1.0"
```
thymikee: would be good to update it, latest is v6 :o

packages/cli/src/commands/link/unlink.js
```diff
@@ -137,7 +136,9 @@ function unlink(args: Array<string>, ctx: ConfigT) {
     })
     .catch(err => {
       logger.error(
-        `It seems something went wrong while unlinking. Error:\n${err.message}`,
+        `It seems something went wrong while unlinking. CLIError:\n${
+          err.message
+        }`,
```
thymikee: How about actually throwing `CLIError` here?

grabbou: yeah, could be.

packages/cli/package.json
grabbou: Yeah, but this is proxy stuff, so I am a bit worried of doing it here.

PR #309, opened 2019-04-10
Do not throw on invalid packages

grabbou:
Summary:
---------

Some packages can have invalid configuration. Our convention was to throw an error in that case. It is very likely some packages can have invalid configuration (since it was super wild and never standardised), so it's better to simply ignore them and print an appropriate warning. 

Test Plan:
----------

Do a broken configuration for any of your dependencies and run `config`. Package should be excluded.

### DIFF
--- a/packages/cli/src/tools/config/__tests__/index-test.js
+++ b/packages/cli/src/tools/config/__tests__/index-test.js
@@ -236,3 +236,19 @@ test('should not add default React Native config when one present', () => {
   const {commands} = loadConfig(DIR);
   expect(commands).toMatchSnapshot();
 });
+
+test('should skip packages that have invalid configuration', () => {
+  writeFiles(DIR, {
+    'node_modules/react-native/package.json': '{}',
+    'node_modules/react-native/react-native.config.js': `module.exports = {
+      invalidProperty: 5
+    }`,
+    'package.json': `{
+      "dependencies": {
+        "react-native": "0.0.1"
+      }
+    }`,
+  });
+  const {dependencies} = loadConfig(DIR);
+  expect(dependencies).toMatchSnapshot();
+});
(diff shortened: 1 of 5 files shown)


packages/cli/src/tools/config/index.js
```diff
@@ -34,14 +36,28 @@ function loadConfig(projectRoot: string = process.cwd()): ConfigT {
...
+            Reason: ${chalk.dim(error.message)}
+          `),
+        );
+        return acc;
+      }
 
       /**
        * This workaround is neccessary for development only before
        * first 0.60.0-rc.0 gets released and we can switch to it
-       * while testing.
```
thymikee: ?

packages/cli/src/tools/config/__tests__/index-test.js
```diff
@@ -236,3 +236,19 @@ test('should not add default React Native config when one present', () => {
...
+      invalidProperty: 5
+    }`,
+    'package.json': `{
+      "dependencies": {
+        "react-native": "0.0.1"
+      }
+    }`,
+  });
+  const {dependencies} = loadConfig(DIR);
+  expect(dependencies).toMatchSnapshot();
```
thymikee: can you also spy on `logger.warn` and assert on the message? :) or snapshot it

packages/cli/src/tools/config/index.js
grabbou: Mistake

packages/cli/src/tools/config/__tests__/index-test.js
grabbou: Should that be a separate test for appropriate warning or here?

thymikee: same test, different `expect` call :)

grabbou:
Updated test case and also reverted the doc block.

PR #310, opened 2019-04-11
fix: open Metro in the correct terminal

### LINKED ISSUE #303: Open Metro in iTerm on macOS
## Describe the Feature
A lot of people use iTerm 2 on macOS. When using `run-android` or `run-ios`, Metro is started in `Terminal.app` even when launched from iterm. Ideally we'd create a new tab or window in iTerm.

## Possible Implementations
* If there is a way to detect whether run-android/ios was invoked from within iTerm, use iTerm.
* Otherwise, provide a way to set this as a configuration (per project, per computer) once and reuse the setting

## Related Issues
Note that `run-android` already has an option to chose the terminal via an environment variable. It looks like it should work but it didn't when I just tested it.

lucasbento:
This PR fixes #303 as well as the `--terminal` argument that was being passed to `--run-android`.

It also adds the `--terminal` argument to `--run-ios`, of which the user is able to specify on which terminal they want to start the Metro bundler, e.g.:

```sh
# iOS
react-native run-ios --terminal Hyper
export REACT_TERMINAL=iTerm react-native run-ios

# Android
react-native run-android --terminal Hyper
export REACT_TERMINAL=iTerm react-native run-android
```

This PR is related to https://github.com/facebook/react-native/pull/24436.


cpojer:
omg I'm so excited, thank you for working on this!

orta:
⭐️

thymikee:
@lucasbento is this still WIP? The test failure seems unrelated (I've seen it on master once in a while, so likely a flaky test with some bad mocking)

lucasbento:
@thymikee: thanks for letting me know!

Still a WIP as I need to open a PR the RN repo and add description to this, would be nice if you could trigger the tests again.

packages/platform-android/src/commands/runAndroid/index.js
```diff
@@ -359,7 +359,7 @@ export default {
       command: '--terminal [string]',
       description:
         'Launches the Metro Bundler in a new window using the specified terminal path.',
-      default: '',
+      default: undefined,
```
grabbou: Let's make it `default: process.env.REACT_TERMINAL || process.env.TERM_PROGRAM` and remove the default assignement on line 231.
```suggestion
      default: process.env.REACT_TERMINAL || process.env.TERM_PROGRAM,
```

packages/platform-android/src/commands/runAndroid/index.js
```diff
@@ -228,7 +228,7 @@ function installAndLaunchOnDevice(
 
 function startServerInNewWindow(
   port,
-  terminal = process.env.REACT_TERMINAL,
+  terminal = process.env.REACT_TERMINAL || process.env.TERM_PROGRAM,
```
grabbou:
```suggestion
  terminal,
```

packages/platform-ios/src/commands/runIOS/index.js
```diff
@@ -29,6 +29,7 @@ type FlagsT = {
   packager: boolean,
   verbose: boolean,
   port: number,
+  terminal: string | typeof undefined,
```
grabbou:
```suggestion
  terminal: ?string,
```

packages/platform-ios/src/commands/runIOS/index.js
```diff
@@ -418,15 +425,17 @@ function printFoundDevices(devices) {
   return output;
 }
 
-function getProcessOptions(launchPackager, port) {
+function getProcessOptions(launchPackager, port, terminal) {
+  const env = {...process.env, RCT_TERMINAL: terminal};
```
grabbou: Any reason to do spreading twice? I would leave it as is and just add `RCT_TERMINAL: terminal` to line 433.

grabbou:
```suggestion
```

packages/platform-ios/src/commands/runIOS/index.js
```diff
@@ -418,15 +425,17 @@ function printFoundDevices(devices) {
...
+
   if (launchPackager) {
     return {
-      env: {...process.env, RCT_METRO_PORT: port},
+      env: {...env, RCT_METRO_PORT: port},
```
grabbou:
```suggestion
      env: {...process.env, RCT_TERMINAL: terminal, RCT_METRO_PORT: port},
```

packages/platform-ios/src/commands/runIOS/index.js
```diff
@@ -418,15 +425,17 @@ function printFoundDevices(devices) {
...
     };
   }
 
   return {
-    env: {...process.env, RCT_NO_LAUNCH_PACKAGER: true},
+    env: {...env, RCT_NO_LAUNCH_PACKAGER: true},
```
grabbou:
```suggestion
    env: {...process.env, RCT_TERMINAL: terminal, RCT_NO_LAUNCH_PACKAGER: true},
```

packages/platform-ios/src/commands/runIOS/index.js
```diff
@@ -499,5 +508,11 @@ export default {
       default: process.env.RCT_METRO_PORT || 8081,
       parse: (val: string) => Number(val),
     },
+    {
+      command: '--terminal [string]',
+      description:
+        'Launches the Metro Bundler in a new window using the specified terminal path.',
+      default: undefined,
```
grabbou: Shouldn't it be same default as in other command? `process.env...`

grabbou:
```suggestion
      default: process.env.REACT_TERMINAL || process.env.TERM_PROGRAM,
```

grabbou: Can also move it to utilities if we don't want to repeat this logic

packages/platform-ios/src/commands/runIOS/index.js
```diff
@@ -179,6 +182,7 @@ async function runOnSimulator(xcodeProject, args, scheme) {
     args.packager,
     args.verbose,
     args.port,
+    args.terminal,
```
grabbou: Looks like `args.port` and `args.terminal` are passed to all these functions just to call `getProcessOptions` at the very end inside `buildProject`.

Why don't we do:
```js
const processOpts = getProcessOpts(args)
```
here and pass the process options explicitly?

packages/platform-ios/src/commands/runIOS/index.js
lucasbento: Just didn't want to specify `RCT_TERMINAL` twice, I'll change it though.

packages/platform-ios/src/commands/runIOS/index.js
lucasbento: Yep, I'll change it!

lucasbento: Good one.

lucasbento:
@grabbou: thanks for the review, I applied the requested changes except for the last one which I need more info.

packages/platform-ios/src/commands/runIOS/index.js
thymikee: @grabbou we call `buildProject` in 2 places. This would cause some unnecessary duplication. 

I think what @lucasbento suggested to combine `args` into single argument of `buildProject` makes sense. It doesn't matter that the order is different between similar functions but feel free to change it to make some more sense, it's only local to this file.

I'm cool with refactoring to this signature:
```js
function buildProject(xcodeProject, udid, scheme, args) {};
```
or generally all functions to something like:
```js
function runOnSimulator(xcodeProject, scheme, args) {};
function runOnDevice(xcodeProject, scheme, args, selectedDevice) {};
function runOnDeviceByUdid(xcodeProject, scheme, args, devices) {};
```
which seams reasonable and I'm happy to have this refactoring as part of this PR :)

grabbou: > Can you elaborate? Do you mean passing args to buildProject and then just passing it along to getProccessOpts?

Yes! The signature proposed by @thymikee is what I had in mind! Just trying to avoid the last few arguments of each of these `run` functions that are not really for `run` function, but for underlying buildProject.

grabbou:
Awesome @lucasbento, thanks for quick turnaround. I'll already answered, looking forward to merging it and shipping in 0.60!

packages/platform-ios/src/commands/runIOS/index.js
lucasbento: Great, I'll work on the `args` refactor.

lucasbento:
@grabbou & @thymikee: just committed the changes, can you review again?

packages/platform-ios/src/commands/runIOS/index.js
```diff
@@ -71,55 +71,39 @@ function runIOS(_: Array<string>, ctx: ConfigT, args: FlagsT) {
...
+  }
+
+  if (devices && devices.length > 0) {
+    // $FlowIssue: args.device is defined in this context
+    return logger.error(`Could not find device with the name: "${args.device}".
 Choose one of the following:${printFoundDevices(devices)}`);
-    } else {
-      logger.error('No iOS devices connected.');
-    }
-  } else if (args.udid) {
```
grabbou: Any reason to change the order of `if/else` in the file? I quite liked the previous one (was a bit more readable to me)

grabbou:
Just did it :) My pleasure!

packages/platform-ios/src/commands/runIOS/index.js
grabbou: I think this has something to do with the negation, which I am not particularly in favour of. 

Also, how about we remove
```
if (args.udid) {
      // $FlowIssue: args.udid is defined in this context
      return runOnDeviceByUdid(scheme, xcodeProject, devices, args);
    }
```
and do something like:
```
if (args.udid || args.device) {
   const selectedDevice = args.udid ? matchingDeviceByUdid(devices, args.udid) : matchingDevice(devices, args.device);

  // logic
}
```

Benefits - it's explicit and we can handle missing device in one place. Right now, we do it in both places.

grabbou: Also, we should update the type of `device` from `?string` to `?string | ?boolean` because (i've just found out), using `--device` (without a value) will result in a "first available simulator" being picked up.

Sorry if this is annoying, but there's a lot of tiny things here that could be improved that I think we could do while we are here. This is not related to issues with your code per se :)

lucasbento: Good points, I'll work on them later on today.

No worries about the review though, it's really nice to improve this part!

If you have any major ideas that don't fit in here you can also create an issue and assign to me, would be happy to work on it.

lucasbento: Does it make sense to have `typeof device === Boolean` though? won't it just be `undefined`?

I think having the command in `options` array as `--device [string]` and type as `?string | ?boolean` can be a little strange.

grabbou: @lucasbento, right now, we can either do `--device "Mike's iPhone"` or `--device`. How we can type that? 

I believe, that given the way Commander works, `--device` will result in `args.device` being `true`, which is why it's been handled this way.

grabbou: Maybe `?(string | true)`

lucasbento: Alright, I'll do it!

lucasbento:
@grabbou: just committed the requested changes!

I'm not able to test the successful flow of `--device` nor `--udid` right now because I only have one iOS phone here and I am hesitant to update Xcode to `10.2` 🙉 

I believe you can already check the code out and I'll test it throughly tomorrow with another device.

packages/platform-ios/src/commands/runIOS/index.js
```diff
@@ -66,65 +71,37 @@ function runIOS(_: Array<string>, ctx: ConfigT, args: FlagsT) {
...
     }
+
     if (devices && devices.length > 0) {
-      // $FlowIssue: args.device is defined in this context
-      logger.error(`Could not find device with the name: "${args.device}".
-Choose one of the following:${printFoundDevices(devices)}`);
-    } else {
-      logger.error('No iOS devices connected.');
+      const device = ((args.device: any): string);
+      const udid = ((args.udid: any): string);
```
thymikee: Can you move device and uuid up in the scope so that line 74 can also use them? Flow should understand it better and you likely won't need to cast like this

packages/platform-ios/src/commands/runIOS/index.js
```diff
@@ -66,65 +71,37 @@ function runIOS(_: Array<string>, ctx: ConfigT, args: FlagsT) {
...
+
+      const message = args.device
```
thymikee: use `device` variable

packages/platform-ios/src/commands/runIOS/index.js
lucasbento: Was going to ask about this, didn't know what it was the best to do and I followed what's in https://flow.org/en/docs/types/casting/#toc-type-casting-through-any.

What do you mean by moving it up in the scope? Do you mean moving the lines 84 and 85 to before 74?

thymikee: yes

lucasbento: But as far as line 74 is concerned the `args.device` and `args.udid` can still be other types rather than `string`, why is it better to put it before it?

thymikee: If the object property is bound to variable Flow knows that it's not gonna mutate in flight (it's a bug in Flow imo, but whatever). So by doing that Flow will understand your if branch better and will know that `uuid` and `device` cannot be undefined in your ternary. At least that's what I would expect

lucasbento: Alright, done!

packages/platform-ios/src/commands/runIOS/index.js
lucasbento: Good catch, fixed!

packages/platform-ios/src/commands/runIOS/index.js
```diff
@@ -66,65 +71,36 @@ function runIOS(_: Array<string>, ctx: ConfigT, args: FlagsT) {
     }),
   );
 
-  if (args.device) {
-    const selectedDevice = matchingDevice(devices, args.device);
+  const device = ((args.device: any): string);
+  const udid = ((args.udid: any): string);
+  if (device || udid) {
```
thymikee: This is not clean, inside we need to branch 2 more times, which is not really more readable and efficient. What would you tell about this refactoring?

1. If there's no `device` nor `udid`, we run on simulator straight away.
1. If there are no devices found, we bail early.
1. We run device/udid branching inside `matchingDevice` fn (inlining `matchingDeviceByUdid` into it)
1. If device matches, run on device
1. If not, print appropriate messages depending on whether `device` or `udid` was passed.

```diff
--- a/packages/platform-ios/src/commands/runIOS/index.js
+++ b/packages/platform-ios/src/commands/runIOS/index.js
@@ -64,6 +64,12 @@ function runIOS(_: Array<string>, ctx: ConfigT, args: FlagsT) {
     }`,
   );

+  const {device, udid} = args;
+
+  if (!device && !udid) {
+    return runOnSimulator(xcodeProject, scheme, args);
+  }
+
   const devices = parseIOSDevicesList(
     // $FlowExpectedError https://github.com/facebook/flow/issues/5675
     child_process.execFileSync('xcrun', ['instruments', '-s'], {
@@ -71,33 +77,31 @@ function runIOS(_: Array<string>, ctx: ConfigT, args: FlagsT) {
     }),
   );

-  const device = ((args.device: any): string);
-  const udid = ((args.udid: any): string);
-  if (device || udid) {
-    const selectedDevice = device
-      ? matchingDevice(devices, device)
-      : matchingDeviceByUdid(devices, udid);
+  if (devices.length === 0) {
+    return logger.error('No iOS devices connected.');
+  }
+
+  const selectedDevice = matchingDevice(devices, device, udid);

   if (selectedDevice) {
     return runOnDevice(selectedDevice, scheme, xcodeProject, args);
   }

-    if (devices && devices.length > 0) {
-      const message = device
-        ? `Could not find device with the name: "${device}". Choose one of the following:\n${printFoundDevices(
-            devices,
-          )}`
-        : `Could not find device with the udid: "${udid}". Choose one of the following:\n${printFoundDevices(
-            devices,
-          )}`;
-
-      return logger.error(message);
+  if (device) {
+    return logger.error(
+      `Could not find device with the name: "${String(
+        device,
+      )}". Choose one of the following:\n${printFoundDevices(devices)}`,
+    );
   }

-    return logger.error('No iOS devices connected.');
+  if (udid) {
+    return logger.error(
+      `Could not find device with the udid: "${udid}". Choose one of the following:\n${printFoundDevices(
+        devices,
+      )}`,
+    );
   }
-
-  return runOnSimulator(xcodeProject, scheme, args);
 }

 async function runOnSimulator(xcodeProject, scheme, args: FlagsT) {
@@ -331,7 +335,10 @@ function xcprettyAvailable() {
   return true;
 }

-function matchingDevice(devices, deviceName) {
+function matchingDevice(devices, deviceName, udid) {
+  if (udid) {
+    matchingDeviceByUdid(devices, udid);
+  }

```

lucasbento: I usually prefer to return as early as possible, even if it needs to have negative conditionals, with a quick look I agree with the diff you sent.

@grabbou thoughts on this? Asking because you pointed that you didn't like the conditionals I introduced on previous commits.

grabbou: I actually disagree this is not clean. The first branch is `runOnDevice` - you can select device by name or id. The second branch is `runOnSimulator`. 

Anyway, I don't have a strong preference here. Let's merge it as soon as possible as I want to ship it for next RC of React Native.

We can tweak it later.

grabbou: PS. I like @thymikee structure better, but at the same time, I feel like @lucasbento keeps doing tweaking that is far beyond his PR. I'd just get it merged to not discourage them from further PRs!

thymikee: @grabbou feel free to merge it as is now. Hopefully @lucasbento will not be discouraged by our nitpicks 😅

grabbou:
We agreed that @thymikee will send a follow up PR with his diff (that he already pasted here in the comment). 

Thanks @lucasbento for your work here! Excited to be shipping this now!

PR #312, opened 2019-04-11
chore: move shared types to root; moar link styles; bye promiseWaterfall

thymikee:
Summary:
---------

This is low priority, but feel free to review. 
What this diff does:

- adds missing typings for Project config and Dependency config for both iOS and Android
- moves `cli/.../tools/types.flow` to root `/types/index.js`, so they're accessible from every package without the need to go to `../../src/...`
- adds `types` alias for global types
- removes `promiseWaterfall` and alike in favor of `async`/`await`

Test Plan:
----------

Green CI


packages/cli/src/commands/link/linkDependency.js
```diff
@@ -34,8 +34,10 @@ const linkDependency = async (
     }
 
     const isInstalled = linkConfig.isInstalled(
+      // $FlowFixMe
       projectConfig,
```
thymikee: couldn't get Flow to understand this (don't say TS gonna fix it, it probably won't). Help appreciated on getting this solved, but I'm also cool with leaving this as is.

packages/cli/src/cliEntry.js
```diff
@@ -12,7 +12,7 @@ import childProcess from 'child_process';
 import commander from 'commander';
 import path from 'path';
 
-import type {CommandT, ConfigT} from './tools/config/types.flow';
+import type {CommandT, ConfigT} from '../../../types/config';
```
grabbou: I wonder if we could somehow tell `flow` to look it up from the `root` w/o specifying the relative path?

packages/cli/src/commands/link/linkDependency.js
grabbou: It will! Flow has issues understanding whether it's Android/iOS its dealing right now. Run into it in other places too, so good to ignore.

packages/cli/src/commands/link/pollParams.js
```diff
@@ -5,12 +5,13 @@
...
  *
  * @format
+ * @flow
  */
 
 import inquirer from 'inquirer';
 
-export default questions =>
-  new Promise((resolve, reject) => {
+export default (questions: any) =>
```
grabbou: I think there's a type for that in `config`. Check `project.params`

thymikee: We do, it's `any` :p. Can import it though

packages/cli/src/commands/link/linkDependency.js
thymikee: Maybe, but that has noting to do with platforms, just figuring out alternatives of types. E.g. If we rewritten this code to switch statement with a lot of repetition, Flow would understand the branching logic, but that would be too much

grabbou: Haha, yeah, I know it has nothing to do with platforms :) Was just an example I used to explain what you said!

grabbou:
Great. Now it's going to be much easier to type the whole codebase with Flow (not that we are actually converting it to TypeScript in the meantime)

PR #314, opened 2019-04-12
chore: use "files" instead of .npmignore for publishing

thymikee:
Summary:
---------

We forgot about adding `.npmignore` or `"files"` entry for new packages, so we unnecessarily published their sources. This diff makes all packages use `"files"` entry as it's more explicit in what we want to publish and it's harder to forget about it when copy-pasting package.json for new packages.

Test Plan:
----------

Run `npm pack --dry-run` for every package. Please verify if I didn't miss anything.


grabbou:
Much better to depend on `files`. Thanks!

PR #318, opened 2019-04-16
Add support for assetPlugins argument to start command

brentvatne:
Summary:
---------

**This pull request adds support for loading `assetPlugins` through flags to the CLI `start` command.**

This option is probably unknown to you because I don't think anybody uses it except Expo and it happens behind the scenes 😆. In Expo projects we depend on the `assetPlugins` configuration option for the asset system in order to inject the md5 hash for each asset into the asset metadata within AssetRegistry. In the meto versioned used by react-native@0.57, support for `assetPlugins` in the query param for bundles was broken along with a refactor for configuration. In order to ship SDK 31 we patched our fork of react-native to make local-cli always include this plugin (https://github.com/expo/react-native/commit/d41bc86fd9fec9c55ae72bd4cdc1508097f7796b) but now that react-native-cli is a separate package it's a bit more complicated -- plus, a fix to the underlying problem is important too. Additionally, using `assetPlugins` in the bundle URL confused users often as it would appear in their error messages as a red herring.

As a follow up to this pull request I will be doing one of the following:

- Sending a pull request to Metro to change the `loadConfig` function to accept the `assetPlugins` option, which will simplify this in the future.
- OR, depending on discussions with CLI folks, we may want to add the `hashAssetFiles` behavior by default in the CLI. We'll need to discuss it further but I think there is a good argument for why this can be useful in general.

Test Plan:
----------

- Create a project with `react-native init`
- Pull in the code from this patch to your local version of react-native-cli
- Add `md5-file` to your project `yarn add md5-file`
- Create a file called `exampleAssetPlugin.js` in the root of your project, put this in it:

```js
const md5File = require('md5-file/promise');

module.exports = function exampleAssetPlugin(asset) {
  return Promise.all(asset.files.map(md5File)).then(hashes => {
    asset.fileHashes = hashes;
    return asset;
  });
};
```

- Run the project with `yarn start --assetPlugins `pwd`/hashAssetFiles` and add the following code to `App.js` (copy an image there and replace the name below with the name of that image):

```js
import { getAssetByID } from 'react-native/Libraries/Image/AssetRegistry';
alert(JSON.stringify(getAssetByID(require('./some-image-in-your-project.png'))));
```

You will see that the asset metadata now includes the `fileHashes` property with md5 hashes.

Other:
-------

If this lands I will backport it to 1.x so that we can release an update that will work with react-native@0.59.

### DIFF
--- a/packages/cli/src/commands/server/runServer.js
+++ b/packages/cli/src/commands/server/runServer.js
@@ -20,6 +20,7 @@ import loadMetroConfig from '../../tools/loadMetroConfig';
 
 export type Args = {|
   assetExts?: string[],
+  assetPlugins?: string[],
   cert?: string,
   customLogReporterPath?: string,
   host?: string,
--- a/packages/cli/src/commands/server/runServer.js
+++ b/packages/cli/src/commands/server/runServer.js
@@ -54,6 +54,12 @@ async function runServer(argv: Array<string>, ctx: ConfigT, args: Args) {
     reporter,
   });
 
+  if (args.assetPlugins) {
+    metroConfig.transformer.assetPlugins = args.assetPlugins.map(plugin =>
+      require.resolve(plugin),
+    );
+  }
+
   const middlewareManager = new MiddlewareManager({
     host: args.host,
     port: metroConfig.server.port,
--- a/packages/cli/src/commands/server/server.js
+++ b/packages/cli/src/commands/server/server.js
@@ -31,6 +31,12 @@ export default {
       parse: (val: string) => val.split(','),
     },
     {
+      command: '--assetPlugins [list]',
+      description:
+        'Specify any additional asset plugins to be used by the packager by full filepath',
+      parse: (val: string) => val.split(','),
+    },
+    {
       command: '--assetExts [list]',
       description:
         'Specify any additional asset extensions to be used by the packager',


grabbou:
Hey Brent,

Thanks for sending this PR.

I have a question - why don't we leverage the Metro configuration, that is, [create a `metro.config.js` file with `assetPlugins` defined](https://facebook.github.io/metro/docs/en/configuration#assetplugins)? This is what we have been doing with other options (which allowed us to remove the mutation behaviour).

brentvatne:
> I have a question - why don't we leverage the Metro configuration, that is, create a metro.config.js file with assetPlugins defined? This is what we have been doing with other options (which allowed us to remove the mutation behaviour).

I explored this option, we considered starting metro with xdl using the `config` option and then passing a path to a custom metro config, but decided that this wasn't ideal because users would not be able to extend the config easily, they'd need to copy the existing config in to a new metro.config.js. I then thought about making it possible to support multiple configs, but @ide suggested that it would likely lead to some confusing merging behaviors.

> If, for some reason, you can't keep this configuration in the user-space and you prefer to use app.json's packagerOpts for that, we could add a first-class support for reading values from it, instead of you, passing each of the settings as a flag

We don't want to keep this in userspace because it isn't something that Expo users should ever think about, so it becomes unnecessary additional overhead / boilerplate. We used to pass it in via url, eg: `x.bundle?assetPlugins=stuff` but some internal metro refactor broke this. This was more visible than I'd have liked even, as folks could see it in the URL.

> We prefer to keep the CLI bundler agnostic in the future and would prefer not adding additional Metro-related flags and favour its standard configuration mechanism instead.

What if this was prefixed with `metro` instead? `metroAssetPlugins` - this way it's clear that it is only supported by a particular bundler. I think that also this same approach may end up being used with webpack or some other bundler because we will still need to somehow ensure that each asset has its md5 attached to its metadata.

grabbou:
> I explored this option, we considered starting metro with xdl using the config option and then passing a path to a custom metro config, but decided that this wasn't ideal because users would not be able to extend the config easily, they'd need to copy the existing config in to a new metro.config.js. I then thought about making it possible to support multiple configs, but @ide suggested that it would likely lead to some confusing merging behaviors.

Yeah, that makes sense. One alternative would be for you to read the config itself and pass already resolved options to the CLI, but that would be unnecessarily confusing.

In that case:

> Sending a pull request to Metro to change the loadConfig function to accept the assetPlugins option, which will simplify this in the future.

This sounds like a fair path going forward. Please let me know when it happens.

grabbou:
Thanks for providing a thorough explanation. Going to merge it as is for now.

brentvatne:
At @thymikee's suggestion I investigated adding our own config file which loads the project's metro config and extends it.

Unfortunately I can't call `loadConfig` from `metro-config` from inside another metro config file because it's async but `metro-config` needs to synchronously export the config object. So, thanks for merging this!

PR #319, opened 2019-04-16
fix: add new line to the help message

ifsnow:
Summary:
---------
There is a problem with the line break of command help as below.

Test Plan:
----------

Not required.

### DIFF
--- a/packages/cli/src/cliEntry.js
+++ b/packages/cli/src/cliEntry.js
@@ -79,7 +79,7 @@ function printHelpInformation(examples, pkg) {
     output = output.concat([chalk.bold('\nExample usage:'), formattedUsage]);
   }
 
-  return output.concat('').join('\n');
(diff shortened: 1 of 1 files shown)


packages/cli/src/cliEntry.js
```diff
@@ -79,7 +79,7 @@ function printHelpInformation(examples, pkg) {
     output = output.concat([chalk.bold('\nExample usage:'), formattedUsage]);
   }
 
-  return output.join('\n');
+  return output.concat('').join('\n');
```
thymikee: How about 
```suggestion
  return output.join('\n').concat('\n');
```
which would make it more obvious that we want to add a final newline?

ifsnow: I think the way you said it is better. Thanks :)

PR #320, opened 2019-04-16
feat: support init inside monorepos

### LINKED ISSUE #277: Init does not work with yarn workspaces
Copy from https://github.com/facebook/react-native/issues/24202 and there's a related issue https://github.com/react-native-community/react-native-cli/issues/271

## 🐛 Bug Report
I'm using `react-native init` to create a simple app inside a monorepo, but the cli seems to have problems with the yarn workspaces (with hoisting) - I'm also using lerna, but i don't think that lerna is causing the problems here.

## To Reproduce
I'm just running:
`react-native init KeepfyApp`

The log is:
```
This will walk you through creating a new React Native project in /home/gabriel/keepfy/root/packages/KeepfyApp
Using yarn v1.15.2
Installing react-native...
yarn add v1.15.2
[1/4] Resolving packages...
[2/4] Fetching packages...
[3/4] Linking dependencies...
**========= SOME WARNINGS NOT INCLUDED HERE =========**

success Saved 1 new dependency.
info Direct dependencies
info All dependencies
└─ react-native@0.59.2
Done in 12.57s.
internal/modules/cjs/loader.js:651
    throw err;
    ^

Error: Cannot find module '/home/gabriel/keepfy/root/packages/KeepfyApp/node_modules/react-native/package.json'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:649:15)
    at Function.Module._load (internal/modules/cjs/loader.js:575:25)
    at Module.require (internal/modules/cjs/loader.js:705:19)
    at require (internal/modules/cjs/helpers.js:14:16)
    at checkNodeVersion (/home/gabriel/.fnm/node-versions/v11.12.0/installation/lib/node_modules/react-native-cli/index.js:306:21)
    at run (/home/gabriel/.fnm/node-versions/v11.12.0/installation/lib/node_modules/react-native-cli/index.js:300:3)
    at createProject (/home/gabriel/.fnm/node-versions/v11.12.0/installation/lib/node_modules/react-native-cli/index.js:249:3)
    at init (/home/gabriel/.fnm/node-versions/v11.12.0/installation/lib/node_modules/react-native-cli/index.js:200:5)
    at Object.<anonymous> (/home/gabriel/.fnm/node-versions/v11.12.0/installation/lib/node_modules/react-native-cli/index.js:153:7)
    at Module._compile (internal/modules/cjs/loader.js:799:30)
```
The `Error: Cannot find module '/home/gabriel/keepfy/root/packages/KeepfyApp/node_modules/react-native/package.json'` is correct, there's no `node_modules` folder there, they are installed in the root folder, the correct path should be `/home/gabriel/keepfy/root/node_modules/react-native/package.json`

## Expected Behavior

I'm expecting the app to be created, but only the `package.json` is created.

## Code Example

```
// lerna.json
{
  "packages": [
    "packages/*"
  ],
  "npmClient": "yarn",
  "useWorkspaces": true,
  "version": "independent"
}

//package.json (root)
{
  "name": "root",
  "private": true,
  "devDependencies": {
    "lerna": "3.13.1",
    "yarn": "1.15.2"
  },
  "workspaces": [
    "packages/*"
  ],
  "scripts": {
    "build": "lerna run build --stream",
    "bootstrap": "lerna bootstrap"
  }
}
```
With those files in a root project folder, run `cd packages && react-native init` and you should see the error.

There's something to note here, i could start a ''correct'' project only once, `react-native init` created the folder and stuff inside, but it installed ignoring the yarn workspaces (because it created the `yarn.lock` inside the app folder).. so i think that time doesn't count.

## Environment
`[skip envinfo]`

Ran inside the created folder `KeepfyApp`:
`react-native info` ->
```
Command `info` unrecognized. Make sure that you have run `npm install` and that you are inside a react-native project.
```
`react-native --version` ->
```
react-native-cli: 2.0.1
react-native: n/a - not inside a React Native project directory
```

If there's another info i can provide to help just ask.

thymikee:
Summary:
---------

Installs template sources inside temporary directory and removes it after operation is done or failed.

Fixes #277

Test Plan:
----------

Tests yet to be added.

### DIFF
--- a/packages/cli/src/commands/init/__tests__/template.test.js
+++ b/packages/cli/src/commands/init/__tests__/template.test.js
@@ -12,6 +12,7 @@ import {
 import * as copyFiles from '../../../tools/copyFiles';
 
 const TEMPLATE_NAME = 'templateName';
+const TEMPLATE_SOURCE_DIR = '/tmp/rncli-init-template-123456';
 
 afterEach(() => {
   jest.restoreAllMocks();
@@ -21,17 +22,18 @@ afterEach(() => {
 test('installTemplatePackage', async () => {
   jest.spyOn(PackageManger, 'install').mockImplementationOnce(() => {});
 
-  await installTemplatePackage(TEMPLATE_NAME, true);
+  await installTemplatePackage(TEMPLATE_NAME, TEMPLATE_SOURCE_DIR, true);
 
   expect(PackageManger.install).toHaveBeenCalledWith([TEMPLATE_NAME], {
     preferYarn: false,
     silent: true,
+    cwd: TEMPLATE_SOURCE_DIR,
   });
 });
 
 test('getTemplateConfig', () => {
   jest.mock(
-    `node_modules/${TEMPLATE_NAME}/template.config`,
+    `${TEMPLATE_SOURCE_DIR}/node_modules/${TEMPLATE_NAME}/template.config`,
     () => ({
       placeholderName: 'someName',
       templateDir: 'someDir',
@@ -42,11 +44,12 @@ test('getTemplateConfig', () => {
   );
   jest.spyOn(path, 'resolve').mockImplementationOnce((...e) => e.join('/'));
 
-  expect(getTemplateConfig(TEMPLATE_NAME)).toEqual({
+  expect(getTemplateConfig(TEMPLATE_NAME, TEMPLATE_SOURCE_DIR)).toEqual({
     placeholderName: 'someName',
     templateDir: 'someDir',
   });
   expect(path.resolve).toHaveBeenCalledWith(
+    TEMPLATE_SOURCE_DIR,
     'node_modules',
     TEMPLATE_NAME,
     'template.config',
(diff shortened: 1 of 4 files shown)


grabbou:
Could you explain a bit the approach, that is, why "Installs template sources inside temporary directory and removes it after operation is done or failed" fixes it for mono-repositories, as opposed to the current solution? Want to have a good understanding before shipping.

thymikee:
Currently `init` works in a way that runs `yarn install` in `cwd` and then we copy-replace this directory with template contents. 

However, when you're inside a Yarn Workspace, this will cause Yarn to throw an error that you tried to add a dependency to a workspace which is not expected. We actually don't want to trigger installing template sources inside user project. So what we do in this PR is we install the template inside a temporary directory and copy to `cwd` from there, then remove the directory.

grabbou:
@thymikee now I understand. Still, I have a trouble understanding why would you like to put `ios` and `android` and other files that depend on `react-native` to the `cwd` that is a workspace root and doesn't have `react-native` in the `package.json`?

Shouldn't users simply not run `init` at the root?

thymikee:
The use case is to run init inside `packages/` dir not from workspace rood:
```
root/
  packages/
    project1/
    $ npx react-native init project2
```

grabbou:
Please bare with me as I missed few `init` improvements recently.

Why do we run `yarn install` in `cwd`, if the `cwd` should clearly be `./packages/project2` - that's where React Native dependency should be? That wouldn't case any monorepo issues.

thymikee:
After this PR the new init will work this way:

1. create `project2` directory in `./packages/`
1. fetch template to a temporary directory (we use `yarn install` to fetch it)
1. copy the template sources from temporary directory to `packages/project2`
1. install deps in `packages/project2` (second `yarn install`)
1. remove the temporary dir
1. done

Hope this makes it clearer?

grabbou:
Yes, thank you!

And which of the steps fails right now? How's the process going to be after this PR?

thymikee:
Step 2 was failing, because we were running `yarn install` in `packages/project2` and then swapping `package.json` with the one from template (by simple copy-replace). This caused Yarn to think that we want to add deps to the workspace and bail out.

PR #325, opened 2019-04-17
fix: prevent unnecessary actions if build failed

ifsnow:
Summary:
---------

```bash
react-native run-android --deviceId=emulator-5554
```

If the build fails, the APK that was built before is installed and executed.

```bash
BUILD FAILED in 5s
320 actionable tasks: 5 executed, 315 up-to-date
error Could not build the app, read the error above for details.
info Running /Users/snow/Library/Android/sdk/platform-tools/adb -s emulator-5554 reverse tcp:8081 tcp:8081
error Not found the correct install APK file!
Could not install the app on the device, read the error above for details.
info Starting the app on emulator-5554 (/Users/snow/Library/Android/sdk/platform-tools/adb -s emulator-5554 shell am start -n me.mycake/me.mycake.MainActivity)...
Starting: Intent { cmp=me.mycake/.MainActivity }
Warning: Activity not started, its current task has been brought to the front
```

I think this is unnecessary. I'm sure execution should be stopped so that we can immediately recognize and face on the error condition, as shown below.

```bash
BUILD FAILED in 4s
320 actionable tasks: 5 executed, 315 up-to-date
error Could not build the app, read the error above for details.
```

Test Plan:
----------

Not required.

### DIFF
--- a/packages/platform-android/src/commands/runAndroid/index.js
+++ b/packages/platform-android/src/commands/runAndroid/index.js
@@ -22,6 +22,7 @@ import {
   isPackagerRunning,
   logger,
   getDefaultUserTerminal,
+  CLIError,
 } from '@react-native-community/cli-tools';
 
 // Verifies this is an Android project
@@ -120,10 +121,7 @@ function runOnSpecificDevice(
   const devices = adb.getDevices(adbPath);
   if (devices && devices.length > 0) {
     if (devices.indexOf(args.deviceId) !== -1) {
-      if (!buildApk(gradlew)) {
-        return;
-      }
-
+      buildApk(gradlew);
       installAndLaunchOnDevice(
         args,
         args.deviceId,
@@ -152,13 +150,12 @@ function buildApk(gradlew) {
     execFileSync(gradlew, ['build', '-x', 'lint'], {
       stdio: [process.stdin, process.stdout, process.stderr],
     });
-
-    return true;
-  } catch (e) {
-    logger.error('Could not build the app, read the error above for details.');
+  } catch (error) {
+    throw new CLIError(
+      'Could not build the app, read the error above for details',
+      error,
+    );
   }
-
-  return false;
 }
 
 function tryInstallAppOnDevice(args, adbPath, device) {
--- a/packages/platform-android/src/commands/runAndroid/index.js
+++ b/packages/platform-android/src/commands/runAndroid/index.js
@@ -120,7 +120,10 @@ function runOnSpecificDevice(
   const devices = adb.getDevices(adbPath);
   if (devices && devices.length > 0) {
     if (devices.indexOf(args.deviceId) !== -1) {
-      buildApk(gradlew);
+      if (!buildApk(gradlew)) {
+        return;
+      }
+
       installAndLaunchOnDevice(
         args,
         args.deviceId,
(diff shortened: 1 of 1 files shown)


packages/platform-android/src/commands/runAndroid/index.js
```diff
@@ -146,9 +149,13 @@ function buildApk(gradlew) {
     execFileSync(gradlew, ['build', '-x', 'lint'], {
       stdio: [process.stdin, process.stdout, process.stderr],
     });
+
+    return true;
   } catch (e) {
     logger.error('Could not build the app, read the error above for details.');
```
thymikee: We could avoid returning a boolean from this function and verifying it inside `runOnSpecificDevice` by throwing the error instead of logging it. Like:
```js
throw new CLIError('Could not build the app, read the error above for details', e)
```

grabbou: Actually, we could do it even better.
```js
throw new CLIError('There was a problem building the app', originalErrorMessage)
```
and user would see the original error when running with `--verbose` flag.

We would have to stop piping stdout/stderr tho.

thymikee: @grabbou try it locally. The code inside CLIError only shows first 2 lines because it does `.split(0, 2)` (I missed that during code review). Feel free to send a fix separately from this PR :)

grabbou: How come: https://github.com/react-native-community/cli/blob/master/packages/tools/src/errors.js#L18 - the `split` should not run when string is passed, not an object.

grabbou: Anyway, we could not pipe `stderr` to `process.stderr` and simply pass it as a second argument to CLIError to show the huge error output in the verbose mode only (actually, we could do the same for stdout too, but this is getting out of scope for this PR)

thymikee: Oh, yea, it will work with a string message, you're right. I thought you were passing whole error (then where the stripping comes in place):

```
BUILD FAILED in 0s
error Failed to build the app with gradlew.
debug Error: Command failed: ./gradlew build -x lint
    at checkExecSyncError (child_process.js:616:11)
```
indstead of:
```
BUILD FAILED in 0s
error Failed to build the app with gradlew.
debug Error: Failed to build the app with gradlew.
    at buildApk (/Users/thymikee/Projects/oss/react-native-cli/packages/platform-android/build/commands/runAndroid/index.js:168:11)
    at runOnSpecificDevice (/Users/thymikee/Projects/oss/react-native-cli/packages/platform-android/build/commands/runAndroid/index.js:149:7)
    at buildAndRun (/Users/thymikee/Projects/oss/react-native-cli/packages/platform-android/build/commands/runAndroid/index.js:135:14)
    at then.result (/Users/thymikee/Projects/oss/react-native-cli/packages/platform-android/build/commands/runAndroid/index.js:102:12)
    at process._tickCallback (internal/process/next_tick.js:68:7)
```

ifsnow: Thanks for your feedbacks. I really agree with using exception. I didn't know the changes in the 2.0 branch.

PR #327, opened 2019-04-17
chore: use jest-circus

thymikee:
Summary:
---------

Just playing around, maybe this will cause flaky test to go away, but if not, with Circus we can retry test a bunch of times.


grabbou:
Really tempted to figure out why the `fs` fails (we are working on a real file system, so it shouldn't happen). Otherwise, we could mock it to `MemoryFS` and use that one instead? This way of mocking file system used to be working for a while.

thymikee:
Closing, this doesn't change anything.

PR #332, opened 2019-04-18
Check `ios-deploy` installation before proceeding to build on device

### LINKED ISSUE #330: UX: throw error about `ios-deploy` not being installed earlier in the process
## Describe the Feature

For some reason I always get errors with `ios-deploy` not being installed so I have to install it again, right now I have to go through these steps to know that it's not installed:

- Run `run-ios` with `--device` argument;
- Wait a few minutes for the build;
- See the error;
- Install `ios-deploy`;
- Rerun `run-ios`.

This could be improved to show the error to the user as soon as it sees that the `--device` argument is fulfilled and there's no `ios-deploy`.

## Possible Implementations

Run `which ios-deploy` as soon as it sees that there's a `--device` or a `--udid` that matches a device, if it outputs something continue, throw the error if not.

lucasbento:
Summary:
---------

This PR fixes #330, it checks if `ios-deploy` is installed or not before it actually starts building.

Test Plan:
----------

- Run `npm uninstall -g ios-deploy`;
- Run the CLI with `--device` argument and wait for it to fail;
- Run `npm i -g ios-deploy`;
- Run the CLI again and see it working.

cc @thymikee

---

### Error about `ios-deploy` not installed
image

### Random error when executing `ios-deploy`
image

### DIFF
--- a/packages/platform-ios/src/commands/runIOS/index.js
+++ b/packages/platform-ios/src/commands/runIOS/index.js
@@ -191,7 +191,7 @@ async function runOnSimulator(xcodeProject, scheme, args: FlagsT) {
 
 async function runOnDevice(selectedDevice, scheme, xcodeProject, args: FlagsT) {
   const isIOSDeployInstalled = child_process.spawnSync(
-    'ioxs-deploy',
+    'ios-deploy',
     ['--version'],
     {encoding: 'utf8'},
   );
--- a/packages/platform-ios/src/commands/runIOS/index.js
+++ b/packages/platform-ios/src/commands/runIOS/index.js
@@ -191,16 +191,17 @@ async function runOnSimulator(xcodeProject, scheme, args: FlagsT) {
 
 async function runOnDevice(selectedDevice, scheme, xcodeProject, args: FlagsT) {
   const isIOSDeployInstalled = child_process.spawnSync(
-    'ios-deploy',
+    'ioxs-deploy',
     ['--version'],
     {encoding: 'utf8'},
   );
 
   if (isIOSDeployInstalled.error) {
-    throw new Error(`Failed to install the app on the device because we couldn't execute the "ios-deploy" command.
-      Please install it by running "${chalk.bold(
+    throw new CLIError(
+      `Failed to install the app on the device because we couldn't execute the "ios-deploy" command. Please install it by running "${chalk.bold(
         'npm install -g ios-deploy',
-      )}" and try again.`);
+      )}" and try again.`,
+    );
   }
 
   const appName = await buildProject(
@@ -227,11 +228,14 @@ async function runOnDevice(selectedDevice, scheme, xcodeProject, args: FlagsT) {
   );
 
   if (iosDeployOutput.error) {
-    throw new Error(`Failed to install the app on the device. We've encountered an error in "ios-deploy" command.
-      Here are more details: ${isIOSDeployInstalled.error.message}`);
+    throw new CLIError(
+      `Failed to install the app on the device. We've encountered an error in "ios-deploy" command: ${
+        iosDeployOutput.error.message
+      }`,
+    );
   }
 
-  return logger.info('App installed on device.');
+  return logger.success('Installed the app on the device.');
 }
 
 function buildProject(xcodeProject, udid, scheme, args: FlagsT) {
--- a/packages/platform-ios/src/commands/runIOS/index.js
+++ b/packages/platform-ios/src/commands/runIOS/index.js
@@ -190,6 +190,18 @@ async function runOnSimulator(xcodeProject, scheme, args: FlagsT) {
 }
 
 async function runOnDevice(selectedDevice, scheme, xcodeProject, args: FlagsT) {
+  const isIOSDeployInstalled = child_process.spawnSync(
+    'ios-deploy',
+    ['--version'],
+    {encoding: 'utf8'},
+  );
+
+  if (isIOSDeployInstalled.error) {
+    return logger.error(
+      '** INSTALLATION FAILED **\nMake sure you have ios-deploy installed globally.\n(e.g "npm install -g ios-deploy")',
+    );
+  }
+
   const appName = await buildProject(
     xcodeProject,
     selectedDevice.udid,
@@ -214,12 +226,12 @@ async function runOnDevice(selectedDevice, scheme, xcodeProject, args: FlagsT) {
   );
 
   if (iosDeployOutput.error) {
-    logger.error(
-      '** INSTALLATION FAILED **\nMake sure you have ios-deploy installed globally.\n(e.g "npm install -g ios-deploy")',
+    return logger.error(
+      '** INSTALLATION FAILED **\nThere was an internal error with ios-deploy.',
     );
-  } else {
-    logger.info('** INSTALLATION SUCCEEDED **');
   }
+
+  return logger.info('** INSTALLATION SUCCEEDED **');
 }
 
 function buildProject(xcodeProject, udid, scheme, args: FlagsT) {
--- a/packages/platform-ios/src/commands/runIOS/index.js
+++ b/packages/platform-ios/src/commands/runIOS/index.js
@@ -197,9 +197,10 @@ async function runOnDevice(selectedDevice, scheme, xcodeProject, args: FlagsT) {
   );
 
   if (isIOSDeployInstalled.error) {
-    return logger.error(
-      '** INSTALLATION FAILED **\nMake sure you have ios-deploy installed globally.\n(e.g "npm install -g ios-deploy")',
-    );
+    throw new Error(`Failed to install the app on the device because we couldn't execute the "ios-deploy" command.
+      Please install it by running "${chalk.bold(
+        'npm install -g ios-deploy',
+      )}" and try again.`);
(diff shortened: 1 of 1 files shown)


packages/platform-ios/src/commands/runIOS/index.js
```diff
@@ -214,12 +226,12 @@ async function runOnDevice(selectedDevice, scheme, xcodeProject, args: FlagsT) {
   );
 
   if (iosDeployOutput.error) {
-    logger.error(
-      '** INSTALLATION FAILED **\nMake sure you have ios-deploy installed globally.\n(e.g "npm install -g ios-deploy")',
+    return logger.error(
+      '** INSTALLATION FAILED **\nThere was an internal error with ios-deploy.',
```
lucasbento: @thymikee: I checked if there was a way of providing a more helpful message but as per what I saw there is a lot of stuff inside of error, I'm open for ideas.

thymikee: Do we have to yell like this? I think simple:
```
Failed to install the app on the device. We've encountered an error in "ios-deploy" command. Here's more details: ${error}
```

Also, let's throw because by returning the exit code is 0, which is unexpected.

packages/platform-ios/src/commands/runIOS/index.js
```diff
@@ -214,12 +226,12 @@ async function runOnDevice(selectedDevice, scheme, xcodeProject, args: FlagsT) {
...
     );
-  } else {
-    logger.info('** INSTALLATION SUCCEEDED **');
   }
+
+  return logger.info('** INSTALLATION SUCCEEDED **');
```
thymikee:
```suggestion
  return logger.success('Installed the app on the device');
```

packages/platform-ios/src/commands/runIOS/index.js
```diff
@@ -190,6 +190,18 @@ async function runOnSimulator(xcodeProject, scheme, args: FlagsT) {
...
 async function runOnDevice(selectedDevice, scheme, xcodeProject, args: FlagsT) {
+  const isIOSDeployInstalled = child_process.spawnSync(
+    'ios-deploy',
+    ['--version'],
+    {encoding: 'utf8'},
+  );
+
+  if (isIOSDeployInstalled.error) {
+    return logger.error(
+      '** INSTALLATION FAILED **\nMake sure you have ios-deploy installed globally.\n(e.g "npm install -g ios-deploy")',
```
thymikee: 1. use `throw CLIError` so that the exit code is not 0.
2. change message to: 

>Failed to install the app on the device because we couldn't execute the "ios-deploy" command. Please install it from the npm registry (e.g. "npm install -g ios-deploy") and try again

packages/platform-ios/src/commands/runIOS/index.js
lucasbento: I also don't really get the deal with the `** INSTALLATION ... **` message, I'll change it.

thymikee:
Side note: I was also thinking about using `npx ios-deploy` instead of letting users install global, but in my local testing it doesn't seem to cache it properly so it slows down the command by a few seconds _every_ time. So I'm more of a fan of installing a global this time. But, only until Yarn v2 is released with `yarn dlx` which supports proper caching. Then we won't need to bother users with installing a global binary :)

packages/platform-ios/src/commands/runIOS/index.js
lucasbento: @thymikee: what do you think about only doing the error details part with `logger.debug`?

thymikee: It's an unexpected error, we don't know what's gonna be inside and we shouldn't expect users to run it and build once again with `--verbose`, but show the message straight away

packages/platform-ios/src/commands/runIOS/index.js
lucasbento: I'm doing:

```js
  const {error} = iosDeployOutput;

  if (error) {
    throw new CLIError(`
      Failed to install the app on the device. We've encountered an error in "ios-deploy" command.
      Here are more details: ${error}.
    `);
  }
```

And Flow is throwing an error saying "Cannot coerce `error` to string because  `Error` [1] should not be coerced.Flow(InferError)", do you have any ideia of how to fix this @thymikee?

lucasbento:
@thymikee: didn't know about `yarn dlx`, awesome!

packages/platform-ios/src/commands/runIOS/index.js
thymikee: yea, just pass `error.stack`, it has message and a stack trace inside

thymikee: oh and btw, once you're done with adjustments, please post screenshots to the OP so we can see how they improved :)

lucasbento: With the `error.stack` the stacktrace won't look very useful at first glance:

image

What do you mean by _the OP_?

lucasbento: If you meant to pass the `error` as the second argument to `CLIError` (just got that from `runAndroid` then the error is prettified but only shows up with `--verbose`:

image

thymikee: By OP I mean Original Post (the PR summary you posted).

Just to be clear, we only want to show full error message straight away if unexpected error happens, not when `ios-deploy` is not installed (there we can put error in verbose mode).

CLIError works in this funny way that it will strip all newlines. Just use regular `throw Error()` and it will work as expected :)

lucasbento:
@thymikee: just did the changes, can you verify again?

Thanks for this 🙌

packages/platform-ios/src/commands/runIOS/index.js
```diff
@@ -190,6 +190,19 @@ async function runOnSimulator(xcodeProject, scheme, args: FlagsT) {
...
 async function runOnDevice(selectedDevice, scheme, xcodeProject, args: FlagsT) {
+  const isIOSDeployInstalled = child_process.spawnSync(
+    'ios-deploy',
+    ['--version'],
+    {encoding: 'utf8'},
+  );
+
+  if (isIOSDeployInstalled.error) {
+    throw new Error(`Failed to install the app on the device because we couldn't execute the "ios-deploy" command.
+      Please install it by running "${chalk.bold(
```
thymikee: we generally don't do indentations in our errors, unless it's something very specific. So in this case it looks like you could use `CLIError` :D

packages/platform-ios/src/commands/runIOS/index.js
```diff
@@ -214,12 +227,11 @@ async function runOnDevice(selectedDevice, scheme, xcodeProject, args: FlagsT) {
   );
 
   if (iosDeployOutput.error) {
-    logger.error(
-      '** INSTALLATION FAILED **\nMake sure you have ios-deploy installed globally.\n(e.g "npm install -g ios-deploy")',
-    );
-  } else {
-    logger.info('** INSTALLATION SUCCEEDED **');
+    throw new Error(`Failed to install the app on the device. We've encountered an error in "ios-deploy" command.
+      Here are more details: ${isIOSDeployInstalled.error.message}`);
```
thymikee: good call with extracting a message. Few improvements:
1. use CLIError (because now it makes sense)
2. change `isIOSDeployInstalled.error.message` to `iosDeployOutput.error.message`
3. change `error in "ios-deploy" command. Here are more details: {message}` to just `error in "ios-deploy"  command: {message}` so it's more concise

packages/platform-ios/src/commands/runIOS/index.js
```diff
@@ -214,12 +227,11 @@ async function runOnDevice(selectedDevice, scheme, xcodeProject, args: FlagsT) {
...
   }
+
+  return logger.info('App installed on device.');
```
thymikee:
```suggestion
  return logger.success('Installed the app on the device');
```

lucasbento: Any reason for omitting the `.`? good point on `success` ✅

thymikee: Nope, let's leave it. Would be cool to set a convention on when we set a dot and when not :D

lucasbento:
@thymikee: all fixed :)

packages/platform-ios/src/commands/runIOS/index.js
```diff
@@ -190,6 +190,20 @@ async function runOnSimulator(xcodeProject, scheme, args: FlagsT) {
 }
 
 async function runOnDevice(selectedDevice, scheme, xcodeProject, args: FlagsT) {
+  const isIOSDeployInstalled = child_process.spawnSync(
+    'ioxs-deploy',
```
thymikee: whops, typo!

lucasbento: aaaaaaa, damn it.

lucasbento: thanks for seeing this!

lucasbento: ready to go 🚀

PR #333, opened 2019-04-18
Support functions in default

grabbou:
Summary:
---------

I was pretty sure this has landed, but turns out - I've never merged the PR: https://github.com/react-native-community/cli/pull/287/files#diff-97059d274ee5d7bbcc5730ddee2a98daR126 

This change is needed to support functions as defaults (lazily called with config to get a default value) - we will need that in other commands in the future.

I am also moving current `getDefaultUserTerminal` to be executed lazily. That way, the value of `default` is a function and it passes the validation.


grabbou:
iOS failure seems to be unrelated.

grabbou:
I cancelled PR, because it had to be rebased.

PR #335, opened 2019-04-18
feat: support platforms that ship with project 

grabbou:
Summary:
---------

React Native (and other packages) can define commands and platforms to be picked by the CLI. When running from source, they are treated as `project`, not as a `dependency`, hence `platforms` are ignored.

This adds support for `platforms` to be defined on per `project` basis too.

Added `stripUnknown` so they can co-exist.

TLDR; Nothing changes, but `Joi` validator will not strip it out now.

### DIFF
--- a/packages/cli/src/tools/config/index.js
+++ b/packages/cli/src/tools/config/index.js
@@ -124,7 +124,7 @@ function loadConfig(projectRoot: string = process.cwd()): ConfigT {
       get assets() {
         return findAssets(projectRoot, userConfig.assets);
       },
-      platforms: {},
+      platforms: userConfig.platforms,
       haste: {
         providesModuleNodeModules: [],
         platforms: [],
--- a/types/index.js
+++ b/types/index.js
@@ -195,7 +195,7 @@ export type UserDependencyConfigT = {
  */
 export type UserConfigT = {
   /**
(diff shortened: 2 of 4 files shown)


types/index.js
```diff
@@ -198,7 +198,7 @@ export type UserConfigT = {
    * Shares some structure with ConfigT, except that haste, root, platforms
    * are calculated and can't be defined
    */
-  ...$Diff<ConfigT, {haste: any, root: any, platforms: any}>,
```
thymikee: Update the comment as well

packages/cli/src/tools/config/index.js
```diff
@@ -124,7 +124,7 @@ function loadConfig(projectRoot: string = process.cwd()): ConfigT {
       get assets() {
         return findAssets(projectRoot, userConfig.assets);
       },
-      platforms: {},
+      platforms: userConfig.platforms,
```
thymikee: This makes sense, why we didn't do it previously?
Btw shouldn't this fall back to empty object?

grabbou: Joi schema automatically makes sure it's empty object by default. So we don't need to.

I believe it was there (we did it for commands). Could've been lost in refactoring.

types/index.js
grabbou: Yup

PR #337, opened 2019-04-19
fix: modify wrong install arguments

ifsnow:
Summary:
---------

In the previous PR(#323), the code was modified with the following comment.

```
@Krizzu
You can also combine arguments -rd, but it's up to you
```

But, I realized that this arguments does not work.

```bash
adb -s ce021712cad7e8990c install -rd test.apk
adb: failed to install test.apk:
Exception occurred while dumping:
java.lang.IllegalArgumentException: No argument expected after "-rd"
```

So, I modified it to divide the arguments.

Test Plan:
----------

Not required.


thymikee:
FYI, this is released in 1.9.2

PR #341, opened 2019-04-20
fix: stop resolving "react-native" from package.json as a config 

thymikee:
Summary:
---------

`"react-native"` key already exists in userspace and is used by Metro to resolve source files of published React Native modules. That's why we cannot use it to store config values insead of `"rnpm"`. I'm not a fan of having configs in package.json, so I got rid of that entirely and rely only on `react-native.config.js` file from now on. 

Test Plan:
----------

Added a test and updated existing ones.

### DIFF
--- a/packages/cli/src/tools/config/__tests__/index-test.js
+++ b/packages/cli/src/tools/config/__tests__/index-test.js
@@ -23,16 +23,15 @@ const removeString = (config, str) =>
 beforeEach(() => {
   cleanup(DIR);
   jest.resetModules();
+  jest.clearAllMocks();
 });
 
 afterEach(() => cleanup(DIR));
 
 test('should have a valid structure by default', () => {
   writeFiles(DIR, {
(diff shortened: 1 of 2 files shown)


grabbou:
Good idea. Already did that in the other PR, but never merged it. I think `package.json` config wouldn't work out anyway since some properties need to be defined as functions with `json` doesn't support.

grabbou:
BTW, is there any reference on what that config is and how they use it?

grabbou:
Mind investigating whether `init` test failure is due to changes in this PR or just flakiness of fs?

thymikee:
Will investigate the failure once I have some time :)

thymikee:
@grabbou see https://github.com/react-native-community/bob/blob/master/README.md#usage on how it's used

thymikee:
@grabbou fixed the test (missing file that got deleted in f1105bc)

PR #345, opened 2019-04-23
docs: initial documentation of the basic features

grabbou:
Summary:
---------

I am trying to describe some of the new (and old, but undocumented) features of the CLI to make it easier for developers to use CLI and hack with its great features.

This PR is a WIP as I am working on it. 

ETA is next 2 to 3 hours - I want to finish and ship it before next RC of React Native so that we can announce this at the same time.

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -21,13 +21,16 @@ module.exports = {
 };
 ```
 
+> Above is an example of a plugin that exports a command named `foo-command` that can be executed with `react-native foo-command` and logs "It worked" and exits. 
+
+
 At the startup, React Native CLI reads configuration from all dependencies listed in `package.json` and reduces them into a single configuration. 
 
 At the end, an array of commands concatenated from all plugins is passed on to the CLI to be loaded after built-in commands.
 
 ## Command interface
 
-```js
+```ts
 type Command = {
   name: string,
   description?: string,
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -134,31 +134,18 @@ provided that `./path-to-commands.js` returns an array of commands.
 
 ### Renaming command options
 
-If your command accepts options, rename `command` to `name`.
+If your command accepts options, rename `command` property of each of them to `name`.
 
-For example,
-```js
-module.exports = {
-  name: 'foo',
-  func: () => console.log('My work'),
-  options: [
-    {
-      command: '--reset-cache, --resetCache',
-      description: 'Removes cached files',
-    }
-  ]
-}
-```
-should be changed to:
-```js
-module.exports = {
-  name: 'foo',
-  func: () => console.log('My work'),
-  options: [
-    {
-      name: '--reset-cache, --resetCache',
-      description: 'Removes cached files',
-    }
-  ]
-}
+```diff
+ module.exports = {
+   name: 'foo',
+   func: () => console.log('My work'),
+   options: [
+     {
+-      command: '--reset-cache, --resetCache',
++      name: '--reset-cache, --resetCache',
+       description: 'Removes cached files',
+     }
+   ]
+ }
 ```
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -27,7 +27,31 @@ At the end, an array of commands concatenated from all plugins is passed on to t
 
 ## Command interface
 
-#### `name: string`
+```js
+type Command = {
+  name: string,
+  description?: string,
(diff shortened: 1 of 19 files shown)


docs/plugins.md
```diff
@@ -1,3 +1,155 @@
...
+## How does it work?
+
+Except for React Native dependencies, where configuration is implicit, each package needs to have a `react-native.config.js` at the root folder in order to be discovered by the CLI as a plugin.
+
+```js
+module.exports = {
+  commands: [
+    {
+      name: 'foo-command',
+      func: () => console.log('It worked')
```
thymikee: provide function interface in a comment or so, it's nice to know what arguments you have available

docs/plugins.md
```diff
@@ -1,3 +1,155 @@
...
+
+#### `name: string`
+
+Name that will be used in order to run the command. 
+
+Note: If you want your command to accept additional arguments, make sure to include them in the name.
+
+For example, `my-command <argument>` will require an argument to be provided and will throw a validation error otherwise. Alternatively, `my-command [argument]` will accept an argument, but will not throw when run without it. In that case, make sure to check for its presence.
+
+#### `func: (argv: string[], config: ConfigT, options: Object) => ?Promise<void>`
```
thymikee: I think it would be more readable to move the type to a code block and leave the section title to be just `#### func`

grabbou: I don't know. Feels to me that referring to "last argument" or "first argument" in that case would be less understandable than it is right now.

docs/plugins.md
grabbou: What do you mean?

thymikee: nevermind, I wrote this before going further

docs/plugins.md
```diff
@@ -1,3 +1,159 @@
...
+
+String that describes this particular usage.
+
+##### `cmd: string`
+
+A command with arguments and options (if applicable) that can be run in order to achieve the desired goal.
+
+## Migrating from `rnpm` configuration
+
+The changes are non-breaking hence the migration should be straight-forward.
```
thymikee:
```suggestion
The changes are mostly cosmetic so the migration should be pretty straight-forward. 
```

docs/plugins.md
grabbou: either way, I did it.

docs/plugins.md
```diff
@@ -1,3 +1,161 @@
...
+For example, `react-native-windows` package is a plugin that provides `react-native run-windows` command and `windows` platform.
+
+Details of this particular integration as well as how to provide an additional platform for React Native were described in a [`dedicated section`](./platforms.md) about platforms.
+
+## How does it work?
+
+Except for React Native dependencies, where configuration is implicit, each package needs to have a `react-native.config.js` at the root folder in order to be discovered by the CLI as a plugin.
+
+```js
+module.exports = {
```
thymikee: How about adding a short description to this file? 

> Below is an example of a project that exports a command named `foo-command` that logs "It worked" and exits.

It feels out of place without it.

docs/plugins.md
```diff
@@ -1,3 +1,161 @@
...
+};
+```
+
+At the startup, React Native CLI reads configuration from all dependencies listed in `package.json` and reduces them into a single configuration. 
+
+At the end, an array of commands concatenated from all plugins is passed on to the CLI to be loaded after built-in commands.
+
+## Command interface
+
+```js
```
thymikee: consider using `ts` to have better types highlighting

docs/plugins.md
```diff
@@ -1,3 +1,161 @@
...
+    cmd: string,
+  }>,
+};
+```
+
+> Note: `ConfigT` is described in [`configuration` section](./configuration.md)
+
+#### `name`
+
+Name that will be used in order to run the command.
```
thymikee:
```suggestion
A name that will be used in order to run the command. 
```

docs/plugins.md
```diff
@@ -1,3 +1,161 @@
...
+  func: () => console.log('My work'),
+  options: [
+    {
+      command: '--reset-cache, --resetCache',
+      description: 'Removes cached files',
+    }
+  ]
+}
+```
+should be changed to:
```
thymikee: I think it would be way easier to digest if we did a simple diff:

```diff
 module.exports = {
   name: 'foo',
   func: () => console.log('My work'),
   options: [
     {
-      command: '--reset-cache, --resetCache',
+      name: '--reset-cache, --resetCache',
       description: 'Removes cached files',
     }
   ]
 }
```

PR #347, opened 2019-04-23
docs: add init documentation

Esemesek:
Adding documentation for `init` command.

### DIFF
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -123,46 +123,7 @@ react-native init <projectName> [options]
 
 Initialize new React Native project.
 
-#### Options
-
-#### `--version [string]`
-
-Uses a valid semver version of React Native as a template.
-
-#### `--template [string]`
-
-Uses a custom template. Accepts either an npm package name or an absolute path to local directory.
-
-Example:
-
-```sh
-react-native init MyApp --template react-native-custom-template
-react-native init MyApp --template file:///Users/name/template-path
-```
-
-A template is any directory or npm package that contains a `template.config.js` file in the root with following of the following type:
-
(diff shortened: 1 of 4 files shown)


docs/init.md
```diff
@@ -0,0 +1,80 @@
...
+```
+
+> Note: If you have both `yarn` and `npm` installed on your machine, React Native CLI will always try to use `yarn`, so even if you use `npx` utility, only `react-native` executable will be installed using `npm` and the rest of the work will be delegated to `yarn`. You can force usage of `npm` adding `--npm` flag to the command.
+
+> Note: for Yarn users, `yarn dlx` command similar to `npx` will be featured in Yarn 2.0: https://github.com/yarnpkg/berry/pull/40 so we’ll be able to use it in a similar fashion.
+
+Installing `react-native` and invoking `init` command:
+
+```sh
+yarn add react-native && react-native init ProjectName
```
thymikee:
```suggestion
yarn init && yarn add react-native && yarn react-native init ProjectName
```

docs/commands.md
```diff
@@ -123,46 +123,7 @@ react-native init <projectName> [options]
...
-
-```ts
-type Template = {
-  // Placeholder used to rename and replace in files
-  // package.json, index.json, android/, ios/
-  placeholderName: string;
-  // Directory with template
-  templateDir: string;
-  // Path to script, which will be executed after init
-  postInitScript?: string;
```
thymikee: I liked that type for config was defined, can you add it back to init.md docs?

Esemesek: Sure

PR #348, opened 2019-04-23
chore: change `upgrade` to use new diff files approach

pvinis:
Summary:
---------

We were relying on github to generate diffs and patches for us. This worked fine when we had one branch for all the releases, but this is not enough anymore with a branch per release. We still use github's compare view, but for patches we grab the ones we generate on the [purge repo](https://github.com/react-native-community/rn-diff-purge).

Test Plan:
----------

More or less the same as https://github.com/react-native-community/react-native-cli/pull/176.
Create a RN project, and try upgrading like so.
Screen Shot 2019-04-23 at 20 21 29

Having conflicts still causes the upgrade to not leave files for the user to resolve.

### DIFF
--- a/packages/cli/src/commands/upgrade/__tests__/upgrade.test.js
+++ b/packages/cli/src/commands/upgrade/__tests__/upgrade.test.js
@@ -159,16 +159,16 @@ test('fetches empty patch and installs deps', async () => {
   (fetch: any).mockImplementation(() => Promise.resolve(''));
   await upgrade.func([newVersion], ctx, opts);
   expect(flushOutput()).toMatchInlineSnapshot(`
-"info Fetching diff between v0.57.8 and v0.58.4...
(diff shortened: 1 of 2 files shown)


thymikee:
> Having conflicts still causes the upgrade to not leave files for the user to resolve.

This depends on how severe the conflicts are. I'm still not sure how to prevent git from bailing on more complex diffs (we're not throwing out changes, it's that git doesn't output them).

Hopefully we'll get past most of the conflicts once we settle with autolinking as a standard used by everyone (will take couple of months though)

pvinis:
One thing I tested was to have a `.babelrc` with an extra `plugin` for example. When the new `babel.config.js` is created instead, then the user should know of the conflict so that they can replicate their config in the js file now. So conflicts will happen and should stay in more cases than just the linking. Autolinking will help, but it's not the only place.

PR #349, opened 2019-04-23
feat: add `platforms` doc

grabbou:
Summary:
---------

Adds a basic description of a platform documentation

### DIFF
--- a/docs/platforms.md
+++ b/docs/platforms.md
@@ -1,3 +1,119 @@
 # Platforms
 
-TBD
+A platform is a React Native package that enables writing and shipping React Native applications to a new target. 
+
+For example, React Native Windows is a platform, because it allows to run React Native apps on Windows. At the same time, React Native itself is also a platform - it allows to run React Native apps on Android, iOS and tvOS.
+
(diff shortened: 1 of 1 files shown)


docs/platforms.md
```diff
@@ -1,3 +1,119 @@
 # Platforms
 
-TBD
+A platform is a React Native package that enables writing and shipping React Native applications to a new target. 
+
+For example, React Native Windows is a platform, because it allows to run React Native apps on Windows. At the same time, React Native itself is also a platform - it allows to run React Native apps on Android, iOS and tvOS.
```
thymikee: not sure if we should mention `tvOS`?

docs/platforms.md
```diff
@@ -1,3 +1,119 @@
...
+
+Each platform can have an additional configuration for the CLI to enable bundling apps and linking  and packages for targets it provides.
```
thymikee:
```suggestion
Each platform can have an additional configuration for the CLI to enable bundling apps and linking packages for targets it provides.
```

docs/platforms.md
grabbou: It is there right now :D

PR #350, opened 2019-04-23
feat: add dependency and project documentation

grabbou:
Summary:
---------

This is incomplete - working on project docs.

Note: I didn't document all properties on purpose - this is MVP for now. Also, a lot of these will be gone once `autolinking` goes stable.


docs/dependency.md
```diff
@@ -1,3 +1,152 @@
...
+    hooks: {
+      prelink: './path-to-a-postlink-hook'
+    }
+  }
+};
+```
+
+### Asking for params while linking has been removed
+
+If your library needs it, do not upgrade over to the new config format.
```
thymikee: Do we provide any alternative?

grabbou: Yes and is listed below (custom solution). I don't think this was super useful and is actually going to be harder to maintain in the future.

grabbou:
@thymikee, please take another look. I've just added `project` documentation, which is the last one before I got to the short summary under `configuration`.

docs/project.md
```diff
@@ -2,4 +2,154 @@
 
 A project is a an app that contains React code and has a dependency on `react-native`.
```
thymikee: "a an" -> "an"

thymikee: maybe make it explicit, that project is usually the app you build

docs/project.md
```diff
@@ -2,4 +2,154 @@
...
+    },
+  },
+  commands: CommandT[]
+};
+```
+
+### reactNativePath
+
+A custom path to React Native, in case `require('react-native')` would throw. Useful when running
+React Native from a (custom) source.
```
thymikee: I wonder do we still need this option? I can imagine `yarn link` may fail but otherwise?

docs/project.md
```diff
@@ -2,4 +2,154 @@
...
+      project: './CustomProject.xcodeproj'
+    }
+  },
+  assets: ['./assets']
+};
+```
+
+You can check all available options below.
+
+## Project interface
```
thymikee: It's pretty confusing that we name it Project, while the type is User Config and contains "project" key inside, isn't it?

docs/project.md
grabbou: Well, we need it to run RNTester because there's no `react-native` package - our default resolver will not work. I wanted to provide a gentle solution to this instead of providing half-baked support.

docs/project.md
grabbou: Right. On the other hand, it works well with other naming patterns we have (dependency, plugin, platform).

grabbou: Renamed the type to start with.

docs/configuration.md
```diff
@@ -1,3 +1,15 @@
...
+
+> Note: Configuring CLI used to be possible via `rn-cli.config.js` (that has been renamed to `metro.config.js`) and never documented `rnpm` entry on the `package.json`. We have provided migration guides when possible.
+
+React Native CLI can be configured by creating a `react-native.config.js` at the root of the project. Depending on a type of a package, this file can have a different set properties.
+
+Check the documentation for:
+- [project](./project.md)
+- [dependency](./dependency.md)
+- [platform](./platforms.md)
+- [plugin](./plugins.md)
```
thymikee:
```suggestion
- [plugins](./plugins.md)
```

docs/configuration.md
```diff
@@ -1,3 +1,15 @@
...
+React Native CLI has a configuration mechanism that allows changing its behavior and providing additional features.
+
+> Note: Configuring CLI used to be possible via `rn-cli.config.js` (that has been renamed to `metro.config.js`) and never documented `rnpm` entry on the `package.json`. We have provided migration guides when possible.
+
+React Native CLI can be configured by creating a `react-native.config.js` at the root of the project. Depending on a type of a package, this file can have a different set properties.
+
+Check the documentation for:
+- [project](./project.md)
+- [dependency](./dependency.md)
+- [platform](./platforms.md)
```
thymikee:
```suggestion
- [platforms](./platforms.md)
```

grabbou:
@thymikee addressed some of your comments and also, added main configuration document. I am hoping this is the end of the journey for today.

grabbou:
@thymikee FYI I moved project and dependency to projects and dependencies too to align with the naming.

PR #353, opened 2019-04-24
feat: improve copying binary files in `init`

Esemesek:
Summary:
---------

Since we plan to add `Pods` to the template, we need to improve handling binary files.

Test Plan:
----------

1. Initialise new project.
2. Check if `*.png` files are not corrupted.
3. Check if `gradle wrapper` works by invoking `./gradlew clean` from `android` directory.

cc @cpojer @thymikee

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -39,6 +39,7 @@
     "glob": "^7.1.1",
     "graceful-fs": "^4.1.3",
     "inquirer": "^3.0.6",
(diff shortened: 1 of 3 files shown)


cpojer:
This broke copying the `.keystore` file as binary again and requires that the `debug.keystore` file on Android be manually copied which is a really bad experience. Could you revert this change and add a test to ensure this doesn't keep breaking?

PR #363, opened 2019-04-29
feat: copy all files in `init` through streams

thymikee:
Summary:
---------

Copy all files during `init` using read/write streams so it doesn't matter if it's a binary file or not.
Reverts #353

Test Plan:
----------

doing it

### DIFF
--- a/packages/cli/src/commands/init/__tests__/editTemplate.test.js
+++ b/packages/cli/src/commands/init/__tests__/editTemplate.test.js
@@ -16,20 +16,28 @@ const FIXTURE_DIR = path.resolve(
 const PLACEHOLDER_NAME = 'PlaceholderName';
 const PROJECT_NAME = 'ProjectName';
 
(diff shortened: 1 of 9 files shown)


packages/cli/src/commands/init/__tests__/editTemplate.test.js
```diff
@@ -74,6 +82,4 @@ test('should edit template', () => {
   expect(
     snapshotDiff(fixtureTree, transformedTree, {contextLines: 1}),
   ).toMatchSnapshot();
-
-  fs.removeSync(testPath);
```
thymikee: FYI @Esemesek it's not safe to do so, because cleanup fn won't be called after failure

grabbou:
Thanks!

Pods are copied now, but `Pods/Headers` is missing (contains symlinks inside). Needs to be fixed before we ship this.

PR #364, opened 2019-04-30
fix: refactor init for less duplication; fix handling versioned templates

thymikee:
Summary:
---------

There's some duplication around the logic of creating from template. This PR aims to resolve it.

@Esemesek please have a close look at this one, especially around installing from version

Test Plan:
----------

Green CI.

### DIFF
--- a/packages/cli/src/commands/init/init.js
+++ b/packages/cli/src/commands/init/init.js
@@ -39,25 +39,45 @@ function adjustNameIfUrl(name, cwd) {
   return name;
 }
 
-async function createFromExternalTemplate(
+async function createFromTemplate({
+  projectName,
+  templateName,
+  version,
+  npm,
+}: {
   projectName: string,
   templateName: string,
+  version?: string,
   npm?: boolean,
-) {
-  logger.debug('Initializing new project from external template');
+}) {
+  logger.debug('Initializing new project');
   logger.log(banner);
-
   const Loader = getLoader();
-
   const loader = new Loader({text: 'Downloading template'});
+
   loader.start();
 
   const templateSourceDir = fs.mkdtempSync(
     path.join(os.tmpdir(), 'rncli-init-template-'),
   );
   try {
-    let {uri, name} = await processTemplateName(templateName);
+    let uri, name;
+    if (version) {
+      if (semver.valid(version) && !semver.gte(version, '0.60.0')) {
+        throw new Error(
+          'Cannot use React Native CLI to initialize project with version lower than 0.60.0',
+        );
+      }
+      const processed = await processTemplateName(`${templateName}@${version}`);
+      uri = processed.uri;
+      name = templateName;
+    } else {
+      const processed = await processTemplateName(templateName);
+      uri = processed.uri;
+      name = processed.name;
+    }
     await installTemplatePackage(uri, templateSourceDir, npm);
+
     loader.succeed();
     loader.start('Copying template');
 
@@ -66,7 +86,7 @@ async function createFromExternalTemplate(
     await copyTemplate(name, templateConfig.templateDir, templateSourceDir);
 
     loader.succeed();
-    loader.start('Preparing template');
+    loader.start('Processing template');
 
     changePlaceholderInTemplate(projectName, templateConfig.placeholderName);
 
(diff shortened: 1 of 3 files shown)


packages/cli/src/commands/init/init.js
```diff
@@ -39,25 +39,45 @@ function adjustNameIfUrl(name, cwd) {
...
   try {
-    let {uri, name} = await processTemplateName(templateName);
+    let uri, name;
+    if (version) {
+      if (semver.valid(version) && !semver.gte(version, '0.60.0')) {
+        throw new Error(
+          'Cannot use React Native CLI to initialize project with version lower than 0.60.0',
+        );
+      }
+      const processed = await processTemplateName(`${templateName}@${version}`);
```
lucasbento: Why not pass the `version` as an argument here and deal with the error on line 67 inside of this function? If you do that you can keep doing `let {uri, name} = await processTemplate...` instead of `if...else`.

grabbou: Good catch! I wouldn't put it inside `processTemplate` myself, but move the validation logic upwards.

```
if (version && semver.valid(version) && !semver.gte(version, '0.60.0')) {
throw new Error(
    'Cannot use React Native CLI to initialize project with version lower than 0.60.0',
);
}
let {uri, name} = await processTemplateName(version ? `${templateName}@${version}` : templateName);

Comment (grabbou) [reply packages/cli/src/commands/init/init.js]: Also, I don't think we should check for `semver.valid`. That can potentially swallow errors when invalid version is passed. 

I would only do `version && semver.gte` here and let `semver.gte` [throw a TypeError when invalid](https://github.com/npm/node-semver/blob/5fb517b2906a0763518e1941a3f4a163956a81d3/semver.js#L305)

Comment (thymikee) [reply packages/cli/src/commands/init/init.js]: This required more work, as it turned out we don't handle versioned dependencies correctly. Added support for that, so we can now call `init --template name@1.2.3` and treat `react-native` with `--version` passed in such a way

File: packages/cli/src/commands/init/init.js [hunk 2]
```diff
@@ -39,93 +39,45 @@ function adjustNameIfUrl(name, cwd) {
...
-
-    loader.start('Installing all required dependencies');
-    await PackageManager.installAll({preferYarn: !npm, silent: true});
-    loader.succeed();
-  } catch (e) {
-    loader.fail();
-    throw new Error(e);
-  } finally {
-    fs.removeSync(templateSourceDir);
+  if (version && !semver.gte(version, '0.60.0-rc.0')) {
```
thymikee: it's now possible to try it with RCs

Esemesek:
Code looks much cleaner but it doesn't quite work. If you run
```sh
node react-native-cli/packages/cli/build init ProjectName
```
you will get an error:
```
error Invalid Version: latest
```

It seems like there is some duplication in `RegExp` checking (https://github.com/react-native-community/cli/blob/chore/refactor-init/packages/cli/src/commands/init/templateName.js#L36-L66).

grabbou:
This error is actually coming from `semver` and is expected. Looks like `semver` treats `latest` as invalid. 

@thymikee maybe we should update the `if` statement to be:
```
if (version === 'latest' || version && version.gte('0.60.0')) {
```

grabbou:
We should probably bring back `semver.valid` again.

Also, note two things:
- tag can be `latest` and right now, `latest` will be `0.59`. We should hardcode that if version is `latest`, we are going to produce an error. This should be removed right before we cut `0.60.0` and tag it as `latest`
- tag can be `next` and in this case, it's okay, because `next` will be `0.60.0-rc.0`.

packages/cli/src/commands/init/init.js
grabbou: Good catch! I hope `gte` will just treat 0.60.0 as higher - I know `npm` had some problems with `^` etc.

Esemesek: `gte` - `greater than or equal`

thymikee:
I've adjusted the regex to accept tags as well, added relevant tests. 
Since RCs of CLI are for early adopters, I'm not a fan of hardcoding this. We explicitly say in the docs that it's for 0.60.0 or greater, so we should be safe.

Esemesek:
bitmoji

PR #365, opened 2019-05-01
feat: support `init` with custom project directory

### LINKED ISSUE #110: react-native init should have a folder option in addition to the current ProjectName one
As written in the title, a `folder` option would be very welcome. It should allow to give a different location to the default one it currently uses.

We should be able to do:
```console
foo@bar:~$ react-native init AwesomeProject .
```
or 
```console
foo@bar:~$ react-native init AwesomeProject awesome05
```

lucasbento:
Summary:
---------

This PR fixes #110 by adding the possibility of specifying the folder path of the project when running the `init` command.

This changes the `fs.mkdirSync` to run with `recursive: true` which is only available on Node.js > 10, I'm open for opinions on how to throw a nice error to the user if the path doesn't exist yet and is not possible to create without `recursive`.

Test Plan:
----------

```sh
npx react-native init AwesomeApp --directory ../
npx react-native init AwesomeApp --directory path/that/does/not/exist/yet
```

I also added a test for this on `init.test.js` that I would love some feedback on!

cc @thymikee maybe you can review this :)

---

### Directory exists and user chooses to replace it
image

### Directory exists and user chooses to not replace it
Screenshot 2019-05-16 at 15 56 01

---

- [x] Prompt the user to replace the project directory after the `init` intro (React logo) shows up;
- [x] Fix test for `init` failing because of `inquirer`.

### DIFF
--- a/__e2e__/init.test.js
+++ b/__e2e__/init.test.js
@@ -22,27 +22,27 @@ test('init --template fails without package name', () => {
   expect(stderr).toContain('missing required argument');
 });
 
-test('init --template', () => {
-  const templateFiles = [
-    '.buckconfig',
-    '.eslintrc.js',
-    '.flowconfig',
-    '.gitattributes',
-    // should be here, but it's not published yet
-    // '.gitignore',
-    '.watchmanconfig',
-    'App.js',
-    '__tests__',
-    'android',
-    'babel.config.js',
-    'index.js',
-    'ios',
-    'metro.config.js',
-    'node_modules',
-    'package.json',
-    'yarn.lock',
-  ];
+const templateFiles = [
+  '.buckconfig',
+  '.eslintrc.js',
+  '.flowconfig',
+  '.gitattributes',
+  // should be here, but it's not published yet
+  // '.gitignore',
+  '.watchmanconfig',
+  'App.js',
+  '__tests__',
+  'android',
+  'babel.config.js',
+  'index.js',
+  'ios',
+  'metro.config.js',
+  'node_modules',
+  'package.json',
+  'yarn.lock',
+];
 
+test('init --template', () => {
   const {stdout} = run(DIR, [
     'init',
     '--template',
@@ -83,4 +83,30 @@ test('init --template file:/tmp/custom/template', () => {
   ]);
 
   expect(stdout).toContain('Run instructions');
+});
+
+test('init --template with custom project path', () => {
+  const projectName = 'TestInit';
+  const customProjectPath = './random/path';
+
+  run(DIR, [
+    'init',
+    '--template',
+    'react-native-new-template',
+    'TestInit',
+    customProjectPath,
+  ]);
+
+  const projectPath = path.join(DIR, customProjectPath);
+
+  // make sure we don't leave garbage
+  expect(fs.readdirSync(projectPath)).toEqual(['TestInit']);
+  expect(fs.readdirSync(path.join(projectPath, projectName))).toEqual(
+    templateFiles,
+  );
+
+  const pkgJson = require(path.join(projectPath, 'TestInit', 'package.json'));
+  expect(pkgJson).toMatchSnapshot(
+    'package.json contains necessary configuration',
+  );
 });
--- a/__e2e__/init.test.js
+++ b/__e2e__/init.test.js
@@ -1,6 +1,7 @@
 // @flow
 import fs from 'fs';
 import path from 'path';
+import inquirer from 'inquirer';
 import {run, getTempDirectory, cleanup, writeFiles} from '../jest/helpers';
 
 const DIR = getTempDirectory('command-init');
@@ -87,35 +88,24 @@ test('init --template file:/tmp/custom/template', () => {
 
 test('init --template with custom project path', () => {
   const projectName = 'TestInit';
-  const customProjectPath = getTempDirectory('command-init-custom-path');
-
-  // Create directory
-  writeFiles(customProjectPath, {});
+  const customPath = 'custom-path';
 
   run(DIR, [
     'init',
     '--template',
     'react-native-new-template',
     projectName,
-    customProjectPath,
+    '--directory',
+    'custom-path',
   ]);
 
   // make sure we don't leave garbage
-  expect(fs.readdirSync(customProjectPath)).toEqual([projectName]);
-  expect(fs.readdirSync(path.join(customProjectPath, projectName))).toEqual(
-    templateFiles,
-  );
+  expect(fs.readdirSync(DIR)).toEqual([customPath]);
+  expect(fs.readdirSync(path.join(DIR, customPath))).toEqual(templateFiles);
 
-  const pkgJson = require(path.join(
-    customProjectPath,
-    projectName,
-    'package.json',
-  ));
+  const pkgJson = require(path.join(DIR, customPath, 'package.json'));
 
   expect(pkgJson).toMatchSnapshot(
     'package.json contains necessary configuration',
   );
-
-  // Remove custom project path
-  cleanup(customProjectPath);
 });
--- a/__e2e__/init.test.js
+++ b/__e2e__/init.test.js
@@ -101,10 +101,4 @@ test('init --template with custom project path', () => {
   // make sure we don't leave garbage
   expect(fs.readdirSync(DIR)).toEqual([customPath]);
   expect(fs.readdirSync(path.join(DIR, customPath))).toEqual(templateFiles);
-
-  const pkgJson = require(path.join(DIR, customPath, 'package.json'));
-
-  expect(pkgJson).toMatchSnapshot(
-    'package.json contains necessary configuration',
-  );
 });
--- a/__e2e__/init.test.js
+++ b/__e2e__/init.test.js
@@ -87,26 +87,35 @@ test('init --template file:/tmp/custom/template', () => {
 
 test('init --template with custom project path', () => {
   const projectName = 'TestInit';
-  const customProjectPath = './random/path';
+  const customProjectPath = getTempDirectory('command-init-custom-path');
+
+  // Create directory
+  writeFiles(customProjectPath, {});
 
   run(DIR, [
     'init',
     '--template',
     'react-native-new-template',
-    'TestInit',
+    projectName,
     customProjectPath,
   ]);
 
-  const projectPath = path.join(DIR, customProjectPath);
-
   // make sure we don't leave garbage
-  expect(fs.readdirSync(projectPath)).toEqual(['TestInit']);
-  expect(fs.readdirSync(path.join(projectPath, projectName))).toEqual(
+  expect(fs.readdirSync(customProjectPath)).toEqual([projectName]);
+  expect(fs.readdirSync(path.join(customProjectPath, projectName))).toEqual(
     templateFiles,
   );
 
-  const pkgJson = require(path.join(projectPath, 'TestInit', 'package.json'));
+  const pkgJson = require(path.join(
+    customProjectPath,
+    projectName,
+    'package.json',
+  ));
+
   expect(pkgJson).toMatchSnapshot(
     'package.json contains necessary configuration',
   );
+
+  // Remove custom project path
+  cleanup(customProjectPath);
 });
--- a/__e2e__/init.test.js
+++ b/__e2e__/init.test.js
@@ -1,7 +1,6 @@
 // @flow
 import fs from 'fs';
 import path from 'path';
-import inquirer from 'inquirer';
 import {run, getTempDirectory, cleanup, writeFiles} from '../jest/helpers';
 
 const DIR = getTempDirectory('command-init');
--- a/__e2e__/init.test.js
+++ b/__e2e__/init.test.js
@@ -94,7 +94,6 @@ test('init --template with custom project path', () => {
     '--template',
     'react-native-new-template',
     projectName,
-    '--directory',
     'custom-path',
   ]);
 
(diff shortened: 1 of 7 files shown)


lucasbento:
Tests are failing because `recursive` is not supported on Node.js 8, I'll see what alternative I can use for that.

lucasbento:
It's now creating the folder before running the command, I also added a `try...catch` block to first try to create the path (with `recursive: true`) and if it's not possible then just throw an error to the user saying `Cannot initialize new project because custom project directory does not exist.`.

grabbou:
> I'm open for opinions on how to throw a nice error to the user if the path doesn't exist yet and is not possible to create without recursive.

@lucasbento, how about:
```
var shell = require('shelljs');
shell.mkdir('-p', fullPath);
```
?

grabbou:
Alternatively, we could split the path into directories and just recursively keep calling `mkdir`, e.g:
```
mkdirSync('/a')
mkdirSync('/a/b')
mkdirSync('/a/b/c')
mkdirSync('/a/b/c/d')
```

packages/cli/src/commands/init/init.js
```diff
@@ -171,7 +192,7 @@ function createProject(projectName: string, options: Options, version: string) {
 }
 
 export default (async function initialize(
-  [projectName]: Array<string>,
+  [projectName, customProjectPath]: Array<string>,
```
grabbou: Maybe `projectPath` and default value `.` (because by default, I guess we create in CWD?)

packages/cli/src/commands/init/init.js
```diff
@@ -26,6 +26,16 @@ type Options = {|
...
 |};
 
+function getProjectPath({
+  projectName,
+  customProjectPath = '',
+}: {
+  projectName: string,
+  customProjectPath: string,
+}) {
+  return path.join(customProjectPath, projectName);
```
grabbou: So this will create a `./custom-path/AppName`. I don't think this is correct approach. I think we should run in a similar manner as `git` command.

By default, `folder` is $projectName.

When you run `react-native init MyProject .`, this will init in CWD. When you run `react-native init MyProject ./custom-location`, this will init in `custom-location`, meaning that `package.json` will be available at `custom-location/package.json`.

Just like when you do git clone repoName.

packages/cli/src/commands/init/init.js
grabbou: If you agree with my comment https://github.com/react-native-community/cli/pull/365/files#r281157763, I believe we should just do:
```
const projectPath = customProjectPath || projectName
```

packages/cli/src/commands/init/init.js
```diff
@@ -155,9 +165,20 @@ async function createFromReactNativeTemplate(
...
-  fs.mkdirSync(projectName);
-  process.chdir(projectName);
+function createProject(
+  projectPath: string,
+  projectName: string,
+  options: Options,
+  version: string,
+) {
+  try {
+    fs.mkdirSync(projectPath, {recursive: true});
```
grabbou: Alternatives:
- use `mkdir -p` 
- manually split path by `path.delimiter` and call `mkdirSync` n times for each path chunk.

__e2e__/init.test.js
```diff
@@ -84,3 +84,38 @@ test('init --template file:/tmp/custom/template', () => {
...
+    projectName,
+    'package.json',
+  ));
+
+  expect(pkgJson).toMatchSnapshot(
+    'package.json contains necessary configuration',
+  );
+
+  // Remove custom project path
+  cleanup(customProjectPath);
```
thymikee: Please wrap this test in a `describe` and move this call to `afterEach` so it's called when the test fails

packages/cli/src/commands/init/init.js
thymikee: We have `mkdirp` in the cli deps, so please use that for now

packages/cli/src/commands/init/init.js
```diff
@@ -155,9 +165,20 @@ async function createFromReactNativeTemplate(
...
+    process.chdir(projectPath);
+  } catch (err) {
+    throw new Error(
+      'Cannot initialize new project because custom project directory does not exist.',
```
thymikee: Hm, I don't think that should happen. `mkdirp` can throw but on lack of permissions. We should forward the error correctly, not mask it without stack trace.

packages/cli/src/commands/init/init.js
```diff
@@ -183,12 +204,14 @@ export default (async function initialize(
    */
   const version: string = minimist(process.argv).version || 'latest';
 
-  if (fs.existsSync(projectName)) {
+  const projectPath = getProjectPath({projectName, customProjectPath});
+
+  if (fs.existsSync(projectPath)) {
     throw new DirectoryAlreadyExistsError(projectName);
```
thymikee: Btw, I don't like this behavior, it's a DX regression compared to previous `init`. While we're at creation of new folders, can we prompt users if they want to overwrite the directory? You can use `inquirer` for that.

grabbou: Yeah, I've actually noticed the same. It's not strictly related to this PR, but I guess it would be a good thing to do. Old init was overwriting directories and that was quite useful generally speaking.

packages/cli/src/commands/init/init.js
grabbou: Yes, you can use CLIError for this use-case. It allows you to re-throw an error with a better message.

packages/cli/src/commands/init/init.js
grabbou: Nice! I didn't remember this. I had a feeling there must be a JavaScript script that mimics this functionality, but quick Google didn't reveal anything.

packages/cli/src/commands/init/init.js
lucasbento: This is a very nice idea, I'll do it!

packages/cli/src/commands/init/init.js
lucasbento: Yeah, noticed `mkdirp` in the deps today in the morning while working on #373, I'll change this :)_

lucasbento:
Just did all the requested changes except for the unit test one that @thymikee pointed out, the tests for `init` are failing, I'll fix them later as I need to find out how to force a `inquirer` reply.

lucasbento:
I'm wondering now, if the directory of the project already exists, should we prompt the user to replace it before or after the react logo with the `Welcome to React Native!` shows up? perhaps it would look better if we show it after.

thymikee:
> perhaps it would look better if we show it after.

I'm in favor of that

lucasbento:
I fixed the tests and put the prompt after the logo but there's a bug of which the folder is deleted even when you specify `No` so I'm working on it.

packages/cli/src/commands/init/init.js
```diff
@@ -160,16 +216,21 @@ export default (async function initialize(
    */
   const version: string = minimist(process.argv).version || 'latest';
 
-  if (fs.existsSync(projectName)) {
-    throw new DirectoryAlreadyExistsError(projectName);
-  }
+  const projectDirectoryExists = await isProjectDirectoryExistent({
```
lucasbento: Not very fan of calling this twice, not sure if it would be a better option to pass the value of this to `createProject` and then to `createFromTemplate`, would appreciate other opinions.

lucasbento:
@grabbou / @thymikee: please review again 🙌

packages/cli/src/commands/init/init.js
```diff
@@ -160,16 +225,25 @@ export default (async function initialize(
...
+  const {directory, directoryExists} = getProjectDirectory({
+    projectName,
+    directory: options.directory,
+  });
 
   try {
-    await createProject(projectName, options, version);
+    await createProject(
+      projectName,
+      {...options, directory},
```
lucasbento: There's a flow error that I haven't been able to fix yet, will investigate it later but I'm open for tips.

image

lucasbento:
@thymikee: cool, thanks!

The problem with having `[directory]` is that it doesn't act as an optional argument but rather just gets what it supplied after the `projectName`.

E.g. `node ../path/to/cli/build/index.js init MyProject --version="1000.0.6"` will have `1000.0.6` as the directory and not `undefined`.

It was like that before and I changed it for the `--directory` argument.

thymikee:
@lucasbento oh that's unfortunate 😞, probably `commander` not being so smart. Gonna revert that change. Thanks for noticing!

packages/cli/src/commands/init/init.js
```diff
@@ -152,6 +199,8 @@ export default (async function initialize(
   _context: ConfigT,
   options: Options,
 ) {
+  const rootFolder = process.cwd();
```
grabbou: I believe there's context.root you can use.

packages/cli/src/tools/installPods.js
```diff
@@ -17,6 +17,10 @@ async function installPods({
   loader?: typeof Ora,
 }) {
   try {
+    if (!(await fs.pathExists('ios'))) {
```
grabbou: You can use `https://nodejs.org/api/fs.html#fs_fs_existssync_path`. The `pathExistsSync` version from `fs-extra` is just its alias.

Ref: https://github.com/jprichardson/node-fs-extra/blob/master/docs/pathExists-sync.md

packages/cli/src/tools/installPods.js
```diff
@@ -17,6 +17,10 @@ async function installPods({
...
+      return;
+    }
+
     process.chdir('ios');
 
     const hasPods = await fs.pathExists('Podfile');
```
grabbou: If [this](https://github.com/react-native-community/cli/pull/365/files#r285092791) gets implemented, likely this should also be updated.

packages/cli/src/commands/init/init.js
```diff
@@ -127,23 +175,22 @@ async function installDependencies({
...
+  const templateName = options.template || 'react-native';
 
   return createFromTemplate({
     projectName,
-    templateName: 'react-native',
-    version,
+    templateName,
+    // version is "latest" by default, but it's easier for us to treat it as
+    // undefined when the "template" param is passed. Might refactor later
+    version: options.template ? undefined : version,
```
grabbou: I don't like this change. I think an `if` statement (calling `createFromTemplate` when `template` is passed) was clearer and more elegant.

What was the reason for this change?

thymikee: less repetition mainly

grabbou: I find it confusing, specially that we leave "might refactor later". I'd suggest "let's change it later" or make it clear right now :)

packages/cli/src/commands/init/init.js
```diff
@@ -44,33 +89,28 @@ function adjustNameIfUrl(name, cwd) {
...
+
+  await setProjectDirectory(directory);
+
   const Loader = getLoader();
   const loader = new Loader({text: 'Downloading template'});
   const templateSourceDir = fs.mkdtempSync(
     path.join(os.tmpdir(), 'rncli-init-template-'),
   );
 
-  if (version && semver.valid(version) && !semver.gte(version, '0.60.0-rc.0')) {
```
thymikee: Turned out we can now entirely extract version handling from `createFromTemplate`, which makes more sense because version is only a parameter of `templateName`. Thanks for making me think about it harder @grabbou :P

grabbou: No worries

packages/cli/src/commands/init/init.js
```diff
@@ -22,12 +24,55 @@ import installPods from '../../tools/installPods';
...
+}
+
+function getProjectDirectory({projectName, directory}): string {
+  return path.relative(process.cwd(), directory || projectName);
+}
+
+async function setProjectDirectory(directory) {
+  const directoryExists = doesDirectoryExist(directory);
+  if (directoryExists) {
+    const {shouldReplaceprojectDirectory} = await inquirer.prompt([
```
Esemesek: Can we default to the `No` option here? We don't want users to overwrite files by a mistake.

thymikee: Makes sense, fixed

Esemesek: Default option is still `Yes`

Screenshot 2019-05-17 at 17 03 21

thymikee: fixd, pushed it to wrong branch

adhikarisandil32:
is there a way to make it work on the existing directory. Let's say my vscode is open in a directory in which I want the project to start, right there. like vite does while creating react app on the web.

`npm create vite@latest ./` and voila there it creates the project, in the current directory.

May be we can have something similar like 

`npx @react-native-community/cli init ProjectName --directory .` that will start the project in the current directory.

szymonrybczak:
hey @adhikarisandil32, if the provided command doesn't work then we'd like to have this fixed. Might submitting a PR? Should be a simple thing 🙌

adhikarisandil32:
> hey @adhikarisandil32, if the provided command doesn't work then we'd like to have this fixed. Might submitting a PR? Should be a simple thing 🙌

if you mean `npx @react-native-community/cli init ProjectName --directory .` command, no it doesn't work, neither does `npx @react-native-community/cli init ProjectName --directory ./`

it shows an error saying **error occoured while trying to create a project directory.**
Screenshot 2024-08-21 112142

szymonrybczak:
Agree @adhikarisandil32! Sorry if I misunderstood! Will try to put a PR soon :)

PR #366, opened 2019-05-04
fix don`t run app on devices when finished run-android build

zusinShinpei:
Summary: 
---------
this PR fixes run-android command that is not running app on device when finished building

Test Plan: 
----------
```
react-native run-android
```


grabbou:
Hey,

Thanks for working on this PR. Unfortunately, I don't think this is a right approach. You should have `adb` available in your `$PATH` - the installation locations can be different from one machine to another.

The current code is good. As an alternative, maybe we could add a `try/catch` for missing `adb` and ask users to install it first (or have it in the `$PATH`)?

Please open an issue first to discuss it to make sure we don't close a PR with a lot of work you put into it.

PR #371, opened 2019-05-06
fix: reporter config option

fson:
*This PR backports the fix from https://github.com/react-native-community/cli/pull/370 to the 1.x branch. I cherry picked 809aef93f20b6 to 1.x.*

Summary:
---------

The `loadConfig` function in `metro-config` reads the `reporter` option
from the second argument (`defaultConfigOverrides`) instead of the
first argument (`argv`), so we need to pass it in that object to
make the `customLogReporterPath` CLI option work.

Test Plan:
----------

1. Created a file `reporter.js` with the following contents:
```js
class JsonReporter {
  update(event) {
    console.log(JSON.stringify(event));
  }
}

module.exports = JsonReporter;
```
2. Ran this command and inspected that the output was JSON:
```
node ~/Projects/react-native-cli/packages/cli/build/index.js start --customLogReporterPath ~/Projects/react-native-cli/reporter.js
```

### DIFF
--- a/packages/cli/src/commands/upgrade/__tests__/upgrade.test.js
+++ b/packages/cli/src/commands/upgrade/__tests__/upgrade.test.js
@@ -170,10 +170,12 @@ success Upgraded React Native to v0.58.4 🎉. Now you can review and commit the
 `);
 });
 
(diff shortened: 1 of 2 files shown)


thymikee:
Thanks! Released 1.9.3 with this patch

PR #373, opened 2019-05-07
feat: automatically run `pod install` when running `init` command

### LINKED ISSUE #368: Automatically run `pod install` when a template ships with a Podfile
New React Native template has CocoaPods integration and other templates might soon have it too.

Unfortunately, the drawback is that you need to run `pod install` yourself after initing a project.

We want to improve this workflow and make CLI run `pod install` automatically for users when CocoaPods is available or attempt installing. 

Since this code is only going to be run when platform is Mac, we can assume `gem` is available and run `sudo gem install cocoapods` as per [the docs](https://guides.cocoapods.org/using/getting-started.html).

We can also allow users to `skip` that step where we only print an information that they should install CocoaPods afterwards in order to run `iOS` application.

lucasbento:
Summary:
---------

This PR fixes #368 of which checks if CocoaPods is installed and runs `pod install`.

Test Plan:
----------

Run `node /path/to/cli/packages/cli/build/index.js init --template /path/to/react-native AwesomeProject`

Looks like this now when user doesn't have CocoaPods installed:

Screenshot 2019-05-07 at 11 15 35

With CocoaPods installed:

Screenshot 2019-05-07 at 11 19 08

---

#### TODOs

- [x] Verify if `Podfile` exists before trying to install and/or check if `cocoapods` is installed;
- [x] Don't do this for non-mac machines;
- [x] Install `cocoapods` without `sudo`, if that doesn't work then do it with `sudo`;
- [x] Verify that `.xcworkspace` exists before printing it out on the run instructions.

### DIFF
--- a/packages/cli/src/commands/init/init.js
+++ b/packages/cli/src/commands/init/init.js
@@ -272,7 +272,7 @@ export default (async function initialize(
   try {
     await createProject(projectName, options, version);
 
-    printRunInstructions(process.cwd(), projectName);
+    await printRunInstructions(process.cwd(), projectName);
   } catch (e) {
     logger.error(e.message);
     fs.removeSync(projectName);
--- a/packages/cli/src/commands/init/init.js
+++ b/packages/cli/src/commands/init/init.js
@@ -2,10 +2,12 @@
 import os from 'os';
 import path from 'path';
 import fs from 'fs-extra';
+import execa from 'execa';
 import Ora from 'ora';
 import minimist from 'minimist';
 import semver from 'semver';
 import inquirer from 'inquirer';
+import commandExists from 'command-exists';
 import type {ConfigT} from 'types';
 import {validateProjectName} from './validate';
 import DirectoryAlreadyExistsError from './errors/DirectoryAlreadyExistsError';
@@ -22,6 +24,7 @@ import * as PackageManager from '../../tools/packageManager';
 import {processTemplateName} from './templateName';
 import banner from './banner';
 import {getLoader} from '../../tools/loader';
+import {CLIError} from '@react-native-community/cli-tools/build/errors';
 
 type Options = {|
   template?: string,
@@ -81,7 +84,6 @@ async function createFromExternalTemplate(
       loader.succeed();
     }
 
-    loader.start('Installing all required dependencies');
     await installDependencies({npm, loader});
   } catch (e) {
     loader.fail();
@@ -91,31 +93,20 @@ async function createFromExternalTemplate(
   }
 }
 
-async function installDependencies({
-  npm,
-  loader,
-}: {
-  npm?: boolean,
-  loader: typeof Ora,
-}) {
-  const {error} = await PackageManager.installAll({
-    preferYarn: !npm,
-    silent: true,
-  });
-  loader.succeed();
+async function installPods(loader: typeof Ora) {
+  process.chdir('ios');
 
-  // TODO: if this is different than debian just return
-  if (!error) {
+  const hasPods = await fs.pathExists('Podfile');
+
+  if (!hasPods) {
     return;
   }
 
-  if (error === PackageManager.ERRORS.failedToRunPodInstall) {
-    return logger.warn(
-      'Failed to run "pod install", please try to run it manually.',
-    );
-  }
+  try {
+    await commandExists('pod');
+  } catch (err) {
+    loader.succeed();
 
-  if (error === PackageManager.ERRORS.noCocoaPods) {
     const {shouldInstallCocoaPods} = await inquirer.prompt([
       {
         type: 'confirm',
@@ -126,28 +117,57 @@ async function installDependencies({
 
     if (shouldInstallCocoaPods) {
       try {
-        await PackageManager.installCocoaPods();
+        await execa('sudo', ['gem', 'install', 'cocoapods'], {
+          stdio: 'pipe',
+        });
       } catch (err) {
         throw new Error(
           'Error occurred while trying to install CocoaPods, please install it manually.',
+          err,
         );
       }
 
       try {
         loader.start('Installing pods');
-        await PackageManager.installPods({silent: true});
-        return loader.succeed();
       } catch (err) {
-        return logger.warn(
+        console.log(err);
+        throw new CLIError(
           'Failed to run "pod install", please try to run it manually.',
+          err,
         );
       }
     }
+  }
 
-    return logger.warn(
-      'Please install CocoaPods and run "pod install" to install all dependencies.',
+  try {
+    await execa('pod', ['install'], {
+      stdio: 'pipe',
+    });
+  } catch (err) {
+    throw new CLIError(
+      'Failed to run "pod install", please try to run it manually.',
+      err,
     );
   }
+}
+
+async function installDependencies({
+  npm,
+  loader,
+}: {
+  npm?: boolean,
+  loader: typeof Ora,
+}) {
+  loader.start('Installing all required dependencies');
+
+  await PackageManager.installAll({
+    preferYarn: !npm,
+    silent: true,
+  });
+
+  await installPods(loader);
+
+  loader.succeed();
 }
 
 async function createFromReactNativeTemplate(
@@ -204,7 +224,6 @@ async function createFromReactNativeTemplate(
       loader.succeed();
     }
 
-    loader.start('Installing all required dependencies');
     await installDependencies({npm, loader});
   } catch (e) {
     loader.fail();
--- a/packages/cli/src/commands/init/init.js
+++ b/packages/cli/src/commands/init/init.js
@@ -127,14 +127,7 @@ async function installPods(loader: typeof Ora) {
         );
       }
 
-      try {
-        loader.start('Installing pods');
-      } catch (err) {
-        throw new CLIError(
-          'Failed to run "pod install", please try to run it manually.',
-          err,
-        );
-      }
+      loader.start('Installing pods');
     }
   }
 
--- a/packages/cli/src/commands/init/init.js
+++ b/packages/cli/src/commands/init/init.js
@@ -164,7 +164,15 @@ async function installDependencies({
     silent: true,
   });
 
-  await installPods(loader);
+  try {
+    await installPods(loader);
+
+    process.chdir('..');
+  } catch (err) {
+    process.chdir('..');
+
+    throw err;
+  }
 
   loader.succeed();
 }
--- a/packages/cli/src/commands/init/init.js
+++ b/packages/cli/src/commands/init/init.js
@@ -130,7 +130,6 @@ async function installPods(loader: typeof Ora) {
       try {
         loader.start('Installing pods');
       } catch (err) {
-        console.log(err);
         throw new CLIError(
           'Failed to run "pod install", please try to run it manually.',
           err,
--- a/packages/cli/src/commands/init/init.js
+++ b/packages/cli/src/commands/init/init.js
@@ -117,14 +117,22 @@ async function installPods(loader: typeof Ora) {
 
     if (shouldInstallCocoaPods) {
       try {
-        await execa('sudo', ['gem', 'install', 'cocoapods'], {
+        // First attempt to install `cocoapods`
+        await execa('gem', ['install', 'cocoapods'], {
           stdio: 'pipe',
         });
       } catch (err) {
-        throw new CLIError(
-          'Error occurred while trying to install CocoaPods, please install it manually.',
-          err,
-        );
+        try {
+          // If that doesn't work then try with sudo
+          await execa('sudo', ['gem', 'install', 'cocoapods'], {
+            stdio: 'pipe',
+          });
+        } catch (err) {
+          throw new CLIError(
+            'Error occurred while trying to install CocoaPods, please run this command again.',
+            err,
+          );
+        }
       }
 
       loader.start('Installing pods');
--- a/packages/cli/src/commands/init/init.js
+++ b/packages/cli/src/commands/init/init.js
@@ -74,7 +74,7 @@ async function createFromExternalTemplate(
     const {postInitScript} = templateConfig;
     if (postInitScript) {
       // Leaving trailing space because there may be stdout from the script
-      loader.start('Executing post init script ');
+      loader.start('Executing post init script');
       await executePostInitScript(name, postInitScript, templateSourceDir);
       loader.succeed();
     }
--- a/packages/cli/src/commands/init/init.js
+++ b/packages/cli/src/commands/init/init.js
@@ -165,14 +165,10 @@ async function installDependencies({
     silent: true,
   });
 
-  try {
+  if (process.platform === 'darwin') {
     await installPods(loader);
 
     process.chdir('..');
-  } catch (err) {
-    process.chdir('..');
-
-    throw err;
   }
 
   loader.succeed();
--- a/packages/cli/src/commands/init/init.js
+++ b/packages/cli/src/commands/init/init.js
@@ -228,7 +228,7 @@ export default (async function initialize(
   try {
     await createProject(projectName, options, version);
 
-    await printRunInstructions(process.cwd(), projectName);
+    printRunInstructions(process.cwd(), projectName);
   } catch (e) {
     logger.error(e.message);
     fs.removeSync(projectName);
--- a/packages/cli/src/commands/init/init.js
+++ b/packages/cli/src/commands/init/init.js
@@ -79,7 +79,7 @@ async function createFromExternalTemplate(
     const {postInitScript} = templateConfig;
     if (postInitScript) {
       // Leaving trailing space because there may be stdout from the script
-      loader.start('Executing post init script');
+      loader.start('Executing post init script ');
       await executePostInitScript(name, postInitScript, templateSourceDir);
       loader.succeed();
     }
--- a/packages/cli/src/commands/init/init.js
+++ b/packages/cli/src/commands/init/init.js
@@ -2,12 +2,9 @@
 import os from 'os';
 import path from 'path';
 import fs from 'fs-extra';
-import execa from 'execa';
 import Ora from 'ora';
 import minimist from 'minimist';
 import semver from 'semver';
-import inquirer from 'inquirer';
-import commandExists from 'command-exists';
 import type {ConfigT} from 'types';
 import {validateProjectName} from './validate';
 import DirectoryAlreadyExistsError from './errors/DirectoryAlreadyExistsError';
@@ -21,10 +18,10 @@ import {
 } from './template';
 import {changePlaceholderInTemplate} from './editTemplate';
 import * as PackageManager from '../../tools/packageManager';
+import installPods from '../../tools/installPods';
 import {processTemplateName} from './templateName';
 import banner from './banner';
 import {getLoader} from '../../tools/loader';
-import {CLIError} from '@react-native-community/cli-tools/build/errors';
 
 type Options = {|
   template?: string,
@@ -98,7 +95,7 @@ async function createFromTemplate({
       loader.succeed();
     }
 
-    await installDependencies({npm, loader});
+    await installDependencies({projectName, npm, loader});
   } catch (e) {
     loader.fail();
     throw new Error(e);
@@ -107,68 +104,12 @@ async function createFromTemplate({
   }
 }
 
-async function installPods(loader: typeof Ora) {
-  process.chdir('ios');
-
-  const hasPods = await fs.pathExists('Podfile');
-
-  if (!hasPods) {
-    return;
-  }
-
-  try {
-    await commandExists('pod');
-  } catch (err) {
-    loader.succeed();
-
-    const {shouldInstallCocoaPods} = await inquirer.prompt([
-      {
-        type: 'confirm',
-        name: 'shouldInstallCocoaPods',
-        message: 'CocoaPods is not installed, do you want to install it?',
-      },
-    ]);
-
-    if (shouldInstallCocoaPods) {
-      try {
-        // First attempt to install `cocoapods`
-        await execa('gem', ['install', 'cocoapods'], {
-          stdio: 'pipe',
-        });
-      } catch (err) {
-        try {
-          // If that doesn't work then try with sudo
-          await execa('sudo', ['gem', 'install', 'cocoapods'], {
-            stdio: 'pipe',
-          });
-        } catch (err) {
-          throw new CLIError(
-            'Error occurred while trying to install CocoaPods, please run this command again.',
-            err,
-          );
-        }
-      }
-
-      loader.start('Installing pods');
-    }
-  }
-
-  try {
-    await execa('pod', ['install'], {
-      stdio: 'pipe',
-    });
-  } catch (err) {
-    throw new CLIError(
-      'Failed to run "pod install", please try to run it manually.',
-      err,
-    );
-  }
-}
-
 async function installDependencies({
+  projectName,
   npm,
   loader,
 }: {
+  projectName: string,
   npm?: boolean,
   loader: typeof Ora,
 }) {
@@ -180,9 +121,7 @@ async function installDependencies({
   });
 
   if (process.platform === 'darwin') {
-    await installPods(loader);
-
-    process.chdir('..');
+    await installPods({projectName, loader});
   }
 
   loader.succeed();
--- a/packages/cli/src/commands/init/init.js
+++ b/packages/cli/src/commands/init/init.js
@@ -2,8 +2,10 @@
 import os from 'os';
 import path from 'path';
 import fs from 'fs-extra';
+import Ora from 'ora';
 import minimist from 'minimist';
 import semver from 'semver';
+import inquirer from 'inquirer';
 import type {ConfigT} from 'types';
 import {validateProjectName} from './validate';
 import DirectoryAlreadyExistsError from './errors/DirectoryAlreadyExistsError';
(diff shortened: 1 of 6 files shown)


packages/cli/src/tools/packageManager.js
```diff
@@ -11,7 +12,12 @@ type Options = {|
 
 let projectDir;
 
-function executeCommand(
+export const ERRORS = {
```
lucasbento: @grabbou: not entirely sure if this is suitable for the current CLI code, open for opinions.

packages/cli/src/tools/packageManager.js
```diff
@@ -61,8 +67,34 @@ export function uninstall(packageNames: Array<string>, options?: Options) {
...
+}
+
+export async function installAll(options?: Options) {
+  shouldUseYarn(options)
+    ? await executeCommand('yarn', ['install'], options)
+    : await executeCommand('npm', ['install'], options);
+
+  // TODO: only do this on macOS
+
+  return installPods(options);
```
lucasbento: Not sure if it makes sense to leave pods inside of `PackageManager` (since it's currently yarn and npm) but it's a process to install dependencies anyway.

packages/cli/src/tools/packageManager.js
```diff
@@ -61,8 +67,34 @@ export function uninstall(packageNames: Array<string>, options?: Options) {
     : executeCommand('npm', ['uninstall', ...packageNames, '--save'], options);
 }
 
-export function installAll(options?: Options) {
-  return shouldUseYarn(options)
-    ? executeCommand('yarn', ['install'], options)
-    : executeCommand('npm', ['install'], options);
+export function installCocoaPods() {
+  return executeCommand('sudo', ['gem', 'install', 'cocoapods'], {
+    silent: false,
```
lucasbento: I needed to do this in order to have the input for the user password since it's using `sudo`, didn't have time yet to check how to do it without having the messages of actually installing the dep, will research a little bit more later.

lucasbento:
@grabbou: this is not ready yet as I need a bit more time but please review what I did so far as I'm a bit doubtful about a few choices.

There are also a few comment TODOs that I need to fix.

lucasbento:
I just noticed that it needs to first verify that the template has `Podfile` on it, gonna do that soon.

packages/cli/src/tools/packageManager.js
grabbou: We've been subclassing errors so far - I guess you could try that and just go with `instanceof` check.

packages/cli/src/tools/packageManager.js
```diff
@@ -61,8 +67,34 @@ export function uninstall(packageNames: Array<string>, options?: Options) {
...
+  } catch (err) {
+    return {error: ERRORS.noCocoaPods};
+  }
+
+  process.chdir('ios');
+
+  try {
+    return await executeCommand('pod', ['install'], options);
+  } catch (err) {
+    return {error: ERRORS.failedToRunPodInstall};
```
grabbou: There's a special CLIError to which you can pass a `stderr` from the `executeCommand` (or the `err.message` in this case). That way, we are not going to swallow an original error message and could see it when running CLI with `--verbose` command.

In case you need a special error here, you would need to subclass `CLIError` to give it a different name

packages/cli/src/tools/packageManager.js
grabbou: Personally, I would leave `PackageManager` as is (since it's also used in different places I guess) and create a different function, `installPods`. 

That way, you will not have to throw your "special" errors, you could just handle the logic inside the `catch` block, e.g.:
```
 let isPodsInstalled = false;
 try {
    await commandExists('pod');
    isPodsInstalled = true;
  } catch (err) {
    isPodsInstalled = await installCocoaPods(); // assuming you can skip it
  }

  if (isPodsInstalled) { installCocoaPodsDependencies() } else {}
```

grabbou: This is obviously a pseudocode so I'd improve it!

packages/cli/src/tools/packageManager.js
grabbou: However, you might not need to subclass errors if you decide to not do it inside PackageManager :)

packages/cli/src/tools/packageManager.js
grabbou: Yeah, good idea!

packages/cli/src/tools/packageManager.js
lucasbento: Alright, gonna work with this in mind.

packages/cli/src/tools/packageManager.js
thymikee: Can we first try without `sudo` and then fallback to use it? For users with custom Ruby (not bundled with the system) it's not necessary

lucasbento: Yeah, I can put that in.

packages/cli/src/commands/init/init.js
```diff
@@ -90,6 +93,83 @@ async function createFromExternalTemplate(
   }
 }
 
+async function installPods(loader: typeof Ora) {
+  process.chdir('ios');
```
thymikee: dir is changed, but it's not updated after this fn ends/fails

packages/cli/src/commands/init/init.js
```diff
@@ -74,14 +79,12 @@ async function createFromExternalTemplate(
     const {postInitScript} = templateConfig;
     if (postInitScript) {
       // Leaving trailing space because there may be stdout from the script
-      loader.start('Executing post init script ');
```
thymikee: please revert

lucasbento: Completely missed the comment.



Thanks for seeing this.

lucasbento:
@grabbou: requested changes applied.

I also added a verification for `Podfile` before trying to install pods.

How can I test the `createFromReactNativeTemplate` function locally? I'm doing `--template /path/to/rn` currently to test `createFromExternalTemplate`.

lucasbento:
Any idea of why the test for `init` is failing? can't reproduce it locally.

Tried on node 8 and 10.

grabbou:
@lucasbento it's flaky, ignore it for now

packages/cli/src/commands/init/init.js
```diff
@@ -90,6 +93,90 @@ async function createFromExternalTemplate(
...
+  const hasPods = await fs.pathExists('Podfile');
+
+  if (!hasPods) {
+    return;
+  }
+
+  try {
+    await commandExists('pod');
+  } catch (err) {
+    loader.succeed();
```
grabbou: This will mark something as succeeded, where do we print the message? I can't see it inside this function, so might be worth moving it here?

packages/cli/src/commands/init/init.js
```diff
@@ -90,6 +93,90 @@ async function createFromExternalTemplate(
...
+
+    const {shouldInstallCocoaPods} = await inquirer.prompt([
+      {
+        type: 'confirm',
+        name: 'shouldInstallCocoaPods',
+        message: 'CocoaPods is not installed, Do you want to install it?',
```
grabbou: nit: lowercase "do"

packages/cli/src/commands/init/init.js
```diff
@@ -90,6 +93,90 @@ async function createFromExternalTemplate(
...
+    ]);
+
+    if (shouldInstallCocoaPods) {
+      try {
+        await execa('sudo', ['gem', 'install', 'cocoapods'], {
+          stdio: 'pipe',
+        });
+      } catch (err) {
+        throw new CLIError(
+          'Error occurred while trying to install CocoaPods, please install it manually.',
```
grabbou: idea: maybe paste a link to official documentation?

grabbou: nit: we can add "and run this command once again"

packages/cli/src/commands/init/init.js
```diff
@@ -90,6 +93,90 @@ async function createFromExternalTemplate(
...
+      } catch (err) {
+        throw new CLIError(
+          'Failed to run "pod install", please try to run it manually.',
+          err,
+        );
+      }
+    }
+  }
+
+  try {
```
grabbou: This shouldn't be here, right? It's outside of the if statement.

packages/cli/src/commands/init/init.js
```diff
@@ -90,6 +93,90 @@ async function createFromExternalTemplate(
...
+          'Error occurred while trying to install CocoaPods, please install it manually.',
+          err,
+        );
+      }
+
+      try {
+        loader.start('Installing pods');
+      } catch (err) {
+        throw new CLIError(
+          'Failed to run "pod install", please try to run it manually.',
```
grabbou: We don't run `pod install` here (check the `try` clause). I think you had a issue while refactoring, because correct code is on line 141 :)

packages/cli/src/commands/init/init.js
```diff
@@ -90,6 +93,90 @@ async function createFromExternalTemplate(
...
+  loader.start('Installing all required dependencies');
+
+  await PackageManager.installAll({
+    preferYarn: !npm,
+    silent: true,
+  });
+
+  try {
+    await installPods(loader);
+
```
grabbou: nit: since we change `chdir` inside `installPods`, I would also bring it back inside it and here, just call `await installPods` w/o any try/catch.

Feels cleaner

packages/cli/src/commands/init/init.js
```diff
@@ -90,6 +93,90 @@ async function createFromExternalTemplate(
...
+}
+
+async function installDependencies({
+  npm,
+  loader,
+}: {
+  npm?: boolean,
+  loader: typeof Ora,
+}) {
+  loader.start('Installing all required dependencies');
```
grabbou: nit: installing Node dependencies?

packages/cli/src/commands/init/init.js
grabbou: Actually, I think we don't need `loader` in this function at all. We can just call `succeed` in the `installDependencies` as we do right now.

Returning from this function w/o error will automatically succeed.

Am I right? Looking at this code, we will succeed if `pod` is available and mark `installing required dependencies` as done.

lucasbento: I believe that's right as I already added `loader.succeed()` on `installDependencies` after calling this function.

Great catch!

packages/cli/src/commands/init/init.js
lucasbento: What would be the link the documentation?

Btw I just found this https://facebook.github.io/react-native/docs/integration-with-existing-apps#3-install-cocoapods which recommends using `brew` to install cocoapods 😛

lucasbento: > nit: we can add "and run this command once again"

Do you mean something like: "Error occurred while trying to install CocoaPods, please run this command again."? If so, is that a good idea? asking because we can't guarantee that this will work on a second time.

packages/cli/src/commands/init/init.js
lucasbento: I, however, need the `loader` being passed as arguments because I use it to specify `Installing pods` after `cocoapods` gets successfully installed.

packages/cli/src/commands/init/init.js
lucasbento: It should. The first `try...catch` is for handling if `cocoapods` is installed, if it is it goes straight to this one that install the pods, if not it goes to the installation of `cocoapods` that then triggers the loader with `Installing pods` and the actual command on line 141.

packages/cli/src/commands/init/init.js
lucasbento: Hmm, if we do this:

```
await installPods(loader);

process.chdir('..');

loader.succeed();
```

The `process.chdir('..');` will never be called if `installPods` fails, is that what you mean?

packages/cli/src/commands/init/init.js
lucasbento: I removed the `loader.succeed()` and it doesn't work without it.

The reason is that it keeps updating the loader and even after I use `inquirer.prompt` the loader erases it and gets on top of it, hence the reason to stop it with a `succeed`.

packages/cli/src/tools/packageManager.js
lucasbento: Done.

packages/cli/src/commands/init/init.js
lucasbento: Fixed.

packages/cli/src/commands/init/init.js
lucasbento: Just changed this, if this is not it then let me know and I'll change it again.

packages/cli/src/commands/init/init.js
lucasbento: Done!

packages/cli/src/commands/init/init.js
lucasbento: Good one, fixed.

packages/cli/src/commands/init/init.js
grabbou: Gotcha, so let's maybe add a comment to mention it's needed for `inquirer.prompt` to work.

packages/cli/src/commands/init/init.js
```diff
@@ -90,6 +93,87 @@ async function createFromExternalTemplate(
...
+        message: 'CocoaPods is not installed, do you want to install it?',
+      },
+    ]);
+
+    if (shouldInstallCocoaPods) {
+      try {
+        // First attempt to install `cocoapods`
+        await execa('gem', ['install', 'cocoapods'], {
+          stdio: 'pipe',
+        });
```
grabbou: @thymikee, I know you suggested installing WITHOUT `sudo`, but according to CocoaPods installation guide, users will need to modify the $PATH themselves in order to make this one work. 

Otherwise `pod` command will not be available.

I'd just go with `sudo` for those who don't have it to make this easier. It's still recommended way.

lucasbento:
Tests are failing but I don't think it's because of any change here.

packages/cli/src/commands/init/init.js
grabbou: Ah, right! Then... shouldn't the "loader.start('Installing pods');" be on line 141, between both try/catch'es?

lucasbento: Can be, I didn't put it there because I didn't want to bug the user with another loading if they are both part of the dependencies, it only shows if the user has been prompted to install cocoapods.

Do you want to show it every time it runs `pod install`?

grabbou: Ah, right! Makes sense. When `pod` is already there, we will transparently install it w/o showing the loader.

I like that

lucasbento: Great!

packages/cli/src/commands/init/printRunInstructions.js
```diff
@@ -9,19 +9,22 @@
...
-  )}.xcodeproj`;
+  const iosProjectDir = path.resolve(projectDir, 'ios');
+
+  const iosPodsFile = path.resolve(iosProjectDir, `${projectName}.xcworkspace`);
+  const isUsingPods = await fs.pathExists(iosPodsFile);
+
   const relativeXcodeProjectPath = path.relative(
     process.cwd(),
-    xcodeProjectPath,
+    isUsingPods
```
lucasbento: @grabbou: it's good to check this out as well, I assume it's fine but it's better to be sure.

I noticed it yesterday that I was assuming that it was `.xcworkspace` without verifying first.

packages/cli/src/commands/init/init.js
thymikee: I'm pretty sure it's automatically handled when you install ruby with brew

grabbou: I think the default `ruby` that most people have is the one that comes with a Mac. For example, that's the one I am using myself.

Anyway, let's check this. We need to be 100% sure that we are shipping the best experience possible here.

CC: @orta, what do you think?

orta: I think this is reasonable, editing the `$PATH` overall is definitely preferable in the long run - but we do recommend using `sudo` in a few places because it works and doesn't affect the end results of a pod installation (in terms of permissions etc). I'm actually on a fresh computer so this is what it'd look like under the hood:

```
Last login: Tue May  7 09:39:43 on ttys002
Ortas-MacBook-Pro:~ orta$ ruby -v
ruby 2.3.7p456 (2018-03-28 revision 63024) [universal.x86_64-darwin18]

Ortas-MacBook-Pro:~ orta$ pod install
-bash: pod: command not found

Ortas-MacBook-Pro:~ orta$ gem install cocoapods
Fetching: ruby-macho-1.4.0.gem (100%)
ERROR:  While executing gem ... (Gem::FilePermissionError)
    You don't have write permissions for the /Library/Ruby/Gems/2.3.0 directory.

Ortas-MacBook-Pro:~ orta$ sudo gem install cocoapods
Password:
Fetching: ruby-macho-1.4.0.gem (100%)
Successfully installed ruby-macho-1.4.0
Fetching: nap-1.1.0.gem (100%)
Successfully installed nap-1.1.0
Fetching: gh_inspector-1.1.3.gem (100%)
Successfully installed gh_inspector-1.1.3
Fetching: fourflusher-2.2.0.gem (100%)
...
Parsing documentation for cocoapods-1.6.1
Installing ri documentation for cocoapods-1.6.1
Done installing documentation for ruby-macho, nap, gh_inspector, fourflusher, escape, colored2, concurrent-ruby, i18n, thread_safe, tzinfo, activesupport, atomos, CFPropertyList, claide, nanaimo, xcodeproj, molinillo, cocoapods-try, netrc, cocoapods-trunk, cocoapods-stats, cocoapods-search, cocoapods-plugins, cocoapods-downloader, cocoapods-deintegrate, fuzzy_match, cocoapods-core, cocoapods after 17 seconds
28 gems installed

Ortas-MacBook-Pro:~ orta$ pod -v
[!] Unknown command: `-v`
Did you mean: env?

Usage:

    $ pod COMMAND

      CocoaPods, the Cocoa library package manager.

Commands:

    + cache         Manipulate the CocoaPods cache
    + deintegrate   Deintegrate CocoaPods from your project
    + env           Display pod environment
    + init          Generate a Podfile for the current directory
    + install       Install project dependencies according to versions from a
                    Podfile.lock
    + ipc           Inter-process communication
    + lib           Develop pods
    + list          List pods
    + outdated      Show outdated project dependencies
    + plugins       Show available CocoaPods plugins
    + repo          Manage spec-repositories
    + search        Search for pods
    + setup         Setup the CocoaPods environment
    + spec          Manage pod specs
    + trunk         Interact with the CocoaPods API (e.g. publishing new specs)
    + try           Try a Pod!
    + update        Update outdated project dependencies and create new Podfile.lock

Options:

    --silent        Show nothing
    --version       Show the version of the tool
    --verbose       Show more debugging information
    --no-ansi       Show output without ANSI codes
    --help          Show help banner of specified command
```

Which is fine.

lucasbento: Shall I keep only with `sudo` then?

grabbou: I think it's fine like this. @orta just demonstrated an example where he did run sudo-less first and with sudo second time, so let's leave it.

orta: Yep, prefer the sudo-less one but expect it to fail

lucasbento:
All requested changes have been made, @grabbou & @thymikee can you take a look?

packages/cli/src/commands/init/init.js
```diff
@@ -104,6 +107,87 @@ async function createFromTemplate({
...
+      } catch (err) {
+        try {
+          // If that doesn't work then try with sudo
+          await execa('sudo', ['gem', 'install', 'cocoapods'], {
+            stdio: 'pipe',
```
thymikee: pipe is the default, you can remove it everywhere

packages/cli/src/commands/init/init.js
```diff
@@ -104,6 +107,87 @@ async function createFromTemplate({
...
+  }
+
+  try {
+    await execa('pod', ['install'], {
+      stdio: 'pipe',
+    });
+  } catch (err) {
+    throw new CLIError(
+      'Failed to run "pod install", please try to run it manually.',
+      err,
```
thymikee: while testing locally I noticed that when `pod install` fails it doesn't output anything to the console and the error is pretty unhelpful (imagine being a beginner: what's "pod install", why is it running, where do I run it?). How about:

```
Failed to install CocoaPods  dependencies for iOS project, which is required by this template. 
Please try again manually: "cd ./${ProjectName}/ios && pod install".
CocoaPods documentation: ${chalk.dim.underline("https://cocoapods.org/")}
```

The fact that `pod` outputs failures to stdout is not helpful as well. I think we should show the `console.log(err.stderr || err.stdout)` straight away before throwing this.

packages/cli/src/commands/init/init.js
```diff
@@ -104,6 +107,87 @@ async function createFromTemplate({
...
+          });
+        } catch (err) {
+          throw new CLIError(
+            'Error occurred while trying to install CocoaPods, please run this command again.',
+            err,
+          );
+        }
+      }
+
+      loader.start('Installing pods');
```
thymikee: this should be moved right before `await execa('pod', ['install'])` is called

packages/cli/src/commands/init/init.js
```diff
@@ -104,6 +107,87 @@ async function createFromTemplate({
...
+  const hasPods = await fs.pathExists('Podfile');
+
+  if (!hasPods) {
+    return;
+  }
+
+  try {
+    await commandExists('pod');
+  } catch (err) {
+    loader.succeed();
```
thymikee: you can call `loader.stop()` to stop the spinner, I think it makes more sense

packages/cli/src/commands/init/init.js
```diff
@@ -144,7 +228,7 @@ export default (async function initialize(
   try {
     await createProject(projectName, options, version);
 
-    printRunInstructions(process.cwd(), projectName);
+    await printRunInstructions(process.cwd(), projectName);
```
thymikee: after using `fs.pathExistsSync` this change won't be necessary

packages/cli/src/commands/init/init.js
```diff
@@ -104,6 +107,87 @@ async function createFromTemplate({
...
+        });
+      } catch (err) {
+        try {
+          // If that doesn't work then try with sudo
+          await execa('sudo', ['gem', 'install', 'cocoapods'], {
+            stdio: 'pipe',
+          });
+        } catch (err) {
+          throw new CLIError(
+            'Error occurred while trying to install CocoaPods, please run this command again.',
```
thymikee: which command? we should point users to what they can do. I'd suggest:
1) showing what command was used
2) link to cocoapods documentation

packages/cli/src/commands/init/printRunInstructions.js
```diff
@@ -9,19 +9,22 @@
...
   const absoluteProjectDir = path.resolve(projectDir);
-  const xcodeProjectPath = `${path.resolve(
-    projectDir,
-    'ios',
-    projectName,
-  )}.xcodeproj`;
+  const iosProjectDir = path.resolve(projectDir, 'ios');
+
+  const iosPodsFile = path.resolve(iosProjectDir, `${projectName}.xcworkspace`);
+  const isUsingPods = await fs.pathExists(iosPodsFile);
```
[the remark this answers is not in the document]
lucasbento: What would the difference be between using `fs-extra` or `fs`? I'm using `fs-extra` because it's already being used on `init`.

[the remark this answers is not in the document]
thymikee: It's an extra dependency to download and I'd like to get rid of it some day, but likely not in forseeable future

packages/cli/src/commands/init/init.js
lucasbento: This is the reason it's not after the condition: https://github.com/react-native-community/cli/pull/373#discussion_r282442627.

thymikee: A comment with the explanation would be nice, because it seems unintuitive

packages/cli/src/commands/init/init.js
```diff
@@ -104,6 +107,87 @@ async function createFromTemplate({
...
+
+  await PackageManager.installAll({
+    preferYarn: !npm,
+    silent: true,
+  });
+
+  if (process.platform === 'darwin') {
+    await installPods(loader);
+
+    process.chdir('..');
```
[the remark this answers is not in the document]
lucasbento: Sorry, I don't follow, what do you mean?

packages/cli/src/commands/init/init.js
lucasbento: I'll add it.

packages/cli/src/commands/init/init.js
lucasbento: > The fact that pod outputs failures to stdout is not helpful as well.

Do you mean this when you run on `--verbose`? If you run on with it enabled don't you already have the output of errors on `pod install`?

packages/cli/src/commands/init/init.js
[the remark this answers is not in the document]
thymikee: At current state, if somebody changes the implementation of `installPods` to not change dir to `./ios`, and will simply forget to change this line (because it's outside of the fn) we'll end up changing directory to something unintended. 

What I want is to collocate `process.chdir` calls, so they happen in the same code/function block. Either:

```js
// process.chdir calls are close to each other:
try {
  process.chdir('./ios');
  await installPods(loader)
} catch(error) {
  throw error
} finally {
  process.chdir('..')
}
```

or 
```js
function installPods(loader) {
 try {
  process.chdir('./ios');
  // do stuff...
 } catch(error) {
  throw error
 } finally {
  process.chdir('..')
 }
}
```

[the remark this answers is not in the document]
lucasbento: Ah I see, that's awesome, I'll do it.

Thanks!

packages/cli/src/commands/init/init.js
thymikee: It shows up with verbose, but really, would you like to run whole `init` again with `--verbose` flag just to see the error? It's a bad idea. I'm leaning towards the `CLIError` to show the error stack by default if it's there (not related to this PR though)

lucasbento: That's a good point, I'll try the `console.log` you suggested, I thought that without `--verbose` they were already suppressed.

thymikee:
One more thing, can we extract the `installPods` function to a helper function? We're going to need it in legacy init as well

lucasbento:
@thymikee sure, where shall I place it?

thymikee:
`packages/cli/src/tools/` should be fine

lucasbento:
@thymikee: changes done, please review again :)

thymikee:
Thank you @lucasbento!

Issue #374, opened 2019-05-07
Cannot read property 'update' of undefined


yinhangfeng:
## Environment

```
  React Native Environment Info:
    System:
      OS: macOS 10.14.4
      CPU: (8) x64 Intel(R) Core(TM) i7-4770HQ CPU @ 2.20GHz
      Memory: 26.78 MB / 16.00 GB
      Shell: 5.3 - /bin/zsh
    Binaries:
      Node: 10.15.0 - ~/.nvm/versions/node/v10.15.0/bin/node
      Yarn: 1.13.0 - /usr/local/bin/yarn
      npm: 6.9.0 - ~/.nvm/versions/node/v10.15.0/bin/npm
      Watchman: 4.9.0 - /usr/local/bin/watchman
    SDKs:
      iOS SDK:
        Platforms: iOS 12.2, macOS 10.14, tvOS 12.2, watchOS 5.2
      Android SDK:
        API Levels: 21, 22, 23, 24, 25, 26, 27, 28
        Build Tools: 22.0.1, 23.0.3, 24.0.3, 25.0.3, 26.0.3, 27.0.3, 28.0.2, 28.0.3
        System Images: android-27 | Google APIs Intel x86 Atom, android-28 | Google APIs Intel x86 Atom
    IDEs:
      Android Studio: 3.4 AI-183.5429.30.34.5452501
      Xcode: 10.2/10E125 - /usr/bin/xcodebuild
    npmPackages:
      react: 16.8.3 => 16.8.3 
      react-native: git+ssh://git@code.timehouse.build:frontend-rn/react-native.git#wind-0.59-stable => 0.59.15 
    npmGlobalPackages:
      react-native-cli: 2.0.1
```

## Description

```
error Cannot read property 'update' of undefined
debug TypeError: Cannot read property 'update' of undefined
    at new Server (/Users/yinhf/work/wind/wired-app/node_modules/metro/src/Server.js:675:29)
    at buildBundle (/Users/yinhf/work/wind/wired-app/node_modules/@react-native-community/cli/build/commands/bundle/buildBundle.js:91:18)
```

Server.js
```
    if (this._config.resetCache) {
      this._config.cacheStores.forEach(store => store.clear());

      this._config.reporter.update({
        type: "transform_cache_reset"
      });
    }
```

## Reproducible Demo

maybe this pr #371 causes the reporter to be empty

thymikee:
Released 1.9.4 with a revert, thanks for reporting!

[...]

PR #376, opened 2019-05-07
fix: config.reporter undefined bug

yinhangfeng:
fix: #374

### DIFF
--- a/packages/cli/src/tools/loadMetroConfig.js
+++ b/packages/cli/src/tools/loadMetroConfig.js
@@ -30,7 +30,7 @@ const getBlacklistRE = () => createBlacklist([/.*\/__fixtures__\/.*/]);
  * @todo(grabbou): As a separate PR, haste.platforms should be added before "native".
  * Otherwise, a.native.js will not load on Windows or other platforms
  */
-export const getDefaultConfig = (ctx: ConfigT) => {
+export const getDefaultConfig = (ctx: ConfigT, options?: ConfigOptionsT) => {
   return {
     resolver: {
       resolverMainFields: ['react-native', 'browser', 'main'],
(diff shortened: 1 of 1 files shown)


thymikee:
Thanks! Mind taking a look at Flow error?

thymikee:
@yinhangfeng I've updated your PR, can you check if it works for you?

yinhangfeng:
> @yinhangfeng I've updated your PR, can you check if it works for you?

it works, thank you

thymikee:
cc @fson

grabbou:
Looks good! I wouldn't touch `getDefaultConfig` function tho, I would just keep the logic where it was before this PR.

grabbou:
@yinhangfeng once again, thank you for sending this PR! Please let me know if you are going to have some time anytime soon to address the feedback. We would like to ship this PR soon and make a new release of the CLI!

If not, it's all right :) We will rebase and continue work on top of your PR.

thymikee:
@grabbou can you share some insights on why you think modifying the `getDefaultConfig` fn is bad here? To me it corresponds nicely, because `reporter` is a part of Metro config and in certain cases it is a default for us, with or without it. It feels even more clunky adding `reporter` conditionally to the already constructed default config.

grabbou:
I don’t like coupling default configuration with CLI options which are
meant to override the default, inferred (from context object) configuration
that this function returns.

There used to be a separate function for overriding config with CLI options
which we got rid of as Metro implemented this internally.

Passing the whole options to the function is in my opinion confusing,
because we do not explicitly handle all other CLI options, except for
reporter. We leave the rest up to Metro.

grabbou:
@thymikee please review after changes.

PR #378, opened 2019-05-07
Use `require.resolve` to resolve package path

janicduplessis:
Summary:
---------

This fixes mono repo support by using `require.resolve` to find the absolute path of dependencies.

Test Plan:
----------

Tested that the path resolution works in a mono repo project where RN and most deps are hoisted.

### DIFF
--- a/packages/cli/src/tools/config/__mocks__/resolveNodeModuleDir.js
+++ b/packages/cli/src/tools/config/__mocks__/resolveNodeModuleDir.js
@@ -4,7 +4,7 @@
 
 const path = require('path');
 
-export default function resolveNodeModule(
+export default function resolveNodeModuleDir(
   root: string,
   packageName: string,
 ): string {
--- a/packages/cli/src/tools/config/resolveNodeModule.js
+++ b/packages/cli/src/tools/config/resolveNodeModule.js
@@ -1,0 +1,19 @@
+/**
+ * @flow
+ */
+import path from 'path';
+
+/**
+ * Finds a path inside `node_modules`
+ */
+export default function resolveNodeModule(
+  root: string,
+  packageName: string,
+): string {
+  return path.dirname(
+    // $FlowIssue: Wrong `require.resolve` type definition
+    require.resolve(path.join(packageName, 'package.json'), {
+      paths: [root],
+    }),
+  );
+}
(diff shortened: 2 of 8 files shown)


thymikee:
Some tests are failing, mind taking a look?

janicduplessis:
@thymikee Fixed, I had to extract it to a module to mock it for tests since `require.resolve` doesn't work with the mock fs and throws if the module doesn't exist. I also moved the call inside the try catch to make sure we keep the same behaviour if a module doesn't exist.

packages/cli/src/tools/config/resolveNodeModule.js
```diff
@@ -0,0 +1,19 @@
+/**
+ * @flow
+ */
+import path from 'path';
+
+/**
+ * Finds a path inside `node_modules`
+ */
+export default function resolveNodeModule(
```
thymikee: Hm, not sure about the name, maybe `resolveNodeModueDir` would be closer to what it does? We're not resolving a module (a JS file) but the directory it's in. What do you think?

janicduplessis: Makes sense, updated

packages/cli/src/tools/config/__mocks__/resolveNodeModuleDir.js
```diff
@@ -0,0 +1,12 @@
...
+ * @flow
+ */
+
+const path = require('path');
+
+export default function resolveNodeModuleDir(
+  root: string,
+  packageName: string,
+): string {
+  return path.join(root, 'node_modules', packageName);
```
grabbou: Shouldn't this use `require.resolve` as per PR desc? Otherwise, it seems to be same as before the PR, unless I am blind at this time of the day!

thymikee: psst, this is a mock

grabbou: so I was right XD I am blind

thymikee: don't worry, I had the same wtf moment 😅

thymikee:
@janicduplessis you can find this fix in `2.0.0-alpha.18`

PR #383, opened 2019-05-11
Add openURLMiddleware

ericlewis:
Summary:
---------

This allow’s react-native to open Chrome windows arbitrarily.

Test Plan:
----------

Post to bundle URL on the /open-url endpoint with a JSON blob: `{"url": "https://www.google.com"}`. See Chrome open to google.com.


packages/cli/src/commands/server/middleware/openURLMiddleware.js
```diff
@@ -0,0 +1,25 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @format
+ */
+
+import launchChrome from '../launchChrome';
```
lucasbento: I would advise to not use `launchChrome` as it only launches chrome and not the default browser.

orta: +1 for this (as a Safari user) - thanks for this PR though!

ericlewis: Ah, I am also a safari user heh. I will make a change for this.

ericlewis:
@thymikee I was looking for an example of this, can you point me to one?

packages/cli/src/commands/server/middleware/openURLMiddleware.js
orta: I think you can use [`opn`](https://github.com/react-native-community/cli/blob/master/packages/cli/src/commands/server/launchChrome.js#L11) directly with the URL and it _should_ just work

ericlewis: @orta correct! that is what i did.

ericlewis:
Should we base launchChrome off of this? Also, I took a look at the tests in open. They aren't all that useful.. I am fine with creating some sort of test ofc. But, I would need to see its value first.

thymikee:
By a test I meant a simple unit test for the middleware to make sure launchBrowser and next is called with proper params, similar for launchBrowser function now. 

E2e test would be cool one day but it's too much of a hassle for this PR.

I'm cool with refactoring the launchChrome now or in a followup

PR #384, opened 2019-05-12
fix: run "config" from react-native binary in autolinking scripts

thymikee:
Summary:
---------

Since `@react-native-community/cli` is not declared as template dependency, there is a possibility for Node to not be able to resolve it. It may happen because package managers can put the package inside `react-native/node_modules/` behind a symlink (like `npm` does in some scenarios) which is invisible to `require.resolve`. 

This diff changes autolinking scripts to use `config` command directly from `react-native` binary instead, as `react-native` is a direct dependency of any template.

Because running node binaries differs between package managers (e.g. Yarn in PnP mode doesn't have `node_modules/` so no `node_modules/.bin/` as well), the best thing we can do is to directly run the command through a package manager. For now Yarn and npm are supported, I'm happy to extend support for pnpm as well in the future if requested.

What happens now:
1) if the project root has Yarn lockfile, use `yarn run react-native config`
2) if not, fallback to `node ./node_modules/.bin/react-native config` (could be replaced with less safe `npx` later)

Test Plan:
----------

Try init with tarball RN master.

### DIFF
--- a/packages/platform-ios/native_modules.rb
+++ b/packages/platform-ios/native_modules.rb
@@ -5,12 +5,10 @@
 #
 def use_native_modules!(root = "..", packages = nil)
   if (!packages)
-    # Resolve the CLI's main index file
-    cli_bin = Pod::Executable.execute_command("node", ["-e", "console.log(require.resolve('@react-native-community/cli/build/index.js'))"], true).strip
     output = ""
     # Make sure `react-native config` is ran from your project root
     Dir.chdir(root) do
-      output = Pod::Executable.execute_command("node", [cli_bin, "config"], true)
+      output = Pod::Executable.execute_command("node", ["./node_modules/.bin/react-native", "config"], true)
     end
 
     json = []
--- a/packages/platform-ios/native_modules.rb
+++ b/packages/platform-ios/native_modules.rb
@@ -5,15 +5,28 @@
 #
 def use_native_modules!(root = "..", packages = nil)
   if (!packages)
+    command = "node"
+    args = ["./node_modules/.bin/react-native", "config"]
+    begin
+      # Check if project uses Yarn
+      Pod::Executable.execute_command("node", ["-e", "console.log(require.resolve('#{root}/yarn.lock'))"], true)
+      command = "yarn"
+      args = ["run", "react-native", "config"]
+    rescue
+    end
+
     output = ""
     # Make sure `react-native config` is ran from your project root
     Dir.chdir(root) do
-      output = Pod::Executable.execute_command("node", ["./node_modules/.bin/react-native", "config"], true)
+      output = Pod::Executable.execute_command(command, args, true)
     end
 
     json = []
     output.each_line do |line|
       case line
+      when /^yarn run/
+      when /^\$/
+      when /^Done/
       when /^warn\s(.+)/
         Pod::UI.warn($1)
       when /^(success|info|error|debug)\s(.+)/
--- a/packages/platform-ios/native_modules.rb
+++ b/packages/platform-ios/native_modules.rb
@@ -11,7 +11,7 @@ def use_native_modules!(root = "..", packages = nil)
       # Check if project uses Yarn
       Pod::Executable.execute_command("node", ["-e", "console.log(require.resolve('#{root}/yarn.lock'))"], true)
       command = "yarn"
-      args = ["run", "react-native", "config"]
+      args = ["run", "--silent", "react-native", "config"]
     rescue
     end
 
@@ -24,9 +24,6 @@ def use_native_modules!(root = "..", packages = nil)
     json = []
     output.each_line do |line|
       case line
-      when /^yarn run/
-      when /^\$/
-      when /^Done/
       when /^warn\s(.+)/
         Pod::UI.warn($1)
       when /^(success|info|error|debug)\s(.+)/
--- a/packages/platform-android/native_modules.gradle
+++ b/packages/platform-android/native_modules.gradle
@@ -54,7 +54,8 @@ class ReactNativeModules {
   private ArrayList<HashMap<String, String>> reactNativeModules
 
   private static String LOG_PREFIX = ":ReactNative:"
-  private static String REACT_NATIVE_CONFIG_CMD = "node ./node_modules/.bin/react-native config"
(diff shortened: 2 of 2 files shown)


grabbou:
Hm, the `react-native` binary is declared inside `CLI`, not inside a `react-native` anymore (we have changed that).

I don't think this PR changes anything rather than using more elegant bin interface (which is nice tho)

thymikee:
`react-native` on [master](https://github.com/facebook/react-native/blob/master/package.json#L4) and [60-stable](https://github.com/facebook/react-native/blob/ec83a8d967251f36e52f5b8bc5dbef573282a482/package.json#L4) is still a binary, that's all I care about in this PR.

One question I have is when deps are installed with Yarn PnP. @arcanis will `node node_modules/.bin/bin-name` work in PnP? If not, what's the alternative?

grabbou:
[Looks like someone must have added it back](https://github.com/facebook/react-native/commit/ab1edae669de4d52dbeeeaa948f6b65e86717479).

For reference, we also define it here: https://github.com/react-native-community/cli/blob/master/packages/cli/package.json#L8

arcanis:
> One question I have is when deps are installed with Yarn PnP. @arcanis will `node node_modules/.bin/bin-name` work in PnP? If not, what's the alternative?

The most portable way is to call the binary through the package manager:

```
$> yarn run bin-name
```

This is supported in all environments (Yarn, PnP, npm).

thymikee:
@arcanis thank you! Forgot about `npm run` completely (after being so used to `yarn bin-name`), will update the PR soon.

thymikee:
Ok, little update:

Because running Node binaries differs between package managers (e.g. Yarn in PnP mode doesn't have `node_modules/` so no `node_modules/.bin/` as well), the best thing we can do is to directly run the command through a package manager. For now Yarn and npm are supported, I'm happy to extend support for pnpm as well in the future if requested.

What happens now:
1) if the project root has Yarn lockfile, use `yarn run react-native config`
2) if not, fallback to `node ./node_modules/.bin/react-native config` (could be replaced with less safe `npx` later)
3) filter out non-essential stuff to leave only JSON from the `config` command, as  package managers (like Yarn) output extra lines to stdout (e.g. "yarn run <version>" or "$ /path/to/command".)

@Salakar and @orta could you please check implementations of `native_modules.{rb,gradle}` and test it in a project?

packages/platform-ios/native_modules.rb
```diff
@@ -5,17 +5,28 @@
...
   if (!packages)
-    # Resolve the CLI's main index file
-    cli_bin = Pod::Executable.execute_command("node", ["-e", "console.log(require.resolve('@react-native-community/cli/build/index.js'))"], true).strip
+    command = "node"
+    args = ["./node_modules/.bin/react-native", "config"]
+    begin
+      # Check if project uses Yarn
+      Pod::Executable.execute_command("node", ["-e", "console.log(require.resolve('#{root}/yarn.lock'))"], true)
+      command = "yarn"
+      args = ["run", "react-native", "config"]
```
arcanis: Protip: If you run `yarn run --silent react-native config` you shouldn't get the headers and footers 🙂

thymikee: Thanks! I tried that and was pretty sure I've still seen the headers/footers with `-s` flag, but clearly must have missed it, because it works now :D

packages/platform-ios/native_modules.rb
```diff
@@ -5,12 +5,20 @@
 #
 def use_native_modules!(root = "..", packages = nil)
   if (!packages)
-    # Resolve the CLI's main index file
-    cli_bin = Pod::Executable.execute_command("node", ["-e", "console.log(require.resolve('@react-native-community/cli/build/index.js'))"], true).strip
+    command = "node"
+    args = ["./node_modules/.bin/react-native", "config"]
+    begin
+      # Check if project uses Yarn
+      Pod::Executable.execute_command("node", ["-e", "console.log(require.resolve('#{root}/yarn.lock'))"], true)
```
grabbou: I wonder if it wouldn't been easier to just get "__dirname" using either JavaScript, Ruby or Gradle and based on that, calculate a relative path?

E.g. `node ../../build/index.js config`

grabbou:
How one could reference an auto linking script from Podfile or Gradle configuration when using Yarn PnP? I think our guides explicitly suggest referring `node_modules` (including our attempts in the default template), which may or may not be an issue.

thymikee:
@grabbou this is a really good question. I've actually already bumped into issues with requiring from node_modules when installing with Yarn vs npm. E.g npm tends to put `@react-native-community/cli` into `node_modules/react-native/node_modules` while we don't account for that (we just blindly assume to find it directly inside root `node_modules/`). 

Also just realized that the way we use Xcode, CocoaPods and Gradle (we require files directly from `../node_modules` in a bunch of places) is not compatible with Yarn PnP and I don't know how to resolve it. I've bumped into https://github.com/facebook/react-native/issues/22163 which got closed by a bot. So yea, we're still not pnp-compliant, but this will at least not make us further away from the goal.

To be 100% sure we work with most common setups (regular Yarn and npm) we could add proxies to our Ruby and Gradle scripts to `react-native` core package. This way we could reference them directly in a safer manner.

grabbou:
Thanks @thymikee for providing this explanation. Makes sense!

> To be 100% sure we work with most common setups (regular Yarn and npm) we could add proxies to our Ruby and Gradle scripts to react-native core package. This way we could reference them directly in a safer manner.

Yeah, I feel like simple:
```ruby
Pod::Executable.execute_command("node", ["-e", "console.log(__dirname)"], true)
```
to get you a current location and making your way to `build/index.js` would be the way to go. What do you think?

thymikee:
The thing is that `__dirname/.../build/index.js` doesn't exist on a disk in PnP mode, it's all stored in `.pnp.js` file and resolved by Node. I highly encourage you to read the brief docs on it: https://yarnpkg.com/lang/en/docs/pnp/

grabbou:
Unfortunately, using `__dirname` approach will not work in this context because link files and the executable CLI file live in two different packages.

If they were living in the same package, this would be the safest approach.

PR #386, opened 2019-05-14
Add `installPods` to legacy init

lucasbento:
Summary:
---------

This PR is related to #368 and adds `installPods` to the legacy init.

Test Plan:
----------

Follow https://github.com/react-native-community/cli/blob/master/CONTRIBUTING.md#testing-init-command.

### DIFF
--- a/packages/cli/src/tools/installPods.js
+++ b/packages/cli/src/tools/installPods.js
@@ -12,7 +12,7 @@ async function installPods({
   loader,
 }: {
   projectName: string,
-  loader: typeof Ora,
+  loader?: typeof Ora,
 }) {
   try {
     process.chdir('ios');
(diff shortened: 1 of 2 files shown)


packages/cli/src/tools/installPods.js
```diff
@@ -26,7 +26,9 @@ async function installPods({
     try {
       await commandExists('pod');
     } catch (e) {
-      loader.stop();
+      if (loader) {
+        loader.stop();
```
grabbou: A bit ugly, but I guess there's nothing we can do about it!

grabbou:
Looks good to me!

packages/cli/src/tools/installPods.js
lucasbento: Thought the same.

thymikee: you could add "onInstallFailure" and "onInstallSuccess" callbacks and delegate to them to make it cleaner, but I don't care about it here. It's not _that_ ugly provided it's a small helper used in 2 places, not a public API surface :)

Issue #51, opened 2018-12-18
Health check command


deadcoder0904:
### Introduction

A way to verify react-native installation is complete and working?

### The Core of It

I was just checking out Flutter & I see there is a thing there called `flutter doctor` which ensures that installation is complete & working. I think it makes sense for React Native to have something like a `react-native doctor` or `react-native health` or something else.

### Discussion points

Might be beneficial for beginners who have a lot of trouble installing Android Studio if they are on anything else than a Mac

### Example of Flutter Doctor

Here's an output of the command `flutter doctor -v` in the terminal -

```bash
[✓] Flutter (Channel master, v1.1.2-pre.39, on Mac OS X 10.14.1 18B75, locale en-IN)
    • Flutter version 1.1.2-pre.39 at /Volumes/Coding/Code/Flutter/installation/flutter
    • Framework revision 0e9ad43416 (34 hours ago), 2018-12-16 16:39:28 -0500
    • Engine revision 4941125829
    • Dart version 2.2.0 (build 2.2.0-dev.1.1 f9ebf21297)

[✓] Android toolchain - develop for Android devices (Android SDK 28.0.3)
    • Android SDK at /Users/deadcoder0904/Library/Android/sdk
    • Android NDK location not configured (optional; useful for native profiling support)
    • Platform android-27, build-tools 28.0.3
    • ANDROID_HOME = /Users/deadcoder0904/Library/Android/sdk
    • Java binary at: /Applications/Android Studio.app/Contents/jre/jdk/Contents/Home/bin/java
    • Java version OpenJDK Runtime Environment (build 1.8.0_152-release-1136-b06)
    • All Android licenses accepted.

[✓] iOS toolchain - develop for iOS devices (Xcode 10.1)
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • Xcode 10.1, Build version 10B61
    • ios-deploy 1.9.4
    • CocoaPods version 1.5.3

[✓] Android Studio (version 3.2)
    • Android Studio at /Applications/Android Studio.app/Contents
    • Flutter plugin version 31.1.1
    • Dart plugin version 181.5656
    • Java version OpenJDK Runtime Environment (build 1.8.0_152-release-1136-b06)

[✓] VS Code (version 1.30.0)
    • VS Code at /Applications/Visual Studio Code.app/Contents
    • Flutter extension version 2.21.1

[!] Connected device
    ! No devices available

! Doctor found issues in 1 category.
```

[...]

PR #389, opened 2019-05-15
Improve CocoaPods prompt question when running  `init`

lucasbento:
Summary:
---------

This PR adds the CocoaPods documentation link to the prompt question when running `init`.

Test Plan:
----------

1. Run `sudo gem uninstall cocoapods` (or without sudo);
1. Run `react-native init`.

### DIFF
--- a/packages/cli/src/tools/installPods.js
+++ b/packages/cli/src/tools/installPods.js
@@ -32,7 +32,11 @@ async function installPods({
         {
           type: 'confirm',
           name: 'shouldInstallCocoaPods',
-          message: 'CocoaPods is not installed, do you want to install it?',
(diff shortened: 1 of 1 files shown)


lucasbento:
@cpojer & @thymikee: please review this.

Should we put something like "This may take a while"? if so, where would be the best?

lucasbento:
I can put a check to see if it's taking long, if it's, show something along the lines of "still working on it", thoughts?

lucasbento:
@thymikee: I can work on mocking `execa` so we can make sure of what is being executed, I'll probably get around this on #51.

cpojer:
Yeah, let's show something only if it takes longer than 60 seconds or so.

PR #390, opened 2019-05-15
feat: Show message if CocoaPods installation takes long

lucasbento:
Summary:
---------

This PR shows a different message if `cocoapods` installation takes more than 30 seconds.

Test Plan:
----------

1. `sudo gem uninstall cocoapods`;
1. Put `await execa('sleep', ['30']);`[here](https://github.com/lucasbento/cli/blob/5eb25ea43788a280266f3659ce1ec11f54525168/packages/cli/src/tools/installPods.js#L59).

---

### When it takes more than 30 seconds

Screenshot 2019-05-15 at 15 44 03

### Less than 30 seconds

Screenshot 2019-05-15 at 15 46 46

### DIFF
--- a/packages/cli/src/tools/installPods.js
+++ b/packages/cli/src/tools/installPods.js
@@ -40,14 +40,24 @@ async function installPods({
         },
       ]);
 
+      // This is only shown if it takes more than 30 seconds for the CocoaPods installation to finish
+      const cocoaPodsInstallationTimeMessage = setTimeout(
+        () => logger.info('Installing CocoaPods, this may take a few minutes'),
+        30000,
+      );
+
       if (shouldInstallCocoaPods) {
         try {
           // First attempt to install `cocoapods`
           await execa('gem', ['install', 'cocoapods']);
+
+          clearTimeout(cocoaPodsInstallationTimeMessage);
         } catch (_error) {
           try {
             // If that doesn't work then try with sudo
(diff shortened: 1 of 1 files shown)


packages/cli/src/tools/installPods.js
```diff
@@ -40,14 +40,24 @@ async function installPods({
...
+        30000,
+      );
+
       if (shouldInstallCocoaPods) {
         try {
           // First attempt to install `cocoapods`
           await execa('gem', ['install', 'cocoapods']);
+
+          clearTimeout(cocoaPodsInstallationTimeMessage);
         } catch (_error) {
```
thymikee: how about adding a `finally` block and clear timeout there?

packages/cli/src/tools/installPods.js
```diff
@@ -40,14 +40,24 @@ async function installPods({
         },
       ]);
 
+      // This is only shown if it takes more than 30 seconds for the CocoaPods installation to finish
+      const cocoaPodsInstallationTimeMessage = setTimeout(
+        () => logger.info('Installing CocoaPods, this may take a few minutes'),
+        30000,
```
thymikee: can we extract this to a variable like `COCOAPODS_INSTALLATION_TIMEOUT` so it's easier to update in future?

lucasbento:
Thanks for changing it @thymikee!

PR #393, opened 2019-05-17
feat: allow `dependencyConfig` to override `podspec` name

Salakar:
Summary:
---------

This change allows dependencies to specify a custom name of the podspec to use when linking / auto-linking ([used here in auto-linking](https://github.com/react-native-community/cli/blob/master/packages/platform-ios/native_modules.rb#L44))

Test Plan:
----------

### DIFF
--- a/docs/dependencies.md
+++ b/docs/dependencies.md
@@ -56,6 +56,7 @@ The following settings are available on iOS and Android:
 ```ts
 type DependencyParamsIOST = {
   project?: string,
+  podspec?: string,
   sharedLibraries?: string[],
 };
 
@@ -70,6 +71,10 @@ type DependencyParamsAndroidT = {
 #### platforms.ios.project
 
 Custom path to `.xcodeproj`
+
+#### platforms.ios.podspec
+
+Custom `podspec` name to use when auto-linking (without the file extension). Your `podspec` file must be located in the root of the dependency package.
 
 #### platforms.ios.sharedLibraries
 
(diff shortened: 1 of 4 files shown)


grabbou:
Looks good, but I have a question: when does the `findPodspecInFolder` returns a wrong results that a custom option is needed?

https://github.com/react-native-community/cli/blob/master/packages/platform-ios/src/config/findPodspecName.js

This code seem to be made with multiple podspecs in mind in order to be able to choose the one appropriately. Maybe it can be done with small fine-tuning.

Salakar:
> @grabbou Looks good, but I have a question: when does the `findPodspecInFolder` returns a wrong results that a custom option is needed?

My reasons for this are:

1) I need to be able to change `podspec` via `react-native.config.js` in my native modules so I can specify different podspecs other than the first one it finds - e.g. `react-native.config.js` can override based on whether a certain feature should be enabled or not (so switching podspecs) - allows native modules to abstract away native code change complexity away from the newer devs.
2) First `podspec` it finds isn't always the correct one, I have a native module with multiple `podspecs`.

grabbou:
Sounds good.

> I need to be able to change podspec via react-native.config.js in my native modules so I can specify different podspecs other than the first one it finds - e.g. react-native.config.js can override based on whether a certain feature should be enabled or not (so switching podspecs) - allows native modules to abstract away native code change complexity away from the newer devs.

So just to clarify, users can already do that w/o this code. Your code adds the ability to define it by developers inside the configuration that ships with the library.

Salakar:
> So just to clarify, users can already do that w/o this code. Your code adds the ability to define it by developers inside the configuration that ships with the library.

Correct, this is for native module developers only.

PR #395, opened 2019-05-20
Improve RNPM warnings

### LINKED ISSUE #346: "rnpm" deprecation warnings encourage library to not be backwards compatible
## Description

If a package defines both `"rnpm"` key in `package.json` and a `react-native.config.js` file, we still produce warnings: 
https://github.com/react-native-community/cli/blob/29ae2531393d42758b49a1b18a31f2ef0f897b64/packages/cli/src/tools/config/index.js#L41-L43

I don't think this is what we want, because if a package wants to get rid of the warning, they'll need to remove "rnpm" entry and in turn bump major version, becoming only compatible with RN 0.60+.

We need to swap the order of loading deps and handle it accordingly.

grabbou:
Summary:
---------

Fixes #346

Test Plan:
----------

Tests are green, so `rnpm` config is still respected.

### DIFF
--- a/packages/cli/src/tools/config/readConfigFromDisk.js
+++ b/packages/cli/src/tools/config/readConfigFromDisk.js
@@ -122,7 +122,7 @@ function readLegacyDependencyConfigFromDisk(
   logger.warn(
     `Package ${chalk.bold(
       path.basename(name),
-    )} is using deprecated "rnpm" config that will stop working from next release. Consider upgrading to the new config format.`,
+     )} is using deprecated "rnpm" config that will stop working from next release. Please notify its maintainers about it.`,
   );
 
   return transformedConfig;
--- a/packages/cli/src/tools/config/readConfigFromDisk.js
+++ b/packages/cli/src/tools/config/readConfigFromDisk.js
@@ -55,7 +55,9 @@ export function readDependencyConfigFromDisk(
     searchPlaces,
   });
 
-  const {config} = explorer.searchSync(rootFolder) || {config: undefined};
+  const {config} = explorer.searchSync(rootFolder) || {
+    config: readLegacyDependencyConfigFromDisk(rootFolder),
+  };
 
   const result = Joi.validate(config, schema.dependencyConfig);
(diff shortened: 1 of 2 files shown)


packages/cli/src/tools/config/readConfigFromDisk.js
```diff
@@ -123,11 +125,5 @@ export function readLegacyDependencyConfigFromDisk(
     )} is using deprecated "rnpm" config that will stop working from next release. Consider upgrading to the new config format.`,
   );
 
-  const result = Joi.validate(transformedConfig, schema.dependencyConfig);
```
thymikee: why removing validation, is to done somewhere else already?

packages/cli/src/tools/config/readConfigFromDisk.js
```diff
@@ -123,11 +125,5 @@ export function readLegacyDependencyConfigFromDisk(
     )} is using deprecated "rnpm" config that will stop working from next release. Consider upgrading to the new config format.`,
```
thymikee: while we're at it, this message is plain wrong, can we change it?
```suggestion
     )} is using deprecated "rnpm" config that will stop working from next release. Please notify its maintainers about it.`,
```

packages/cli/src/tools/config/readConfigFromDisk.js
grabbou: Yes, this function is now not exported to the outside world. The return value of this function will be validated in the `readConfigFromDisk` function.

grabbou: [Here](https://github.com/react-native-community/cli/pull/395/files#diff-f06b7f8213679acb5163a64a52db952bR62), to be precise

packages/cli/src/tools/config/readConfigFromDisk.js
thymikee: this suggestion caused a lint error, please run `yarn lint --fix`

PR #403, opened 2019-05-25
fix: Ensure that bundle URL origin matches debugger page origin

ide:
Summary:
---------

This structurally sets up the Chrome debugger not to have CORS issues if the bundle URL domain differs from the Chrome debugger page domain. This bug was initially fixed in https://github.com/facebook/react-native/pull/17720 and regressed in https://github.com/react-native-community/cli/pull/194. This commit fixes the original bug without changing the behavior introduced in #194.

This commit makes the debugger page use its own URL origin when loading the bundle URL. This addresses the case where the native client may have a bundle URL with a domain like xip.io. Fundamentally, unless the development server allows cross-origin requests, the bundle URL's domain must match the domain used by the Chrome debugger page, which is what this commit achieves.

Test Plan:
----------

Enabled Chrome debugging after the client loaded a URL with a non-localhost domain (ex: LAN IP) and verified the Chrome debugging page doesn't have CORS errors.

### DIFF
--- a/packages/cli/src/commands/server/debugger-ui/index.html
+++ b/packages/cli/src/commands/server/debugger-ui/index.html
@@ -219,7 +219,15 @@
   connectToDebuggerProxy();
 
   async function getBlobUrl(url) {
-    return await window.deltaUrlToBlobUrl(url.replace('.bundle', '.delta'));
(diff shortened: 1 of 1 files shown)


thymikee:
Released v1.9.6 with this fix included

PR #409, opened 2019-05-31
chore(docs): update CONTRIBUTING.md

lithin:
Summary:
---------

I've been struggling to get CLI working locally, especially using a combination of init, start, and run-android commands. I figured out in the end the steps necessary for it and thought it might be maybe useful for others to have them as well.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -61,21 +61,25 @@ react-native init --version ${RN_VERSION}
 npm config set registry https://registry.npmjs.org/
 ```
 
-## Running RN using local version of CLI
+## Running `start` command
 
-To start a new project with local version of CLI and RN without publishing or proxy, you can follow these steps:
+In order for symlinks to work correctly when running `start` locally, set REACT_NATIVE_APP_ROOT as the root folder of your cli project:
 
-1. Check out `react-native` repo. Then update template in local `react-native/template/package.json`, replacing dependency version of `react-native` with the absolute path of the react native repo, for example: "react-native": "file:///Users/username/react-native"
+```
+REACT_NATIVE_APP_ROOT=path/to/cli node path/to/cli/packages/cli/build/index.js start
+```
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -61,6 +61,21 @@ react-native init --version ${RN_VERSION}
 npm config set registry https://registry.npmjs.org/
 ```
 
+## Running RN using local version of CLI
+
+To start a new project with local version of CLI and RN without publishing or proxy, you can follow these steps:
+
+1. Check out `react-native` repo. Then update template in local `react-native/template/package.json`, replacing dependency version of `react-native` with the absolute path of the react native repo, for example: "react-native": "file:///Users/username/react-native"
+
+1. Check out `react-native-cli` repo, cd into it and run `yarn` and `yarn build`.
```
thymikee: This step is redundant, since we advise to o it at the very beginning of this guide. Also `yarn build` runs on postinstall hook, so it's not necessary as well

CONTRIBUTING.md
```diff
@@ -61,6 +61,21 @@ react-native init --version ${RN_VERSION}
 npm config set registry https://registry.npmjs.org/
 ```
 
+## Running RN using local version of CLI
+
+To start a new project with local version of CLI and RN without publishing or proxy, you can follow these steps:
+
+1. Check out `react-native` repo. Then update template in local `react-native/template/package.json`, replacing dependency version of `react-native` with the absolute path of the react native repo, for example: "react-native": "file:///Users/username/react-native"
```
thymikee: This step only makes sense when running RN from the source, while this section is pretty much generic to any project

thymikee: Also it's worth adding that you can get the absolute path by running `pwd` inside react-native directory

lithin:
@thymikee Thanks for the review - I've updated it but I wasn't sure how exactly you'd want to split these.

PR #410, opened 2019-05-31
fix: no active iOS device case in `logIOS`

yuri-kulikov:
Summary:
---------

The `findAvailableDevice()` method can only return `null` or `Object`

```js
function findAvailableDevice(devices) {
      if (device.availability === '(available)' && device.state === 'Booted') {
        return device;
      }
    }
  }

  return null;
}
```

So this code
```js
if (device === undefined) {
  _cliTools().logger.error('No active iOS device found');

  return;
}
```
is useless at the moment

Test Plan:
----------

### When I have no active iOS devices
Before fix:
Screenshot 2019-05-31 at 15 57 20

After fix:
Screenshot 2019-05-31 at 15 58 45


thymikee:
Thank you!

PR #413, opened 2019-06-03
fix: added an additional handling of symlink's in copyAndReplace.js for APP template

tbrat:
Summary:
---------
During developing the app template for ReactNative I have faced with an issue while preparing the template after npm installed it as a dependency into the application:

(node:34062) UnhandledPromiseRejectionWarning: Error: EISDIR: illegal operation on a directory, read

After the investigation, I noticed that symlinks are handled incorrectly. isDirectory method returns false for symlink pointed to the directory.

Test Plan:
----------

Create an application with local template

### DIFF
--- a/packages/cli/src/tools/copyAndReplace.js
+++ b/packages/cli/src/tools/copyAndReplace.js
@@ -4,7 +4,7 @@
  * This source code is licensed under the MIT license found in the
  * LICENSE file in the root directory of this source tree.
  *
- * @format
+ * @flow
  */
 
 import fs from 'fs';
@@ -26,10 +26,10 @@ const binaryExtensions = ['.png', '.jar', '.keystore'];
  *        Function(path, 'identical' | 'changed' | 'new') => 'keep' | 'overwrite'
  */
 function copyAndReplace(
-  srcPath,
-  destPath,
-  replacements,
-  contentChangedCallback,
+  srcPath: string,
+  destPath: string,
+  replacements: {[key: string]: string},
+  contentChangedCallback: (destPath: string, contentChanged: string) => void,
 ) {
   if (isDirectory(srcPath)) {
     if (!fs.existsSync(destPath)) {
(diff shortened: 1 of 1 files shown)


packages/cli/src/tools/copyAndReplace.js
```diff
@@ -131,4 +131,23 @@ function copyBinaryFile(srcPath, destPath, cb) {
...
+ * extended directory expectation.
+ * npm installs the directory of a local node module as a link to the folder of origin location.
+ * such the folder should be allowed to copy and replace as a template
+ */
+function isDirectory(srcPath) {
+  let srcStat = fs.lstatSync(srcPath);
+
+  if (srcStat.isSymbolicLink()) {
+    try {
+      return fs.readDirSync(srcPath) || false;
```
thymikee: wouldn't it be better to use `fs.realpathSync`?

tbrat: @thymikee 
It could be. In such the case, it is required to make an additional API call to FileSystem to get the stat of the real path entry (dir or file).
The current approach doesn't require it.

thymikee:
Ah, I just noticed you're talking about the legacy init! We encourage to try the new init, see https://github.com/react-native-community/cli/blob/master/docs/init.md

packages/cli/src/tools/copyAndReplace.js
thymikee: Also, there's no `readDirSync` but `readdirSync`, are you sure this code works as intended? I've updated it, but would be cool if you verified it works for you

tbrat: @thymikee 
Thanks a lot for fixing of my fault :)

grabbou:
What is the status of this PR? Are we planning to merge it? It's been a while and I see that this targets the legacy `init`.

thymikee:
I think it doesn't hurt to support that in old init as well, isn't it? @grabbou feel free to merge it or close it

grabbou:
Unfortunately, we do have a conflict here. I am not sure how our Typescript migration affects legacy `init`, but in case it's still possible to ship it, I'd say let's submit this PR one more time!

Thank you for your work on this PR - unfortunately, I am going to close it since it needs to be rebased.

PR #415, opened 2019-06-05
chore: remove deprecated entries in DevTools middleware

thymikee:
Summary:
---------

No need for this code to be there anymore.

Test Plan:
----------

None really, removing dead branches


Esemesek:
bitmoji

PR #416, opened 2019-06-05
feat(breaking): link only assets by default; add `--all` flag

thymikee:
Summary:
---------

Change the default behavior of `link` to only link assets.

Since autolinking, manual linking of packages is quickly becoming obsolete, however there's no easy way to autolink assets (can be automated with CocoaPods podspecs, but not sure about Gradle cc @Salakar. EDIT: should be doable, @Salakar is on it). That's why I think it makes sense to default `link` to process assets only, unless the `--all` flag is passed explicitly.

Additional changes that accumulated while testing: 

- fix linking hooks failing to execute
- fix handling assets from legacy "rnpm" config
- overhaul linking logs and error messages
- disable flaky e2e tests in lts-tests phase

Test Plan:
----------

Run `link`, `link --all`, `link package-name` on sample projects.

### DIFF
--- a/packages/cli/src/commands/link/linkAll.js
+++ b/packages/cli/src/commands/link/linkAll.js
@@ -19,20 +19,21 @@ type Options = {
 };
 
 async function linkAll(config: ConfigT, options: Options) {
-  try {
-    if (options.linkDeps) {
-      logger.debug('Linking all dependencies');
-      logger.info(
-        `Linking dependencies using "${chalk.bold(
-          'link',
-        )}" command is now legacy and likely unnecessary. We encourage you to try ${chalk.bold(
-          'autolinking',
-        )} that comes with React Native v0.60 default template. Autolinking happens at build time – during CocoaPods install or Gradle install phase. More information: ${chalk.dim.underline(
-          'https://github.com/react-native-community/cli/blob/master/docs/autolinking.md',
-        )}`,
-      );
-      Object.keys(config.dependencies).forEach(async key => {
-        const dependency = config.dependencies[key];
+  if (options.linkDeps) {
+    logger.debug('Linking all dependencies');
+    logger.info(
+      `Linking dependencies using "${chalk.bold(
+        'link',
+      )}" command is now legacy and likely unnecessary. We encourage you to try ${chalk.bold(
+        'autolinking',
+      )} that comes with React Native v0.60 default template. Autolinking happens at build time – during CocoaPods install or Gradle install phase. More information: ${chalk.dim.underline(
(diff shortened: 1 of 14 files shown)


packages/cli/src/tools/config/readConfigFromDisk.js
```diff
@@ -7,32 +7,69 @@ import Joi from '@hapi/joi';
...
 /**
  * Reads a project configuration as defined by the user in the current
  * workspace.
  */
 export function readConfigFromDisk(rootFolder: string): UserConfigT {
   const explorer = cosmiconfig('react-native', {searchPlaces});
 
-  const {config} = explorer.searchSync(rootFolder) || {config: undefined};
+  const {config} = explorer.searchSync(rootFolder) || {
+    config: readLegacyConfigFromDisk(rootFolder),
```
thymikee: This is fixing a regression where `rnpm` (mostly used for `assets` definition) is not being read by new CLI. This will also make the library maintainers aware of the deprecations directly.

packages/cli/src/tools/config/readConfigFromDisk.js
```diff
@@ -118,7 +155,6 @@ function readLegacyDependencyConfigFromDisk(
       : {},
   };
 
-  // @todo: paste a link to documentation that explains the migration steps
```
thymikee: removed the todo, since migration guide is displayed when reading project config (so it displays for projects and library maintainers, but not to user's deps which would be annoying in the transition period)

packages/cli/src/commands/link/linkAll.js
```diff
@@ -4,39 +4,46 @@
...
-      }
-      await linkDependency(config.platforms, config.project, dependency);
-      if (dependency.hooks.postlink) {
-        await dependency.hooks.postlink();
-      }
-    });
-    await linkAssets(config.platforms, config.project, assets);
+    if (options.linkDeps) {
+      logger.debug('Linking all dependencies');
+      Object.keys(config.dependencies).forEach(async key => {
```
Esemesek: I think we should print some kind of deprecation warning here.

thymikee: Maybe not deprecation, but a warning message about an alternative would be nice way to let people know about autolinking 👍

thymikee:
"try autolinking" info:
Screenshot 2019-06-05 at 18 31 11
"legacy rnpm config' warning
Screenshot 2019-06-05 at 18 31 05

grabbou:
This is a really good idea and solution to fixing that confusion! I like that!

PR #417, opened 2019-06-05
feat: Only pass `hasteImplModulePath` if the `hasteImpl` file is available in RN

cpojer:
Summary:
---------
We are removing `hasteImpl` related stuff in https://github.com/facebook/react-native/pull/24811. However, the cli always passes a hasteImpl on to Metro and on to `jest-haste-map`. If a path to file that doesn't exist is passed, Metro will fail.

This changes the default config to detect whether there is a `hasteImpl.js` file available inside of RN and to only pass it when it is. This makes the cli compatible both with RN <0.60 and >0.61.

Test Plan:
----------

I patched the cli locally and it seemed to work when `hasteImpl` was removed.

### DIFF
--- a/packages/cli/src/tools/loadMetroConfig.js
+++ b/packages/cli/src/tools/loadMetroConfig.js
@@ -5,6 +5,7 @@
 import path from 'path';
 import {createBlacklist} from 'metro';
 import {loadConfig} from 'metro-config';
+import {existsSync} from 'fs';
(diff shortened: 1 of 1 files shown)


thymikee:
Thanks! <3

grabbou:
I usually don't comment on historical PRs to not cause unnecessary notifications, but this is amazing!

PR #419, opened 2019-06-07
feat: add `optimize` command for image assets

### LINKED ISSUE #382: Add "optimize" command for images like Expo does
## Describe the Feature

See: https://blog.expo.io/image-compression-with-expo-cli-d32d15cc8b73

chakrihacker:
Summary:
---------
Add an option to optimize Image assets
closes #382 

Changes:
---------

.reactnative-shared folder will be added when optimize command is added and assets.json file is created which will store all the hashes of compressed images. This helps to avoid git issues and opimizing assets second time

Credits:
---------
Most of the functionality is from this PR https://github.com/expo/expo-cli/pull/555


packages/cli/src/commands/optimize/optimizeCommandLineArgs.js
```diff
@@ -0,0 +1,30 @@
+import path from 'path';
```
thymikee: missing `@flow` pragma

packages/cli/src/commands/optimize/optimize.js
```diff
@@ -0,0 +1,73 @@
+/**
+ * @flow
+ */
+
+const fs = require('fs').promises;
```
thymikee: Please use regular `fs`, `fs.promises` is experimental API and only available in Node 10+, while we target Node 8.3+

packages/cli/src/commands/optimize/optimize.js
```diff
@@ -0,0 +1,73 @@
...
+const path = require('path');
+import sharp from 'sharp';
+import type {ConfigT} from 'types';
+import optimizeCommandLineArgs from './optimizeCommandLineArgs';
+import {logger} from '@react-native-community/cli-tools';
+
+function isImage(filePath: string) {
+  const regex = /\.(png|jpg|jpeg)/;
```
thymikee: This regex is very permissive. Please use something like `/\.png|jpg|jpeg$/i` (notice `$` to mark end of the string and added `i` to avoid lowercasing later)

packages/cli/src/commands/optimize/optimize.js
```diff
@@ -0,0 +1,73 @@
...
+  return filePath.toLowerCase().match(regex);
+}
+
+async function optimize(args: Array<string>, ctx: ConfigT) {
+  logger.log('Optimizing assets...');
```
thymikee:
```suggestion
  logger.info('Optimizing assets...');
```

packages/cli/src/commands/optimize/optimize.js
```diff
@@ -0,0 +1,73 @@
...
+    const metadata = await sharp(buffer).metadata();
+    const newName = createOutputFilename(image);
+
+    const {mTimeMS: lastModified, ino: inode} = fs.statSync(image);
+    if (metadata.format === 'jpeg') {
+      sharp(image)
+        .jpeg({quality: 75})
+        .toFile(newName);
+    } else {
+      const outputFile = await sharp(image)
```
thymikee: unused variable

packages/cli/src/commands/optimize/optimize.js
```diff
@@ -0,0 +1,73 @@
...
+        .png({quality: 75})
+        .toFile(newName);
+    }
+  }
+}
+
+/**
+ * Get all image assets in react native project
+ */
+const getAllImageAssets = async (dir, fileList = []) => {
```
thymikee: We have a utility for [walking directories](https://github.com/react-native-community/cli/blob/master/packages/cli/src/tools/walk.js), can you use that instead?

packages/cli/src/commands/optimize/optimize.js
```diff
@@ -0,0 +1,73 @@
...
+  const images = await getAllImageAssets(rootPathForImages);
+  for (const image of images) {
+    const buffer = await sharp(image).toBuffer();
+    const metadata = await sharp(buffer).metadata();
+    const newName = createOutputFilename(image);
+
+    const {mTimeMS: lastModified, ino: inode} = fs.statSync(image);
+    if (metadata.format === 'jpeg') {
+      sharp(image)
+        .jpeg({quality: 75})
```
thymikee: The quality is configurable through `--quality` flag, please make it so

packages/cli/src/commands/optimize/optimize.js
```diff
@@ -0,0 +1,73 @@
...
+        fileList.push(filepath);
+      }
+    }
+  }
+
+  return fileList;
+};
+
+const createOutputFilename = image => {
+  const imagePath = image.split('/');
```
thymikee:
```suggestion
  const imagePath = image.split(path.sep);
```

packages/cli/src/commands/optimize/optimize.js
```diff
@@ -0,0 +1,73 @@
...
+  const filename = imagePath.pop();
+  const [base, extension] = filename.split('.');
+  const output = base + '-output.' + extension;
+  return imagePath.join('/') + '/' + output;
```
thymikee: Can we assume that `sharp` outputs POSIX paths correctly on Windows as well?

packages/cli/src/commands/optimize/optimize.js
```diff
@@ -0,0 +1,73 @@
...
+import {logger} from '@react-native-community/cli-tools';
+
+function isImage(filePath: string) {
+  const regex = /\.(png|jpg|jpeg)/;
+  return filePath.toLowerCase().match(regex);
+}
+
+async function optimize(args: Array<string>, ctx: ConfigT) {
+  logger.log('Optimizing assets...');
+  const rootPathForImages = path.join(ctx.root, 'src');
```
thymikee: Having `src` is quite opinionated. People may keep images in `<projectRoot>/assets` dir for example. Maybe we should require a directory, or ask users if they didn't pass any?

packages/cli/src/commands/optimize/optimize.js
```diff
@@ -0,0 +1,73 @@
+/**
+ * @flow
+ */
+
+const fs = require('fs').promises;
+const path = require('path');
```
thymikee:
```suggestion
import path from 'path';
```

packages/cli/src/commands/optimize/optimize.js
chakrihacker: implementing *include* option, so users can pass their custom directory names

packages/cli/src/commands/optimize/optimizeCommandLineArgs.js
chakrihacker: added flow pragma

packages/cli/src/commands/optimize/optimize.js
chakrihacker: not using promises

packages/cli/src/commands/optimize/optimize.js
chakrihacker: added **$** for regex

packages/cli/src/commands/optimize/optimize.js
chakrihacker: using available log options

packages/cli/src/commands/optimize/optimize.js
chakrihacker: quality is configurable

packages/cli/src/commands/optimize/optimize.js
chakrihacker: removed unused variable

packages/cli/src/commands/optimize/optimize.js
chakrihacker: using imports

chakrihacker:
@thymikee can you please help with failing test

thymikee:
Rebase to latest master and it's gonna be fine :)

chakrihacker:
That's super fast response

chakrihacker:
@thymikee any update on this??

thymikee:
@chakrihacker not yet, will let you know once I get to review this again :)

packages/cli/src/commands/optimize/optimize.js
```diff
@@ -0,0 +1,142 @@
...
+      },
+    ]);
+    if (saveOriginals) {
+      options.save = true;
+    }
+  }
+  await optimizeAsync(ctx.root, options);
+}
+
+const optimizeAsync = async (projectDir, options = {}) => {
```
grabbou: nit: the `Async` naming convention is Expo convention. While interesting, we don't name our async function this way in the CLI. I'd suggest to rename it to keep the code aligned with the styling.

thymikee:
>why .reactnative-shared?

@chakrihacker FYI we have a `~/.react-native-cli` for storing temporary stuff.

>as a React Native developer, what is the use-case scenario for me? How can I benefit from running this command? (how to integrate)

I think a nice next step for this would doing what Expo does – prompt once in a while that you have unoptimized assets and you can fix it with this command. But no need to make this diff any bigger than necessary for now.

grabbou:
> I think a nice next step for this would doing what Expo does – prompt once in a while that you have unoptimized assets and you can fix it with this command. But no need to make this diff any bigger than necessary for now.

So the command will essentially overwrite the assets I have locally, right? Got confused by `assets.json`, I thought it's some kind of a special file I need to pass to React Native now.

chakrihacker:
> have you thought about chatting with Expo team whether this could be a base for their command?

No I haven't. 

> why .reactnative-shared?

In order to track changes in git, which will help when other collaborators are working on same project

> as a React Native developer, what is the use-case scenario for me? How can I benefit from running this command? (how to integrate)

Optimizing images can help decrease the app size significantly

chakrihacker:
> FYI we have a ~/.react-native-cli for storing temporary stuff.

I am afraid we can't use that as the assets.json file needs to be committed to maintain consistency

@thymikee

packages/cli/src/commands/optimize/optimize.js
chakrihacker: I will change that

grabbou:
This PR has been hanging out here for a while - what is the conclusion here?

chakrihacker:
I think we need a discussion as expo uses config file and should we implement it.

brentvatne:
fwiw in `expo-cli` we are in the process of extracting `optimize` out into it's own standalone lib, so you don't need to install `expo-cli` to use it even and you can just run `npx expo-optimize` in any react-native project. if you want to embed it in `react-native-cli` then you can use the same pattern that `expo-cli` is using in this pr https://github.com/expo/expo-cli/pull/1192, rather than having to vendor the code and maintain it separately

chakrihacker:
That's good news, closing this pr

PR #421, opened 2019-06-07
feat: Pass along CLI port configuration for Android

nhunzaker:
Hello!

Over on the RN repo, I recently merged a PR to make the dev server port configurable on Android:

https://github.com/facebook/react-native/pull/23616

A change needs to be made to the CLI in order to pass this port information over to the Android build as a gradle build argument.

This PR includes that change. Although I'm having some trouble testing it since, what looks like, changes for a 2.0 release of this project.

From what I'm seeing, the `run-android` command appears to be missing when I link this repo directly:

image

Would it be better to send this out against the 1.0 branch?

### DIFF
--- a/packages/platform-android/src/commands/runAndroid/runOnAllDevices.js
+++ b/packages/platform-android/src/commands/runAndroid/runOnAllDevices.js
@@ -39,6 +39,10 @@ function runOnAllDevices(
     const tasks = args.tasks || ['install' + toPascalCase(args.variant)];
     const gradleArgs = getTaskNames(args.appFolder, tasks);
 
+    if (args.port != null) {
+      gradleArgs.push('-PreactNativeDevServerPort=' + args.port);
+    }
+
     logger.info('Installing the app...');
     logger.debug(
       `Running command "cd android && ${cmd} ${gradleArgs.join(' ')}"`,


thymikee:
Thanks! Since this part was heavily refactored in 2.0, it would be lovely if you sent another one to 1.x branch. 

> From what I'm seeing, the run-android command appears to be missing when I link this repo directly

The iOS and Android specific commands are now in separate packages and are discovered lazily like any other platforms. However they should be visible in help output. Can you share how do you test it as I cannot repro?

packages/platform-android/src/commands/runAndroid/runOnAllDevices.js
```diff
@@ -39,6 +39,10 @@ function runOnAllDevices(
     const tasks = args.tasks || ['install' + toPascalCase(args.variant)];
     const gradleArgs = getTaskNames(args.appFolder, tasks);
 
+    if (args.port != null) {
```
Esemesek: We should probably use stricter equal operator `!==`. Or is it used on purpose here?

nhunzaker: I figured the value could be undefined if the argument wasn't provided, but I also wasn't sure if it could ever be null too.

I bet I could make this `"port" in args` or `args.port !== undefined`

thymikee: `args.port` can be: `undefined`, `NaN` (we're passing string input through `Number()`), `number`

nhunzaker:
@thymikee Hey! I'm testing using a repo I made to link RN and the CLI to a test project:
https://github.com/nhunzaker/ReactNativeTestPlan

It's probably not perfect. However to be sure, I also cloned down the project locally and followed these commands:

```
git clone git@github.com:react-native-community/cli.git
cd react-native-community
yarn install
yarn build
node packages/cli/build/index.js --help
```

The final command reporting:

```
Usage: index [options] [command]

Options:
  --version                     Print CLI version
  --verbose                     Increase logging verbosity
  -h, --help                    output usage information

Commands:
  start [options]               starts the webserver
  bundle [options]              builds the javascript bundle for offline use
  ram-bundle [options]          builds javascript as a "Random Access Module" bundle for offline use
  link [options] [packageName]  links assets and optionally native modules
  unlink <packageName>          unlink native dependency
  install <packageName>         install and link native dependencies
  uninstall <packageName>       uninstall and unlink native dependencies
  upgrade [options] [version]   Upgrade your app's template files to the specified or latest npm version using `rn-diff-purge` project. Only valid semver versions are allowed.
  info                          Get relevant version info about OS, toolchain and libraries
  config                        Print CLI configuration
  init [options] <projectName>  Initialize a new React Native project named <projectName> in a directory of the same name.
```

thymikee:
Ahm, yea, that's expected because we only load them if there's `react-native` in the dependencies. If you add `@react-native-community/cli` to your RN project, or `yarn link` it, you'll be able to see them :)

thymikee:
@nhunzaker Based on a fact that this commit is gonna land on RN 0.60 or 0.61, does it make sense to port it to 0.59 anyway?

nhunzaker:
hey! Sorry for the delay. I'd be happy to prep a PR targeting 0.59 if you'd like.

thymikee:
No worries. Only if RN 0.59 supports it. I know your commit was cherry-picked into 0.60, but not sure about 0.59? If it's not there, then our work here is done :D

nhunzaker:
[Looks like it wasn't](https://github.com/facebook/react-native/commits/0.59-stable). Fine by me to keep this as 0.60+. I've got my sights on other RN issues anyway :P

packages/platform-android/src/commands/runAndroid/runOnAllDevices.js
grabbou: I guess `if(args.port)` would be less confusing in that case :D

packages/platform-android/src/commands/runAndroid/runOnAllDevices.js
```diff
@@ -39,6 +39,10 @@ function runOnAllDevices(
...
+      gradleArgs.push('-PreactNativeDevServerPort=' + args.port);
```
grabbou: Quick q: Why this is called `Preact`?

thymikee: It's how you can pass parameters to Gradle, something like `-P reactNativeDevServerPort` should also work I guess? But I didn't bother to check it. cc @nhunzaker

nhunzaker: `-P` for "Set project property for the build script"

Directly from the gradlew usage stuff:

```
-P, --project-prop        Set project property for the build script (e.g. -Pmyprop=myvalue).
```

Having no space is definitely confusing.

grabbou: Ah, gotcha, thanks! I read it "Preact" and asked myself what's going on :joy: Thank you for explaining it!

PR #426, opened 2019-06-13
Log `console` invocations to the terminal.

cpojer:
Summary:
---------

In development, React Native on master now sends all `console` calls to the CLI/Metro. This means we can show logs directly in the terminal, which means debugging simple things a ton easier.

Test Plan:
----------

Screen Shot 2019-06-13 at 15 49 43

### DIFF
--- a/packages/cli/src/commands/server/middleware/logToConsoleMiddleware.js
+++ b/packages/cli/src/commands/server/middleware/logToConsoleMiddleware.js
@@ -1,0 +1,37 @@
+'use strict';
+
+import chalk from 'chalk';
+
+let cache = [];
+let timer;
+
+const log = ({level, data, id}) => {
+  const logFunction = level !== 'trace' && console[level] ? level : 'log';
+  const color =
+    level === 'error' ? 'red' : level === 'warn' ? 'yellow' : 'white';
+
+  console[logFunction].apply(console, [
+    chalk.inverse[color].bold(` ${level.toUpperCase()} `),
+    ...data,
+  ]);
+};
+
+// Hold messages and flush them to reduce the amount of out-of-order logs
+const flush = () => {
+  timer = null;
+  cache.sort((a, b) => a.id - b.id).forEach(log);
+  cache = [];
+};
+
+export default (req, res, next) => {
+  if (req.url === '/log-to-console') {
+    cache.push(JSON.parse(req.rawBody));
+    if (!timer) {
+      timer = setTimeout(flush, 200);
+    }
+
+    res.end('OK');
+  } else {
+    next();
+  }
+};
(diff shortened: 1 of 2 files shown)


packages/cli/src/commands/server/middleware/logToConsoleMiddleware.js
```diff
@@ -0,0 +1,37 @@
+'use strict';
+
+import chalk from 'chalk';
+
+let cache = [];
+let timer;
+
+const log = ({level, data, id}) => {
```
Esemesek: You should probably use our custom logger.

cpojer: What's the custom logger and what does it do? This basically just forwards `console` calls from the device onto the terminal without any special magic except Jest-like highlights.

Esemesek: We have those Jest-like highlight already implemented. https://github.com/react-native-community/cli/blob/master/packages/tools/src/logger.ts

matt-oakes: I believe it would be this: https://github.com/react-native-community/cli/blob/master/packages/tools/src/logger.ts

cpojer: I don't really think this adds any value here. The current logger is highlighting information issued by the CLI, and here we are highlighting information from an RN app. I think it makes sense for those to look differently because one is from the tool and the other from the user. A different style will make people pay attention differently.

thymikee: Agree with @cpojer on that. We're gonna overhaul the UI Metro gives us some time later, but not today.

JanithaR:
This wouldn't stop us from seeing these same messages on Chrome console does it?

cpojer:
No, it just adds a feature but doesn't disable any other ways of logging.

roshangm1:
What will be the result of logging an object or big array ? I used to do logging in terminal using `react-native log-android`. The output weren't pretty.

RubenSandwich:
@cpojer 

This is a great solution of using the tooling already required to display logs, instead of having to use another tool for this info. (A debug window.)

To riff off of @roshangm1's comment. Here is an idea of displaying arrays and objects to the console using some basic ASCII graphics. (The format below is loosely based upon [console.table](https://developer.mozilla.org/en-US/docs/Web/API/Console/table) and [csvkit](https://csvkit.readthedocs.io/en/latest/tutorial/2_examining_the_data.html#csvsort-order-matters).)

```

console.log([true, false]);

(index) | Values
0       | true
1       | false

console.log({firstName: "Erica", lastName: "Smith"});

(index)    | Values
firstName  | "Erica"
lastName   | "Smith"

console.log([{firstName: "Monica", lastName: "Johnson"}, {firstName: "Erica", lastName: "Smith"}]);

(index) | firstName | lastName
0       | "Monica"  | "Johnson"
1       | "Erica"   | "Smith"

console.log([22, {firstName: "Erica", lastName: "Smith"}, 34]);

(index) | Values
0       | 22
1       | firstName: "Erica"
        | lastName: "Smith"
2       | 34

```

PR #427, opened 2019-06-13
chore: update init docs to be more descriptive

emin93:
Summary:
---------

Hi everyone,

I am liking the new CLI changes a lot, really good work there, thank you all for the effort! 👍

While [refactoring the TS template to be compatible with the new CLI](https://github.com/emin93/react-native-template-typescript/issues/38), I saw there are a few things documented differently than the actual implementation.

In this PR I changed the documentation to match the current implementation. It's likely, that the implementation has to be changed as well in order to support both ways.

Test Plan:
----------

Created a new project with the template config as described in the docs. As this failed, I had a look at the current implementation and found out:

- Template config file name shouldn't have a `.js` extension: https://github.com/react-native-community/cli/blob/master/packages/cli/src/commands/init/template.js#L35
- Post install script is not a node script, therefore the sample `.js` extension is misleading as it's rather a shell script: https://github.com/react-native-community/cli/blob/master/packages/cli/src/commands/init/template.js#L74

As said above, maybe it's the implementation that needs to be fixed and not the documentation, just let me know.

### DIFF
--- a/docs/init.md
+++ b/docs/init.md
@@ -66,7 +66,7 @@ react-native init ProjectName
 
 # Creating custom template
 
-Every custom template needs to have configuration file called `template.config` in the root of the project:
+Every custom template needs to have configuration file called `template.config.js` in the root of the project:
 
 ```js
 module.exports = {
--- a/docs/init.md
+++ b/docs/init.md
@@ -66,7 +66,7 @@ react-native init ProjectName
 
 # Creating custom template
 
-Every custom template needs to have configuration file called `template.config.js` in the root of the project:
+Every custom template needs to have configuration file called `template.config` in the root of the project:
 
 ```js
 module.exports = {
@@ -77,7 +77,7 @@ module.exports = {
   templateDir: './template',
 
   // Path to script, which will be executed after initialization process, but before installing all the dependencies specified in the template.
-  postInitScript: './script.js',
+  postInitScript: './script.sh',
 };
 ```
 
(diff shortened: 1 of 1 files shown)


docs/init.md
```diff
@@ -77,7 +77,7 @@ module.exports = {
   templateDir: './template',
 
   // Path to script, which will be executed after initialization process, but before installing all the dependencies specified in the template.
-  postInitScript: './script.js',
+  postInitScript: './script.sh',
```
thymikee: Nah, let's use JS. You can make your `.js` file executable using shebang: https://github.com/Esemesek/react-native-new-template/blob/master/script.js
or update this file to:
```
postInitScript: 'node ./script.js'
```

emin93: I tried `node ./script.js` but that gave me an error. I'm fine with adding a shebang in the script, but I think it makes sense to document that somewhere at least.

emin93: Ah sorry, I totally oversaw that you added the shebang already in the sample script.

thymikee: I'm fine with making it more obvious though :)

docs/init.md
```diff
@@ -66,7 +66,7 @@ react-native init ProjectName
 
 # Creating custom template
 
-Every custom template needs to have configuration file called `template.config.js` in the root of the project:
```
thymikee: Um, why this change? We resolve path to `template.config` but then using Node require resolution:

https://github.com/react-native-community/cli/blob/f316f9269bdd71d492efd8b73104f38c568c9e52/packages/cli/src/commands/init/template.js#L31-L40

emin93: True, seems like I did a mistake yesterday while creating a new project.

PR #428, opened 2019-06-14
fix: Add fallback path check when not found app file for iOS

zhongwuzw:
Summary:
---------

I talked to @grabbou few days ago, I met a `CFBundleIdentifier problem`, I give it a look, and the reason is `-derivedDataPath` not work as expected. Leads to `app` file not found. 

I can reproduce it on latest `cli` and `react-native`, Xcode 10.2.1. I may think it's a `xcodebuild` command bug. 

Related issue: https://github.com/facebook/react-native/issues/25259

Test Plan:
----------

`react-native run-ios` can works as expected.

### DIFF
--- a/packages/platform-ios/src/commands/runIOS/index.js
+++ b/packages/platform-ios/src/commands/runIOS/index.js
@@ -329,7 +329,13 @@ function getBuildPath(configuration, appName, isDevice, scheme) {
     device = 'iphonesimulator';
   }
 
-  return `build/${scheme}/Build/Products/${configuration}-${device}/${appName}.app`;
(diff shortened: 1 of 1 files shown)


thymikee:
@zhongwuzw FYI I ported this to 1.x branch and published as 1.9.9 so you can use it in RN 0.59

zhongwuzw:
@thymikee ❤️ I'll check 1.9.9 later. :)

grabbou:
Good to see it fixed, tho I still have no idea why the other path would work. I guess that's just some kind of iOS magic.

PR #433, opened 2019-06-19
chore: refactor packageManager.js for less redundancy

jamesgeorge007:
This is an attempt made to reduce redundant code.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,6 +28,7 @@
     "@react-native-community/eslint-config": "^0.0.5",
     "@types/jest": "^24.0.11",
     "@types/node": "^11.13.0",
+    "@types/node-fetch": "^2.3.7",
     "babel-jest": "^24.6.0",
     "babel-plugin-module-resolver": "^3.2.0",
     "chalk": "^2.4.2",
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "test": "jest",
     "test:ci:unit": "jest packages --ci --coverage",
     "test:ci:e2e": "jest e2e --ci -i",
-    "lint": "eslint --fix --ext .js,.ts . --cache --report-unused-disable-directives",
+    "lint": "eslint --ext .js,.ts . --cache --report-unused-disable-directives",
     "test:ci:cocoapods": "ruby packages/platform-ios/native_modules.rb",
     "flow-check": "flow check",
     "postinstall": "yarn build",
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "test": "jest",
     "test:ci:unit": "jest packages --ci --coverage",
     "test:ci:e2e": "jest e2e --ci -i",
-    "lint": "eslint --ext .js,.ts . --cache --report-unused-disable-directives",
+    "lint": "eslint --fix --ext .js,.ts . --cache --report-unused-disable-directives",
     "test:ci:cocoapods": "ruby packages/platform-ios/native_modules.rb",
     "flow-check": "flow check",
     "postinstall": "yarn build",
--- a/packages/cli/src/tools/packageManager.js
+++ b/packages/cli/src/tools/packageManager.js
@@ -12,21 +12,22 @@ type Options = {|
 let projectDir;
 
 const yarnConfig = {
-	'dependencies': ['add'],
-	'devDependencies': ['add', '-D'],
-	'uninstallDependencies': ['remove']
+  dependencies: ['add'],
+  devDependencies: ['add', '-D'],
+  uninstallDependencies: ['remove'],
 };
 
 const npmConfig = {
-	'dependencies': ['install', '--save', '--save-exact'],
-	'devDependencies': ['install', '--save-dev', '--save-exact'],
-	'uninstallDependencies': ['uninstall', '--save']
+  dependencies: ['install', '--save', '--save-exact'],
+  devDependencies: ['install', '--save-dev', '--save-exact'],
+  uninstallDependencies: ['uninstall', '--save'],
 };
 
 function configurePackageManager(
-	pmIsYarn: boolean,
-	packageNames: Array<string>,
-	options?: Options, installType: string
+  pmIsYarn: boolean,
+  packageNames: Array<string>,
+  options?: Options,
+  installType: string,
 ) {
   const pm = pmIsYarn ? 'yarn' : 'npm';
   const pmConfig = pm === 'npm' ? npmConfig : yarnConfig;
@@ -61,18 +62,33 @@ export function setProjectDir(dir: string) {
 }
 
 export function install(packageNames: Array<string>, options?: Options) {
-  configurePackageManager(shouldUseYarn(options), packageNames, options, 'dependencies');
+  configurePackageManager(
+    shouldUseYarn(options),
+    packageNames,
+    options,
+    'dependencies',
+  );
 }
 
 export function installDev(packageNames: Array<string>, options?: Options) {
-  configurePackageManager(shouldUseYarn(options), packageNames, options, 'devDependencies');
+  configurePackageManager(
+    shouldUseYarn(options),
+    packageNames,
+    options,
+    'devDependencies',
+  );
 }
 
 export function uninstall(packageNames: Array<string>, options?: Options) {
-  configurePackageManager(shouldUseYarn(options), packageNames, options, 'uninstallDependencies');
+  configurePackageManager(
+    shouldUseYarn(options),
+    packageNames,
+    options,
+    'uninstallDependencies',
+  );
 }
 
 export function installAll(options?: Options) {
-  const pm = shouldUseYarn(options) ? 'yarn': 'npm';
+  const pm = shouldUseYarn(options) ? 'yarn' : 'npm';
   return executeCommand(pm, ['install'], options);
 }
--- a/packages/cli/src/tools/packageManager.js
+++ b/packages/cli/src/tools/packageManager.js
@@ -76,5 +76,8 @@ export function uninstall(packageNames: Array<string>, options?: Options) {
 }
 
 export function installAll(options?: Options) {
-  return configurePackageManager(options, 'install');
+  const pm = shouldUseYarn(options) ? 'yarn' : 'npm';
+  const pmConfig = packageManagers[pm];
+
+  return executeCommand(pm, pmConfig.install, options);
 }
--- a/packages/cli/src/tools/packageManager.js
+++ b/packages/cli/src/tools/packageManager.js
@@ -26,12 +26,12 @@ const npmConfig = {
 function configurePackageManager(
 	pmIsYarn: boolean,
 	packageNames: Array<string>,
-	options?: Options, args: string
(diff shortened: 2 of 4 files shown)


jamesgeorge007:
cc @thymikee

jamesgeorge007:
It seems the CI is not happy due to the order in which those options appear :thinking:

Esemesek:
Order doesn't really matter here:
```sh
yarn add package -D
```
and
```sh
yarn add -D package
```
is exactly the same command.
Can you please fix those tests?

jamesgeorge007:
> Order doesn't really matter here:

@Esemesek I was saying about these CI warnings:-

```md
● should use yarn if project is using yarn

    expect(jest.fn()).toHaveBeenCalledWith(expected)

    Expected mock function to have been called with:
      ["add", "react", "react-native"]
    as argument 2, but it was called with
      ["add", "react", "react-native", "react", "react-native"].

    Difference:

    - Expected
    + Received

      Array [
        "add",
        "react",
        "react-native",
    +   "react",
    +   "react-native",
      ]
```

I guess the CI build failed with some of the recent commits landed on the master branch :thinking:

Esemesek:
@jamesgeorge007 CI is not failing on the master branch. It is failing on your branch because you swapped the order of `--flags` and `arguments`. You can either, change your code [here](https://github.com/react-native-community/cli/pull/433/files#diff-022240c53fa3b0d949a5bf06e9d03203R35), or fix tests to expect different order. I would prefer the first solution, because your code would have exactly the same behaviour as tests describe.

packages/cli/src/tools/packageManager.js
```diff
@@ -11,6 +11,32 @@ type Options = {|
...
+function configurePackageManager(
+  pmIsYarn: boolean,
+  packageNames: Array<string>,
+  options?: Options,
+  installType: string,
+) {
+  const pm = pmIsYarn ? 'yarn' : 'npm';
+  const pmConfig = pm === 'npm' ? npmConfig : yarnConfig;
+
+  let args = pmConfig[installType];
```
jamesgeorge007: What might be a better approach to conditionally assign `args`?

jamesgeorge007: `packageNames` seem to come up in the second place within the `args` array.

Esemesek: You could do some thing like this:

```js
const [executable, flags] = pmConfig[installType];
const args = [executable, ...packageNames, ...flags];
```

packages/cli/src/tools/packageManager.js
```diff
@@ -11,6 +11,32 @@ type Options = {|
 
 let projectDir;
 
+const yarnConfig = {
+  dependencies: ['add'],
+  devDependencies: ['add', '-D'],
+  uninstallDependencies: ['remove'],
```
thymikee: I'd prefer this shape and naming:
```js
const packageManagers = {
  yarn: {
  	add: ['add'],
  	addDev: ['add', '-D'],
  	remove: ['remove'],
  	install: ['install'],
  },
  npm: ...
}
```
It's gonna make it easier for us to add more (like `pnpm`) in the future

packages/cli/src/tools/packageManager.js
```diff
@@ -36,33 +62,33 @@ export function setProjectDir(dir: string) {
...
 
 export function install(packageNames: Array<string>, options?: Options) {
-  return shouldUseYarn(options)
-    ? executeCommand('yarn', ['add', ...packageNames], options)
-    : executeCommand(
-        'npm',
-        ['install', ...packageNames, '--save', '--save-exact'],
-        options,
-      );
+  configurePackageManager(
```
thymikee:
```suggestion
  return configurePackageManager(
```

packages/cli/src/tools/packageManager.js
```diff
@@ -36,33 +62,33 @@ export function setProjectDir(dir: string) {
...
 
 export function installDev(packageNames: Array<string>, options?: Options) {
-  return shouldUseYarn(options)
-    ? executeCommand('yarn', ['add', '-D', ...packageNames], options)
-    : executeCommand(
-        'npm',
-        ['install', ...packageNames, '--save-dev', '--save-exact'],
-        options,
-      );
+  configurePackageManager(
```
thymikee:
```suggestion
  return configurePackageManager(
```

packages/cli/src/tools/packageManager.js
```diff
@@ -36,33 +62,33 @@ export function setProjectDir(dir: string) {
...
+    options,
+    'devDependencies',
+  );
 }
 
 export function uninstall(packageNames: Array<string>, options?: Options) {
-  return shouldUseYarn(options)
-    ? executeCommand('yarn', ['remove', ...packageNames], options)
-    : executeCommand('npm', ['uninstall', ...packageNames, '--save'], options);
+  configurePackageManager(
```
thymikee:
```suggestion
  return configurePackageManager(
```

packages/cli/src/tools/packageManager.js
```diff
@@ -11,6 +11,32 @@ type Options = {|
...
+  dependencies: ['install', '--save', '--save-exact'],
+  devDependencies: ['install', '--save-dev', '--save-exact'],
+  uninstallDependencies: ['uninstall', '--save'],
+};
+
+function configurePackageManager(
+  pmIsYarn: boolean,
+  packageNames: Array<string>,
+  options?: Options,
+  installType: string,
```
thymikee:
```suggestion
  action: 'add' | 'addDev' | 'remove' | 'install',
```

packages/cli/src/tools/packageManager.js
```diff
@@ -36,33 +62,33 @@ export function setProjectDir(dir: string) {
...
+    'uninstallDependencies',
+  );
 }
 
 export function installAll(options?: Options) {
-  return shouldUseYarn(options)
-    ? executeCommand('yarn', ['install'], options)
-    : executeCommand('npm', ['install'], options);
+  const pm = shouldUseYarn(options) ? 'yarn' : 'npm';
+  return executeCommand(pm, ['install'], options);
```
thymikee: use `config.install`

packages/cli/src/tools/packageManager.js
```diff
@@ -36,33 +62,33 @@ export function setProjectDir(dir: string) {
...
 export function install(packageNames: Array<string>, options?: Options) {
-  return shouldUseYarn(options)
-    ? executeCommand('yarn', ['add', ...packageNames], options)
-    : executeCommand(
-        'npm',
-        ['install', ...packageNames, '--save', '--save-exact'],
-        options,
-      );
+  configurePackageManager(
+    shouldUseYarn(options),
```
thymikee: I think it makes sense for `shouldUseYarn` to be moved to `configurePackageManager`. We already have `options`, so we could change the fn signature to be:

```js
function configurePackageManager(
  packageNames: Array<string>,
  options?: Options,
  action: 'add' | 'addDev' | 'remove' | 'install',
)
```

packages/cli/src/tools/packageManager.js
```diff
@@ -11,6 +11,35 @@ type Options = {|
...
+};
+
+function configurePackageManager(
+  packageNames: Array<string>,
+  options?: Options,
+  action: string,
+) {
+  const pm = shouldUseYarn(options) ? 'yarn' : 'npm';
+  const {yarn, npm} = packageManagers;
+  const pmConfig = pm === 'npm' ? npm : yarn;
```
thymikee: The nice thing about having package managers as keys now, is that we can directly access their config once we know which one to use:
```suggestion
  const pmConfig = packageManagers[pm];
```
line 35 is unnecessary now

packages/cli/src/tools/packageManager.js
```diff
@@ -11,6 +11,35 @@ type Options = {|
...
+    addDev: ['install', '--save-dev', '--save-exact'],
+    remove: ['uninstall', '--save'],
+    install: ['install'],
+  },
+};
+
+function configurePackageManager(
+  packageNames: Array<string>,
+  options?: Options,
+  action: string,
```
thymikee: please make it an enum as I already noted: https://github.com/react-native-community/cli/pull/433#discussion_r296023034

packages/cli/src/tools/packageManager.js
jamesgeorge007: Oh didn't notice it :sweat_smile:

packages/cli/src/tools/packageManager.js
```diff
@@ -11,6 +11,34 @@ type Options = {|
...
+function configurePackageManager(
+  packageNames: Array<string>,
+  options?: Options,
+  action: 'add' | 'addDev' | 'remove',
+) {
+  const pm = shouldUseYarn(options) ? 'yarn' : 'npm';
+  const pmConfig = packageManagers[pm];
+
+  const [executable, ...flags] = pmConfig[action];
+  const args = [executable, ...packageNames, ...flags];
```
thymikee:
```suggestion
  const args = [executable, ...flags, ...packageNames];
```

jamesgeorge007:
@thymikee Still the CI build fails as the order in which those `args` appear isn't the same.

packages/cli/src/tools/packageManager.js
```diff
@@ -36,33 +64,20 @@ export function setProjectDir(dir: string) {
...
 }
 
 export function installAll(options?: Options) {
-  return shouldUseYarn(options)
-    ? executeCommand('yarn', ['install'], options)
-    : executeCommand('npm', ['install'], options);
+  const pm = shouldUseYarn(options) ? 'yarn' : 'npm';
+  const pmConfig = packageManagers[pm];
+
+  return executeCommand(pm, pmConfig.install, options);
```
jamesgeorge007: The respective method doesn't include the usage of `packageNames` and I guess having a separate implementation makes sense :thinking:

thymikee: packageNames may be an empty array, nothing bad about it

jamesgeorge007:
@thymikee What about the failing CI builds :thinking:

thymikee:
You'll need to adjust tests. npm doesn't care about the order of parameters, so we're fine

package.json
```diff
@@ -12,7 +12,7 @@
     "test": "jest",
     "test:ci:unit": "jest packages --ci --coverage",
     "test:ci:e2e": "jest e2e --ci -i",
-    "lint": "eslint --ext .js,.ts . --cache --report-unused-disable-directives",
+    "lint": "eslint --fix --ext .js,.ts . --cache --report-unused-disable-directives",
```
jamesgeorge007: Should there be yet another script with the `--fix` option?

thymikee: Please revert this `--fix` flag. It can be added in local development, but on CI it doesn't make too much sense, because in case of auto-fixable issues, it's gonna report success, the code will be changed (auto-fixed), but won't be reflected in the source control.

jamesgeorge007:
@thymikee Are we all set to go?

Issue #429, opened 2019-06-14
Support pre-/postlink hooks in autolinking


thymikee:
## Describe the Feature

Currently `prelink` and `postlink` are not executed when autolinking happens. That's because the linking logic moved to CocoaPods and Gradle scripts. 

## Possible Implementations

Hooks should be available in `config` already: https://github.com/react-native-community/cli/blob/7dec6541a88ba3496d43a20a832b3150166522aa/types/index.js#L123-L127

Hope @orta and @Salakar have ideas on how to support these on per-platform level.

## Related Issues
https://github.com/react-native-community/releases/issues/116#issuecomment-502158419

thymikee:
I was thinking for this topic for a while recently and I'm considering killing prelink/postlink hooks in the CLI. It occurs to me that the whole concept is not so good after all, even if helpful at times. Here's the list of pain points I have with these "hooks":
- **incoherent DX** – some do background work, some ask questions, some run even when already linked,
- **potential for security vulnerabilities** – can do virtually anything without users realizing, e.g. running arbitrary commands or tracking scripts (could be fixed with a direct question whether to run the hook, but still),
- **incompatible with autolinking design** – they run pre/post "linking" phase which was triggered for all all platforms at once using a command, while autolinking is per-platform during buildtime, which means to stay backwards compatible CLI would run Android-related changes when doing "pod install" (no way to distinguish as it's all in single script file)
- **often misused to hack/workaround CLI limitations/bugs** – potential for over-complicating user code by e.g. editing native build files, which makes it even harder for users to upgrade, since the code was added automatically, causing FUD if user is not experienced with native tooling

Saying all that, a vast majority of the existing postlink hooks is either being obsolete with autolinking and general improvements to the CLI and default RN template or can be trimmed down in scope and quite easily swapped by a comprehensive docs, which I'm probably willing to contribute for some more popular packages. Feel free to change my mind!

grabbou:
In their current form and implementation, I considered them deprecated and that was one of the reasons I didn't push on implementing them as a part of `autolinking` which should solve the primary use-case for them. 

I am generally in favour of leaving them as a thing of the past and not adding support for the future linking.

One thing that I would like to make sure though is whether consumers of the CLI, such as Ignite (CC: @jamonholmgren) are taking advantage of that and if yes, what does it look like. 

We should probably come up with a better mechanism if needed to support hooking into the lifecycle.

jamonholmgren:
[...]

PR #436, opened 2019-06-21
feat: filter out non-native deps from config output; fix missing hooks

thymikee:
Summary:
---------

This diff aims to filter out unnecessary non-native dependencies from the `config` command output and display dependency linking "hooks" properly (needed for https://github.com/react-native-community/cli/issues/429), which were not there because we stored them as non-serializable functions.

Before:
```sh
$ yarn react-native config
{
  "dependencies": {
    "dep-1": { ... },
    "dep-2": { ... },
    "native-dep-1": { 
   	  "hooks": {}
    },
  }
  ...
}
```

After:
```sh
$ yarn react-native config
{
  "dependencies": {
    "native-dep-1": { 
   	  "hooks": {
		"postlink": "node script.js"
      }
    },
  }
  ...
}
```

Test Plan:
----------

Updated tests accordingly.


packages/cli/src/tools/config/index.js
```diff
@@ -2,36 +2,90 @@
...
+  );
+}
 
 /**
  * Loads CLI configuration
  */
 function loadConfig(projectRoot: string = process.cwd()): ConfigT {
   const userConfig = readConfigFromDisk(projectRoot);
 
+  const initialConfig: ConfigT = {
```
thymikee: this block was moved without modifications

packages/cli/src/tools/config/index.js
```diff
@@ -2,36 +2,90 @@
...
+            : finalConfig.platforms[platform].dependencyConfig(
+                root,
+                config.dependency.platforms[platform] || {},
+              );
+          return dependency;
+        },
+        {},
+      ),
+      assets: findAssets(root, config.dependency.assets),
+      hooks: config.dependency.hooks,
```
thymikee: the only change in this moved block is this line – we pass the `hooks` string directly, without mapping to functions (which are not serializable, hence not displayed after JSON.stringify runs on config)

packages/cli/src/commands/config/config.js
```diff
@@ -2,10 +2,31 @@
...
+  });
+  return filtered;
+}
+
 export default {
   name: 'config',
   description: 'Print CLI configuration',
   func: async (argv: string[], ctx: ConfigT) => {
-    console.log(JSON.stringify(ctx, null, 2));
+    console.log(JSON.stringify(filterConfig(ctx), null, 2));
```
grabbou: This is a good idea to do it here - we still keep the lazy behaviour in other places

packages/cli/src/tools/config/index.js
grabbou: Yeah, right!

grabbou: I guess an alternative take would be to provide a special `toString` implementation as a result of `makeHook` call to make it serializable?

thymikee: I think it's better to keep the whole config serializable anyway, less things to remember

packages/cli/src/commands/config/config.js
thymikee: Yup, that was the intention. Should I leave a comment to make it obvious?

grabbou: I think it's fine like this :)

packages/cli/src/tools/config/index.js
grabbou: Yeah, I was suggesting to make `makeHook` return a serializable function (`toString` returns a path). 

They are going to be gone (most likely) in the future anyway, so not bothering to much with that one.

packages/cli/src/commands/config/config.js
thymikee: Feel free to approve it then? 🙂

grabbou: Be my guest

Issue #412, opened 2019-06-03
bug on custom config for podspec on autolinking


sibelius:
## Environment

```jsx
React Native Environment Info:
    System:
      OS: macOS 10.14.5
      CPU: (4) x64 Intel(R) Core(TM) i7-5557U CPU @ 3.10GHz
      Memory: 106.45 MB / 16.00 GB
      Shell: 5.0.7 - /usr/local/bin/bash
    Binaries:
      Node: 10.15.0 - ~/.nvm/versions/node/v10.15.0/bin/node
      Yarn: 1.16.0 - /usr/local/bin/yarn
      npm: 6.9.0 - ~/.nvm/versions/node/v10.15.0/bin/npm
      Watchman: 4.9.0 - /usr/local/bin/watchman
    SDKs:
      iOS SDK:
        Platforms: iOS 12.2, macOS 10.14, tvOS 12.2, watchOS 5.2
      Android SDK:
        API Levels: 22, 23, 24, 25, 26, 27, 28
        Build Tools: 23.0.1, 23.0.2, 23.0.3, 24.0.0, 24.0.2, 24.0.3, 25.0.0, 25.0.1, 25.0.2, 25.0.3, 26.0.1, 26.0.2, 27.0.3, 28.0.0, 28.0.3
        System Images: android-25 | Android TV Intel x86 Atom, android-25 | Google APIs Intel x86 Atom, android-25 | Google APIs Intel x86 Atom_64, android-26 | Google APIs Intel x86 Atom, android-27 | Google APIs Intel x86 Atom, android-27 | Google Play Intel x86 Atom, android-N | Android TV Intel x86 Atom, android-N | Android Wear ARM EABI v7a, android-N | Android Wear Intel x86 Atom, android-N | Intel x86 Atom, android-N | Intel x86 Atom_64, android-P | Google Play Intel x86 Atom
    IDEs:
      Android Studio: 3.4 AI-183.6156.11.34.5522156
      Xcode: 10.2.1/10E1001 - /usr/bin/xcodebuild
    npmPackages:
      react-native: 0.59.8 => 0.59.8 
    npmGlobalPackages:
      react-native-cli: 2.0.1
      react-native-git-upgrade: 0.2.7
      react-native-rename: 2.4.1
```

## Description

I'm following this doc https://github.com/react-native-community/cli/blob/master/docs/dependencies.md to provide a custom path for podspec for react-native-app-auth

[...]

PR #442, opened 2019-06-24
feat(breaking): use 'podspecPath' instead of 'podspec', make it smarter

thymikee:
Summary:
---------

This diff removes the `project.ios.podspec` in favor of more obvious `project.ios.podspecPath` which makes it possible for users to provide a custom path to the `.podspec` file. I chose this, as it aligns better with what we have already in iOS config (namely `projectPath` and `pbxprojPath`).

I also made searching for the podspec a bit smarter – by not only looking in the library root, but also in its `sourceDir`. With this, I could probably make this change non-breaking, because it covers a lot more libs with podspec resolution. However, the `podspec` as it was (just the name of a podspec) kinda doesn't make much sense and is pretty limited. I think path makes it more obvious.

Also the only user of `podspec` option I know of is @Salakar – let me know how this works for you.

Fixes https://github.com/react-native-community/cli/issues/412

cc @orta @alloy 

Test Plan:
----------

Adjusted existing tests to cover this functionality.


packages/platform-ios/src/config/index.js
```diff
@@ -47,13 +47,19 @@ export function projectConfig(
...
+    sourceDir,
     folder,
     pbxprojPath: path.join(projectPath, 'project.pbxproj'),
     podfile: findPodfilePath(projectPath),
-    podspec: userConfig.podspec || findPodspecName(folder),
+    podspecPath:
+      userConfig.podspecPath ||
+      // podspecs are usually placed in the root dir of the library or in the
+      // iOS project path
+      findPodspec(folder) ||
```
grabbou: I wonder whether the order is correct. `sourceDir` is within `folder`, so theoretically speaking, `findPodspec(folder)` will always return the same results as `findPodspec(sourceDir)` and never return `null`.

thymikee: I thought the same but that's not how it works, the glob for podspec is `glob('*.podspec', {cwd})` which only traverses the cwd and goes no further. I concluded this as desired behavior for performance reasons (plus it happens that there are node_modules inside that cwd, and unrelated podspecs as well)

grabbou: Okay, so it only looks at the root level and doesn't go any deeper. Smart.

alloy:
Yup, agreed, much better to use full path 👍

PR #444, opened 2019-06-25
docs: document how to disable autolinking for unsupported packages

thymikee:
Summary:
---------

There are packages (like [appcenter](https://github.com/microsoft/appcenter-sdk-react-native/) or [react-native-fbsdk](https://github.com/facebook/react-native-fbsdk#usage)) that are not able to support autolinking on Android just yet, because of too complicated setup. This diff documents how we can disable certain platform of a package to not be autolinked.

By the way I've adjusted some types (please double check). The types naming is pretty confusing even after all this time :|

Test Plan:
----------

Tested on `appcenter` and `react-native-fbsdk`.

Related issue in appcenter: https://github.com/microsoft/appcenter-sdk-react-native/issues/613

### DIFF
--- a/packages/platform-android/native_modules.gradle
+++ b/packages/platform-android/native_modules.gradle
@@ -150,7 +150,7 @@ class ReactNativeModules {
       packageImports = packageImports + packages.collect {
         "// ${it.name}\n${it.packageImportPath}"
       }.join('\n')
-      packageClassInstances = ",\n      " + packages.collect { it.packageInstance }.join(',')
+      packageClassInstances = ",\n      " + packages.collect { it.packageInstance }.join(",\n      ")
     }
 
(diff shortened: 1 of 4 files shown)


packages/platform-android/native_modules.gradle
```diff
@@ -150,7 +150,7 @@ class ReactNativeModules {
       packageImports = packageImports + packages.collect {
         "// ${it.name}\n${it.packageImportPath}"
       }.join('\n')
-      packageClassInstances = ",\n      " + packages.collect { it.packageInstance }.join(',')
+      packageClassInstances = ",\n      " + packages.collect { it.packageInstance }.join(",\n      ")
```
thymikee: this updates the generated PackageList.java to be nicely formatted around package instance invocations

PR #447, opened 2019-06-29
feat: add warnings for `run-ios` and `run-android` when manually linked libraries detected

### LINKED ISSUE #445: Warn about `link`-linked packages when using autolinking
## Describe the Feature

In order for autolink to work properly, the libraries linked with `link` need to be unlinked, otherwise the build will likely fail because of duplicate headers declaration or similar.

## Possible Implementations

We could dry-run `link` before `run-ios` and `run-android` commands, collect the information about already linked modules and output a descriptive warning. Some packages are still not compatible with autolinking, so we can't really make it a hard error. Unless we can somehow tell the users to unlink every package and link the unsupported ones again (so it uses Podfile), but that will be tricky.

thymikee:
Summary:
---------

In order for autolink to work properly, the libraries linked with `link` (or just manually) need to be unlinked, otherwise the build will likely fail because of duplicate declarations.

Example of a warning:

Screenshot 2019-06-29 at 11 59 52

I used `error` to make it more visible, but it doesn't stop the builds. 

Possible improvements:

- don't run if Podfile is not detected (e.g. in Carthage or just non-CocoaPods project)
- maybe run a prompt asking whether to continue the build (don't run on CIs) to draw attention? It may be irritating, so we'd need to run it once in a while, like new version check.
- deduplicate and move the `warnAboutManuallyLinkedLibs` helper to `cli-tools` – involves converting to TypeScript (since ConfigT is not converted yet, I left a TODO for later)

Fixes #445 

Test Plan:
----------

TBD. Init a project, manually link a library, run `run-ios` and `run-android`, observe the warning being displayed.

### DIFF
--- a/packages/platform-android/src/link/warnAboutManuallyLinkedLibs.js
+++ b/packages/platform-android/src/link/warnAboutManuallyLinkedLibs.js
@@ -1,0 +1,51 @@
+// @flow
+
+import chalk from 'chalk';
+import {logger} from '@react-native-community/cli-tools';
+import type {ConfigT} from 'types';
+import getLinkConfig from './index';
+
+// TODO: move to cli-tools once platform-ios and platform-android are migrated
+// to TS and unify with iOS implementation
+export default function warnAboutManuallyLinkedLibs(
+  config: ConfigT,
+  platform: string = 'android',
+  linkConfig: $Call<typeof getLinkConfig> = getLinkConfig(),
+) {
+  let deps = [];
+
+  for (let key in config.dependencies) {
+    const dependency = config.dependencies[key];
+    try {
+      const projectConfig = config.project[platform];
+      const dependencyConfig = dependency.platforms[platform];
+      if (projectConfig && dependencyConfig) {
+        const x = linkConfig.isInstalled(
+          projectConfig,
+          dependency.name,
+          dependencyConfig,
+        );
+        deps = deps.concat(x ? dependency.name : []);
+      }
+    } catch (error) {
+      logger.debug('error');
+    }
+  }
+
+  const installedModules = [...new Set(deps)];
+
+  if (installedModules.length) {
+    logger.error(
+      `React Native CLI uses autolinking for native dependencies, but following modules are linked manually: \n${installedModules
+        .map(
+          x =>
+            `  - ${chalk.bold(x)} ${chalk.dim(
+              `(to unlink run: "react-native unlink ${x}")`,
+            )}`,
+        )
+        .join(
+          '\n',
+        )}\nThis is likely to happen when upgrading React Native from version lower than 0.60 to 0.60 or later. Please unlink them as they are likely to cause build failures. You can do so with "react-native unlink" command as shown above. If a library is not compatible with autolinking yet, please ignore this warning and notify the library maintainers.`,
+    );
+  }
+}
--- a/packages/platform-android/src/link/warnAboutManuallyLinkedLibs.js
+++ b/packages/platform-android/src/link/warnAboutManuallyLinkedLibs.js
@@ -28,7 +28,7 @@ export default function warnAboutManuallyLinkedLibs(
         deps = deps.concat(x ? dependency.name : []);
       }
     } catch (error) {
-      logger.debug('error');
+      logger.debug('Checking manually linked modules failed.', error);
     }
   }
 
--- a/packages/platform-ios/src/config/index.js
+++ b/packages/platform-ios/src/config/index.js
@@ -9,10 +9,13 @@
  */
 
 import path from 'path';
+import {memoize} from 'lodash';
 import findProject from './findProject';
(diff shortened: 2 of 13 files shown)


packages/platform-ios/src/config/index.js
```diff
@@ -9,11 +9,14 @@
  */
 
 import path from 'path';
+import {memoize} from 'lodash';
 import findProject from './findProject';
 import findPodfilePath from './findPodfilePath';
 import findPodspec from './findPodspec';
 import type {UserConfigT} from 'types';
 
+const memoizedFindProject = memoize(findProject);
```
thymikee: `findProject` is expensive (runs FS operations through `glob`) and often called with the same `folder` argument, hence memoizing. Without this change, displaying the warning before builds would essentially be unusable, since it runs ~10s for as small as 4 native modules. With memoization the time is less than 800ms, yielding ~12x improvement that gets bigger with more deps.

packages/platform-ios/src/link/isInstalled.js
```diff
@@ -4,19 +4,33 @@
...
+) {
+  let project;
+
+  if (memo.has(projectConfig.pbxprojPath)) {
+    project = memo.get(projectConfig.pbxprojPath);
+  } else {
+    project = xcode.project(projectConfig.pbxprojPath).parseSync();
+    memo.set(projectConfig.pbxprojPath, project);
+  }
+  // const project = xcode.project(projectConfig.pbxprojPath).parseSync();
```
Krizzu: comment left out

packages/platform-android/src/link/warnAboutManuallyLinkedLibs.js
```diff
@@ -0,0 +1,51 @@
...
+    } catch (error) {
+      logger.debug('Checking manually linked modules failed.', error);
+    }
+  }
+
+  const installedModules = [...new Set(deps)];
+
+  if (installedModules.length) {
+    logger.error(
+      `React Native CLI uses autolinking for native dependencies, but following modules are linked manually: \n${installedModules
```
cpojer: "but the following"

packages/platform-android/src/link/warnAboutManuallyLinkedLibs.js
```diff
@@ -0,0 +1,51 @@
...
+        .map(
+          x =>
+            `  - ${chalk.bold(x)} ${chalk.dim(
+              `(to unlink run: "react-native unlink ${x}")`,
+            )}`,
+        )
+        .join(
+          '\n',
+        )}\nThis is likely to happen when upgrading React Native from version lower than 0.60 to 0.60 or later. Please unlink them as they are likely to cause build failures. You can do so with "react-native unlink" command as shown above. If a library is not compatible with autolinking yet, please ignore this warning and notify the library maintainers.`,
```
cpojer: "This is likely happening when upgrading React Native from below 0.60 to 0.60 or above. Going forward, you can unlink this dependency via `react-native unlink <dependency>` and it will included in your app automatically. If a library isn't compatible with autolinking, disregard this message and notify the library maintainers.

Read more about autolinking: […]"

packages/platform-ios/src/link/warnAboutManuallyLinkedLibs.js
```diff
@@ -0,0 +1,51 @@
...
+    } catch (error) {
+      logger.debug('Checking manually linked modules failed.', error);
+    }
+  }
+
+  const installedModules = [...new Set(deps)];
+
+  if (installedModules.length) {
+    logger.error(
+      `React Native CLI uses autolinking for native dependencies, but following modules are linked manually: \n${installedModules
```
cpojer: "but the following"

packages/platform-ios/src/link/warnAboutManuallyLinkedLibs.js
```diff
@@ -0,0 +1,51 @@
...
+        .map(
+          x =>
+            `  - ${chalk.bold(x)} ${chalk.dim(
+              `(to unlink run: "react-native unlink ${x}")`,
+            )}`,
+        )
+        .join(
+          '\n',
+        )}\nThis is likely to happen when upgrading React Native from version lower than 0.60 to 0.60 or later. Please unlink them as they are likely to cause build failures. You can do so with "react-native unlink" command as shown above. If a library is not compatible with autolinking yet, please ignore this warning and notify the library maintainers.`,
```
cpojer: Same as above.

cpojer:
I left a few comments to improve the messaging a little bit. A link to autolinking docs would be useful so people know what is going on.

When saying "notify library maintainers" it would be great to link to the URL listed in the library's `package.json`.

thymikee:
Thanks! 

> When saying "notify library maintainers" it would be great to link to the URL listed in the library's package.json.

I'm doing something similar here: https://github.com/react-native-community/cli/pull/450

PR #450, opened 2019-06-30
chore: group deprecated rnpm warnings, add migration guide

thymikee:
Summary:
---------

One warning with all deprecated packages seems nicer and less intrusive. Also added links to the packages home pages for user convenience when informing maintainers.

Screenshot 2019-06-30 at 12 45 15

Also added a migration guide to the configuration docs.

cc @satya164 @turnrye 	

Test Plan:
----------

Nope

### DIFF
--- a/packages/cli/src/tools/config/index.js
+++ b/packages/cli/src/tools/config/index.js
@@ -187,9 +187,7 @@ function loadConfig(projectRoot: string = process.cwd()): ConfigT {
       `The following packages use deprecated "rnpm" config that will stop working from next release:\n${depsWithWarnings
         .map(
           ([name, link]) =>
-            `  - ${chalk.bold(name)} ${chalk.dim(
-              `(${chalk.underline(link)})`,
-            )}`,
+            `  - ${chalk.bold(name)}: ${chalk.dim(chalk.underline(link))}`,
         )
         .join(
           '\n',
--- a/packages/cli/src/tools/config/index.js
+++ b/packages/cli/src/tools/config/index.js
@@ -95,13 +95,23 @@ function loadConfig(projectRoot: string = process.cwd()): ConfigT {
     },
   };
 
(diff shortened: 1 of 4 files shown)


packages/cli/src/tools/config/index.js
```diff
@@ -172,6 +182,23 @@ function loadConfig(projectRoot: string = process.cwd()): ConfigT {
...
+      `The following packages use deprecated "rnpm" config that will stop working from next release:\n${depsWithWarnings
+        .map(
+          ([name, link]) =>
+            `  - ${chalk.bold(name)} ${chalk.dim(
+              `(${chalk.underline(link)})`,
+            )}`,
+        )
+        .join(
+          '\n',
+        )}\nPlease notify its maintainers about it. You can find more details at ${chalk.dim.underline(
```
satya164:
```suggestion
        )}\nPlease notify their maintainers about it. You can find more details at ${chalk.dim.underline(
```

packages/cli/src/tools/config/index.js
```diff
@@ -172,6 +182,23 @@ function loadConfig(projectRoot: string = process.cwd()): ConfigT {
     initialConfig,
   );
 
+  if (depsWithWarnings.length) {
+    logger.warn(
+      `The following packages use deprecated "rnpm" config that will stop working from next release:\n${depsWithWarnings
+        .map(
+          ([name, link]) =>
+            `  - ${chalk.bold(name)} ${chalk.dim(
+              `(${chalk.underline(link)})`,
```
cpojer: I recommend using: `name: url` instead of the parenthesis.

thymikee: so:
```
  - [bold appcenter]: [dimmed,underlined link]
```
?

cpojer: I like it.

PR #453, opened 2019-07-01
chore: fix grammar in warning

thymikee:
Summary:
---------

cc @cpojer


thymikee:
I'll add another fixes to this diff, so hold on :D

PR #454, opened 2019-07-02
chore(docs): simplify autolinking

thymikee:
Summary:
---------

Used http://www.hemingwayapp.com/ to improve the readability.

### DIFF
--- a/docs/autolinking.md
+++ b/docs/autolinking.md
@@ -1,10 +1,21 @@
 # Autolinking
 
-React Native libraries usually come with platform-specific (native) code. Autolinking is a mechanism that allows your project to discover and use this code.
+React Native libraries often come with platform-specific (native) code. Autolinking is a mechanism that allows your project to discover and use this code.
 
-Add a library using your favorite package manager and run the build. That's it. No more editing build config files to use native code.
+Add a library using your favorite package manager and run the build:
 
-> Autolinking is a replacement for [`react-native link`](./linking.md). You're expected to `unlink` native dependencies (if any) for it to work well. It will also ease React Native upgrades in the future.
+```sh
+# install
+yarn add react-native-webview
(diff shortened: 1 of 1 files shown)


docs/autolinking.md
```diff
@@ -1,20 +1,14 @@
 # Autolinking
 
-Autolinking is a mechanism built into CLI that allows adding a dependency with native components for React Native to be as simple as:
+React Native libraries usually come with platform-specific (native) code. Autolinking is a mechanism that allows your project to discover and use this code.
```
cpojer: "often" instead of "usually"

docs/autolinking.md
```diff
@@ -1,20 +1,14 @@
...
 
-```sh
-yarn add react-native-webview
-```
+Add a library using your favorite package manager and run the build. That's it. No more editing build config files to use native code.
```
cpojer: This sounds confusing. Can you specify which package managers you are referring to? An example is helpful too.

docs/autolinking.md
thymikee: Ah, that was my first attempt and I changed it :P

docs/autolinking.md
```diff
@@ -1,20 +1,14 @@
...
 
-> Autolinking is a replacement for [`react-native link`](./linking.md) that brings new features (such as ability to easily integrate native dependencies on iOS) and fixes some of the long-standing issues.
+> Autolinking is a replacement for [`react-native link`](./linking.md). You're expected to `unlink` native dependencies (if any) for it to work well. It will also ease React Native upgrades in the future.
```
cpojer: "If you have been using React Native before version 0.60, please `unlink` native dependencies if you have any from a previous install. <Example>".

docs/autolinking.md
thymikee: Yea, how about that: fb156f4? (funny hash)

PR #458, opened 2019-07-03
Fixed iOS simulator naming problem

iKoala:
Summary:
---------

react-native cli --simulator option is not working with some default simulator name in Xcode 10.2.1.
(e.g. "iPad Pro (9.7-inch)" )
Changed the regex in `findMatchingSimulator.js` to parse --simulator option string to more like Xcode default simulator naming convention, `${simulatorName} - ${iosVersion}` (e.g. "iPad Pro (9.7-inch) - 12.2" )

Test Plan:
----------

### DIFF
--- a/packages/platform-ios/src/commands/runIOS/__tests__/findMatchingSimulator-test.js
+++ b/packages/platform-ios/src/commands/runIOS/__tests__/findMatchingSimulator-test.js
@@ -602,7 +602,7 @@ describe('findMatchingSimulator', () => {
             ],
           },
         },
-        'iPhone 6s (10.0)',
+        'iPhone 6s - 10.0',
       ),
     ).toEqual({
       udid: 'CBBB8FB8-77AB-49A9-8297-4CCFE3189C22',
@@ -671,7 +671,7 @@ describe('findMatchingSimulator', () => {
             ],
           },
         },
-        'iPhone 6s (10.0)',
+        'iPhone 6s - 10.0',
       ),
     ).toEqual(null);
   });
--- a/packages/platform-ios/src/commands/runIOS/findMatchingSimulator.js
+++ b/packages/platform-ios/src/commands/runIOS/findMatchingSimulator.js
@@ -26,9 +26,21 @@ function findMatchingSimulator(simulators, simulatorString) {
   }
   const devices = simulators.devices;
 
+  // Failed to run with --simulator option in Xcode 10.2.1 update for simulator name with brackets
+  // e.g "iPad Pro (9.7-inch)"
   const parsedSimulatorName = simulatorString
     ? simulatorString.match(/(.*)? (?:\((.*)?\))?/)
     : [];
+
+  // Changed to support new Xcoce simulator name and iOS/tvOS version support
+  // e.g. --simulator='iPhone XS'
+  // e.g. --simulator='iPhone XS - 12.0'
+  // e.g. --simulator='iPad Pro (9.7-inch)'
+  // e.g. --simulator='iPad Pro (9.7-inch) - 12.2'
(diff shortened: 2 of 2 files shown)


packages/platform-ios/src/commands/runIOS/findMatchingSimulator.js
```diff
@@ -26,9 +26,21 @@ function findMatchingSimulator(simulators, simulatorString) {
...
+  // e.g "iPad Pro (9.7-inch)"
+  // const parsedSimulatorName = simulatorString
+  //   ? simulatorString.match(/(.*)? (?:\((.*)?\))?/)
+  //   : [];
+
+  // Changed to support new Xcode simulator name and iOS/tvOS version support
+  // e.g. --simulator='iPhone XS'
+  // e.g. --simulator='iPhone XS - 12.0'
+  // e.g. --simulator='iPad Pro (9.7-inch)'
+  // e.g. --simulator='iPad Pro (9.7-inch) - 12.2'
```
thymikee: These should be moved to the test instead of a comment

packages/platform-ios/src/commands/runIOS/findMatchingSimulator.js
```diff
@@ -26,9 +26,21 @@ function findMatchingSimulator(simulators, simulatorString) {
   }
   const devices = simulators.devices;
 
+  // Failed to run with --simulator option in Xcode 10.2.1 update for simulator name with brackets
+  // e.g "iPad Pro (9.7-inch)"
+  // const parsedSimulatorName = simulatorString
+  //   ? simulatorString.match(/(.*)? (?:\((.*)?\))?/)
+  //   : [];
```
thymikee: Please remove the old code

packages/platform-ios/src/commands/runIOS/findMatchingSimulator.js
```diff
@@ -26,9 +26,21 @@ function findMatchingSimulator(simulators, simulatorString) {
...
+
+  // Changed to support new Xcode simulator name and iOS/tvOS version support
+  // e.g. --simulator='iPhone XS'
+  // e.g. --simulator='iPhone XS - 12.0'
+  // e.g. --simulator='iPad Pro (9.7-inch)'
+  // e.g. --simulator='iPad Pro (9.7-inch) - 12.2'
   const parsedSimulatorName = simulatorString
-    ? simulatorString.match(/(.*)? (?:\((.*)?\))?/)
+    ? simulatorString.match(/(.*)? - (?:(\d+\.\d+)?$)?/)
```
thymikee: Shouldn't hyphen be optional?

packages/platform-ios/src/commands/runIOS/__tests__/findMatchingSimulator-test.js
```diff
@@ -671,7 +671,7 @@ describe('findMatchingSimulator', () => {
             ],
           },
         },
-        'iPhone 6s (10.0)',
```
thymikee: Instead of changing this test, can you please add a new case? I'd like to be sure we don't introduce any regression

iKoala: @thymikee got it, I really didn't think about the support for old Xcode simulator name.
But I agree with you, I think we should make it more compatible with previous version Xcode simulator name because there are quite a lot of developers who still has the old simulator setup.
I will close this pull request first and create another one to make it more compatible with old versions, thanks for your time.

iKoala:
I will close this pull request first and create another one to make it more compatible with old Xcode versions, thanks for your time.

Issue #367, opened 2019-05-06
Improve `run-ios` experience when selecting a device/simulator


lucasbento:
## Describe the Feature

This is similar to what has been brought up on [#142](https://github.com/react-native-community/cli/issues/142#issuecomment-460011052), basically add a simpler way to specify which device/simulator you want to run your app on.

## Possible Implementations

```sh
react-native run-ios --device="iPhone X" # will pick up the first iPhone X
react-native run-ios --device="iPhone X 12.1"
react-native run-ios --list-devices # will list devices for iOS
react-native --list-devices # list connected devices, maybe for both platforms
```

lucasbento:
cc @satya164, perhaps you have some thoughts on this as you are working on #142.

lucasbento:
Similar work has been done on https://github.com/wswebcreation/start-ios-simulator.

thymikee:
Can you elaborate more on how to make the experience better? We already have `--device` flag and by default we run first device/some simulator (not sure about the logic here yet). It would be bad if we made the `run-ios` command always interactive imo.

lucasbento:
I'm also against making the `run-ios` command always interactive, this is more like an idea to add a different way of selecting a device to run on _and_ to improve the `--device` argument to accept part of device names.

Currently if you specify `--udid` or `--device` wrong you will get an error with a list of available devices which is currently the only way that you can see their info in the CLI.

adamTrz:
Not stale

adamTrz:
This will be fixed by #1676 once that's merged.

PR #460, opened 2019-07-03
Fixed iOS simulator naming problem (name with brackets) in latest Xcode update

iKoala:
Summary:
---------
react-native cli --simulator option is not working with some default simulator name in Xcode 10.2.1.
(e.g. "iPad Pro (9.7-inch)" )
Changed the regex in `findMatchingSimulator.js` to parse --simulator option string to find version number only (e.g. "iPad Pro (9.7-inch) (12.2)" )

Test Plan:
----------

### DIFF
--- a/packages/platform-ios/src/commands/runIOS/__tests__/findMatchingSimulator-test.js
+++ b/packages/platform-ios/src/commands/runIOS/__tests__/findMatchingSimulator-test.js
@@ -676,6 +676,226 @@ describe('findMatchingSimulator', () => {
     ).toEqual(null);
   });
 
+  it('should return iPad(name with bracket) simulator if simulator name is in the list', () => {
+    expect(
+      findMatchingSimulator(
+        {
+          devices: {
+            'iOS 12.0': [
+              {
+                state: 'Shutdown',
+                availability: '(unavailable, runtime profile not found)',
+                name: 'iPhone 4s',
+                udid: 'B9B5E161-416B-43C4-A78F-729CB96CC8C6',
+              },
+              {
+                state: 'Shutdown',
+                availability: '(available)',
+                name: 'iPhone 5',
+                udid: '1CCBBF8B-5773-4EA6-BD6F-C308C87A1ADB',
+              },
+              {
+                state: 'Shutdown',
+                availability: '(available)',
+                name: 'iPhone 6',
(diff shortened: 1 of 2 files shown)


thymikee:
Looks like it doesn't match some simulators? https://regexr.com/4grth

iKoala:
> Looks like it doesn't match some simulators? https://regexr.com/4grth

Yes, if there isn't any matched version string, it will use the whole string as simulator name which is handled in the next code block: 
https://github.com/react-native-community/cli/blob/2cd6c7fd6083f2101989dbfbb7b4f31a8ecc9aeb/packages/platform-ios/src/commands/runIOS/findMatchingSimulator.js#L32-L37

Although I think it's weird, it's the original design idea.
You can check with the previous regex before bugfix which gives the same result: 
https://regexr.com/4gs27
but it's bugged with simulator name "iPad Pro (9.7-inch)" because it parsed "9.7-inch" as version string and failed to find the simulator.

My PR mainly handle the bug for simulator name "iPad Pro (9.7-inch)" (without version specified) 
https://regexr.com/4gs2m
which won't parse "9.7-inch" as version string.

thymikee:
Do you think we could further simplify this code somehow?

iKoala:
> Do you think we could further simplify this code somehow?

Maybe, but considering to avoid regression, personally I think it doesn't worth the effort to simplify the code in `findMatchingSimulator.js`, instead it's better to improve usability of the `run-ios` command in the following ways:

1. Separate simulator name and iOS/tvOS version string with `--simulator` and new `--os-version` options
2. Do not manipulate simulator name provided by `--simulator` option from users to make it less confusing for users and make use of the Xcode simulator command "Copy Name". This also complies with the latest Xcode simulator naming changes. (but we still need to parse the simulator name for version string to support old version of Xcode simulator naming)
3. `--uuid` option should support simulator but not only running on device, i don't understand why `--uuid` option doesn't support simulator searching
4. Better verbose message and give `--simulator`, `--uuid` option example when failed to find specified simulator/device
5. (ref. #367 ) Add `run-ios --list-devices` command so that users can check the simulator/device name and versions which used by cli internally

thymikee:
Thank you very much! We'd appreciate any improvements to the workflow. Hope you could help on that as well 🙂

Issue #423, opened 2019-06-09
Standardize how we make HTTP requests?


matei-radu:
During the review of [#268](https://github.com/react-native-community/cli/pull/268), we noticed that we are using multiple fetch implementations across the CLI. These are:
 
1. `fetch` function from [`cli/src/tools/fetch.js `](https://github.com/react-native-community/cli/blob/master/packages/cli/src/tools/fetch.js). This is a very, very minimalistic GET-only wrapper around the standard `https` module. 
2. `httpsGet` function from [`cli/src/tools/releaseChecker/getLatestRelease.js`](https://github.com/react-native-community/cli/blob/master/packages/cli/src/tools/releaseChecker/getLatestRelease.js). This is, again, a GET-only wrapper around `https` that exists because the above mentioned `fetch` does not satisfies the requirements ([see discussion](https://github.com/react-native-community/cli/pull/268#discussion_r283701172)).
3. `node-fetch` in [`tools/src/isPackagerRunning.js`](https://github.com/react-native-community/cli/blob/644380f769ec1c9b53259c0b45c66f3c36270d0e/packages/tools/src/isPackagerRunning.ts).

#### Should we standardize the way we make HTTP requests in this project?

I propose to discuss and decide on one implementation to use across the entire CLI to keep things consistent. Of course, it doesn't have to be one of the above implementations if you want to suggest a different one (e.g. [`axios`](https://github.com/axios/axios) or the standard `https` Node module) 😃

I'll start! I believe we can stick with `node-fetch` and replace the other implementations with it. We should also mention somewhere in the Contributing doc to use `node-fetch` when making network request so that we avoid this inconsistency happening again in the future.

thymikee:
Thanks for bringing this up. I'm also for using `node-fetch` since we have it already. However, I'd build a small abstraction around it, so we can switch the implementation easily in the future, and call it `tools/networking.js` or so. Then it will be pretty obvious what we use to handle network requests :)

PR #463, opened 2019-07-03
chore: Standarise fetch in CLI

Esemesek:
Summary:
---------

Standardise  `fetch` calls in CLI.

Fixes https://github.com/react-native-community/cli/issues/423

Test Plan:
----------

- [x] - Make sure that tests pass
- [x] - Make sure that commands using new fetch function work on real world example


packages/cli/src/commands/init/__tests__/templateName.test.js
```diff
@@ -1,12 +1,18 @@
...
+import {fetch} from '@react-native-community/cli-tools';
 
-jest.mock('../../../tools/fetch', () => ({fetch: jest.fn()}));
+jest.mock('@react-native-community/cli-tools', () => ({fetch: jest.fn()}));
 
 const RN_NPM_PACKAGE = 'react-native';
 const ABS_RN_PATH = '/path/to/react-native';
 
+const mockFetch = (value: any, status: number) => {
+  (fetch: any).mockImplementation(() =>
```
thymikee: this was mockImplementationOnce

packages/cli/src/commands/init/templateName.js
```diff
@@ -81,7 +81,13 @@ async function tryTemplateShorthand(templateName: string) {
       `https://registry.yarnpkg.com/${reactNativeTemplatePackage}/latest`,
     );
 
-    if (JSON.parse(response).name) {
+    if (response.status < 200 || response.status > 299) {
+      throw new Error(`Failed to load page, status code: ${response.status}`);
+    }
+
+    const body = await response.json();
```
thymikee: can't we do the `response.json` inside fetch function?

packages/cli/src/commands/upgrade/__tests__/upgrade.test.js
```diff
@@ -217,7 +225,7 @@ test('fetches regular patch, adds remote, applies patch, installs deps, removes
   );
 }, 60000);
 test('fetches regular patch, adds remote, applies patch, installs deps, removes remote when updated from nested directory', async () => {
-  (fetch: any).mockImplementation(() => Promise.resolve(samplePatch));
+  mockFetch(samplePatch, 200);
```
thymikee: looks like we could use single mock in testing tools

packages/cli/src/commands/init/templateName.js
Esemesek: Yes, but sometimes we check for status of the request. Maybe we should add generic handler if the response status code is not 2xx?

packages/cli/src/commands/init/templateName.js
```diff
@@ -81,7 +81,13 @@ async function tryTemplateShorthand(templateName: string) {
       `https://registry.yarnpkg.com/${reactNativeTemplatePackage}/latest`,
     );
 
-    if (JSON.parse(response).name) {
+    if (response.status < 200 || response.status > 299) {
+      throw new Error(`Failed to load page, status code: ${response.status}`);
```
thymikee: It's a package, not page. Add some information about it

packages/cli/src/commands/init/templateName.js
thymikee: we can check for status inside fetch and then return something like `{data: response.json(), status}`

packages/cli/src/commands/init/templateName.js
Esemesek: This error message is copy pasted from previous error message that was here before. We can add some generic error handling in this case. I started a conversation in one of the previous comment. Closing this one.

packages/cli/src/commands/init/templateName.js
Esemesek: That makes sense. On it.

thymikee:
ping

Esemesek:
@thymikee I'll finish this likely today C:

packages/cli/src/commands/upgrade/upgrade.js
```diff
@@ -46,7 +45,14 @@ const getPatch = async (currentVersion, newVersion, config) => {
   logger.info(`Fetching diff between v${currentVersion} and v${newVersion}...`);
 
   try {
-    patch = await fetch(`${rawDiffUrl}/${currentVersion}..${newVersion}.diff`);
+    const {status, data} = await fetch(
+      `${rawDiffUrl}/${currentVersion}..${newVersion}.diff`,
+    );
+
+    if (status < 200 || status > 299) {
+      throw new Error(`Failed to load page, status code: ${status}`);
```
thymikee: Users of `upgrade` command have no idea what page failed to load, status code doesn't tell much as well. Can you make this error a bit more descriptive, and maybe show the error message as well? Also, use `CLIError`.

PR #466, opened 2019-07-03
Changed the URL to be valid for configuration.md

RohanTalip:


packages/cli/src/tools/config/index.js
```diff
@@ -196,7 +196,7 @@ function loadConfig(projectRoot: string = process.cwd()): ConfigT {
         .join(
           '\n',
         )}\nPlease notify their maintainers about it. You can find more details at ${chalk.dim.underline(
-        'https://react-native-community/cli/docs/configuration.md#migration-guide',
+        'https://github.com/react-native-community/cli/blob/master/docs/configuration.md',
```
Esemesek:
```suggestion
        'https://github.com/react-native-community/cli/blob/master/docs/configuration.md#migration-guide',
```

Esemesek: We should point to the section with migrate guide.

RohanTalip: Oops, done.

Issue #484, opened 2019-07-05
Metro Bundler immediately closes after react-native run-android


theweavrs:
## Environment
```
System:
    OS: Windows 10
    CPU: (8) x64 Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz
    Memory: 8.99 GB / 15.89 GB
  Binaries:
    Node: 10.16.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.16.0 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
    npm: 6.9.0 - C:\Program Files\nodejs\npm.CMD
  SDKs:
    Android SDK:
      Android NDK: 18.1.5063045
  IDEs:
    Android Studio: Version  3.3.0.0 AI-182.5107.16.33.5264788
```
```
react-native-cli: 2.0.1
react-native: 0.60.0
```
## Description
Doing `yarn react-native run-android` opens the `metro bundler` but it immediately closes. Don't know if its a bug with the `cli` or the `metro` itself. This only happens on Windows, I think.

Things I have tried:
1. Restarting my system
2. Clearing cache etc.

## Reproducible Demo
1. Create a new project using `react-native init`
2. Run `yarn react-native run-android`

thymikee:
We'll need help debugging what's causing this. What does it mean that "it closes"? Can you share more details?

theweavrs:
@thymikee well I can't see the error on it as it only blinks. How can I debug it?

thymikee:
Can you post a video? :D Set up some console.logs in `commands/server/runServer.js`

theweavrs:
@thymikee runServer gives nothing so it's not even getting there.

thymikee:
Btw, does running `yarn react-native start` works without issues? If so, check files related to this diff: https://github.com/react-native-community/cli/pull/310

theweavrs:
Running `launchPackager.bat` directly gives this error:
`error: unknown option `--projectRoot'`

theweavrs:
Doing `yarn react-native start` gives me the console.logs I had set so I am guessing, `runServer` is getting called in that case.

theweavrs:
Running `node .\node_modules\react-native\cli.js --projectRoot` also gives this error:
error: unknown option `--projectRoot'

thymikee:
[...]

PR #489, opened 2019-07-06
Fix Metro Bundler crashing on run-android on Windows

theweavrs:
Summary:
---------
This is a temporary fix for #484 until [this PR](https://github.com/facebook/react-native/pull/25517) get approved in RN repo and a new RN version is released. It should be removed afterwards.

I basically replace the content of `launchPackager.bat` to its fixed version. Everything runs smoothly afterwards.

Test Plan:
----------
Tested this on a new project created using `react-native init` and manually editing the appropriate files to implement the fix. It seems to be working on my Windows machine. Should be tested before merge though.

### DIFF
--- a/packages/platform-android/src/commands/runAndroid/index.js
+++ b/packages/platform-android/src/commands/runAndroid/index.js
@@ -251,6 +251,23 @@ function startServerInNewWindow(port, terminal, reactNativePath) {
     : `export RCT_METRO_PORT=${port}`;
 
   /**
+   * Quick & temporary fix for packager crashing on Windows due to using removed --projectRoot flag
+   * in script. So we just replace the contents of the script with the fixed version.
+   */
+  const launchPackagerScriptContent = `:: Copyright (c) Facebook, Inc. and its affiliates.
+  ::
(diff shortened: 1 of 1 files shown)


packages/platform-android/src/commands/runAndroid/index.js
```diff
@@ -296,6 +314,11 @@ function startServerInNewWindow(port, terminal, reactNativePath) {
   if (/^win/.test(process.platform)) {
     procConfig.detached = true;
     procConfig.stdio = 'ignore';
+    //Temporary fix for #484. See comment on line 254
+    fs.writeFileSync(launchPackagerScript, launchPackagerScriptContent, {
```
ruhaim: fs is not defined. Run CLI with --verbose flag for more details.
ReferenceError: fs is not defined

theweavrs: Are you editing the node_modules files directly?

ruhaim: ok, noted
Thank you for the comment

PR #483, opened 2019-07-05
Support versioned/tagged shorthand templates

Taym95:
I tried to fix #475

PR #495, opened 2019-07-07
chore: remove unstable `init --template` shorthand support

thymikee:
Summary:
---------

This only work with `latest` tag on packages (see more details in #483) and hence cause confusion. For now, only full npm package names are supported. We can introduce some shorthand syntax, but under different flag. Legacy `init` is not affected, users can still use shorthand (and only that) notation there.

Test Plan:
----------

Updated tests.

### DIFF
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -139,7 +139,6 @@ Uses a custom directory instead of `<projectName>`.
 Uses a custom template. Accepts following template sources:
 
 - an npm package name
-- a shorthand name for packages prefixed with `react-native-template-`
 - an absolute path to a local directory
 - an absolute path to a tarball created using `npm pack`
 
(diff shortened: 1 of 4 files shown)


thymikee:
cc @emin93, users can still install new TS template with:
```
npx react-native init Project --template react-native-template-typescript
```

emin93:
Thanks for the info. I think this will cause some trouble for users that are used to the shorthand syntax (causing them to check out the TypeScript package and failing because there is no `template.config.js`). But as long as the old init keeps the shorthand syntax, I think I can live with it. Do you know when this will be released?

thymikee:
There's nothing new added in this diff, only removing stuff that caused confusion, because we added them thinking it's gonna be working fine once RN 0.60 is released.

However, since `typescript` template is the most popular once, I good with adding a warning notice if somebody tries that :) Which tag should I point to so it's always valid?

emin93:
Oh nice, that would be helpful, thanks 😄. For now it's the `next` tag, but I'm not yet if it's going to stay that way.

Issue #492, opened 2019-07-06
Allow configuring metro root


janicduplessis:
## Describe the Feature
AFAIK it is currently impossible to configure the metro root from the cli config file. This is the only thing missing to make it work properly in a monorepo setup where the metro root is the repo root and projects use entry points like `apps/app/index`. This setup allows running multiple apps on a single metro instance.

## Possible Implementations
This could use a value from `userConfig` if provided https://github.com/react-native-community/cli/blob/9e3134b2c33521229ed377899ea96ec3b92d74b6/packages/cli/src/tools/config/index.js#L70

Another alternative that might be better is make it so the metro config for `projectRoot` (metro.config.js) actually works. Seems like it is currently ignored.

## Related Issues
N/A

PR #496, opened 2019-07-07
fix: don't set `projetRoot` for Metro

thymikee:
Summary:
---------

Fixes https://github.com/react-native-community/cli/issues/492 by not overriding user's `metro.config.js`. The `ctx.root` is always set to `process.cwd()` currently, so it didn't make sense anyway.

cc @janicduplessis 	

Test Plan:
----------

Verified on a sample project.


janicduplessis:
Nice, I'll test this diff to confirm it fixes my use case.

thymikee:
Published in `@react-native-community/cli@2.1.2`

PR #497, opened 2019-07-07
fix: fix init command to work with scoped templates without version

BenLorantfy:
Summary:
---------
Fixes init command to work with scoped template names that don't have a version suffix. A regex issue was causing template names like this to return an empty string as the template name:

`@react-native-firebase/template`

Test Plan:
----------

Added a unit test and also tried running init.


thymikee:
Published in `@react-native-community/cli@2.1.2`

Issue #506, opened 2019-07-08
Invalid Podfile Error - dependency with no podspec


RyanThomas73:
## Environment

System:
    OS: macOS 10.14.4
    CPU: (8) x64 Intel(R) Core(TM) i7-3740QM CPU @ 2.70GHz
    Memory: 391.61 MB / 8.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 11.13.0 - /var/folders/st/dpvzyxq913s2mxv8dzb_fcv979pvgz/T/yarn--1562612651190-0.6482734780758797/node
    Yarn: 1.15.2 - /var/folders/st/dpvzyxq913s2mxv8dzb_fcv979pvgz/T/yarn--1562612651190-0.6482734780758797/yarn
    npm: 6.7.0 - /usr/local/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  SDKs:
    iOS SDK:
      Platforms: iOS 12.2, macOS 10.14, tvOS 12.2, watchOS 5.2
    Android SDK:
      API Levels: 19, 21, 22, 23, 24, 25, 26
      Build Tools: 23.0.1, 25.0.2, 25.0.3, 26.0.3
[...]

PR #507, opened 2019-07-08
fix: use_native_modules! warns and skips dependencies without a podspec

RyanThomas73:
Summary:
---------
fix: use_native_modules! displays a descriptive warning and then skips any dependency packages without a podspec (#506)

Test Plan:
----------
1. Create a new react-native project using react-native v0.60.0
2. Install a dependency into `node_modules/` which the `react-native config` command will return a `platform.ios.podspecPath` value of `null/empty`. (e.g. `react-native-text-input-mask` https://github.com/react-native-community/react-native-text-input-mask).
3. Run the `pod install` command from the `ios/` folder.

**Expected Behavior:** The `use_native_modules!` function no longer attempts to load a podspec file using the `nil` `podspec_path` value causing the error message described in #506. The `use_native_modules!` function now skips the dependency without a podspec and continues. At completion, the `pod install` command prints a warning and suggested actions for any/each dependency encountered that did not have the required `podspec` file.

image

### DIFF
--- a/packages/platform-ios/native_modules.rb
+++ b/packages/platform-ios/native_modules.rb
@@ -27,6 +27,17 @@ def use_native_modules!(root = "..", packages = nil)
     next unless package_config = package["platforms"]["ios"]
 
     podspec_path = package_config["podspecPath"]
+
+    # Add a warning to the queue and continue to the next dependency if the podspec_path is nil/empty
+    if podspec_path.nil? || podspec_path.empty?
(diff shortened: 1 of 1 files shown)


RyanThomas73:
Clipboard - July 8, 2019 4_17 PM

RyanThomas73:
Screenshot of what it looks like if I switch the `Pod::UI.warn(..)` call to raising an error instead.

image

RyanThomas73:
Screenshot of what it would look like if I change the code to raising an error in lieu of using the `Pod::UI.warn(..)` call.

image

PR #510, opened 2019-07-09
chore: remove special iOS and Android platform handling for react-native

thymikee:
Summary:
---------

RN 0.60.0 shipped with [`react-native.config.js`](https://github.com/facebook/react-native/blob/0.60-stable/react-native.config.js) that registers commands and `ios` and `android` platform. There's no need to special case this anymore. 

FYI, there's no need to rush in merging it. It's not breaking, unless you're on RC (which didn't publish `react-native.config.js` by mistake)

Test Plan:
----------

`yarn react-native config` in a project should return the same results as currently (with `commands` and `platforms` and `haste` entries).

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -24,8 +24,6 @@
   },
   "dependencies": {
     "@hapi/joi": "^15.0.3",
-    "@react-native-community/cli-platform-android": "^2.1.1",
-    "@react-native-community/cli-platform-ios": "^2.2.0",
     "@react-native-community/cli-tools": "^2.0.2",
     "chalk": "^1.1.1",
     "command-exists": "^1.2.8",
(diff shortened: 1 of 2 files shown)


thymikee:
Ugh, a lot of unit tests rely on this code 😅

grabbou:
Ah, fixing them is not hard @thymikee, just make sure to define a `react-native` mock in a way that it has both platforms. Look what we do for `windows` mock.

Otherwise, tests are failing, because there's no `ios` and `android` platforms, which they are testing.

grabbou:
@thymikee, did you have any progress on this? I might want to give it a try. I have been dreaming of full decoupling for years!

grabbou:
Closing in favor of https://github.com/react-native-community/cli/pull/767

Issue #521, opened 2019-07-11
Make it easier to use libraries which don't support autolinking


satya164:
Maybe we could improve the experience when using libraries which don't support autolinking and/or android x.

This can involve as little as detecting if a library uses old support libraries and showing a warning with instructions on how to use `jetifier`. Similar for autolinking (we could have a list of libraries/support some key in `package.json` to allow a library to say if it can be autolinked).

Another idea is to include `jetifier` in the RN template temporarily during the migration period. It doesn't seem have any downsides and will improve the UX.

thymikee:
> Another idea is to include jetifier in the RN template temporarily during the migration period. It doesn't seem have any downsides and will improve the UX.

Like it. I think it proved to be a good tool. Can you send a PR to `react-native`? That's where the template lives.

> Similar for autolinking (we could have a list of libraries/support some key in package.json to allow a library to say if it can be autolinked).

If we ever gonna do that, I'd go with a list of unsupported libraries that we know of (like appcenter and gesture-handler). I'm hesitating, because it will cause these libraries to migrate later (I know that appcenter is working hard on being compatible though, but no signs from RNGH)

satya164:
What do you think about a key in `package.json` to tell the CLI that it shouldn't try to autolink a library?

thymikee:
They can control it by publishing with `react-native.config.js` that disables autolinking. However that's buggy and I'd like to work on that – seems like a good path forward.

thymikee:
AndroidX is tackled here: https://github.com/react-native-community/cli/pull/526

grabbou:
Right now, a library can disable link entirely. That said, not just auto-linking.

What kind of libraries do not support autolinking, except for CocoaPods-less iOS projects? I'd like to better understand this use case.

RioFiveJack:
hi, i would like to ask if it's possible to override package's `react-native.config.js` ? sometimes i want to have different packages constructor for my app ?

thymikee:
@RioFiveJack https://github.com/react-native-community/cli/blob/master/docs/dependencies.md#platformsandroidpackageinstance

RioFiveJack:
@thymikee thanks for quick reply , i've read it before, but could i put on my project folder and override node_modules packages setting ? if the package also have `react-native.config.js`

thymikee:
Your root config will override the package's one.

grabbou:
Can we get an example of a library that doesn't support autolinking here to have a good testing environment?

[...]

PR #522, opened 2019-07-11
WIP feat: allow dependencies to disable autolinkin

thymikee:
Summary:
---------

A dependency (library) that's not compatible with autolinlking can mark themselves as so using this config:

```js
module.exports = {
  dependency: {
    platforms: {
      android: null,
    },
  },
};

```

Tries to fix https://github.com/react-native-community/cli/issues/521

Test Plan:
----------

Not tested very thoroughly, need to go for now.

### DIFF
--- a/types/index.js
+++ b/types/index.js
@@ -179,8 +179,8 @@ export type UserDependencyConfigT = {
   // Additional dependency settings
   dependency: {
     platforms: {
-      android?: DependencyParamsAndroidT,
-      ios?: ProjectParamsIOST,
+      android?: DependencyParamsAndroidT | null,
+      ios?: ProjectParamsIOST | null,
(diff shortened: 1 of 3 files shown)


types/index.js
```diff
@@ -179,8 +179,8 @@ export type UserDependencyConfigT = {
   // Additional dependency settings
   dependency: {
     platforms: {
-      android?: DependencyParamsAndroidT,
-      ios?: ProjectParamsIOST,
+      android?: DependencyParamsAndroidT | null,
+      ios?: ProjectParamsIOST | null,
```
grabbou: Shouldn't this be `?`

grabbou:
This will unfortunately disable linking on both. I guess we need a different flag.

PR #526, opened 2019-07-12
feat: run `jetifier` before `run-android`

thymikee:
Summary:
---------

Telling people to run `jetifier` seems to be less effective than just running it.

People can opt-out using `--no-jetifier` flag if needed.

cc @satya164 @mikehardy 	

Test Plan:
----------

image

### DIFF
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -326,6 +326,12 @@ Launches the Metro Bundler in a new window using the specified terminal path.
 Run custom gradle tasks. If this argument is provided, then `--variant` option is ignored.
 Example: `yarn react-native run-android --tasks clean,installDebug`.
 
+#### `--no-jetifier`
+
+> default: false
+
+Do not run [jetifier](https://www.npmjs.com/package/jetifier) – the AndroidX transition tool. By default it runs before Gradle to ease working with libraries that don't support AndroidX yet.
+
 ### `run-ios`
 
 Usage:
--- a/packages/platform-android/package.json
+++ b/packages/platform-android/package.json
@@ -6,6 +6,7 @@
   "dependencies": {
     "@react-native-community/cli-tools": "^2.4.1",
     "chalk": "^2.4.2",
+    "jetifier": "^1.6.2",
     "logkitty": "^0.5.0",
     "slash": "^2.0.0",
     "xmldoc": "^0.4.0"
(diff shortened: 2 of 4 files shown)


mikehardy:
Works for me - I'm doing my best to keep the jetifier api as is now and I guess at this point I'll consider it frozen with no flags strictly required except -r for 0.59. @m4tt72 I had a feeling something like this could happen since we are an infrastructure Lego piece, which was part of my desire for the zeroconf api. If we add flags in the future to unlock speed etc they should be optional I suppose

mikehardy:
If there is a change that would help (a default export or whatever) we could take that in jetifier? Also if cli detects it is working on rn59 or lower it should run I  reverse with -r, is that possible

thymikee:
> If there is a change that would help (a default export or whatever) we could take that in jetifier?

Sure, that would be nice, but please mark it's a breaking change :D 

> Also if cli detects it is working on rn59 or lower it should run I reverse with -r, is that possible

The v2 of the CLI is intended to be used with RN >=0.60, we don't expect it to run with 0.59 projects. However, we could add that to `1.x` branch separately if that makes sense? Feel free to send a PR if it doesn't break anything, but I don't think it's as necessary as with 0.60 projects, people can always stay with older versions of libraries when on RN 0.59.

mikehardy:
"people can always stay with older versions of libraries" -> until they are faced with a gnarly dependency tangle or a bugfix they must have - just to say, I think over the next year or so there will be many poor souls trapped on RN59 and I did the '-r' out of deep sympathy even though it seems like it shouldn't matter. I don't have the CLI know-how to do a PR and I'm traveling now, but keep it in mind in case it comes up - something like that on the 1.x branch might be a useful thing to a few

If adding a default export in jetifier is a breaking change and you've got it hooked up here already, I'll leave it as is I think

thymikee:
Sure, I'll see how it performs somewhere after the weekend :). As for the breaking changes, that's why we have semver. But I don't think it's necessary to change it, it works anyway and it's a temporary solution

mikehardy:
okay - I'll probably be out of touch next week so I may not get back to you but it seems like we're on good spot now. As for semver I generally agree and don't fear major versions I just don't even want that barrier for people (going and checking release notes etc, having to force the version change) during this messy upgrade if I can help it - it's hard enough as is. I'm still struggling in my main work project even :rofl:

PR #527, opened 2019-07-12
fix: make "install CocoaPods" step in `init` true by default

sw-yx:
fix minor bug where shouldInstallCocoaPods is falsy by default

Summary:
---------

when i run this cli for the first time with no cocoapods, the prompt asks me to install with a "Y/n" but when i hit enter it is taken as falsy and so it shows "No". but then it goes on to install anyway. seems like a weird quirk with `inquirer`

image

Test Plan:
----------

image

### DIFF
--- a/packages/cli/src/tools/installPods.js
+++ b/packages/cli/src/tools/installPods.js
@@ -45,6 +45,7 @@ async function installPods({
           )} ${chalk.reset.bold(
             "is not installed. It's necessary for iOS project to run correctly. Do you want to install it?",
           )}`,
+          default: true
         },
       ]);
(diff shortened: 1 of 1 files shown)


packages/cli/src/tools/installPods.js
```diff
@@ -45,6 +45,7 @@ async function installPods({
           )} ${chalk.reset.bold(
             "is not installed. It's necessary for iOS project to run correctly. Do you want to install it?",
           )}`,
+          default: true
```
thymikee:
```suggestion
          default: true,
```

PR #533, opened 2019-07-15
fix: shorter and correct paths in init run instructions

### LINKED ISSUE #524: CLI completion message for iOS requires an extra path element
## Environment

N/A

## Description

First off, love this messaging!

Screen Shot 2019-07-12 at 9 49 57 AM

This line

>    Open ios/AwesomeProject.xcworkspace in Xcode or run "xed -b iOS"

Should probably not make the assumption you're in your root dir because none of the others do.

Taym95:
Fix #524


thymikee:
Can we make all the paths inside `printRunInstructions.js` output relative?

Taym95:
Yes sure!

PR #546, opened 2019-07-16
Init duplication

### LINKED ISSUE #441: Better placeholder name for the template
## Description

Cross-posting https://github.com/microsoft/react-native-windows/issues/2663.

## Reproducible Demo

Running `react-native init helloworld_test` will result in `helloworld_test_test` name being used in `package.json`, because:
- our placeholder name is `HelloWorld`
- the regex is case insensitive (?) and will replace `helloworld` in `helloworld_test` (not looking for a whole word, but part of it).

Would be good to use a better placeholder name, something more unique.

Esemesek:
Summary:
---------

Fixes #441 

This is a fix for weird duplication when initializing with the project with `helloworld` name in it. This fix is only for the `new init` because the `old init` logic for name verification exists in `global-cli` package which is no longer maintained.

Test Plan:
----------

Added UT and tested it in real life initialization.

### DIFF
--- a/packages/cli/src/commands/init/errors/HelloWorldError.js
+++ b/packages/cli/src/commands/init/errors/HelloWorldError.js
@@ -1,0 +1,8 @@
+// @flow
+export default class HelloWorldError extends Error {
+  constructor() {
+    super(
+      'Project name shouldn\'t contain "HelloWorld" name in it, because it is CLI\'s default placeholder name. You can find more information https://github.com/react-native-community/cli/issues/441',
+    );
+  }
+}
--- a/packages/cli/src/commands/init/errors/ReservedNameError.js
+++ b/packages/cli/src/commands/init/errors/ReservedNameError.js
@@ -2,7 +2,7 @@
 export default class ReservedNameError extends Error {
   constructor() {
     super(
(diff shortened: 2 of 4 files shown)


packages/cli/src/commands/init/errors/HelloWorldError.js
```diff
@@ -0,0 +1,8 @@
+// @flow
+export default class HelloWorldError extends Error {
+  constructor() {
+    super(
+      'Project name shouldn\'t contain "HelloWorld" name in it, because it is CLI\'s default placeholder name. You can find more information https://github.com/react-native-community/cli/issues/441',
```
thymikee: I'd don't think we need this link to the issue, feel free to remove it

packages/cli/src/commands/init/errors/ReservedNameError.js
```diff
@@ -2,7 +2,7 @@
 export default class ReservedNameError extends Error {
   constructor() {
     super(
-      'Not a valid name for a project. Please do not use the reserved word "React".',
+      'Not a valid name for a project. Please do not use the reserved words "React".',
```
thymikee: Why this change?

Issue #537, opened 2019-07-16
autoLinking doesn't work with monorepo


yozman:
## Environment

```sh
info Fetching system and libraries information...
System:
    OS: macOS 10.14.5
    CPU: (4) x64 Intel(R) Core(TM) i5-4690 CPU @ 3.50GHz
    Memory: 4.87 GB / 20.00 GB
    Shell: 5.3 - /bin/zsh
  Binaries:
    Node: 12.4.0 - ~/.nvm/versions/node/v12.4.0/bin/node
    Yarn: 1.16.0 - ~/.nvm/versions/node/v12.4.0/bin/yarn
    npm: 6.9.0 - ~/.nvm/versions/node/v12.4.0/bin/npm
  SDKs:
    iOS SDK:
      Platforms: iOS 12.2, macOS 10.14, tvOS 12.2, watchOS 5.2
    Android SDK:
      API Levels: 27, 28, 29
      Build Tools: 27.0.3, 28.0.3, 29.0.0
      System Images: android-27 | Intel x86 Atom_64, android-28 | Intel x86 Atom_64, android-29 | Intel x86 Atom_64
  IDEs:
    Android Studio: 3.4 AI-183.6156.11.34.5522156
    Xcode: 10.2.1/10E1001 - /usr/bin/xcodebuild
  npmPackages:
    react: 16.8.6 => 16.8.6 
    react-native: 0.60.0 => 0.60.0
```

## Description

I have a project use monorepo with yarn workspace feature.
it's fail when I `cd ios && pod install`. 
```sh
Detected React Native module pods for RNRippleView, react-native-video, and react-native-webview
Analyzing dependencies
Fetching podspec for `DoubleConversion` from `../node_modules/react-native/third-party-podspecs/DoubleConversion.podspec`
Fetching podspec for `Folly` from `../node_modules/react-native/third-party-podspecs/Folly.podspec`
Fetching podspec for `RNRippleView` from `../`
[!] No podspec found for `RNRippleView` in `../`
```
then I run `react-native config`

<details>
<summary>Details</summary>

```json
{
  "root": "/Users/flyboy/.dee/yozman/yoboto-app",
  "reactNativePath": "/Users/flyboy/.dee/yozman/yoboto-app/node_modules/react-native",
  "dependencies": {
    "react-native-touchable-ripple": {
      "root": "/Users/flyboy/.dee/yozman/yoboto-app/packages/react-native-touchable-ripple",
      "name": "react-native-touchable-ripple",
[...]

PR #547, opened 2019-07-17
fix link dependencies use monorepo

yozman:
Summary:
---------

fix <https://github.com/react-native-community/cli/issues/537>

Test Plan:
----------

### DIFF
--- a/packages/platform-ios/native_modules.rb
+++ b/packages/platform-ios/native_modules.rb
@@ -56,7 +56,7 @@ def use_native_modules!(root = "..", packages = nil)
 
     # Use relative path
     absolute_podspec_path = File.dirname(podspec_path)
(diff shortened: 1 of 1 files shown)


thymikee:
@maciejsimka is onto fixing that, maybe let's wait for him

yozman:
@thymikee 
> Node_modules may not always be there, e.g. when working with local native modules.

I think it should be a npm package folder with package.json in this case.

> @maciejsimka is onto fixing that, maybe let's wait for him

ok

maciejsimka:
[Here's](https://github.com/react-native-community/cli/pull/550) what I came up with. @yozman would you mind verifying if this works in your case?

yozman:
@maciejsimka 
nice job, works for me

PR #563, opened 2019-07-18
fix: Package class name regex tweaks on Android

dratwas:
Summary:
---------
Current regex doesn't support code like this:
```kotlin
class SomePackage() : ReactPackage {}
```

but kotlin does ;)

In this PR i added support for this

Test Plan:
----------
Try autolinking for Andorid with package class like this:
```kotlin
class SomePackage() : ReactPackage {

    override fun createNativeModules(reactContext: ReactApplicationContext): MutableList<NativeModule> {
        return Collections.emptyList()
    }

    override fun createViewManagers(reactContext: ReactApplicationContext): MutableList<SimpleViewManager<View>> {
       return Collections.emptyList()
    }
}
```

### DIFF
--- a/packages/platform-android/src/config/__fixtures__/android.js
+++ b/packages/platform-android/src/config/__fixtures__/android.js
@@ -91,3 +91,145 @@ exports.noPackage = {
     },
   },
 };
+
+exports.findPackagesClassNameKotlinValid = [
+  `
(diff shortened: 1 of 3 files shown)


thymikee:
I'd really appreciate extracting this regex to a function and adding a small unit test suite :)

PR #565, opened 2019-07-19
feat: install pods after upgrade

### LINKED ISSUE #212: Run `pod install` after upgrade
## Describe the Feature

`upgrade` command could use [installPods](https://github.com/react-native-community/cli/blob/master/packages/cli/src/tools/installPods.js) helper, and run: 
```sh
pod repo update
pod install
``` 
after installing JS deps. 

Here's how `init` uses it:
https://github.com/react-native-community/cli/blob/a3d468e54dd3de51c2a0c788c1c8962d1c6abfc1/packages/cli/src/commands/init/init.js#L164-L166

It's likely that `installPods` will need another option introduced to decide whether update the pods or not. It's not necessary during `init` because all pods are local, but it's usually necessary when having 3rd party libraries. We can also be smart about it and check for presence of 3rd party RN libs by accessing `config.dependencies`.

cc @orta @fson what do you think?

maciejsimka:
Summary:
---------

Added calls to `installPods` in `upgrade` command to automatically
update pods after applying patch and installing JS dependencies.
Changed `installPods` helper to also accept argument determining whether
`pod repo update` should be ran before `install` and refactored it
slightly.
Added warnings when patch was not successfully applied to remind users
about running `pod install`.

Resolves #212

Test Plan:
----------

Tested changes manually by initialising new project on 0.60.2 and upgrading to 0.60.3 (happy path) and 0.60.4 (applying diff fails), updated snapshots with warning about manually updating pods if upgrade fails.

### DIFF
--- a/packages/cli/src/commands/init/init.js
+++ b/packages/cli/src/commands/init/init.js
@@ -162,7 +162,7 @@ async function installDependencies({
   });
 
   if (process.platform === 'darwin') {
-    await installPods({projectName, loader});
+    await installPods({projectName, loader, shouldUpdatePods: false});
   }
 
   loader.succeed();
--- a/packages/cli/src/commands/upgrade/__tests__/upgrade.test.js
+++ b/packages/cli/src/commands/upgrade/__tests__/upgrade.test.js
@@ -296,7 +296,7 @@ test('cleans up if patching fails,', async () => {
     [fs] unlink tmp-upgrade-rn.patch
     $ execa git status -s
     error Patch failed to apply for unknown reason. Please fall back to manual way of upgrading
-    warn After resolving conflicts don't forget to change into \\"ios\\" directory and run \\"pod install\\"
+    warn After resolving conflicts don't forget to run \\"pod install\\" inside \\"ios\\" directory
     info You may find these resources helpful:
     • Release notes: https://github.com/facebook/react-native/releases/tag/v0.58.4
     • Manual Upgrade Helper: https://react-native-community.github.io/upgrade-helper/?from=0.57.8&to=0.58.4
--- a/packages/cli/src/commands/upgrade/__tests__/upgrade.test.js
+++ b/packages/cli/src/commands/upgrade/__tests__/upgrade.test.js
@@ -107,6 +107,9 @@ beforeEach(() => {
   fs.unlinkSync = jest.fn((...args) => mockPushLog('[fs] unlink', args));
   logs = [];
   (execa: any).mockImplementation(mockExecaDefault);
+  Object.defineProperty(process, 'platform', {
+    value: 'darwin',
+  });
 });
 
 afterEach(() => {
--- a/packages/cli/src/commands/upgrade/__tests__/upgrade.test.js
+++ b/packages/cli/src/commands/upgrade/__tests__/upgrade.test.js
@@ -192,24 +192,24 @@ test('fetches regular patch, adds remote, applies patch, installs deps, removes 
     opts,
   );
   expect(flushOutput()).toMatchInlineSnapshot(`
-        "info Fetching diff between v0.57.8 and v0.58.4...
-        [fs] write tmp-upgrade-rn.patch
-        $ execa git rev-parse --show-prefix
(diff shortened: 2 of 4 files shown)


packages/cli/src/commands/upgrade/__tests__/upgrade.test.js
```diff
@@ -166,16 +166,16 @@ test('fetches empty patch and installs deps', async () => {
...
-    info Diff has no changes to apply, proceeding further
-    info Installing \\"react-native@0.58.4\\" and its peer dependencies...
-    $ execa npm info react-native@0.58.4 peerDependencies --json
-    $ yarn add react-native@0.58.4 react@16.6.3
-    $ execa git add package.json
-    $ execa git add yarn.lock
-    $ execa git add package-lock.json
-    success Upgraded React Native to v0.58.4 🎉. Now you can review and commit the changes"
-  `);
+        "info Fetching diff between v0.57.8 and v0.58.4...
```
thymikee: Can you revert the indentation? It's a bug in Jest, fixed but not released yet

packages/cli/src/commands/upgrade/upgrade.js
```diff
@@ -163,6 +164,20 @@ const installDeps = async (newVersion, projectDir) => {
   }
 };
 
+const installCocoaPodsDeps = async (projectDir, thirdPartyIOSDeps) => {
```
thymikee: this step is likely to fail, e.g. because of conflicts in Podfile. Can we wrap it in try/catch and output a helpful message?

packages/cli/src/commands/upgrade/upgrade.js
```diff
@@ -328,6 +350,11 @@ async function upgrade(argv: Array<string>, ctx: ConfigT, args: FlagsT) {
           'Please run "git diff" to review the conflicts and resolve them',
         );
       }
+      if (process.platform === 'darwin') {
+        logger.warn(
+          'After resolving conflicts don\'t forget to change into "ios" directory and run "pod install"',
```
thymikee:
```suggestion
          'After resolving conflicts don\'t forget to run "pod install" inside "ios" directory',
```

packages/cli/src/commands/upgrade/upgrade.js
maciejsimka: done, not sure if more detailed message is necessary here?

thymikee:
unit tests are failing, mind taking a look?

maciejsimka:
sure, I'm looking into it. weird thing is they pass locally.

thymikee:
CI runs on Linux, but I can't change it. @grabbou @kelset @Titozzz is it possible to switch it to run on macOS container?

packages/cli/src/commands/init/init.js
```diff
@@ -162,7 +162,7 @@ async function installDependencies({
   });
 
   if (process.platform === 'darwin') {
-    await installPods({projectName, loader});
+    await installPods({projectName, loader, shouldUpdatePods: false});
```
Salakar: Could we have an option to enable this for template configs - there's an issue with any custom templates that use pod specs where the template will fail to init because it could not find the pod specs (because a repo update is required)

See: https://github.com/invertase/react-native-firebase/issues/2351

thymikee: `installPods` checks if the template has a Podfile, so it's gonna work in your template as well

thymikee:
Thank you @maciejsimka ❤️

packages/cli/src/commands/init/init.js
mikehardy: Strangely, this did not work for a react-native-firebase v6 user just now, in that they had stale cocoapods repo while the react-native-firebase v6 starter template was looking for some pretty recent pods, and pod install failed because pod update needed to run. A manual run of pod update for the user resolved it so I think somewhere in here is the root cause but I'm not sure what the interaction is.

I believe this is the template Podfile: https://github.com/invertase/react-native-firebase/blob/master/packages/template/project/ios/Podfile

mikehardy: got a repro and logged it up here https://github.com/react-native-community/cli/issues/752

PR #571, opened 2019-07-22
Fix getVersionToUpgradeTo

sourabhv:
Summary:
---------

Semver's `gt` is defined as `gt(v1, v2): v1 > v2`

This PR fixes the order of versions from `currentVersion, newVersion` which will always return `false` to `newVersion, currentVersion`.

Test Plan:
----------


thymikee:
The check prevents you from upgrading to an older version and it works fine. `newVersion` _is expected_ to be higher than `currentVersion`, but in case it isn't (current is greater than new), it throws.

PR #572, opened 2019-07-22
fix: bring `--projectRoot` argument back to `start`

liorJuice:
Summary:
---------
#496 by @thymikee fixed projectRoot from metro config got overriden by ctx.root. I think the real error was taking the root from ctx instead of args. previously, an option '--projectRoot' was enabled. the projectRoot from args should override the metro config, if it exists, like the other options of the cli. this PR makes the `--projectRoot` option work again. and, if not specified will use the config option.

Notify:
----------
this should also help with storybook that previously used the `--projectRoot` option to start metro from the storybook root folder, instead of including it in your app (which is very non optimal and the recommendation as of now ). @shilman, @Gongreg is that right?

Test Plan:
----------

create a react native project. set a projectRoot option in the metro config to a new file called a.js, check that it work. try using the cli with --projectRoot option set to a new file called b.js, it should override the metro config and load from b.js.

### DIFF
--- a/packages/cli/src/commands/server/runServer.js
+++ b/packages/cli/src/commands/server/runServer.js
@@ -48,6 +48,7 @@ async function runServer(argv: Array<string>, ctx: ConfigT, args: Args) {
     port: args.port,
     resetCache: args.resetCache,
     watchFolders: args.watchFolders,
+    projectRoot: args.projectRoot,
     sourceExts: args.sourceExts,
     reporter,
   });
--- a/packages/cli/src/commands/server/runServer.js
+++ b/packages/cli/src/commands/server/runServer.js
@@ -35,6 +35,7 @@ export type Args = {|
   verbose?: boolean,
   watchFolders?: string[],
   config?: string,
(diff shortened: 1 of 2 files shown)


Gongreg:
Hey @liorJuice.

Currently we recommend --projectRoot as an alternative to including the storybook in the app itself (But first we recommend to add in in the app).

The better approach is to use customExtensions metro config for dev/prod build, but those are more difficult to explain.

If someone spent time to write a guide for proper way to set up additional code in your dev build, then we could remove suggesting --projectRoot.

liorJuice:
@Gongreg isn't the `sb init` command _used to_ add `start-storybook -p 7007 | npm run start --projectRoot storybook` to the project script by default? i thought that was changed because `--projectRoot` ceased to work.

Gongreg:
Oh, then we definitely should add back --projectRoot .

thymikee:
Thanks! Let's add it back for now, until we come up with better interface. Mind fixing type errors?

PR #574, opened 2019-07-22
Add ability to reload all connected devices and simulators

TheSavior:
Summary:
---------

At Facebook we have a hotkey hooked up to the ability to reload all connected devices and simulators. This is incredibly useful! However, we make this work by deploying an app to everyone's laptop that listens for a global hotkey to tell metro to reload. Since it isn't feasible to do that in open source, this seems like a reasonable approach. 

This PR adds the ability to hit the server host /reload, for example: `http://localhost:8081/reload` and metro will send a command to reload all connected apps.

Personally, I plan on wiring up a hotkey with better touch tool to curl this endpoint. I imagine people can set up alfred workflows or wire it into other apps too.

Test Plan:
----------

reloadall

### DIFF
--- a/packages/cli/src/commands/server/runServer.js
+++ b/packages/cli/src/commands/server/runServer.js
@@ -98,6 +98,11 @@ async function runServer(argv: Array<string>, ctx: ConfigT, args: Args) {
   middlewareManager.attachDevToolsSocket(wsProxy);
   middlewareManager.attachDevToolsSocket(ms);
 
+  middlewareManager.getConnectInstance().use('/reload', (req, res) => {
+    ms.broadcast('reload', null);
+    res.end('OK');
+  });
+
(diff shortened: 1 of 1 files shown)


thymikee:
FYI @Esemesek has an idea to put a "Reload" button into the debugger UI, which would be a good starter to overhaul it :)

TheSavior:
It looks like I might need some help on the iOS side to fix the reliability of the socket connection from the app to metro which makes this work. As it is today it seems like this doesn’t work reliably in open source, at least not for me. Let’s hold off a bit on being loud about this maybe. Once we have a release with this I can hopefully ask for some help in discord

TheSavior:
The connection reliability was fixed in this commit https://github.com/facebook/react-native/commit/4ab9da134c988db832b1a2daa90ce38bf8c419eb

We can talk about this as loudly as we'd like now. :)

Issue #576, opened 2019-07-23
Installing pods in CI (specifically CircleCI)


raen79:
Hey guys,

I run into no trouble installing pods locally, but when attempting to install the pods in my CI, using the same XCode, Ruby, and Gemfile version, I run into an error that I see has been reported here before:
`Invalid Podfile file: 784: unexpected token at `

Is there anything you can think of that could cause this when running `pod install`. The only other issue that was opened about it: https://github.com/react-native-community/cli/issues/448 does not provide any solutions to this particular problem.

You can find a full trace here (with `verbose`):
```
/Users/distiller/project/node_modules/.bin/react-native config
  {
    "root": "/Users/distiller/project",
    "reactNativePath": "/Users/distiller/project/node_modules/react-native",
    "dependencies": {
      "appcenter": {
        "root": "/Users/distiller/project/node_modules/appcenter",
        "name": "appcenter",
        "platforms": {
          "ios": null,
          "android": null
        },
        "assets": [],
        "hooks": {
          "postlink": "node node_modules/appcenter/scripts/postlink"
        },
        "params": []
      },
      "@react-native-community/async-storage": {
        "root": "/Users/distiller/project/node_modules/@react-native-community/async-storage",
        "name": "@react-native-community/async-storage",
        "platforms": {
          "ios": {
            "sourceDir": "/Users/distiller/project/node_modules/@react-native-community/async-storage/ios",
            "folder": "/Users/distiller/project/node_modules/@react-native-community/async-storage",
            "pbxprojPath": "/Users/distiller/project/node_modules/@react-native-community/async-storage/ios/RNCAsyncStorage.xcodeproj/project.pbxproj",
            "podfile": null,
            "podspecPath": "/Users/distiller/project/node_modules/@react-native-community/async-storage/RNCAsyncStorage.podspec",
            "projectPath": "/Users/distiller/project/node_modules/@react-native-community/async-storage/ios/RNCAsyncStorage.xcodeproj",
            "projectName": "RNCAsyncStorage.xcodeproj",
            "libraryFolder": "Libraries",
            "sharedLibraries": [],
            "plist": []
          },
          "android": {
            "sourceDir": "/Users/distiller/project/node_modules/@react-native-community/async-storage/android",
            "folder": "/Users/distiller/project/node_modules/@react-native-community/async-storage",
            "packageImportPath": "import com.reactnativecommunity.asyncstorage.AsyncStoragePackage;",
[...]

PR #578, opened 2019-07-23
fix: stream config output in ruby script

raen79:
Attempt at streaming output in ruby in the same way it is currently being done in the gradle setup.

Fixes https://github.com/react-native-community/cli/issues/576

### DIFF
--- a/packages/platform-ios/native_modules.rb
+++ b/packages/platform-ios/native_modules.rb
@@ -5,17 +5,15 @@
 #
 def use_native_modules!(root = "..", config = nil)
   if (!config)
-    command = "./node_modules/.bin/react-native"
-    args = ["config"]
-    output = ""
+    json = []
+
     # Make sure `react-native config` is ran from your project root
     Dir.chdir(root) do
-      output = Pod::Executable.execute_command(command, args, true)
-    end
-
-    json = []
-    output.each_line do |line|
-      json << line
+      IO.popen("./node_modules/.bin/react-native config") do |data|
+        while line = data.gets
+          json << line
+        end
+      end
     end
 
     config = JSON.parse(json.join("\n"))


thymikee:
@raen79 are you able to verify this works on your CI?

raen79:
Yes I tested it in my local environment and CI and it worked. Do you need a screenshot or something like that?

thymikee:
I trust you, thanks. Verified locally that the output is read line-by-line, but having terminal buffer overflow was just an assumption of mine, glad my guess was correct.

thymikee:
@alloy if you feel like there's a more idiomatic Ruby way to solve it, please leave a note :)

alloy:
No this is fine, as far as idiomatic Ruby goes, but this change does degrade normal CocoaPods UX somewhat:

* `pod install --verbose` will no longer include the command and output
* when an error occurs there will be no graceful handling

Additionally, CocoaPods’ [Executable](https://github.com/CocoaPods/CocoaPods/blob/master/lib/cocoapods/executable.rb) module _should_ be streaming, so I'm a little suspect about this change 🤔 

@raen79 Did you also apply the change @thymikee suggested about not pretty-printing the config JSON?

thymikee:
Would be lovely to make it more CocoaPods friendly then in a followup.

raen79:
@alloy I did not apply the changes as I saw he opened his own PR separately. I do agree this would make debugging more difficult in some cases, but would allow for CI builds to work which I think would otherwise be a major drawback from upgrading to react native 0.6+.

alloy:
Sure, totally agreed, CI should work of course—but this isn’t an either/or issue, it just needs to be debugged further. Is there a reproduction available? (Not that I have any time right now, but any further work will need a repro.)

raen79:
I can't provide my source code of course, but from what I see any project run on CircleCI would fail with the previous versions when running `pod install`.

PR #588, opened 2019-07-25
adding projectRoot to options and resolve it

liorJuice:
Summary:
---------
Following my PR from yesterday #572 , I forgot to include the --projectRoot option descriptor in the PR, which makes it non usable or documented in --help. This PR fixes it.

i also went ahead and added path.resolve to the parsing of it, and of --watchFolders.

Test Plan:
----------
try using the cli start with custom project root set in metro config, it should load from the custom root. try using the cli start with --projectRoot that point to another directory (you can also check for relative path), it should override the metro config. try using the cli with a --watchFolders with relative paths, the folders you specified should be resolved and merged to those in the metro config.

### DIFF
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -408,6 +408,10 @@ Starts the server that communicates with connected devices
 
 Specify port to listen on
 
+#### `--projectRoot [list]`
+
+Path to a custom project root
+
 #### `--watchFolders [list]`
 
 Specify any additional folders to be added to the watch list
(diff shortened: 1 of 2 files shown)


liorJuice:
done

docs/commands.md
```diff
@@ -408,6 +408,10 @@ Starts the server that communicates with connected devices
 
 Specify port to listen on
 
+#### `--projectRoot [list]`
```
mmerickel: this isn't actually a list is it? looks like it's just forwarded to `path.resolve()` which doesn't take a list

thymikee: Good point, it's a string. Mind sending a PR?

mmerickel: https://github.com/react-native-community/cli/pull/632

PR #590, opened 2019-07-25
docs(autolinking): Monorepos

MANTENN:
where to use `use_native_modules`, update every pod path, and updated
example

Summary:
---------
This is for developers who are not familiar with CocoaPods.

### DIFF
--- a/docs/autolinking.md
+++ b/docs/autolinking.md
@@ -47,7 +47,7 @@ root/
     ios/
 ```
 
-you'll need to set a custom root. Pass it as an argument to `use_native_modules!`--set inside a target, adjust the `native_modules`--and every pods--path accordingly:
+you'll need to set a custom root. Pass it as an argument to `use_native_modules!` function inside the targets and adjust the relatively required `native_modules` path accordingly:
 
 ```rb
 # example/ios/Podfile
--- a/docs/autolinking.md
+++ b/docs/autolinking.md
@@ -56,7 +56,6 @@ target 'RNapp' do
   pod 'React', :path => '../../node_modules/react-native/'
   use_native_modules!("../..")
 end
-# ...
 ```
 
 ## Platform Android
--- a/docs/autolinking.md
+++ b/docs/autolinking.md
@@ -53,7 +53,7 @@ you'll need to set a custom root. Pass it as an argument to `use_native_modules!
 # example/ios/Podfile
 require_relative '../../node_modules/@react-native-community/cli-platform-ios/native_modules'
 target 'RNapp' do
-  pod 'React', :path => '../../node_modules/react-native/'
+  # React pods and custom pods here...
   use_native_modules!("../..")
 end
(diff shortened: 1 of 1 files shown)


docs/autolinking.md
```diff
@@ -47,13 +47,16 @@ root/
     ios/
 ```
 
-you'll need to set a custom root. Pass it as an argument to `use_native_modules!` and adjust the `native_modules` path accordingly:
+you'll need to set a custom root. Pass it as an argument to `use_native_modules!`--set inside a target, adjust the `native_modules`--and every pods--path accordingly:
```
thymikee: This reads really hard. Please adjust:
```suggestion
you'll need to set a custom root. Pass it as an argument to `use_native_modules!` function inside the targets and adjust the relatively required `native_modules` path accordingly:
```

docs/autolinking.md
```diff
@@ -47,13 +47,16 @@ root/
...
 
 ```rb
 # example/ios/Podfile
 require_relative '../../node_modules/@react-native-community/cli-platform-ios/native_modules'
+target 'RNapp' do
+  pod 'React', :path => '../../node_modules/react-native/'
```
thymikee:
```suggestion
  # React pods and custom pods here...
```

docs/autolinking.md
```diff
@@ -47,13 +47,16 @@ root/
...
+  use_native_modules!("../..")
+end
 # ...
```
thymikee:
```suggestion
```

docs/autolinking.md
MANTENN: Why would you remove the pod? I added it as an example.

Sorry, I am new to this repo, and I am not familiar with the decisions made early on.

thymikee: Because React pod is not essential to this setting. It's a partially valid example (React pod requires a lot more to be valid) so I chose to add a comment, to make it more obvious. However I can see it may help somebody to realize all paths need to be adjusted. Feel free to leave this line and add a comment 
```rb
# rest of React pods and custom pods go here...
```

MANTENN: Oh, ok. I see, I didn't even notice the comment.

MANTENN: It's a bit more verbose when an example is provided in the documentation. However, as you suggested the example would lack context, and your comment would be necessary. 

Does the documentation utilize any alerts--information, warning, or error?

This would be more quicker to parse for a reader because the reader would know the information alert would serve as context, or an example--at least from my experience.

MANTENN: I assume, I'd have to create a new pull request. Otherwise I can't modify the suggested changes, or propose one.

thymikee: You can disregard the initial suggestion. No need for a separate PR. Make it readable for a typical user and move on :)

thymikee:
Thank you!

PR #596, opened 2019-07-29
feat: set a default `symbolicator.customizeFrame` value

motiz88:
Summary:
---------

Sets a default `symbolicator.customizeFrame` value. This configuration option was added to Metro in https://github.com/facebook/metro/pull/435.

This is part of a redesign of https://github.com/facebook/react-native/pull/24662.

Test Plan:
----------

Equivalent config change tested as part of https://github.com/facebook/react-native/pull/25839.

### DIFF
--- a/packages/cli/src/tools/loadMetroConfig.js
+++ b/packages/cli/src/tools/loadMetroConfig.js
@@ -25,6 +25,13 @@ const getWatchFolders = () => {
 
 const getBlacklistRE = () => createBlacklist([/.*\/__fixtures__\/.*/]);
 
+const INTERNAL_CALLSITES_REGEX = new RegExp(
+  [
+    '/Libraries/Renderer/implementations/.+\\.js$',
+    '/Libraries/BatchedBridge/MessageQueue\\.js$',
+  ].join('|'),
+);
+
 /**
  * Default configuration
  *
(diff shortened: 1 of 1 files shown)


thymikee:
Since this is a new config option, older Metro versions (used by RN 0.60) will spawn a validation warning, like this:

image

Which may confuse people.

Can you add a check to only include this if newer version of Metro is used?

thymikee:
Also, I'm so happy you work on this space to make errors more approachable ❤️

motiz88:
@thymikee  Is there a scenario where people use a new CLI version with an old Metro version? Would it be enough to hold off on merging this until we cut a new Metro release and bump the dependency here?

thymikee:
If you could update Metro without breaking changes, then we could bump it internally here, provided it doesn't clash with RN. It still internally relies on 3 Metro packages: https://github.com/facebook/react-native/blob/59db059dbddb8101212f3739eecf0db494cfab41/package.json#L103-L105, which makes us (the CLI) to be up to date with what's there. 

If you could also make it so that all Metro deps are required from the CLI (which makes sense), that would make us in charge of what's breaking or not. It would also make RN bundler-agnostic at the core. And in theory we could bump major Metro versions without a breaking change.

cpojer:
We are having quite a bunch of big changes in Metro related to changes in RN. We are hoping to bump Metro to 0.56 this week and update the CLI and RN. Could we merge this change as part of that? RN master/0.61 will only work with Metro 0.56 and above.

thymikee:
Ok, let's wait for Metro 0.56, merge this alongside and put 2.x into a branch for maintenance mode. 

cc @Esemesek if you feel like moving some CLI stuff to TS, this is the time, as structural changes will be hard to port between branches.

motiz88:
With #630 now merged, I think we can merge this as well?

thymikee:
@motiz88 yep! Could you rebase and fix Flow?

motiz88:
The failure in e2e-tests also exists on master: https://circleci.com/gh/react-native-community/cli/8256

thymikee:
@motiz88 FYI this is available in `v3.0.0-alpha.2`

PR #597, opened 2019-07-29
imp: Migrate `platform-android` to Typescript

Esemesek:
Summary:
---------

All of the files inside `platform-android` were migrated to TS. Some of the types from common Flow definitions were moved to `platform-android/types.ts`. Some of them are strictly connected to the android package, but some of them will need to be moved to a common package with types. I've added `@ts-ignore` in some places, they will need to be adjusted in the future.

TODO:

- [x] - Fix linter
- [ ] - ~~Adjust common types~~ - will add a separate package while extracting next package

Test Plan:
----------

Unit test and CI are passing


packages/platform-android/package.json
```diff
@@ -5,15 +5,22 @@
...
     "build",
     "native_modules.gradle"
-  ]
+  ],
+  "devDependencies": {
+    "@types/execa": "^0.9.0",
+    "@types/fs-extra": "^8.0.0",
+    "@types/glob": "^7.1.1",
+    "@types/xmldoc": "^1.1.4"
+  }
```
thymikee: we need "references" to "@react-native-community/cli-tools" now, see: https://github.com/facebook/jest/blob/481334b645348fc7471baa8b1a41d99c8e1d6017/packages/jest-changed-files/tsconfig.json#L7-L9

packages/platform-android/src/commands/runAndroid/__tests__/runOnAllDevices.test.ts
```diff
@@ -22,6 +22,7 @@ describe('--appFolder', () => {
   });
 
   it('uses task "install[Variant]" as default task', () => {
+    // @ts-ignore
```
thymikee: this is JS file, why it needs ts-ignore?

packages/platform-android/src/commands/runAndroid/index.ts
```diff
@@ -299,7 +307,7 @@ function startServerInNewWindow(port, terminal, reactNativePath) {
    */
   const scriptsDir = path.dirname(launchPackagerScript);
   const packagerEnvFile = path.join(scriptsDir, packagerEnvFilename);
-  const procConfig: Object = {cwd: scriptsDir};
+  const procConfig: any = {cwd: scriptsDir};
```
thymikee: do we need `any`?

packages/platform-android/src/commands/runAndroid/runOnAllDevices.ts
```diff
@@ -54,19 +53,21 @@ function runOnAllDevices(
...
+        packageName,
+        adbPath,
+        args.mainActivity,
+      );
+    },
+  );
 }
 
-function createInstallError(error) {
+function createInstallError(error: any) {
```
thymikee:
```suggestion
function createInstallError(error: unknown) {
```

packages/platform-android/src/commands/runAndroid/tryLaunchAppOnDevice.ts
```diff
@@ -4,14 +4,14 @@
...
- * @flow
  */
 
 import {spawnSync} from 'child_process';
 import {logger, CLIError} from '@react-native-community/cli-tools';
 
 function tryLaunchAppOnDevice(
-  device?: string,
+  device?: string | void,
+  // @ts-ignore
```
thymikee: can you add a comment on what's wrong?

packages/platform-android/src/commands/runAndroid/tryRunAdbReverse.ts
```diff
@@ -4,15 +4,17 @@
...
 
 import {execFileSync} from 'child_process';
 import {logger} from '@react-native-community/cli-tools';
 import getAdbPath from './getAdbPath';
 
 // Runs ADB reverse tcp:8081 tcp:8081 to allow loading the jsbundle from the packager
-function tryRunAdbReverse(packagerPort: number | string, device?: string) {
+function tryRunAdbReverse(
+  packagerPort: number | string,
+  device?: string | void,
```
thymikee:
```suggestion
  device?: string,
```

packages/platform-android/src/commands/runAndroid/__tests__/runOnAllDevices.test.ts
Esemesek: This is a TS file. Check again C:

packages/platform-android/src/link/patches/__tests__/applyParams.test.ts
```diff
@@ -9,12 +9,12 @@ import applyParams from '../applyParams';
...
   it('apply params to the string', () => {
-    expect(applyParams('${foo}', {foo: 'foo'}, 'react-native')).toEqual(
+    expect(applyParams('${foo}', {foo: 'foo'} as any, 'react-native')).toEqual(
       'getResources().getString(R.string.reactNative_foo)',
     );
   });
 
   it('use null if no params provided', () => {
-    expect(applyParams('${foo}', {}, 'react-native')).toEqual('null');
+    expect(applyParams('${foo}', {}, 'react-native')).toEqual('');
```
thymikee: why this changed?

packages/platform-android/src/link/warnAboutManuallyLinkedLibs.ts
```diff
@@ -1,18 +1,16 @@
...
+import {Config} from '../types';
 
 // TODO: move to cli-tools once platform-ios and platform-android are migrated
 // to TS and unify with iOS implementation
 export default function warnAboutManuallyLinkedLibs(
-  config: ConfigT,
+  config: Config,
   platform: string = 'android',
-  linkConfig: $Call<typeof getLinkConfig> = getLinkConfig(),
+  linkConfig: any = getLinkConfig(),
```
thymikee: here's the type for that: https://github.com/react-native-community/cli/blob/eddfd4d20a62d11133687843b52f7095c5c88118/types/index.js#L83-L93

And you already have it here: https://github.com/react-native-community/cli/pull/597/files?file-filters%5B%5D=.json&file-filters%5B%5D=.ts#diff-0e7f694a165ac213be4e26f76b7179d3R42

packages/platform-android/src/types.ts
```diff
@@ -0,0 +1,158 @@
...
+      projectConfig: ProjectConfig,
+      otherDependencies: Array<DependencyConfig>,
+    ) => void,
+    copyAssets: (assets: string[], projectConfig: ProjectConfig) => void,
+    unlinkAssets: (assets: string[], projectConfig: ProjectConfig) => void,
+  };
+}
+
+export interface Config {
+  // Root where the configuration has been resolved from
```
thymikee: how about using TSDoc comments so that it shows up in dialogs?

packages/platform-android/src/link/patches/__tests__/applyParams.test.ts
Esemesek: Oh, my bad. I have changed the logic of `applyParams` function and forgot to revert it back.

packages/platform-android/src/commands/runAndroid/runOnAllDevices.ts
Esemesek: This ain't working, but I have got a different solution.

packages/platform-android/src/commands/runAndroid/tryRunAdbReverse.ts
Esemesek: This also isn't working, as we sometimes pass `undefined` to the function.

packages/platform-android/src/commands/runAndroid/index.ts
Esemesek: Changed type annotation here to more strict.

packages/platform-android/src/commands/runAndroid/tryLaunchAppOnDevice.ts
Esemesek: Added a comment, what should be done to get rid of the suppression

packages/platform-android/src/link/patches/applyParams.ts
```diff
@@ -14,10 +14,11 @@ export default function applyParams(
   params: ProjectParamsAndroid,
   prefix: string,
 ) {
+  // @ts-ignore String.replace function must return string, not string | null.
   return str.replace(/\$\{(\w+)\}/g, (_pattern: string, param: string) => {
     const name = `${toCamelCase(prefix)}_${param}`;
 
     // @ts-ignore
-    return params[param] ? `getResources().getString(R.string.${name})` : '';
+    return params[param] ? `getResources().getString(R.string.${name})` : null;
```
thymikee: Looks like returning `'null'` string would do here?

Esemesek: According to tests yes :P

packages/platform-android/src/types.ts
Esemesek: I've added them. It is not much, but it works.

thymikee:
Anything more to add here?

Esemesek:
@thymikee I was thinking about making an additional package just for common types, but it might be better to make this package while migrating `platform-ios`.

thymikee:
Yep, let's do it then, not necessary now

PR #601, opened 2019-07-31
fix: fallback to default terminal if wrong one passed

thymikee:
Summary:
---------

It may happen, that the terminal detected by CLI or passed by the user is wrong and not really an executable. A good example is `vscode`, which sets `TERM_PROGRAM` env var to `vscode`, which is neither a command, nor a macOS app name. This PR introduces a fallback behavior for such cases, so that the terminal will be opened by a default `open` program. If that doesn't work, we produce a warning with directions on what to do now:

Screenshot 2019-07-31 at 21 50 31

Test Plan:
----------

Packager should start in new terminal window, when running `run-android` from `vscode` or any other terminal.

### DIFF
--- a/packages/platform-android/src/commands/runAndroid/index.js
+++ b/packages/platform-android/src/commands/runAndroid/index.js
@@ -9,7 +9,6 @@
 
 import path from 'path';
 import execa from 'execa';
-import {spawnSync, spawn, execFileSync} from 'child_process';
 import chalk from 'chalk';
 import fs from 'fs';
 import type {ConfigT} from 'types';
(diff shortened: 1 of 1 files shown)


packages/platform-android/src/commands/runAndroid/index.js
```diff
@@ -400,7 +416,7 @@ export default {
       name: '--terminal [string]',
       description:
         'Launches the Metro Bundler in a new window using the specified terminal path.',
-      default: getDefaultUserTerminal,
+      default: getDefaultUserTerminal(),
```
Esemesek: How was it working previously?

thymikee: it either didn't or used `open`/`sh` to open with a default program

samuelgranja:
Hello,

When I see "Checks" tab it seems to me that built has failed :/
Could you confirm, please?

We are struggling with this issue since we upgraded RN to 0.60.4 and it is not running the packager unless we explicitly specify `--terminal terminal`

thymikee:
@samuelgranja please describe your issue in detail. Ideally create a new one, with a repro we can download and investigate. Otherwise, it's not really actionable.

Jeangel:
Hi @thymikee, why this approach is only implemented for android? With the new version `2.8.1`/`2.8.2` it works like a charm on android but in iOS it still uses `vscode` or `Apple_Terminal`. Could you confirm why only for android?

thymikee:
@Jeangel I guess one of the reasons may be that nobody sent a PR 😉

PR #604, opened 2019-08-01
fix: CLI stuck on "Starting JS server" on Windows

theweavrs:
Summary:
---------
This fixes [RN Issue #25904](https://github.com/facebook/react-native/issues/25904) where CLI gets stuck on `Starting JS Server...` in Windows.

This was due to using the `sync` command of `execa` which doesn't allow multiple commands to run side by side. In the issue, if you close the Metro Bundler, the `run-android` command continues.

Test Plan:
----------
Tested on a new RN Project

cc @thymikee

### DIFF
--- a/packages/platform-android/src/commands/runAndroid/index.ts
+++ b/packages/platform-android/src/commands/runAndroid/index.ts
@@ -356,7 +356,7 @@ function startServerInNewWindow(
       encoding: 'utf8',
       flag: 'w',
     });
-    return execa.sync('cmd.exe', ['/C', launchPackagerScript], {
+    return execa('cmd.exe', ['/C', launchPackagerScript], {
       ...procConfig,
       detached: true,
       stdio: 'ignore',
--- a/packages/platform-android/src/commands/runAndroid/index.ts
+++ b/packages/platform-android/src/commands/runAndroid/index.ts
@@ -356,6 +356,8 @@ function startServerInNewWindow(
       encoding: 'utf8',
       flag: 'w',
     });
+
+    // Awaiting this causes the CLI to hang indefinitely, so this must execute without await.
     return execa('cmd.exe', ['/C', launchPackagerScript], {
       ...procConfig,
       detached: true,


packages/platform-android/src/commands/runAndroid/index.ts
```diff
@@ -356,7 +356,7 @@ function startServerInNewWindow(
       encoding: 'utf8',
       flag: 'w',
     });
-    return execa.sync('cmd.exe', ['/C', launchPackagerScript], {
+    return execa('cmd.exe', ['/C', launchPackagerScript], {
```
thymikee: this is now async, so technically we should make the `startServerInNewWindow` async and await it inside `isPackagerRunning`

theweavrs: Will the other platforms run `execa` command directly?

thymikee: I haven't encounter issues with `execa.sync` on other platforms (at least macOS), so it's likely Windows-only quirk.

theweavrs: I think if `execa.sync` works on other platforms, `execa` directly _should_ work too.

thymikee: Feel free to convert other cases in this function to async, I don't mind :). I avoided doing so, because I didn't want to propagate `async` all the way up, but since it's now necessary, there's no excuse anymore :)

theweavrs: Weird happening. Using async/await brings back this issue again. Can we revert back to `spawn` for Windows only or just use `execa` without async/await?

thymikee: Let's go with `execa` then. But add a comment that it's not awaited on purpose, because the script on Windows doesn't finish or something

theweavrs: Done.

PR #606, opened 2019-08-02
Fix tests on windows

theweavrs:
Summary:
---------
This fixes all the failing tests on Windows.

Test Plan:
----------
Ran `yarn test` on Windows and everything passed.

### DIFF
--- a/packages/cli/src/commands/init/init.js
+++ b/packages/cli/src/commands/init/init.js
@@ -58,7 +58,7 @@ async function setProjectDirectory(directory) {
       throw new DirectoryAlreadyExistsError(directory);
     }
 
-    await fs.emptyDir(directory);
+    fs.emptyDirSync(directory);
   }
 
   try {
--- a/packages/cli/src/commands/init/template.js
+++ b/packages/cli/src/commands/init/template.js
@@ -4,10 +4,7 @@ import path from 'path';
 import * as PackageManager from '../../tools/packageManager';
 import {logger} from '@react-native-community/cli-tools';
 import copyFiles from '../../tools/copyFiles';
-
-const RegexEscape = function(regex) {
-  return regex.replace(/[-\/\\^$*+?.()|[\]{}]/g, '\\$&');
-};
+import {replacePathSepForRegex} from 'jest-regex-util';
 
 export type TemplateConfig = {
   placeholderName: string,
@@ -59,7 +56,7 @@ export async function copyTemplate(
   logger.debug(`Copying template from ${templatePath}`);
   let regexStr = path.resolve(templatePath, 'node_modules');
   await copyFiles(templatePath, process.cwd(), {
-    exclude: [new RegExp(RegexEscape(regexStr))],
+    exclude: [new RegExp(replacePathSepForRegex(regexStr))],
   });
 }
 
--- a/packages/cli/src/commands/init/template.js
+++ b/packages/cli/src/commands/init/template.js
@@ -5,6 +5,10 @@ import * as PackageManager from '../../tools/packageManager';
 import {logger} from '@react-native-community/cli-tools';
 import copyFiles from '../../tools/copyFiles';
 
+const RegexEscape = function(regex) {
+  return regex.replace(/[-\/\\^$*+?.()|[\]{}]/g, '\\$&');
+};
+
 export type TemplateConfig = {
   placeholderName: string,
   templateDir: string,
@@ -53,9 +57,9 @@ export async function copyTemplate(
   );
 
   logger.debug(`Copying template from ${templatePath}`);
-
+  let regexStr = path.resolve(templatePath, 'node_modules');
   await copyFiles(templatePath, process.cwd(), {
-    exclude: [new RegExp(path.resolve(templatePath, 'node_modules'))],
+    exclude: [new RegExp(RegexEscape(regexStr))],
   });
 }
 
--- a/packages/cli/src/commands/init/template.js
+++ b/packages/cli/src/commands/init/template.js
@@ -4,7 +4,16 @@ import path from 'path';
 import * as PackageManager from '../../tools/packageManager';
 import {logger} from '@react-native-community/cli-tools';
 import copyFiles from '../../tools/copyFiles';
-import {replacePathSepForRegex} from 'jest-regex-util';
+
+const replacePathSepForRegex = (string: string) => {
+  if (path.sep === '\\') {
+    return string.replace(
+      /(\/|(.)?\\(?![[\]{}()*+?.^$|\\]))/g,
+      (_match, _, p2) => (p2 && p2 !== '\\' ? p2 + '\\\\' : '\\\\'),
+    );
+  }
+  return string;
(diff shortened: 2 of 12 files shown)


thymikee:
Looks like we need to sort the output in the init e2e test

theweavrs:
The tests that check `templateFiles` fail due to incorrect order of files in the directory. Other fail due to incorrect output.

thymikee:
Last failure left

packages/cli/src/tools/__tests__/copyFiles.test.js
```diff
@@ -47,10 +51,10 @@ test('copies text and binary files from source to destination', async () => {
 
 test('copies files from source to destination excluding directory', async () => {
   const src = path.resolve(__dirname, './__fixtures__');
+  let regexStr = path.join(src, 'extraDir');
   await copyFiles(src, DIR, {
-    exclude: [new RegExp(path.join(src, 'extraDir'))],
+    exclude: [new RegExp(RegexEscape(regexStr))],
```
thymikee: We use it as `exclude: [new RegExp(path.resolve(templatePath, 'node_modules'))]` in the source. Maybe it's not working on Windows? Could you verify on a local template and see if `node_modules` are copied? They shouldn't

theweavrs: Yes they are being copied. Tried on local and remote template.

theweavrs: I can't `init` using a template. Remote or local.

**For local I get this error:**
```
error Error: Cannot find module '\C:\Users\thecodrr\Source\Repos\cli\packages\cli\build\react-native-template-typescript-master-cli-v2\package.json'
error EBUSY: resource busy or locked, rmdir 'C:\Users\thecodrr\Source\Repos\cli\packages\cli\build\TestInit2'
Error: EBUSY: resource busy or locked, rmdir 'C:\Users\thecodrr\Source\Repos\cli\packages\cli\build\TestInit2'
    at Object.rmdirSync (fs.js:684:3)
    at rmdirSync (C:\Users\thecodrr\Source\Repos\cli\node_modules\fs-extra\lib\remove\rimraf.js:276:13)
    at Object.rimrafSync [as removeSync] (C:\Users\thecodrr\Source\Repos\cli\node_modules\fs-extra\lib\remove\rimraf.js:252:7)
    at Object.initialize [as func] (C:\Users\thecodrr\Source\Repos\cli\packages\cli\build\commands\init\init.js:295:26)
    at process._tickCallback (internal/process/next_tick.js:68:7)
    at Function.Module.runMain (internal/modules/cjs/loader.js:832:11)
    at startup (internal/bootstrap/node.js:283:19)
    at bootstrapNodeJSCore (internal/bootstrap/node.js:622:3)
```

For Remote:
```
debug Installing template from typescript
yarn add v1.16.0
info No lockfile found.
[1/4] Resolving packages...
[2/4] Fetching packages...
[3/4] Linking dependencies...
success Saved lockfile.
success Saved 1 new dependency.
info Direct dependencies
└─ typescript@3.5.3
info All dependencies
└─ typescript@3.5.3
Done in 7.86s.
debug Getting config from C:\Users\thecodrr\AppData\Local\Temp\rncli-init-template-jV11JR\node_modules\typescript\template.config.js
error Error: Cannot find module 'C:\Users\thecodrr\AppData\Local\Temp\rncli-init-template-jV11JR\node_modules\typescript\template.config'
error EBUSY: resource busy or locked, rmdir 'C:\Users\thecodrr\Source\Repos\cli\packages\cli\build\TestInit2'
Error: EBUSY: resource busy or locked, rmdir 'C:\Users\thecodrr\Source\Repos\cli\packages\cli\build\TestInit2'
    at Object.rmdirSync (fs.js:684:3)
    at rmdirSync (C:\Users\thecodrr\Source\Repos\cli\node_modules\fs-extra\lib\remove\rimraf.js:276:13)
    at Object.rimrafSync [as removeSync] (C:\Users\thecodrr\Source\Repos\cli\node_modules\fs-extra\lib\remove\rimraf.js:252:7)
    at Object.initialize [as func] (C:\Users\thecodrr\Source\Repos\cli\packages\cli\build\commands\init\init.js:295:26)
    at process._tickCallback (internal/process/next_tick.js:68:7)
```
In remote, the temp directory is deleted before it can copy the template (don't know why). And in local, the path is always wrong.

theweavrs: No they are not being deleted

theweavrs: Tests should be fixed now.

thymikee: I'm not sure if I follow. The point of this is to exclude certain dirs, like `node_modules`. If you have a local template with `node_modules` inside, they _shouldn't_ be copied to the new project. From the code you added, it seems that this escaping needs to be done in the source, not in the test only, because this is exactly how we use it. Can you move the escape fn there?

theweavrs: I did create an escape function in the source; check the latest commits.

Sure, I will rename it.

theweavrs: There is no need for copying as `jest-regex-util` is already present, we can just use the function from there.

packages/cli/src/commands/init/template.js
```diff
@@ -4,10 +4,7 @@ import path from 'path';
 import * as PackageManager from '../../tools/packageManager';
 import {logger} from '@react-native-community/cli-tools';
 import copyFiles from '../../tools/copyFiles';
-
-const RegexEscape = function(regex) {
-  return regex.replace(/[-\/\\^$*+?.()|[\]{}]/g, '\\$&');
-};
+import {replacePathSepForRegex} from 'jest-regex-util';
```
thymikee: It's in our development deps. Jest libs will usually be there but there's not guarantee. If we import a package, we need to declare it as our dependency. So, either copy the oneliner from jest-regex-util or add it to the deps of packages/cli/package.json. I'd prefer to copy to avoid users having multiple versions of jest packages :)

theweavrs: Oh okay, got it. I will make a copy.

packages/cli/src/commands/init/init.js
```diff
@@ -58,7 +58,7 @@ async function setProjectDirectory(directory) {
       throw new DirectoryAlreadyExistsError(directory);
     }
 
-    await fs.emptyDir(directory);
+    fs.emptyDirSync(directory);
```
thymikee: I wonder, why this change?

theweavrs: Using await causes hang on Windows, that's why.

thymikee: Geez, Windows is so weird

theweavrs: It indeed is. 😂😂

PR #607, opened 2019-08-02
feat: Add prefers-color-scheme dark mode

PaitoAnderson:
Summary:
---------

Automatically set the appearance based on the global system dark mode, and remove the manual Dark Theme toggle.

Test Plan:
----------

Set macOS Appearance to Light and Dark in System Preferences -> General and watch the debugging UI background change from white to black. Tested myself in Chrome 76 and Safari 12.1.2 on macOS 10.14.6.

Screen Shot 2019-08-02 at 10 22 32 AM

### DIFF
--- a/packages/cli/src/commands/server/debugger-ui/index.html
+++ b/packages/cli/src/commands/server/debugger-ui/index.html
@@ -35,7 +35,6 @@
   const Page = window.Page = {
 
     state: {
-      isDark: localStorage.getItem('darkTheme') === 'on',
       isPriorityMaintained: localStorage.getItem('maintainPriority') === 'on',
       status: {type: 'disconnected'},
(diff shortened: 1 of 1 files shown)


thymikee:
Um, but this removes the dark mode for folks that are on light theme and still want to use dark in the debugger. Not mentioning Windows or various Linux distros that may not have this functionality. Can you bring the button down and set the default base on that CSS query?

PaitoAnderson:
@thymikee Fair enough, I've changed it to just default to the user's system color scheme the first time it's opened.

I did test Windows 10 as it does have a dark mode, works there too using Chrome 76. I haven't used Linux in years so I don't know if the popular distros are shipping a dark mode or not.

Thanks for reviewing!

PR #609, opened 2019-08-03
feat: better error messages with human readable outputs

thecodrr:
Summary:
---------
_This is a work in progress and might take more than a little time._

These checks should add a little more context (and actionable output) when the script fails and reduce the number of false issues.  All these handle _potential points of failure_ e.g. invalid arguments, typos in provided file/dir paths or directories that no longer exist.

Adding a list here of all the commands that need to checked.

### CLI
- [x] init [done]
- [x] bundle - [done]
- [x] config [no need]
- [x] info [no need]
- [x] install [done]
- [x] link [done]
- [x] upgrade [done]

### Android
- [x] log-android [no need]
- [x] run-android [done but not tested]
- [x] link [no need]
- [x] config [no need]

Test Plan:
----------
Only way to test is by failing all the checks one by one. 😆 Maybe more jest tests can be added to assure these comply?


packages/cli/src/commands/init/template.js
```diff
@@ -4,6 +4,7 @@ import path from 'path';
 import * as PackageManager from '../../tools/packageManager';
 import {logger} from '@react-native-community/cli-tools';
 import copyFiles from '../../tools/copyFiles';
+import fs from 'fs-extra';
```
thymikee:
```suggestion
import fs from 'fs';
```

packages/cli/src/commands/init/template.js
```diff
@@ -36,7 +37,11 @@ export function getTemplateConfig(
   );
 
   logger.debug(`Getting config from ${configFilePath}.js`);
-
+  if (!fs.existsSync(configFilePath)) {
+    throw new Error(
+      `Invalid template '${templateName}' specified. Make sure the template is CLI v2 compliant i.e. has a 'template.config.js' file.`,
```
thymikee: Add a link to the init docs

packages/cli/src/commands/bundle/saveAssets.js
```diff
@@ -22,6 +22,12 @@ function saveAssets(assets, platform, assetsDest) {
     return Promise.resolve();
   }
 
+  if (!fs.existsSync(assetsDest)) {
+    let error = `The specified assets destination folder: "${assetsDest}" does not exist, stopping...`;
```
thymikee:
```suggestion
    let error = `The specified assets destination folder "${assetsDest}" does not exist.`;
```

packages/cli/src/commands/bundle/saveAssets.js
```diff
@@ -22,6 +22,12 @@ function saveAssets(assets, platform, assetsDest) {
...
+    logger.error(error);
+    return Promise.reject();
```
thymikee: can we move the error to `reject`?

packages/cli/src/commands/init/templateName.js
```diff
@@ -1,6 +1,7 @@
 // @flow
 import path from 'path';
 import {URL} from 'url';
+import fs from 'fs-extra';
```
thymikee:
```suggestion
import fs from 'fs';
```

packages/cli/src/commands/init/templateName.js
```diff
@@ -9,14 +10,37 @@ const VERSIONED_PACKAGE = /(@?.+)(@)(.+)/;
 
 function handleFileProtocol(filePath: string) {
   const uri = new URL(filePath).pathname;
-
+  if (!fs.existsSync(uri)) {
+    throw new Error(
+      `Failed to retrieve package name. The specified template directory path: ${uri} does not exist or is invalid.`,
```
thymikee:
```suggestion
      `Failed to retrieve package name. The specified template directory path "${uri}" does not exist or is invalid.`,
```

packages/cli/src/commands/init/templateName.js
```diff
@@ -9,14 +10,37 @@ const VERSIONED_PACKAGE = /(@?.+)(@)(.+)/;
...
+    );
+  }
+  const packageJsonPath = path.join(uri, 'package.json');
+  if (!fs.existsSync(packageJsonPath)) {
+    throw new Error(
+      'Failed to retrieve package name. We expect the template directory to include package.json file, but no such file was found.',
```
thymikee:
```suggestion
      'Failed to retrieve package name. We expect the template directory to include "package.json" file, but it was not found.',
```

packages/cli/src/commands/init/templateName.js
```diff
@@ -9,14 +10,37 @@ const VERSIONED_PACKAGE = /(@?.+)(@)(.+)/;
...
+    );
+  }
+  const packageJson = require(packageJsonPath);
+  if (!packageJson.name) {
+    throw new Error(
+      `Failed to retrieve package name. We expect the package.json to include package name, e.g.: "template-name", but received: "${JSON.stringify(
```
thymikee:
```suggestion
      `Failed to retrieve template name. We expect the "package.json" of the template to include the "name" property, but we found "${packageJson.name}" which is invalid.`
```

packages/cli/src/commands/init/templateName.js
```diff
@@ -9,14 +10,37 @@ const VERSIONED_PACKAGE = /(@?.+)(@)(.+)/;
...
     uri,
-    name: require(path.join(uri, 'package.json')).name,
+    name: packageJson.name,
   };
 }
 
 function handleTarball(filePath: string) {
+  if (!fs.existsSync(filePath)) {
+    throw new Error(
+      `Failed to retrieve tarball name. The specified tarball path: ${filePath} does not exist or is invalid.`,
```
thymikee:
```suggestion
      `Failed to retrieve tarball name. The specified tarball path "${filePath}" does not exist or is invalid.`,
```

packages/cli/src/commands/init/templateName.js
```diff
@@ -9,14 +10,37 @@ const VERSIONED_PACKAGE = /(@?.+)(@)(.+)/;
...
   return {
     uri,
-    name: require(path.join(uri, 'package.json')).name,
+    name: packageJson.name,
   };
 }
 
 function handleTarball(filePath: string) {
+  if (!fs.existsSync(filePath)) {
+    throw new Error(
```
thymikee:
```suggestion
    throw new CLIError(
```

packages/platform-android/src/commands/runAndroid/index.ts
```diff
@@ -247,7 +274,7 @@ function getInstallApkName(
     return apkName;
   }
 
-  throw new Error('Not found the correct install APK file!');
+  throw new Error('Could not find the correct install APK file!');
```
thymikee:
```suggestion
  throw new CLIError('Could not find the correct install APK file.');
```

theweavrs:
@thymikee Thank you!

A question: I have not been able to test the `platform-android` code. Is there a simple way I can run it and check the error outputs?

thymikee:
You need to `yarn link @react-native-community/cli-platform-android` in the test project

theweavrs:
@thymikee I am done with this, please review.

packages/cli/src/commands/bundle/assetPathUtils.js
```diff
@@ -33,7 +33,7 @@ function getAndroidAssetSuffix(scale: number): string {
     case 4:
       return 'xxxhdpi';
     default:
-      throw new Error('no such scale');
+      return '';
```
grabbou: Do we handle that error somewhere else? Or in other words, what is the expected action here?

packages/cli/src/commands/bundle/saveAssets.js
```diff
@@ -14,14 +14,19 @@ import fs from 'fs';
...
 
 function saveAssets(assets, platform, assetsDest) {
   if (!assetsDest) {
     logger.warn('Assets destination folder is not set, skipping...');
     return Promise.resolve();
   }
 
+  if (!fs.existsSync(assetsDest)) {
+    let error = `The specified assets destination folder "${assetsDest}" does not exist.`;
+    throw new CLIError(error);
```
grabbou: Is `let error` needed here on line 26?

packages/cli/src/commands/bundle/assetPathUtils.js
thecodrr: Yes, the error is being handled on [#L57](https://github.com/react-native-community/cli/blob/d82a457509e1ca8fab08a15afec130e6773d7417/packages/cli/src/commands/bundle/assetPathUtils.js#L57). Before making this change, that was error was never being hit because we were throwing here. I couldn't use undefined/null because `flow` gave errors then so I had to fallback to empty string.

packages/cli/src/commands/bundle/saveAssets.js
thecodrr: No, I don't think so. I will remove this.

thecodrr:
@grabbou thanks. Yes this is done and ready to be reviewed. I may have left some errors out (intentionally) because they don't make sense or slow down the performance.

grabbou:
Uhm, looks like we have moved on with Typescript and other PRs in the meantime (unintentionally), making conflicts with this PR.

I do really like it and was wondering if you'd be open to rebasing and preparing it for merging once again?

thecodrr:
@grabbou Okay I will begin the work ASAP.

thymikee:
@thecodrr I took a stab at it and rebased your work. Please review if I didn't mess anything up :)

thecodrr:
@thymikee Everything looks perfect. : )

thecodrr:
@thymikee ping

packages/cli/src/commands/bundle/saveAssets.ts
```diff
@@ -30,6 +30,12 @@ function saveAssets(
     return Promise.resolve();
   }
 
+  if (!fs.existsSync(assetsDest)) {
```
adrianha: hi @thecodrr @thymikee @grabbou,
I have a situation that I need to run bundle and copy assets into new created folder based on `assetsDest` and I was relying on https://github.com/react-native-community/cli/pull/609/files#diff-e59437386ac414095e6ed43a3da37303R93 to create folder if not exists.
Any specific reason that it needs to check `assetsDest` first before copying assets?

thecodrr: Since we are calling `mkdirp` in the copy function, maybe this is no longer needed?

adrianha: yes, i think it should be safe using `mkdirp` before copy files

thecodrr: this check can be safely removed then. What do you think @thymikee?

thymikee: Please do

adrianha: Thanks @thecodrr @thymikee will open the PR

PR #611, opened 2019-08-04
Add reload button in debugger UI

theweavrs:
Summary:
---------

This acts as a useful extension of `reload` endpoint implemented in #574.

**Side note:** Maybe we can rename the button to something more descriptive than 'Reload' as that can be a little confusing...? Suggestions welcome

Test Plan:
----------
I tested the endpoint directly in Chrome DevTools and by deploying a sample app using `run-android` and using the button to reload. It works.

cc @thymikee


packages/cli/src/commands/server/debugger-ui/index.html
```diff
@@ -253,6 +260,9 @@
   .content {
     padding: 10px;
   }
+  .reload-btn {
+    padding: 5px 10px 5px 10px;
```
thymikee:
```suggestion
    padding: 5px 10px;
```

packages/cli/src/commands/server/debugger-ui/index.html
```diff
@@ -282,6 +292,9 @@
     <p>Press <kbd id="shortcut" class="shortcut">⌘⌥I</kbd> to open Developer Tools. Enable <a href="https://stackoverflow.com/a/17324511/232122" target="_blank">Pause On Caught Exceptions</a> for a better debugging experience.</p>
     <p>You may also install <a href="https://github.com/facebook/react-devtools/tree/master/packages/react-devtools" target="_blank">the standalone version of React Developer Tools</a> to inspect the React component hierarchy, their props, and state.</p>
     <p>Status: <span id="status">Loading...</span></p>
+    <button class="reload-btn" onclick="onReloadClicked()">
+      Reload
```
thymikee: maybe
```suggestion
      Reload JS
```
what do you think @Esemesek?

theweavrs: Or "Reload JS Bundle"? Or perhaps "Reload App"?

thymikee: I think I'm fine with "Reload app"

PR #612, opened 2019-08-04
feat: fallback to default browser if Chrome is not found

thecodrr:
Summary:
---------

It can be annoying for users (like me) who do not have Google Chrome installed to enable 'Remote Debugging' only to receive an error. This rectifies that by falling back to the default installed browser making the whole experience more fluid.

Test Plan:
----------
Tested on a sample app by enabling 'Remote Debugging'.

### DIFF
--- a/packages/cli/src/commands/server/launchDebugger.js
+++ b/packages/cli/src/commands/server/launchDebugger.js
@@ -28,7 +28,7 @@ function commandExistsUnixSync(commandName) {
 
 function commandExistsWindowsSync(commandName) {
   try {
-    var stdout = execSync('where ' + commandName, {stdio: []});
+    const stdout = execSync('where ' + commandName, {stdio: []});
     return !!stdout;
   } catch (error) {
     return false;
@@ -87,7 +87,7 @@ function launchDebugger(url: string) {
     launchBrowser(url);
     return;
   }
-  launchChrome();
+  launchChrome(url);
 }
 
 export default launchDebugger;
--- a/packages/cli/src/commands/server/launchDebugger.js
+++ b/packages/cli/src/commands/server/launchDebugger.js
@@ -11,7 +11,7 @@
 import open from 'open';
 import {execSync} from 'child_process';
 import {logger} from '@react-native-community/cli-tools';
-import launchBrowser from './launchBrowser';
+import launchDefaultBrowser from './launchDefaultBrowser';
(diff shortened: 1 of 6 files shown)


packages/cli/src/commands/server/launchDebugger.js
```diff
@@ -24,6 +26,28 @@ function commandExistsUnixSync(commandName) {
   }
 }
 
+function commandExistsWindowsSync(commandName) {
+  try {
+    var stdout = execSync('where ' + commandName, {stdio: []});
```
thymikee:
```suggestion
    const stdout = execSync('where ' + commandName, {stdio: []});
```

theweavrs: Fixed.

packages/cli/src/commands/server/middleware/getDevToolsMiddleware.js
```diff
@@ -6,29 +6,29 @@
  *
  * @format
  */
-import launchChrome from '../launchChrome';
 import {logger} from '@react-native-community/cli-tools';
 import {exec} from 'child_process';
+import launchDebugger from './../launchDebugger';
```
thymikee:
```suggestion
import launchDebugger from '../launchDebugger';
```

theweavrs: I don't see why it shouldn't but nothing wrong in testing. I use Brave and it works without an issue but that uses Chrome DevTools. I will test this on Firefox etc. These are essentially websockets and http requests so in theory it should work

packages/cli/src/commands/server/launchDebugger.js
```diff
@@ -52,4 +76,18 @@ function launchChrome(url: string) {
...
 
-export default launchChrome;
+function launchDebugger(url: string) {
+  if (!commandExists(getChromeAppName())) {
+    logger.info(
+      `For a better debugging experience please install Google Chrome from: ${chalk.underline.dim(
+        'https://www.google.com/chrome/',
+      )}`,
+    );
+    //fallback to default browser
```
grabbou: nit: space here before the inline command. I also think we can rid of that comment and rename the function to something like: `launchDefaultBrowser`

theweavrs: Done.

PR #613, opened 2019-08-04
feat: watch mode in react-native start

thecodrr:
Summary:
---------

Watch mode is basically a key listener for various commands. It will enable developers to execute various commands just by tapping a key in the `packager` window.

**Commands & Keys:**
1. Reload app - r
2. Open Dev Menu - d

**Questions:**
1. Should the commands be remappable?

cc @thymikee 

Test Plan:
----------
Tested on Windows on a sample RN 0.60.4 app. Should be tested on Linux & macOS (to check if ctrl is caught or not).

**Instructions for testing:**
1. Load the app using `react-native run-*`
2. Make sure Metro Bundler is running. Use `npm start` or `react-native start` if it's not.
3. Focus Metro Bundler process window.
4. Press 'r' to reload & 'd' to open dev menu.

Current Looks:
----------
```cmd
┌──────────────────────────────────────────────────────────────────────────────┐
│                                                                              │
│  Running Metro Bundler on port 8081.                                         │
│                                                                              │
│  Keep Metro running while developing on any JS projects. Feel free to        │
│  close this tab and run your own Metro instance if you prefer.               │
│                                                                              │
│  https://github.com/facebook/react-native                                    │
│                                                                              │
└──────────────────────────────────────────────────────────────────────────────┘

Looking for JS files in
   C:\Users\thecodrr\Source\Repos\Atomic

Loading dependency graph, done.

To reload the app press "r"
To open developer menu press "d"

DELTA  [android, dev] ./index.js ▓▓▓▓▓▓▓▓▓▓▓▓▓▓ 100% (276/276), success.

info Developer menu opened.
info Reloading app...
```


thymikee:
Hey, thanks for the PR! @lucasbento is working on a similar functionality, something like a watch mode for `start`. Can you consult this with him?

theweavrs:
Sure though I have no idea how to contact him :D

theweavrs:
@thymikee so i contacted @lucasbento and he hasn't started working on the watch mode. I do like the idea of a watch mode and I think I can implement it with a little help. Implementing for Android would be relatively easier than iOS (no way to test it). 

We already have reload, we need hot reloading toggle, dev menu toggle, debugging mode toggle. Maybe we can add these commands similar to the reload command implementation, via an http hook?

We can do Android without a hassle using direct adb commands...but iOS would be a problem then. 

Would love thoughts on this.

lucasbento:
> Maybe we can add these commands similar to the reload command implementation, via an http hook?

Yeah, that's definitely the idea.
I'm not sure about adding other dev menu toggles or other stuff like that (in terms of being useful to the users) but we can definitely discuss it further.

thymikee:
When having 2 platforms opened, it's common to only want to debug on a single device, as it won't work with 2 at once. We'd need to detect multiple devices connected and hide the hint in such case

theweavrs:
> When having 2 platforms opened, it's common to only want to debug on a single device, as it won't work with 2 at once. We'd need to detect multiple devices connected and hide the hint in such case

Isn't multiple device detection already there when running `run-*`? Maybe we can use that...?

theweavrs:
> > Maybe we can add these commands similar to the reload command implementation, via an http hook?
> 
> Yeah, that's definitely the idea.
> I'm not sure about adding other dev menu toggles or other stuff like that (in terms of being useful to the users) but we can definitely discuss it further.

Currently, shaking to open dev menu is a solid pain in the ***. Having a key binding would make it more seemless.

thymikee:
I'd prefer having the menu displayed by the `start` command, since you could start your iOS/Android project from the IDE, not the `run-*`.

theweavrs:
> > Maybe we can add these commands similar to the reload command implementation, via an http hook?

> Yeah, that's definitely the idea.

If we end up using http hooks, we can also add the respective buttons for commands in the debugger-ui.

> I'd prefer having the menu displayed by the start command, since you could start your iOS/Android project from the IDE, not the run-*.

Yes, the command menu would appear when `start` is used as it makes more sense since the commands will operate from there. We would also need to output a reaction to all the related commands when a key is pressed in the packager window.

theweavrs:
@thymikee @lucasbento 

I am done implementing Reload & Open Dev Menu commands, they were pretty simple. Is there already a way to "Enable Remote Debugging" using websockets? I am looking through React Native code but can't find a proper way, maybe you guys know something? I have a feeling we might need to add it in React Native.

lucasbento:
> Edit: on second thought, I don't think we need to have a way of Enabling Remote Debugging since we already have a way to open dev menu. Better to keep it simple and usable.

Agreed.

thecodrr:
This is open for review and ready to be merged. @thymikee

packages/cli/src/commands/server/watchMode.js
```diff
@@ -0,0 +1,50 @@
...
+  // right after its done loading.
+  hookStdout(
+    output =>
+      output.includes('Loading dependency graph, done.') &&
+      printWatchModeInstructions(),
+  );
+
+  stdin.on('keypress', (key, data) => {
+    const {ctrl, name} = data;
+    if (ctrl === true && name === 'c') {
```
Esemesek: What about other shortcuts like `ctrl + d` / `ctrl + z`?

packages/cli/src/commands/server/watchMode.js
```diff
@@ -0,0 +1,50 @@
...
+      process.exit();
+    } else if (name === 'r') {
```
Esemesek: Switch case might be more appropriate if you remove the `ctrl + c` handling.

packages/cli/src/commands/server/server.js
```diff
@@ -108,5 +108,9 @@ export default {
       description: 'Path to the CLI configuration file',
       parse: (val: string) => path.resolve(val),
     },
+    {
+      name: '--disableWatchMode',
```
Esemesek: Not sure about the name here. I think we should name it `interactiveMode` and rename this flag to `--disableInteractiveMode`. WDYT @thymikee ?

packages/cli/src/commands/server/watchMode.js
```diff
@@ -0,0 +1,50 @@
+/**
```
Esemesek: Since we are migrating the whole codebase to Typescript, can we also have this PR written in it?

packages/cli/src/commands/server/server.js
thymikee: I'd prefer to have `--interactive` flag and use a built-in feature of commanderjs that let us negate the fag with `no-` prefix, so `--no-interactive` would disable it then.

packages/cli/src/commands/server/watchMode.js
thecodrr: Overriding `stdin` makes all the usual commands not work any more that's why I am defining my own `ctrl + c` for exiting. I will take a look at those shortcuts you mention and implement them.

packages/cli/src/commands/server/server.js
thecodrr: Will do.

packages/cli/src/commands/server/watchMode.js
thecodrr: I will start the work on this.

lucasbento:
Bump @thecodrr, how's this going?

thecodrr:
@lucasbento I have everything done. Just have to push. Will do it today.

thecodrr:
@thymikee this is ready for review.

docs/commands.md
```diff
@@ -505,6 +505,10 @@ Path to custom SSL cert
 
 Path to the CLI configuration file
 
+#### `--disableWatchMode [string]`
```
grabbou: I believe this is a bit outdated, right?

packages/cli/src/commands/server/server.ts
```diff
@@ -106,5 +106,9 @@ export default {
       description: 'Path to the CLI configuration file',
       parse: (val: string) => path.resolve(val),
     },
+    {
+      name: '--interactive',
+      description: 'Disables interactive mode',
```
grabbou: I think something is wrong with the name here. How am I supposed to disable it? `--interactive false`? We can either rephrase it so that its presence negates (e.g. `--no-interactive`) or leave it as is and add `default: true` as well as `description: Enables interactive mode, true by default`.

packages/cli/src/commands/server/watchMode.ts
```diff
@@ -0,0 +1,56 @@
...
+        case 'z':
+          process.emit('SIGTSTP');
+          break;
+      }
+    } else if (name === 'r') {
+      messageSocket.broadcast('reload', null);
+      logger.info('Reloading app...');
+    } else if (name === 'd') {
+      messageSocket.broadcast('devMenu', null);
+      logger.info('Developer menu opened.');
```
grabbou: "Opening developer menu" will be a bit better and more aligned with the rest we have

grabbou:
Thank you! This is great! I think it's ready to go from the coding perspective! My only questions are around the flag - I feel like the documentation / description of what the flag is supposed to do are a bit ouf of sync.

Would you mind looking at that, addressing it and we ship it?

This is going to be a great PR!

packages/cli/src/commands/server/server.ts
thecodrr: This was @thymikee's idea:
> I'd prefer to have --interactive flag and use a built-in feature of commanderjs that let us negate the fag with no- prefix, so --no-interactive would disable it then.

thecodrr:
@grabbou done, please review.

grabbou:
Thanks! I am going to do some internal testing right now and ship it today. Then, I will release another alpha (just in case there's a bug, don't want to break people on Friday) and we will promote it to stable around next Wednesday!

Happy to see this being merged!

thecodrr:
@grabbou okay awesome! Can't wait.

grabbou:
I have performed some of the changes myself, because the flags were a bit messed up. It seems to be working properly.

Unfortuantely, the interactive mode doesn't start. It produces the following error:
> TypeError [ERR_INVALID_ARG_TYPE]: The "chunk" argument must be one of type string or Buffer. Received type function
    at validChunk (_stream_writable.js:263:10)
    at WriteStream.Writable.write (_stream_writable.js:297:21)
    at WriteStream.process.stdout.write.output [as write] (/Users/grabbou/Repositories/react-native-cli/packages/cli/build/tools/hookStdout.js:12:11)

That comes from the `hookStdout`.

grabbou:
Okay, I have resolved this via 4e8c5ea and added a cleanup utility via c780f5f. I am performing tests right now.

grabbou:
Metro 0.58 removed "Loading dependency graph" message. It is now the CLI banner. I am attaching a working picture as a proof.

thymikee:
How it behaves when you have some more output (logs, bundle loading info) from Mtro?

grabbou:
> How it behaves when you have some more output (logs, bundle loading info) from Mtro?

@thymikee our watch mode instructions stay on top, right below the logo (as demonstrated by the OP in the first message). They will not be re-printed. Rather than that, I am not sure how the amount of logs would affect the UX.

lucasbento:
Jumping on the conversation here:

I think it's fine to merge like this and work on a better UI later on, developing something later on top of the UI of `doctor` would fit this really well.

grabbou:
Sounds like a plan to me @lucasbento! Thanks for chiming in! I would like to better understand the UI controls used in doctor and see how we can improve the overall experience.

satya164:
I think printing the port number was useful. Is there any other way to know which port it is running on with the new format? Especially with Expo, different instances can use different port numbers.

grabbou:
We could hack around the configuration and see if the port is there, but my guess is that it's inferred within a Metro server. I would recommend rising an issue there if there's anything that you feel could be improved.

thecodrr:
We could print the instructions after every output but that would be a bit too much. Printing them once at the top is the best I could think of as they won't be needed after the first few times. Besides, the watch mode is too simple currently (only 2 commands) so, no need to print the instructions again and again.

grabbou:
I am going to merge it. We can tweak it later.

lucasbento:
Awesome work @thecodrr 🙌

thecodrr:
@lucasbento thanks a lot! : )

PR #618, opened 2019-08-05
fix: use forward slashes instead of path.sep in iOS link

### LINKED ISSUE #615: react-native link on Windows uses incorrect path separator for xcode
## Environment

info
    System:
      OS: Windows 10
      CPU: (4) x64 Intel(R) Core(TM) i7-7500U CPU @ 2.70GHz
      Memory: 5.49 GB / 15.89 GB
    Binaries:
      Yarn: 1.17.3 - C:\Users\gaspa\AppData\Roaming\npm\yarn.CMD
      npm: 6.4.1 - C:\Program Files\nodejs\npm.CMD
    IDEs:
      Android Studio: Version  3.3.0.0 AI-182.5107.16.33.5314842

## Description

Running react-native link on windows will use backslashes in the xcode project file project.pbxcode in the HEADER_SEARCH_PATHS.

Resulting in strings like these:
"$(SRCROOT)..\node_modules\neact-native-fs/**",

These will not be parsed correctly by xcode.

The issue is in /packages/platform-ios/src/link/getHeaderSearchPath.js. It should not use the platform path.sep, instead it should always uses forward slashes, regardless of the current platform since xcode only runs on MacOS.

## Reproducible Demo

create new project
add library that needs linking, such as react-native-fs
run react-native link react-native-fs on a windows machine
observe result in project.pbxcode file under HEADER_SEARCH_PATHS.

Taym95:
Fix #615

### DIFF
--- a/packages/platform-ios/src/link/getHeaderSearchPath.js
+++ b/packages/platform-ios/src/link/getHeaderSearchPath.js
@@ -22,14 +22,14 @@ import {last, union} from 'lodash';
  */
 const getOuterDirectory = directories =>
   directories.reduce((topDir, currentDir) => {
-    const currentFolders = currentDir.split(path.sep);
-    const topMostFolders = topDir.split(path.sep);
+    const currentFolders = currentDir.split('/');
+    const topMostFolders = topDir.split('/');
 
     if (
       currentFolders.length === topMostFolders.length &&
       last(currentFolders) !== last(topMostFolders)
     ) {
-      return currentFolders.slice(0, -1).join(path.sep);
+      return currentFolders.slice(0, -1).join('/');
     }
 
     return currentFolders.length < topMostFolders.length ? currentDir : topDir;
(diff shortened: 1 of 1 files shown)


thymikee:
@qwertzguy can you verify this works well for you?

packages/platform-ios/src/link/getHeaderSearchPath.js
```diff
@@ -53,8 +53,8 @@ export default function getHeaderSearchPath(sourceDir, headers) {
   const directories = union(headers.map(path.dirname));
 
   return directories.length === 1
-    ? `"$(SRCROOT)${path.sep}${path.relative(sourceDir, directories[0])}"`
-    : `"$(SRCROOT)${path.sep}${path.relative(
+    ? `"$(SRCROOT)/${path.relative(sourceDir, directories[0])}"`
```
qwertzguy: Won't `path.relative` use the OS specific separator?

qwertzguy: I think you can use `path.posix` to switch to the posix implementation regardless of the OS.

thymikee: @qwertzguy https://github.com/react-native-community/cli/pull/635

PR #620, opened 2019-08-05
imp: Migrate `platform-ios` to TypeScript

Esemesek:
Summary:
---------

This PR includes moving the whole `platform-ios` module to Typescript. There were a lot of issues while moving this module because most of it was untyped and the main library that we rely on (`xcode`) is untyped. Also, there is a lot of explicit casting to `any` type, empty stub definitions for `xcode` module and plenty of `@ts-ignore`.

Changes:
- Rename `.js` to `.ts`
- Fixed all the Typescript issues in `platform-ios`
- New module for all the common types `cli-types`

TODO:
~~- [ ] - Fix as many types issues as possible~~
~~- [ ] - Improve typings for `xcode`~~

^ Since those will take much more effort and time to make, I think we can make issues for them after we merge this PR.

Test Plan:
----------

👍- All the check are passing.


packages/platform-ios/src/commands/runIOS/index.ts
```diff
@@ -21,21 +23,22 @@ import {
...
-  projectPath: string,
-  device: ?(string | true),
-  udid: ?string,
-  packager: boolean,
-  verbose: boolean,
-  port: number,
-  terminal: ?string,
+  simulator: string;
+  configuration: string;
+  scheme: string | void;
```
tido64: Wouldn't this be more accurate here (and below)?
```suggestion
  scheme?: string;
```

packages/platform-ios/src/commands/runIOS/index.ts
```diff
@@ -277,7 +294,7 @@ function buildProject(xcodeProject, udid, scheme, args: FlagsT) {
     );
     let buildOutput = '';
     let errorOutput = '';
-    buildProcess.stdout.on('data', data => {
+    buildProcess.stdout.on('data', (data: any) => {
```
tido64: I think this should be `node.Buffer`.

packages/platform-ios/src/commands/runIOS/index.ts
```diff
@@ -243,7 +255,12 @@ async function runOnDevice(selectedDevice, scheme, xcodeProject, args: FlagsT) {
   return logger.success('Installed the app on the device.');
 }
 
-function buildProject(xcodeProject, udid, scheme, args: FlagsT) {
+function buildProject(
+  xcodeProject: ProjectInfo,
+  udid: string | void,
```
tido64:
```suggestion
  udid: string | undefined,
```

packages/platform-ios/src/commands/runIOS/index.ts
```diff
@@ -372,7 +394,11 @@ function xcprettyAvailable() {
   return true;
 }
 
-function matchingDevice(devices, deviceName, udid) {
+function matchingDevice(
+  devices: Array<Device>,
+  deviceName: string | true | void,
```
tido64:
```suggestion
  deviceName: string | true | undefined,
```

packages/platform-ios/src/commands/runIOS/index.ts
```diff
@@ -372,7 +394,11 @@ function xcprettyAvailable() {
...
+  udid: string | void,
```
tido64:
```suggestion
  udid: string | undefined,
```

packages/platform-ios/src/commands/runIOS/index.ts
```diff
@@ -390,30 +416,46 @@ function matchingDevice(devices, deviceName, udid) {
   );
 }
 
-function matchingDeviceByUdid(devices, udid) {
+function matchingDeviceByUdid(devices: Array<Device>, udid: string | void) {
```
tido64:
```suggestion
function matchingDeviceByUdid(devices: Array<Device>, udid: string | undefined) {
```

packages/platform-ios/src/link-pods/__tests__/isInstalled.test.ts
```diff
@@ -18,18 +16,21 @@ describe('pods::isInstalled', () => {
   it('returns false if pod is missing', () => {
     const project = {podfile: path.join(PODFILES_PATH, 'PodfileSimple')};
     const podspecName = {podspecPath: '/path/NotExisting'};
+    // @ts-ignore FIXME: Improve types
```
tido64: What was the issue with these?

packages/platform-ios/src/link-pods/addPodEntry.ts
```diff
@@ -13,8 +12,10 @@ import getPodspecName from '../config/getPodspecName';
 export default function addPodEntry(
   podLines: Array<string>,
   linesToAddEntry?:
-    | Array<{line: number, indentation: number}>
-    | {line: number, indentation: number},
+    | Array<{line: number; indentation: number}>
+    | {line: number; indentation: number}
+    | null,
+  // @ts-ignore Required argument cannot follow optional argument
```
tido64: Instead of using `?`, you can just append `| undefined`.

packages/platform-ios/src/link/common/isInstalled.ts
```diff
@@ -4,17 +4,21 @@
...
+  IOSDependencyConfig,
+} from '@react-native-community/cli-types';
 
 export default function isInstalled(
-  projectConfig: ProjectConfigIOST,
-  name?: string,
-  dependencyConfig: DependencyConfigIOST,
+  projectConfig: IOSProjectConfig,
+  // FIXME: name is never used
+  _name?: string,
```
tido64:
```suggestion
  _name: string | undefined,
```

packages/platform-ios/src/link/getTargets.ts
```diff
@@ -4,20 +4,19 @@
...
-export default function getTargets(project) {
+export default function getTargets(project: any) {
   const {
     firstProject: {targets},
   } = project.getFirstProject();
   const nativeTargetSection = project.pbxNativeTargetSection();
   return targets
-    .filter(target => nativeTargetSection[target.value] !== undefined)
-    .map(target => {
+    .filter((target: any) => nativeTargetSection[target.value] !== undefined)
```
tido64: Would something like this work?
```suggestion
    .filter((target: { value: string; }) => nativeTargetSection[target.value] !== undefined)
```

packages/platform-ios/src/link/removeFromStaticLibraries.ts
```diff
@@ -15,7 +14,11 @@ import removeFromPbxReferenceProxySection from './removeFromPbxReferenceProxySec
  *
  * Similar to `node-xcode` addStaticLibrary
  */
-export default function removeFromStaticLibraries(project, path, opts) {
+export default function removeFromStaticLibraries(
+  project: any,
+  path: string,
+  opts: {[key: string]: any},
```
tido64: Will you do a follow up PR to fix all instance of `any`? There's an awful lot of them…

Esemesek: This is the plan. We have so many instances of `any` because library that we rely on - `https://www.npmjs.com/package/xcode` - has no typings. I want to write them in the future and make this module better. For now, this PR is big enough and didn't want to include the next big thing.

packages/platform-ios/src/link-pods/__tests__/isInstalled.test.ts
Esemesek: The `isInstalled` function is not written correctly. We are passing whole big objects to the function, but not all properties are used. I've added `ts-ignore` in the test to prevent adding unused mock data.

Esemesek:
@tido64 I resolved all the comments that were fixed. There is still plenty of work to do. We have to add types to the `xcode` module and then make types more strict. Thank you for walking through this quite long and boring PR C:

PR #621, opened 2019-08-05
[iOS][auto-linking] fix: Add missing scriptPhases config

Salakar:
Summary:
---------

This PR fixes the `scriptPhases` config support that was partially implemented as part of the first version of auto linking for iOS.

Test Plan:
----------

Locally linked and tested using React Native Firebase custom script phases. ~Will post screenshots tomorrow, it's late here 😴~

**Custom script phase config used for testing:**
image

**Logging showing custom script phase was added:**
image

**Custom script phase is shown in the generated workspace:**
image


packages/cli/src/tools/config/schema.js
```diff
@@ -52,6 +52,7 @@ export const dependencyConfig = t
                 podspecPath: t.string(),
                 sharedLibraries: t.array().items(t.string()),
                 libraryFolder: t.string(),
+                scriptPhases: t.array().items(t.object()),
```
thymikee: we could be a bit more specific there to help users avoid typos

packages/platform-ios/native_modules.rb
```diff
@@ -75,6 +76,7 @@ def use_native_modules!(root = "..", config = nil)
           phase["execution_position"] = phase["execution_position"].to_sym
         end
 
+        phase = Hash[phase.map { |k, v| [k.to_sym, v] }]
```
Salakar: cc @orta @alloy - I wasn't sure if this was the best way of doing it but I needed to symbolize the keys for it to be accepted for `script_phase` (otherwise errors with invalid args)

orta: I don't know if there's a _better_ way, but this is fine IMO

PR #631, opened 2019-08-13
Update copyProjectTemplateAndReplace.js

pvinis:
Summary:
---------

Quick fix for the `_prettierrc.js` file. We should probably fix it with a proper lookup of all files in depth 0 starting with `_`.

Test Plan:
----------


thymikee:
Can you update the e2e test as well?

PR #632, opened 2019-08-14
fix docstring

mmerickel:
https://github.com/react-native-community/cli/pull/588#discussion_r313600324

### DIFF
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -408,7 +408,7 @@ Starts the server that communicates with connected devices
 
 Specify port to listen on
 
-#### `--projectRoot [list]`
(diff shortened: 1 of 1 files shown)


docs/commands.md
```diff
@@ -408,7 +408,7 @@ Starts the server that communicates with connected devices
 
 Specify port to listen on
 
-#### `--projectRoot [list]`
+#### `--projectRoot [path]`
```
thymikee:
```suggestion
#### `--projectRoot [string]`
```

mmerickel: I chose path as it’s more specific and was used down below on cert and key options.

PR #640, opened 2019-08-19
Remove Delta Bundler from debugger UI.

cpojer:
Summary:
---------
Just cleaning up some code, loading regular bundles is fine.

Test Plan:
----------

Try out the debugger-ui, it works.


cpojer:
Would be nice if somebody could double check this in open source.

packages/cli/src/commands/server/debugger-ui/index.html
```diff
@@ -185,27 +181,8 @@
       } else if (object.method === '$disconnected') {
         shutdownJSRuntime();
         Page.setState({status: {type: 'disconnected'}});
-      } else if (object.method === 'executeApplicationScript') {
```
thymikee: is it safe to remove this?

cpojer: Yes, it will simply fall through and call `postMessage` below so it's going to do the same as before, just with less code ;)

cpojer:
There is a lot of complexity involved with building native delta bundler clients across platforms. Now we have Fast Refresh (RN 0.61) which reduces the amount of times somebody will refresh (HMR is basically a variant of delta bundling) and at FB we are doing bundle splitting in development which reduced time spent in Metro by about 10x and also reduced memory usage on device by the same amount.

tl;dr: we believe reloads on the client are now fast enough not to need the extra complexity. The server parts of delta bundling within Metro will stay the same and are invaluable, of course.

PR #641, opened 2019-08-20
Update banner colors

cpojer:
Summary:
---------

Improved colors

Test Plan:
----------

yolo

### DIFF
--- a/packages/cli/src/commands/init/banner.js
+++ b/packages/cli/src/commands/init/banner.js
@@ -38,6 +38,6 @@ const learnOnceMessage =
 
 export default `${chalk.blue(reactLogoArray.join('\n'))}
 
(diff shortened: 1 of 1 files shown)


packages/cli/src/commands/init/banner.js
```diff
@@ -38,6 +38,6 @@ const learnOnceMessage =
 
 export default `${chalk.blue(reactLogoArray.join('\n'))}
 
-${chalk.yellow.bold(welcomeMessage)}
-${chalk.gray(learnOnceMessage)}
+${chalk.blue.bold(welcomeMessage)}
+${chalk.dim(learnOnceMessage)}
```
thymikee: somehow I missed it was gray before, good change

Issue #561, opened 2019-07-18
When we init a new project, some template strings are not replaced


pvinis:
## Environment

```
npx react-native info
info Fetching system and libraries information...
System:
    OS: macOS 10.14.5
    CPU: (8) x64 Intel(R) Core(TM) i5-8259U CPU @ 2.30GHz
    Memory: 613.23 MB / 16.00 GB
    Shell: 5.3 - /bin/zsh
  Binaries:
    Node: 12.3.1 - ~/.nodenv/versions/12.3.1/bin/node
    Yarn: 1.16.0 - /usr/local/bin/yarn
    npm: 6.9.0 - ~/.nodenv/versions/12.3.1/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  SDKs:
    iOS SDK:
      Platforms: iOS 12.2, macOS 10.14, tvOS 12.2, watchOS 5.2
    Android SDK:
      API Levels: 28, 29
      Build Tools: 28.0.3, 29.0.1
      System Images: android-29 | Google APIs Intel x86 Atom
  IDEs:
    Android Studio: 3.4 AI-183.6156.11.34.5692245
    Xcode: 10.2.1/10E1001 - /usr/bin/xcodebuild
```

## Description

Two files, one for ios and one for android have a template string as the name of the app. In both files it's `Hello App Display Name`.

I tried running `react-native init Bla` using the latest `react-native-cli` and it worked well, replacing these strings with `Bla`.
Using the latest `@react-native-community/cli`, the strings were not replaced.

## Reproducible Demo
- Run `npx react-native init Bla`
- Open file `Bla/ios/Bla/Info.plist` and `Bla/android/app/src/main/res/values/string.xml` and notice that in both files there is the string `Hello App Display Name` as the project's name.

This should normally be `Bla` in our case.

<img width="97" alt="Screenshot 2019-07-11 at 18 11 19" src="https://user-images.githubusercontent.com/100233/61468203-2f3c5500-a97d-11e9-99e6-3c3af7f24739.png">

Taym95:
I would like to give this issue a try!

pvinis:
What details do you need @Taym95?

PR #650, opened 2019-08-27
fix: Change title for newly initialised projects

Esemesek:
Summary:
---------

Fixes: #561 

Apart from fixing a bug, I've added the ability to specify title placeholder for custom templates. Additionally, you can specify custom `title` for the project by passing `--title` flag.

Test Plan:
----------

- [x] - Tested manually
- [x] - UT passing

### DIFF
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -134,6 +134,10 @@ Uses a valid semver version of React Native as a template.
 
 Uses a custom directory instead of `<projectName>`.
 
+### `--title [string]`
+
+Uses a custom title instead of `<projectName>`.
+
 #### `--template [string]`
 
 Uses a custom template. Accepts following template sources:
@@ -162,6 +166,8 @@ type Template = {
   templateDir: string;
   // Path to script, which will be executed after init
   postInitScript?: string;
+  // Placeholder used to rename app title inside values.xml and Info.plist
+  titlePlaceholder?: string;
 };
 ```
 
@@ -170,6 +176,7 @@ Example `template.config.js`:
 ```js
 module.exports = {
   placeholderName: 'ProjectName',
+  titlePlaceholder: 'Project Display Name',
   templateDir: './template',
   postInitScript: './script.js',
 };
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -136,7 +136,7 @@ Uses a custom directory instead of `<projectName>`.
 
 ### `--title [string]`
 
-Uses a custom title instead of `<projectName>`.
+Uses a custom app title instead of `<projectName>`.
 
 #### `--template [string]`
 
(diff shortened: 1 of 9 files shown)


docs/commands.md
```diff
@@ -134,6 +134,10 @@ Uses a valid semver version of React Native as a template.
 
 Uses a custom directory instead of `<projectName>`.
 
+### `--title [string]`
+
+Uses a custom title instead of `<projectName>`.
```
thymikee:
```suggestion
Uses a custom app title instead of `<projectName>`.
```

packages/cli/src/commands/init/editTemplate.js
```diff
@@ -4,6 +4,8 @@ import path from 'path';
 import walk from '../../tools/walk';
 import {logger} from '@react-native-community/cli-tools';
 
+const DEFAULT_TITLE_PLACEHOLDER = 'Hello App Display Name';
```
thymikee: we'll be able to remove it once we add the title to default template, right?

packages/cli/src/commands/init/init.js
```diff
@@ -125,7 +129,12 @@ async function createFromTemplate({
     loader.succeed();
     loader.start('Processing template');
 
-    changePlaceholderInTemplate(projectName, templateConfig.placeholderName);
+    changePlaceholderInTemplate(
+      projectName,
+      templateConfig.placeholderName,
+      templateConfig.titlePlaceholder,
+      projectTitle,
```
thymikee: an object argument would make more sense now

packages/cli/src/commands/init/index.js
```diff
@@ -24,5 +24,9 @@ export default {
       name: '--directory [string]',
       description: 'Uses a custom directory instead of `<projectName>`.',
     },
+    {
+      name: '--title [string]',
+      description: 'Uses a custom title name for application',
```
thymikee:
```suggestion
      description: 'Uses a custom app title name for application',
```

packages/cli/src/commands/init/editTemplate.js
Esemesek: In theory - yes, but in practice, some of the templates might still use this string, so would keep it for a couple of versions. You can see that I typed `templateConfig` with optional type for `titlePlaceholder`.

thymikee: let's leave a TODO comment then

Esemesek: Right C:

packages/cli/src/commands/init/init.js
Esemesek: Done

PR #651, opened 2019-08-27
fix: Infinite spinner when installing cocoapods

Esemesek:
Summary:
---------

When a user doesn't have cocoapods installed and doesn't have sudo permissions they were prompted with infinite spinner. To fix that I've added system password prompt.

cc @orta @cpojer 

This is how it looks like:

Kapture 2019-08-27 at 14 20 53

Or MP4 here: https://dump.video/i/4DnHbM.mp4

Test Plan:
----------
- [x] - Unit tests passing
- [x] - Tested manually

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -55,6 +55,7 @@
     "semver": "^5.0.3",
     "serve-static": "^1.13.1",
     "shell-quote": "1.6.1",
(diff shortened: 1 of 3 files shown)


lucasbento:
Can we merge this @thymikee?

PR #653, opened 2019-08-28
feat: Allow passing in MainPackageConfig to PackageList

akshetpandey:
Summary:
---------

This allow customizing configuration for the MainReactPackage.
Right now this is limited to changing how FrescoModule gets initialized
but in the future this can allow changes in behavior for other core modules

Test Plan:
----------

Since this is a template only change, it shouldn't require any tests.
I have verified that this is not a breaking change for existing users.

### DIFF
--- a/docs/autolinking.md
+++ b/docs/autolinking.md
@@ -66,6 +66,7 @@ The [native_modules.gradle](https://github.com/react-native-community/cli/blob/m
    1. A first Gradle plugin (in `settings.gradle`) runs `applyNativeModulesSettingsGradle` method. It uses the package metadata from `react-native config` to add Android projects.
    1. A second Gradle plugin (in `app/build.gradle`) runs `applyNativeModulesAppBuildGradle` method. It creates a list of React Native packages to include in the generated `/android/build/generated/rn/src/main/java/com/facebook/react/PackageList.java` file.
 1. At runtime, the list of React Native packages generated in step 1.2 is registered by `getPackages` method of `ReactNativeHost` in `MainApplication.java`.
+   1. You can optionally pass in an instance of `MainPackageConfig` when initializing `PackageList` if you want to override the default configuration of `MainReactPackage`.
 
 ### Example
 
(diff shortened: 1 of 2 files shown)


akshetpandey:
One use case for this is here: [MainApplication.java](https://github.com/akshetpandey/react-native-cronet/blob/master/example/android/app/src/main/java/com/example/MainApplication.java#L29)

thymikee:
I'm good with this, but would also like @Salakar and @dulmandakh to take a look. In the meantime, can you please update the documentation?

akshetpandey:
I am not sure if PackageList is documented directly.

The closest thing I could find was here [autolinking.md#platform-android](https://github.com/react-native-community/cli/blob/8c2d5b5f337db02c8868c627d9e4a3b39e0af329/docs/autolinking.md#platform-android)

I can change the example here: [helloworld/MainApplication.java](https://github.com/facebook/react-native/blob/769e35ba5f4c31ef913035a5cc8bc0e88546ca55/template/android/app/src/main/java/com/helloworld/MainApplication.java#L22-L28) but that is in the react-native repo

thymikee:
There's https://github.com/react-native-community/cli/blob/8c2d5b5f337db02c8868c627d9e4a3b39e0af329/docs/dependencies.md#platformsandroidpackageinstance. We need to restructure the docs so that they're more accessible, but not sure about the approach yet

akshetpandey:
@thymikee : I am not sure what I should do with that documentation. We probably need to to figure out the right API we want to expose for users to manually configure their modules.

One option is, to ask users to subclass MainPackageConfig and pass in an instance of that when they create PackageList in MainApplication, or whenever they are initializing react-native. If that makes sense I can create a new PR to add that documentation.

While we figure that out, can we merge this PR?

akshetpandey:
@thymikee : Sorry for bothering you, but is there anything I can do to get this merged. This blocks v1 release of https://github.com/akshetpandey/react-native-cronet for me.

thymikee:
@akshetpandey yep, a brief mention in the docs please :D See https://github.com/react-native-community/cli/pull/653#issuecomment-526100009 if you missed it

akshetpandey:
I added some documentation in [docs/autolinking.md](https://github.com/react-native-community/cli/blob/master/docs/autolinking.md#platform-android) as that seems to be a more appropriate place than docs/dependencies.md

thymikee:
Thank you!

PR #662, opened 2019-09-03
fix semver check was breaking for non-semver versions

jmeistrich:
Summary:
---------

The semver check was failing for version numbers that don't fit semver, like for Xcode version 10.3.


packages/cli/src/commands/doctor/checkInstallation.js
```diff
@@ -17,6 +17,6 @@ const isSoftwareInstalled = async command => {
 };
 
 const doesSoftwareNeedToBeFixed = ({version, versionRange}) =>
-  version === 'Not Found' || !semver.satisfies(version, versionRange);
+  version === 'Not Found' || !semver.satisfies(semver.coerce(version), versionRange);
```
thymikee: needs a `yarn lint --fix` run to fix Prettier

PR #663, opened 2019-09-03
feat(doctor): Reuse installCocoaPods() and add Homebrew option

tido64:
Summary:
---------

Consolidates how to install CocoaPods in a single function, and added the option to install it from Homebrew.


packages/cli/src/tools/installPods.js
```diff
@@ -41,6 +41,59 @@ function runSudo(command: string): Promise<void> {
...
+        }
+        break;
+      case withHomebrew:
+        command = 'brew install cocoapods';
+        await execa('brew', ['install', 'cocoapods']);
+        break;
+    }
+  } catch (error) {
+    loader.fail();
+    logger.log(error.stderr);
```
Esemesek: `logger.error(error.stderr);`

PR #669, opened 2019-09-04
Remove unused import (child_process.exec)

jackkinsella:
One-line removal of an un-used import. It isn't much but removing un-used code always helps in getting things understandable.

### DIFF
--- a/packages/global-cli/index.js
+++ b/packages/global-cli/index.js
@@ -38,7 +38,6 @@
 
 const fs = require('fs');
 const path = require('path');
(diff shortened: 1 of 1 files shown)


thymikee:
Oh, we try to discourage devs from using global react-native-cli. We (current maintainers) also don't have the npm publish rights. We plan to rewrite this module to be a proxy to local cli (`@react-native-community/cli`), so I'm inclined  to close it. Thanks for caring about it! :)

PR #672, opened 2019-09-04
doctor: add error messages for failures

jmeistrich:
Summary:
---------

This adds error messages for failing checks.

### DIFF
--- a/packages/cli/src/commands/doctor/doctor.js
+++ b/packages/cli/src/commands/doctor/doctor.js
@@ -27,7 +27,7 @@ const printIssue = ({
   label: string,
   needsToBeFixed: boolean,
   isRequired: boolean,
-  description: String,
+  description: string,
 }) => {
   const symbol = needsToBeFixed
     ? isRequired
@@ -96,7 +96,7 @@ export default (async function runDoctor(
         return {
           label: healthcheck.label,
           needsToBeFixed: Boolean(needsToBeFixed),
-          description: needsToBeFixed,
+          description: String(needsToBeFixed),
           runAutomaticFix: healthcheck.runAutomaticFix,
           isRequired,
           type: needsToBeFixed
--- a/packages/cli/src/commands/doctor/doctor.js
+++ b/packages/cli/src/commands/doctor/doctor.js
@@ -22,10 +22,12 @@ const printIssue = ({
   label,
   needsToBeFixed,
   isRequired,
+  description,
 }: {
   label: string,
   needsToBeFixed: boolean,
   isRequired: boolean,
+  description: String,
 }) => {
   const symbol = needsToBeFixed
     ? isRequired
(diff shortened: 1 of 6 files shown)


thymikee:
@jmeistrich can I ask you to rebase? We've added types to doctor command and now conflicts arisen

thymikee:
CI is failing because of known race conditions (with yet unknown cause) in config test

thymikee:
Thank you!

jmeistrich:
Happy to help!

packages/cli/src/commands/doctor/doctor.js
```diff
@@ -22,18 +22,20 @@ const printIssue = ({
...
+  description: string,
 }) => {
   const symbol = needsToBeFixed
     ? isRequired
       ? chalk.red('✖')
       : chalk.yellow('●')
     : chalk.green('✓');
 
-  logger.log(` ${symbol} ${label}`);
+  logger.log(` ${symbol} ${label}${needsToBeFixed ? ': ' + description : ''}`);
```
lucasbento: @jmeistrich @thymikee: what's the reason for adding this?

With this change, this is how it looks:

image

thymikee: That's the reason – to show why the check failed. We treat it as first pass, feel free to iterate on how to present it better :)

lucasbento: Perhaps it would be nice to show this only with `--verbose`?

thymikee: Nah, why hiding the reason behind failure? Running doctor again may take some time as well.

lucasbento: Good point.

I would like to, at least, get rid of `should be installed` as there isn't much being provided there, thoughts?

thymikee: "is expected to be installed globally" maybe?

lucasbento: > is expected to be installed globally

Cool, I like it.

`ios-deploy` is not required to have installed though, the user only needs it if they want to run on the device through the CLI, how would we put that in there?

thymikee: We could set messages per package after running `checkSoftwareInstalled`. This way we can get rid of this a function returning either true or a string. Would you like to refactor?

lucasbento: You mean that each health check file can provide its own `description`?

thymikee: Something like this could be added to `getDiagnostics()` output. Currently `needsToBeFixed` can be a boolean or string. But we could make it a bit more explicit and add extra information like error message/descirption. Or we can keep the current behavior and only do something like:
```ts
    needsToBeFixed: await checkSoftwareInstalled('pod') || "custom message",
  }),
```

lucasbento: If I got what you mean right, it would be nice to do:

```js
    description: 'required for installing dependencies',
    needsToBeFixed: await checkSoftwareInstalled('pod'),
  }),
```

Is that it?

thymikee: Could be, I'm ok with that.

lucasbento: Cool, I'm finishing up some stuff to open a PR, will update #694 with that to refactor it, thanks @thymikee.

PR #675, opened 2019-09-04
Doctor/windows

jmeistrich:
Summary:
---------

This fixes multiple issues on Windows:
1. It was crashing when checking Android SDKs if no Android SDKs installed.
2. It was checking for watchman, which is not part of the install guide for Windows.
3. There is a bug in envinfo which does not get Android Studio info correctly: https://github.com/tabrindle/envinfo/pull/119. This has a workaround until that is fixed.

Test Plan:
----------

Run `doctor` on a Windows machine with no Android SDKs installed to see #1, and then install Android SDKs and see that it was still showing an error.

### DIFF
--- a/packages/cli/src/commands/doctor/healthchecks/androidSDK.js
+++ b/packages/cli/src/commands/doctor/healthchecks/androidSDK.js
@@ -17,32 +17,34 @@ export default {
     // Windows. This can be removed when envinfo fixes it.
     // See the PR: https://github.com/tabrindle/envinfo/pull/119
     if (sdks === 'Not Found' && process.platform !== 'darwin') {
-      sdks = await new Promise(resolve => {
-        exec(
-          process.env.ANDROID_HOME
-            ? `${process.env.ANDROID_HOME}/tools/bin/sdkmanager --list`
-            : 'sdkmanager --list',
-          (err, stdout) => {
-            if (err) {
-              resolve('Not Found');
-            } else {
-              const matches = [];
-              const regex = /build-tools;([\d|.]+)[\S\s]/g;
-              let match = null;
-              while ((match = regex.exec(stdout)) !== null) {
-                matches.push(match[1]);
-              }
-              resolve(
+      try {
+        sdks = await new Promise((resolve, reject) => {
+          exec(
+            process.env.ANDROID_HOME
+              ? `${process.env.ANDROID_HOME}/tools/bin/sdkmanager --list`
+              : 'sdkmanager --list',
+            (err, stdout) => {
+              if (err) {
+                reject();
+              } else {
+                const matches = [];
+                const regex = /build-tools;([\d|.]+)[\S\s]/g;
+                let match = null;
+                while ((match = regex.exec(stdout)) !== null) {
+                  matches.push(match[1]);
+                }
                 matches.length > 0
-                  ? {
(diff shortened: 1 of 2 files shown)


packages/cli/src/commands/doctor/healthchecks/androidSDK.js
```diff
@@ -2,15 +2,58 @@ import chalk from 'chalk';
...
+    // See the PR: https://github.com/tabrindle/envinfo/pull/119
+    if (sdks === 'Not Found' && process.platform !== 'darwin') {
+      sdks = await new Promise(resolve => {
+        exec(
+          process.env.ANDROID_HOME
+            ? `${process.env.ANDROID_HOME}/tools/bin/sdkmanager --list`
+            : 'sdkmanager --list',
+          (err, stdout) => {
+            if (err) {
+              resolve('Not Found');
```
tido64: Should this be `reject()`?

packages/cli/src/commands/doctor/healthchecks/androidSDK.js
```diff
@@ -2,15 +2,58 @@ import chalk from 'chalk';
...
+              );
+            }
+          },
+        );
+      });
+    }
+
+    return {
+      needsToBeFixed:
+        (sdks === 'Not Found' && installMessage) ||
```
tido64: Can we make `'Not Found'` a constant or make `sdks` undefined instead?

packages/cli/src/commands/doctor/healthchecks/androidSDK.js
Esemesek: I was wondering same thing. This probably should be `reject`.

packages/cli/src/commands/doctor/healthchecks/androidSDK.js
```diff
@@ -2,15 +2,58 @@ import chalk from 'chalk';
...
+              let match = null;
+              while ((match = regex.exec(stdout)) !== null) {
+                matches.push(match[1]);
+              }
+              resolve(
+                matches.length > 0
+                  ? {
+                      'Build Tools': matches,
+                    }
+                  : 'Not Found',
```
Esemesek: `Not Found` is resolved here as well.

jmeistrich:
I changed it to reject on failure and set sdks to 'Not Found' in the catch so it's just in one place. Is this better?

I don't think it should make `sdks` undefined because the goal is to fix the output of the broken call in envinfo, and once that's fixed this block can just be removed. envinfo sets it to 'Not Found' so this should just replicate that.

packages/cli/src/commands/doctor/healthchecks/androidSDK.js
```diff
@@ -2,15 +2,60 @@ import chalk from 'chalk';
 import {logManualInstallation} from './common';
 import versionRanges from '../versionRanges';
 import {doesSoftwareNeedToBeFixed} from '../checkInstallation';
+import {exec} from 'child_process';
```
thymikee: Can we use `execa` for spawning process? It has a built-in promise interface and seems to work better with Windows by default

jmeistrich: You got it. Updated to use execa.

PR #676, opened 2019-09-05
feat: launch emulator during run-android

### LINKED ISSUE #142: run-android should also open default simulator, like run-ios
Right now, run-android will fail when there's no emulator. It can be annoying sometimes. I think it would be great to have this feature supported.

@ferrannp was investigating this issue already

jaysbytes:
Summary:
---------

This PR resolves #142
TLDR; 
During`react-native run-android` the  first of user's emulators will be launched if there is no any running android device/emulator

`emulator -list-advs` to get a list of emulators
`emulator @FIRST_ON_LIST` to run the first available emulator

Test Plan:
----------

* Generate a project with React-native >60.0
* install CLI from my fork
* run `react-native run-android` without launched emulator -> emulator is launching
* run `react-native run-android` with launched emulator -> nothing new happens
* run `react-native run-android` with device connected -> nothing new happens
* run `react-native run-android` without launched emulator, emulator has saved snapshot -> emulator is launching

### DIFF
--- a/packages/platform-android/src/commands/runAndroid/runOnAllDevices.ts
+++ b/packages/platform-android/src/commands/runAndroid/runOnAllDevices.ts
@@ -12,6 +12,7 @@ import {logger, CLIError} from '@react-native-community/cli-tools';
 import adb from './adb';
 import tryRunAdbReverse from './tryRunAdbReverse';
 import tryLaunchAppOnDevice from './tryLaunchAppOnDevice';
+import tryLaunchEmulator from './tryLaunchEmulator';
 import {Flags} from '.';
 
 function getTaskNames(
@@ -27,13 +28,24 @@ function toPascalCase(value: string) {
   return value[0].toUpperCase() + value.slice(1);
 }
 
-function runOnAllDevices(
+async function runOnAllDevices(
   args: Flags,
   cmd: string,
   packageNameWithSuffix: string,
   packageName: string,
   adbPath: string,
 ) {
+  const devices = adb.getDevices(adbPath);
+  if (devices.length === 0) {
+    logger.info('Trying to launch emulator...');
+    const result = await tryLaunchEmulator(adbPath);
+    if (result) {
+      logger.info('Emulator launch succeeded!');
+    } else {
+      logger.warn('Emulator launch failed.');
+    }
+  }
+
   try {
     const tasks = args.tasks || ['install' + toPascalCase(args.variant)];
     const gradleArgs = getTaskNames(args.appFolder, tasks);
@@ -51,7 +63,6 @@ function runOnAllDevices(
   } catch (error) {
     throw createInstallError(error);
   }
-  const devices = adb.getDevices(adbPath);
 
   (devices.length > 0 ? devices : [undefined]).forEach(
     (device: string | void) => {
--- a/packages/platform-android/src/commands/runAndroid/runOnAllDevices.ts
+++ b/packages/platform-android/src/commands/runAndroid/runOnAllDevices.ts
@@ -47,7 +47,7 @@ async function runOnAllDevices(
         `Failed to launch emulator. Reason: ${chalk.dim(result.error || '')}.`,
       );
       logger.warn(
-        'Please launch an emulator manually or connect a device. Otherwise app launch may fail.',
+        'Please launch an emulator manually or connect a device. Otherwise app may fail to launch.',
       );
     }
   }
--- a/packages/platform-android/src/commands/runAndroid/runOnAllDevices.ts
+++ b/packages/platform-android/src/commands/runAndroid/runOnAllDevices.ts
@@ -35,12 +35,13 @@ async function runOnAllDevices(
   packageName: string,
   adbPath: string,
 ) {
-  const devices = adb.getDevices(adbPath);
+  let devices = adb.getDevices(adbPath);
   if (devices.length === 0) {
     logger.info('Trying to launch emulator...');
     const result = await tryLaunchEmulator(adbPath);
     if (result.success) {
       logger.info('Emulator launched!');
+      devices = adb.getDevices(adbPath);
     } else {
       logger.error('Emulator launch failed.');
       if (result.error !== undefined) {
--- a/packages/platform-android/src/commands/runAndroid/runOnAllDevices.ts
+++ b/packages/platform-android/src/commands/runAndroid/runOnAllDevices.ts
@@ -37,16 +37,15 @@ async function runOnAllDevices(
 ) {
   let devices = adb.getDevices(adbPath);
   if (devices.length === 0) {
-    logger.info('Trying to launch emulator...');
+    logger.info('Launching emulator...');
     const result = await tryLaunchEmulator(adbPath);
     if (result.success) {
-      logger.info('Emulator launched!');
+      logger.info('Successfully launched emulator.');
       devices = adb.getDevices(adbPath);
     } else {
-      logger.error('Emulator launch failed.');
-      if (result.error !== undefined) {
-        logger.error(result.error);
-      }
+      logger.error(
+        `Failed to launch emulator. Reason: ${chalk.dim(result.error || '')}.`,
+      );
       logger.warn(
         'Please launch an emulator manually or connect a device. Otherwise app launch may fail.',
       );
--- a/packages/platform-android/src/commands/runAndroid/runOnAllDevices.ts
+++ b/packages/platform-android/src/commands/runAndroid/runOnAllDevices.ts
@@ -39,10 +39,16 @@ async function runOnAllDevices(
   if (devices.length === 0) {
     logger.info('Trying to launch emulator...');
     const result = await tryLaunchEmulator(adbPath);
-    if (result) {
-      logger.info('Emulator launch succeeded!');
+    if (result.success) {
+      logger.info('Emulator launched!');
     } else {
-      logger.warn('Emulator launch failed.');
+      logger.error('Emulator launch failed.');
+      if (result.error !== undefined) {
+        logger.error(result.error);
+      }
+      logger.warn(
+        'Please launch an emulator manually or connect a device. Otherwise app launch may fail.',
+      );
     }
   }
 
--- a/packages/platform-android/src/commands/runAndroid/tryLaunchEmulator.ts
+++ b/packages/platform-android/src/commands/runAndroid/tryLaunchEmulator.ts
@@ -12,10 +12,16 @@ const getEmulators = () => {
 
 const launchEmulator = async (emulatorName: string, adbPath: string) => {
   return new Promise((resolve, reject) => {
-    const cp = execa('emulator', [`@${emulatorName}`], {
-      detached: true,
-      stdio: ['ignore', 'pipe', 'ignore'],
-    });
+    const cp = execa(
+      process.env.ANDROID_HOME
+        ? `${process.env.ANDROID_HOME}/emulator/emulator`
+        : 'emulator',
+      [`@${emulatorName}`],
+      {
+        detached: true,
+        stdio: ['ignore', 'pipe', 'ignore'],
+      },
+    );
 
     const timeout = 30;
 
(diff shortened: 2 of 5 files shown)


packages/platform-android/src/commands/runAndroid/tryLaunchEmulator.ts
```diff
@@ -0,0 +1,61 @@
+import {execFileSync, spawn} from 'child_process';
```
thymikee: Please use execa

packages/platform-android/src/commands/runAndroid/runOnAllDevices.ts
```diff
@@ -27,13 +28,24 @@ function toPascalCase(value: string) {
...
   packageNameWithSuffix: string,
   packageName: string,
   adbPath: string,
 ) {
+  const devices = adb.getDevices(adbPath);
+  if (devices.length === 0) {
+    logger.info('Trying to launch emulator...');
+    const result = await tryLaunchEmulator(adbPath);
+    if (result) {
+      logger.info('Emulator launch succeeded!');
```
thymikee: Emulator launched

packages/platform-android/src/commands/runAndroid/runOnAllDevices.ts
```diff
@@ -27,13 +28,24 @@ function toPascalCase(value: string) {
...
+    } else {
+      logger.warn('Emulator launch failed.');
```
thymikee: Add "Please open an emulator manually, Or connect a device."
Also please add a debug log with the error message

packages/platform-android/src/commands/runAndroid/tryLaunchEmulator.ts
```diff
@@ -0,0 +1,61 @@
...
+  return new Promise((resolve, reject) => {
+    const cp = spawn('emulator', [`@${emulatorName}`], {
+      detached: true,
+      stdio: ['ignore', 'pipe', 'ignore'],
+    });
+
+    // Reject command after timeout
+    const rejectTimeout = setTimeout(() => {
+      cp.stdout.destroy();
+      reject();
```
thymikee: Please pass something like "new Error('timeout exceeded')" to reject

packages/platform-android/src/commands/runAndroid/tryLaunchEmulator.ts
```diff
@@ -0,0 +1,61 @@
...
+const launchEmulator = async (emulatorName: string, adbPath: string) => {
+  return new Promise((resolve, reject) => {
+    const cp = spawn('emulator', [`@${emulatorName}`], {
+      detached: true,
+      stdio: ['ignore', 'pipe', 'ignore'],
+    });
+
+    // Reject command after timeout
+    const rejectTimeout = setTimeout(() => {
+      cp.stdout.destroy();
```
thymikee: Do we need this?

packages/platform-android/src/commands/runAndroid/tryLaunchEmulator.ts
```diff
@@ -0,0 +1,61 @@
...
+};
+
+export default async function tryLaunchEmulator(adbPath: string) {
+  const emulators = getEmulators();
+  if (emulators.length > 0) {
+    try {
+      await launchEmulator(emulators[0], adbPath);
+      return true;
+    } catch (e) {
+      console.log('Emulator error', e);
```
thymikee: logger.error

packages/platform-android/src/commands/runAndroid/tryLaunchEmulator.ts
```diff
@@ -0,0 +1,61 @@
...
+    setTimeout(() => {
+      if (Adb.getDevices(adbPath).length > 0) {
+        resolve();
+      }
+    }, 5000);
+
+    cp.unref();
+
+    cp.stdout.addListener('data', message => {
+      if (message.toString().indexOf('boot completed') >= 0) {
```
thymikee: Includes()

packages/platform-android/src/commands/runAndroid/tryLaunchEmulator.ts
```diff
@@ -0,0 +1,61 @@
...
+
+    // Reject command after timeout
+    const rejectTimeout = setTimeout(() => {
+      cp.stdout.destroy();
+      reject();
+    }, 30 * 1000);
+
+    // When emulator is started from snapshot, it does not emit boot completed message.
+    // It starts immediately so we can check if device is present
+    setTimeout(() => {
```
thymikee: Collect all the timeouts and cancel them when needed

packages/platform-android/src/commands/runAndroid/tryLaunchEmulator.ts
jaysbytes: Yes, without this I was unable to detach the child process from its parent. The main process `react-native run-android` has not beeing finished until emulator was turned off. Destroying the stdout is disconnecting a pipe between processes and they can be managed independently

Titozzz:
Something that bothers me, there is no point in declaring a function as async just to await its return, you are just overwrapping stuff as it already returns a promise.

jaysbytes:
Thanks, @Titozzz  :) It was redundant, fixes.

packages/platform-android/src/commands/runAndroid/tryLaunchEmulator.ts
```diff
@@ -0,0 +1,80 @@
...
+    const emulatorsOutput = execa.sync('emulator', ['-list-avds']).stdout;
+    return emulatorsOutput.split('\n').filter(name => name !== '');
+  } catch {
+    return [];
+  }
+};
+
+const launchEmulator = async (emulatorName: string, adbPath: string) => {
+  return new Promise((resolve, reject) => {
+    const cp = execa('emulator', [`@${emulatorName}`], {
```
thymikee: `emulator` command may not be available if someone doesn't have `$ANDROID_HOME` added to shell path, which is not mandatory (the variable needs to exists tho). 
```suggestion
    const cp = execa(
      process.env.ANDROID_HOME 
        ? `${process.env.ANDROID_HOME}/emulator/emulator` 
        : 'emulator', 
      [`@${emulatorName}`],
    {
```
Would be nice to also catch if both branches failed for some reason, and give a message like: 
> Couldn't launch Android Emulator. Please make sure $ANDROID_HOME env variable is set or the command "emulator" exists.

packages/platform-android/src/commands/runAndroid/tryLaunchEmulator.ts
```diff
@@ -0,0 +1,80 @@
...
+      detached: true,
+      stdio: ['ignore', 'pipe', 'ignore'],
+    });
+
+    const timeout = 30;
```
thymikee: I think we can make it 15, which is already pretty long, but I can see there are slower environments to develop with. What do you think?
```suggestion
    const timeout = 15;
```

jaysbytes: I'm not sure about the timeouts here @thymikee . On My machine time to `emulator` prints
> emulator: INFO: boot completed
is 14.25 seconds. 15 may be to less. 
and to `adb devices` to return that device is online (or is a `device` actually) it takes about one second less, like 13 seconds. 
I was wondering if we could just check for adb devices in an interval using `adb.getDevices(adbPath)`. It could be one interval for both cold and snapshot boot.
Then we could clear interval once it is booted or after the delay

packages/platform-android/src/commands/runAndroid/tryLaunchEmulator.ts
jaysbytes: How we can check if it failed because of a missing path? check exit code or check if binary exists before even launching?

thymikee: I think the check I proposed should be just enough

packages/platform-android/src/commands/runAndroid/tryLaunchEmulator.ts
thymikee: Ah I assumed this is not the time to boot the emulator but to just start it. Do you think we need to wait for this event? I'd be cool with just starting it and failing only if it couldn't be started for some reason

jaysbytes: if the emulator is not booted but just started it will not be available to install an app. We can't assume that build takes some time (when emulator could boot properly) because in case of cached build it takes seconds. I think the best approach will be to ignore that event and in the interval, loop checks if there is a device on the list. It would simplify the code.

thymikee: Yea, I just realised that after I wrote the message. Do whatever is simpler and reliable

thymikee:
cc @satya164 would you like to give it a try? :)

jaysbytes:
@thymikee thanks for the help with shipping of this awesome feature 😀

a-c-sreedhar-reddy:
Wouldn't it be better if available emulators are shown and ask to select one?

jsjoeio:
Has this feature been included in a recent release? 

I didn't see it mentioned in any of the [latest release notes](https://github.com/react-native-community/cli/releases)

engineerbaraazakariya:
> Has this feature been included in a recent release?
> 
> I didn't see it mentioned in any of the [latest release notes](https://github.com/react-native-community/cli/releases)

I found this here:
https://github.com/react-native-community/cli/pull/676

Shan7anu:
@jsjoeio Yes. It's available in the latest [release](https://github.com/react-native-community/cli/releases/tag/v3.0.0).


C:\Users\Shan7anu\Desktop\RNProjects\Sample>**react-native run-android**
```
info Running jetifier to migrate libraries to AndroidX. You can disable it using "--no-jetifier" flag.
Jetifier found 864 file(s) to forward-jetify. Using 4 workers...
info Starting JS server...
info Launching emulator...
info Successfully launched emulator.
info Installing the app...
```

jsjoeio:
Woohoo! Thanks for confirming @Shan7anu!

shon3005:
hello @jayu thanks for adding this feature. unfortunately i seem to be unable to get this to work.

for context, android studio is not running at the time i run `npx react-native run-android` in terminal.
expected behavior:
1. android studio boots up
2. emulator automatically launches

actual behavior:
1. Build fails with an error `com.android.builder.testing.api.DeviceException: No connected devices!`


alternatively, i opened up android studio. emulator has not launched yet. i run `npx react-native run-android` in terminal.
expected behavior:
1. emulator automatically launches

actual behavior:
1. Build fails with an error `com.android.builder.testing.api.DeviceException: No connected devices!`


workaround:
I resorted to manually running my emulator. only then was i able to run the command `npx react-native run-android` smoothly.

honeybadger4768:
> hello @jayu thanks for adding this feature. unfortunately i seem to be unable to get this to work.
> 
> for context, android studio is not running at the time i run `npx react-native run-android` in terminal. additional information is that i get this error when i run the above command: `Failed to launch emulator. Reason: The emulator (Pixel_6_API_31) quit before it finished opening. You can try starting the emulator manually from the terminal with: /Library/Android/sdk/emulator/emulator @Pixel_6_API_31.`
> 
> expected behavior:
> 
>     1. android studio boots up
> 
>     2. emulator automatically launches
> 
> 
> actual behavior:
> 
>     1. Build fails with an error `com.android.builder.testing.api.DeviceException: No connected devices!`
> 
> 
> alternatively, i opened up android studio. emulator has not launched yet. i run `npx react-native run-android` in terminal. expected behavior:
> 
>     1. emulator automatically launches
> 
> 
> actual behavior:
> 
>     1. Build fails with an error `com.android.builder.testing.api.DeviceException: No connected devices!`
> 
> 
> workaround: I resorted to manually running my emulator. only then was i able to run the command `npx react-native run-android` smoothly.

I'm using Linux Mint and getting a similar error

```
$ npx react-native run-android
info JS server already running.
info Launching emulator...
error Failed to launch emulator. Reason: The emulator (telefon) quit before it finished opening. You can try starting the emulator manually from the terminal with: /home/furkan/Android/Sdk/emulator/emulator @telefon.
warn Please launch an emulator manually or connect a device. Otherwise app may fail to launch.
```
```
$ emulator -list-avds
telefon
```

PR #679, opened 2019-09-07
Warn user if pod install was not run

Titozzz:
Summary:
---------

Let's detect if the user forgot to run pod install after installing a native library 😄 

Test Plan:
----------

yarn add some-library-with-ios-native-code

react-native run-ios will print an error message until you run pod install warning you that some-library-with-ios-native-code has not been found in the lockfile so it's probably not properly linked

### DIFF
--- a/packages/platform-ios/src/link-pods/readPodfileLock.ts
+++ b/packages/platform-ios/src/link-pods/readPodfileLock.ts
@@ -1,0 +1,8 @@
+import fs from 'fs';
+import {logger} from '@react-native-community/cli-tools';
+
+export default function readPodfileLock(podfileLockPath: string) {
+  logger.debug(`Reading ${podfileLockPath}`);
+  const podLockContent = fs.readFileSync(podfileLockPath, 'utf8');
+  return podLockContent.split(/\r?\n/g);
+}
--- a/packages/platform-ios/src/commands/runIOS/index.js
+++ b/packages/platform-ios/src/commands/runIOS/index.js
@@ -16,6 +16,7 @@ import findXcodeProject from './findXcodeProject';
 import parseIOSDevicesList from './parseIOSDevicesList';
 import findMatchingSimulator from './findMatchingSimulator';
 import warnAboutManuallyLinkedLibs from '../../link/warnAboutManuallyLinkedLibs';
+import warnAboutPodInstall from '../../link/warnAboutPodInstall';
 import {
   logger,
   CLIError,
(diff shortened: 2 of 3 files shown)


Titozzz:
FYI I directly wrote typescript even if the files around were js. Hope that's not an issue. I tried keeping the surface of the PR as thin as possible to make for easier code conflicts

Titozzz:
maybe @orta you want to make sure I parse the podfile.lock correctly?

orta:
It's little bit janky - is there possibly already a proper yaml parser in the dependency tree? Sometimes the ruby yaml writer can be either `key: "string"` or `key: string` and you may end up with pod names which you think are `"\"Pod+Name\""` and not `"Pod+Name"`

packages/platform-ios/src/link-pods/readPodfileLock.ts
```diff
@@ -0,0 +1,8 @@
+import fs from 'fs';
+import {logger} from '@react-native-community/cli-tools';
+
+export default function readPodfileLock(podfileLockPath: string) {
+  logger.debug(`Reading ${podfileLockPath}`);
+  const podLockContent = fs.readFileSync(podfileLockPath, 'utf8');
```
thymikee: this needs a try/catch with a nice error message on what happened

Titozzz:
Thanks will do @orta

Titozzz:
@orta I depend on "SPEC CHECKSUM" as the keys are easy to get from here, is that OK?

thymikee:
@Titozzz we've just merged the PR migrating platforms-ios to TS. Would you mind rebasing? :)

orta:
@Titozzz  that should be fine, as you're mainly just looking for the existence 👍

Titozzz:
Will do tomorrow

packages/platform-ios/src/link-pods/getDependenciesFromPodfileLock.ts
```diff
@@ -0,0 +1,19 @@
...
+export default function getDependenciesFromPodfileLock(
+  podfileLockPath: string,
+) {
+  logger.debug(`Reading ${podfileLockPath}`);
+  let fileContent;
+  try {
+    fileContent = fs.readFileSync(podfileLockPath, 'utf8');
+  } catch (err) {
+    logger.error(
+      `Could not find ${podfileLockPath}, did you run pod \`install\``,
```
thymikee:
```suggestion
      `Could not find ${podfileLockPath}. Did you run \`pod install\` in iOS directory?`,
```

Issue #683, opened 2019-09-09
☂️Umbrella: TypeScript migration


thymikee:
## Describe the Feature

Some time ago we started migrating CLI to TS package by package. The infrastructure is already set up, so we can work with TS and Flow files without issues.

Currently all packages except for `packages/cli` are converted. We need your help in converting the rest of the codebase 💪.

## Work plan

If you'd like to help with migrating any of the packages mentioned below, please leave a comment and we'll assign it to you.

### Commands

Commands live in separate directories. It makes sense to tackle them one by one.

- [x] commands/bundle (@Quadriphobs1)
- [x] commands/config (@netochaves)
- [x] commands/doctor (@thib92)
- [x] commands/info (@FLGMwt)
- [x] commands/init (@gvarandas)
- [x] commands/install (@DanielUche)
- [x] commands/link (@anikethsaha)
- [x] commands/server (@Esemesek)
- [x] commands/upgrade (@thymikee)
- [x] commands/index.js (@thymikee)

### Tools

Tools are helpers we use throughout  the `cli` package. They have corresponding test files, which can be found in `commands/tools/__tests__` directory. Those should also be converted.

- [x] commands/tools/assertRequiredOptions.js (@konojunya)
- [x] commands/tools/assign.js (@radko93)
- [x] commands/tools/brewInstall.js (@thymikee)
- [x] commands/tools/config/ (@ptrocki)
- [x] commands/tools/copyAndReplace.js (@luizcieslak)
- [x] commands/tools/copyFiles.js (@sasurau4)
- [x] commands/tools/findSymlinkedModules.js (@pveyes)
- [x] commands/tools/generator/ (@KangYoosam)
- [x] commands/tools/gracefulifyFs.js (@assuncaocharles)
- [x] commands/tools/install.js (@thymikee)
- [x] commands/tools/installPods.js (@thymikee)
- [x] commands/tools/isValidPackageName.js (@radko93)
- [x] commands/tools/loadMetroConfig.js (@radko93)
- [x] commands/tools/loader.js (@thymikee)
- [x] commands/tools/merge.js (@Naturalclar)
- [x] commands/tools/packageManager.js (@moriyuu)
- [x] commands/tools/releaseChecker/ (@FLGMwt)
- [x] commands/tools/replacePathSepForRegex.js (@FLGMwt)
- [x] commands/tools/walk.js (@FLGMwt)
- [x] commands/tools/yarn.js (@FLGMwt)

### Root index files in `src/`

[...]

PR #684, opened 2019-09-09
chore: migrate upgrade command to TS and remove legacy implementation

thymikee:
Summary:
---------

Part of https://github.com/react-native-community/cli/issues/683

Had issues with inquirer module typings, they somehow clash with Node. Not sure how to resolve, help wanted :)

Btw decided to remove legacy upgrade implementation, it's about time.

Test Plan:
----------

Green CI


packages/cli/src/commands/upgrade/upgrade.ts
```diff
@@ -98,10 +93,14 @@ const getPatch = async (currentVersion, newVersion, config) => {
...
-const getVersionToUpgradeTo = async (argv, currentVersion, projectDir) => {
+const getVersionToUpgradeTo = async (
+  argv: Array<string>,
+  currentVersion: string,
+  projectDir: string,
+) => {
   const newVersion = argv[0]
     ? semver.valid(argv[0]) ||
-      (semver.coerce(argv[0]) ? semver.coerce(argv[0]).version : null)
+      (semver.coerce(argv[0]) ? semver.coerce(argv[0])!.version : null)
```
thymikee: This looks like a bug in TS, `semver.coerce(argv[0])` is clearly not null

package.json
```diff
@@ -47,6 +47,9 @@
     "micromatch": "^3.1.10",
     "mkdirp": "^0.5.1",
     "string-length": "^2.0.0",
-    "typescript": "^3.4.5"
+    "typescript": "^3.6.2"
+  },
+  "devDependencies": {
+    "@babel/plugin-proposal-class-properties": "^7.5.5"
```
Esemesek: Why we need it?

thymikee: tests are failing without this transform, not yet sure why

packages/cli/src/commands/upgrade/upgrade.ts
Esemesek: Not really. You assumed that `semver.coerce` is a pure function. If it had side effects it could return null second time. I am pretty sure if you assigned `semver.coerce(argv[0])` to a variable and then reuse everything would be okay.

thymikee: Fair point

packages/cli/src/commands/upgrade/__tests__/upgrade.test.ts
```diff
@@ -215,18 +208,22 @@ test('fetches regular patch, adds remote, applies patch, installs deps, removes
...
+      },
+    ),
   ).toMatchSnapshot(
     'RnDiffApp is replaced with app name (TestApp and com.testapp)',
   );
 }, 60000);
 test('fetches regular patch, adds remote, applies patch, installs deps, removes remote when updated from nested directory', async () => {
   mockFetch(samplePatch, 200);
-  (execa: any).mockImplementation(mockExecaNested);
+  ((execa as unknown) as jest.Mock).mockImplementation(mockExecaNested);
```
Esemesek: I usually make it `((execa as any) as jest.Mock)`, which might be a little bit smaller. It doesn't really matter here :P

PR #687, opened 2019-09-09
[#683] Convert merge.js to typescript

Naturalclar:
Summary:
---------

Part of #683 
Converted `commands/tools/merge.js` to typescript

Test Plan:
----------

Expect tests for `getDependencyConfig`, which uses `merge` to pass


packages/cli/src/tools/merge.ts
```diff
@@ -0,0 +1,15 @@
+import deepmerge from 'deepmerge';
```
Esemesek: Really weird that `git` haven't picked up this change as a `mv`.

packages/cli/src/tools/merge.ts
```diff
@@ -0,0 +1,15 @@
...
+
+/**
+ * `deepmerge` concatenates arrays by default instead of overwriting them.
+ * We define custom merging function for arrays to change that behaviour
+ */
+export default function merge(
```
Esemesek: There should be a better way to `infer` those parameters. For now, I think it is okay to leave it because it might cause errors in other parts (especially those which are not migrated to TS yet).

PR #685, opened 2019-09-09
chore: migrate info command to TS

FLGMwt:
Summary:
---------

@thymikee I cribbed your ignore comments from #683. If this is the right approach, I'll claim a few more.

Test Plan:
----------

Green CI

PR #688, opened 2019-09-10
Typescriptify releasechecker

FLGMwt:
Builds on #685, that should be merged first

Summary:
---------

Migrate releaseChecker, yarn, walk, and replacePathSepForRegex to TS.

Test Plan:
----------

Green CI

### DIFF
--- a/packages/cli/src/commands/info/info.ts
+++ b/packages/cli/src/commands/info/info.ts
@@ -9,7 +9,6 @@
 import envinfo from 'envinfo';
 import {logger} from '@react-native-community/cli-tools';
 import {Config} from '@react-native-community/cli-types';
-// @ts-ignore - JS file
 import releaseChecker from '../../tools/releaseChecker';
 
 const info = async function getInfo(_argv: Array<string>, ctx: Config) {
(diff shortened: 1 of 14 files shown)


packages/cli/src/commands/info/info.ts
```diff
@@ -3,20 +3,15 @@
  *
  * This source code is licensed under the MIT license found in the
  * LICENSE file in the root directory of this source tree.
- *
- * @flow
  */
 
+// @ts-ignore untyped
```
Esemesek: `envinfo` is untyped? :<

packages/cli/src/tools/releaseChecker/releaseCacheManager.ts
```diff
@@ -42,17 +40,18 @@ function saveCache(name: string, cache: Cache) {
 function getCacheRootPath() {
   const cachePath = path.resolve(os.homedir(), '.react-native-cli', 'cache');
   if (!fs.existsSync(cachePath)) {
-    mkdirp(cachePath);
+    mkdirp(cachePath, () => {});
```
Esemesek: Just wondering @thymikee. Shouldn't we use `mkdirp.sync` here? Also, it seemes `@types/mkdirp` aren't accurate because there is no need for noop function (https://github.com/substack/node-mkdirp/blob/master/index.js).

thymikee: Yup, we should use `mkdirp.sync` here

FLGMwt:
Thanks for the follow up!

PR #689, opened 2019-09-10
chore: migrate config command to TS

netochaves:
Summary:
---------

Part of #683 
Converted `commands/config` to typescript

Test Plan:
----------
Green CI

### DIFF
--- a/packages/cli/src/commands/config/config.ts
+++ b/packages/cli/src/commands/config/config.ts
@@ -1,9 +1,6 @@
-/**
- * @flow
- */
-import {type ConfigT} from 'types';
+import {Config, Dependency} from '@react-native-community/cli-types';
 
-function isValidRNDependency(config) {
+function isValidRNDependency(config: Dependency) {
   return (
     Object.keys(config.platforms).filter(key => Boolean(config.platforms[key]))
       .length !== 0 ||
@@ -13,7 +10,7 @@ function isValidRNDependency(config) {
   );
 }
 
-function filterConfig(config) {
+function filterConfig(config: Config) {
   const filtered = {...config};
   Object.keys(filtered.dependencies).forEach(item => {
     if (!isValidRNDependency(filtered.dependencies[item])) {
@@ -26,7 +23,7 @@ function filterConfig(config) {
 export default {
   name: 'config',
   description: 'Print CLI configuration',
-  func: async (argv: string[], ctx: ConfigT) => {
+  func: async (_argv: string[], ctx: Config) => {
     console.log(JSON.stringify(filterConfig(ctx), null, 2));
   },
 };
--- a/packages/cli-types/src/index.ts
+++ b/packages/cli-types/src/index.ts
@@ -74,6 +74,22 @@ interface PlatformConfig<
   };
 }
 
+export interface Dependency {
+  name: string;
+  root: string;
+  platforms: {
+    android?: AndroidDependencyConfig | null;
+    ios?: IOSDependencyConfig | null;
+    [key: string]: any;
+  };
+  assets: string[];
+  hooks: {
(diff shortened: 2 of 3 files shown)


packages/cli/src/commands/config/config.ts
```diff
@@ -1,9 +1,6 @@
-/**
- * @flow
- */
-import {type ConfigT} from 'types';
+import {Config} from '@react-native-community/cli-types';
 
-function isValidRNDependency(config) {
+function isValidRNDependency(config: any) {
```
Esemesek: I think this config can be typed as:
```ts
config: AndroidDependencyConfig | IOSDependencyConfig | null
```

packages/cli/src/commands/config/config.ts
```diff
@@ -26,7 +23,7 @@ function filterConfig(config) {
 export default {
   name: 'config',
   description: 'Print CLI configuration',
-  func: async (argv: string[], ctx: ConfigT) => {
+  func: async (ctx: Config) => {
```
Esemesek: What happened to the first argument of the function? Have you made sure everything works as expected with this change?

thymikee: `Config` is a second argument. This change breaks that. If `argv` is not used, prepend it with an underscore: `_argv`

netochaves: OK. I din't notice that, I'll fix it.

thymikee:
Looks like TS is failing

packages/cli/src/commands/config/config.ts
netochaves: I'm having trouble to set the type of this parameter, I try this approach but it's not working

packages/cli/src/commands/config/config.ts
```diff
@@ -1,9 +1,12 @@
...
-import {type ConfigT} from 'types';
+import {
+  Config,
+  AndroidDependencyConfig,
+  IOSDependencyConfig,
+} from '@react-native-community/cli-types';
 
-function isValidRNDependency(config) {
+function isValidRNDependency(
+  config: AndroidDependencyConfig | IOSDependencyConfig | null,
```
assuncaocharles: Just an idea, but maybe we want to the Dependency in ```@react-native-community/cli-types``` as an interface like:

```typescript
export interface Dependency {
  [key: string]: {
    name: string;
    root: string;
    platforms: {
      android?: AndroidDependencyConfig | null;
      ios?: IOSDependencyConfig | null;
      [key: string]: any;
    };
    assets: string[];
    hooks: {
      prelink?: string;
      postlink?: string;
    };
    params: InquirerPrompt[];
  };
}
```
Don't forgetting to update the Config interface to ```dependencies: Dependacy; ``` and here we can use:

```typescript
import {
  Dependency
} from '@react-native-community/cli-types';

function isValidRNDependency(
  config: Dependency
) {
  return (
    Object.keys(config.platforms).filter(key => Boolean(config.platforms[key]))
      .length !== 0 ||
    (config.hooks && Object.keys(config.hooks).length !== 0) ||
    (config.assets && config.assets.length !== 0) ||
    (config.params && config.params.length !== 0)
  );
}
```

It would solve most of the problems breaking in this file

thymikee: Yea, makes sense to extract this. We used this type in Flow files: https://github.com/react-native-community/cli/blob/fde227d8867e921f6df500c448ab49250f3bcfb6/types/index.js#L172-L175
@Esemesek can you take it over?

netochaves: Done.

assuncaocharles: Nice!

PR #690, opened 2019-09-10
[#683] convert copyFiles from flow to typescript

sasurau4:
Summary:
---------

Part of #683
Converted `commands/tools/copyFiles.js` to typescript

Test Plan:
----------

- `yarn test`


Esemesek:
@sasurau4 Can you fix TS errors?

PR #691, opened 2019-09-10
chore: convert assertRequiredOptions to TS

konojunya:
Summary:
---------

Part of #683 
Converted `commands/tools/assertRequiredOptions.js` to typescript

Test Plan:
----------

`yarn test`

### DIFF
--- a/packages/cli/src/tools/assertRequiredOptions.ts
+++ b/packages/cli/src/tools/assertRequiredOptions.ts
@@ -6,14 +6,16 @@
  *
  * @format
  */
-
 import {Option} from 'commander';
 import {camelCase} from 'lodash';
 
 // Commander.js has a 2 years old open issue to support <...> syntax
 // for options. Until that gets merged, we run the checks manually
 // https://github.com/tj/commander.js/issues/230
-export default function assertRequiredOptions(options, passedOptions) {
+export default function assertRequiredOptions(
+  options: any[],
+  passedOptions: any,
+) {
   options.forEach(opt => {
     const option = new Option(opt.name);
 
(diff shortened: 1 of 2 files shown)


packages/cli/src/tools/assertRequiredOptions.ts
```diff
@@ -6,14 +6,16 @@
...
-
 import {Option} from 'commander';
 import {camelCase} from 'lodash';
 
 // Commander.js has a 2 years old open issue to support <...> syntax
 // for options. Until that gets merged, we run the checks manually
 // https://github.com/tj/commander.js/issues/230
-export default function assertRequiredOptions(options, passedOptions) {
+export default function assertRequiredOptions(
+  options: any[],
```
thymikee: use `Command['options']`. You'll find `Command` in `@react-native-community/cli-types`

packages/cli/src/tools/assertRequiredOptions.ts
```diff
@@ -6,14 +6,16 @@
...
+  passedOptions: any,
```
thymikee: let's make this a `Record<string, unknown>` or something similar, but not `any` :)

konojunya:
@thymikee fixup :)

packages/cli/src/tools/assertRequiredOptions.ts
```diff
@@ -6,14 +6,22 @@
...
  * @format
  */
-
 import {Option} from 'commander';
 import {camelCase} from 'lodash';
+import {Command} from '@react-native-community/cli-types';
+
+type Options = NonNullable<Command['options']>;
+type PassedOptions = {
+  [x: string]: unknown;
```
thymikee: Let's be descriptive where we can
```suggestion
  [key: string]: unknown;
```

PR #692, opened 2019-09-10
[#683] Convert commands/packageManager.ts to TS

moriyuu:
Summary:
---------

Part of #683
Converted `commands/packageManager.ts` to typescript

Test Plan:
----------

`yarn test`

### DIFF
--- a/packages/cli/src/tools/packageManager.ts
+++ b/packages/cli/src/tools/packageManager.ts
@@ -1,16 +1,15 @@
 // @flow
 import execa from 'execa';
 import {logger} from '@react-native-community/cli-tools';
-// $FlowFixMe - converted to TS
 import {getYarnVersionIfAvailable, isProjectUsingYarn} from './yarn';
 
-type Options = {|
-  preferYarn?: boolean,
-  silent?: boolean,
-  cwd?: string,
-|};
+type Options = {
+  preferYarn?: boolean;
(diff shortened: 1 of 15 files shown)


packages/cli/src/tools/packageManager.ts
```diff
@@ -1,16 +1,15 @@
 // @flow
```
thymikee: Please remove Flow annotation

packages/cli/src/tools/packageManager.ts
```diff
@@ -29,8 +27,8 @@ const packageManagers = {
 
 function configurePackageManager(
   packageNames: Array<string>,
-  options?: Options,
   action: 'install' | 'installDev' | 'installAll' | 'uninstall',
+  options?: Options,
```
thymikee: @Esemesek you ok with that? Makes sense imho

thymikee:
Please make sure lint passes

moriyuu:
@thymikee Should I convert `tools/__tests__/packageManager-test.js` too?

thymikee:
yes please!

packages/cli/src/tools/packageManager.ts
Esemesek: Yup. It makes a lot of sense.

PR #693, opened 2019-09-10
chore: convert findSymlinkedModules to TS

pveyes:
Summary:
---------

Part of #683. 

Test Plan:
----------

`yarn flow-check && yarn test`

### DIFF
--- a/packages/cli/src/tools/loadMetroConfig.js
+++ b/packages/cli/src/tools/loadMetroConfig.js
@@ -3,10 +3,10 @@
  * @flow
  */
 import path from 'path';
-import { createBlacklist } from 'metro';
-import { loadConfig } from 'metro-config';
-import { existsSync } from 'fs';
-import { type ConfigT } from 'types';
+import {createBlacklist} from 'metro';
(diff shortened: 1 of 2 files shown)


packages/cli/src/tools/loadMetroConfig.js
```diff
@@ -3,10 +3,11 @@
  * @flow
  */
 import path from 'path';
-import {createBlacklist} from 'metro';
-import {loadConfig} from 'metro-config';
-import {existsSync} from 'fs';
-import {type ConfigT} from 'types';
+import { createBlacklist } from 'metro';
```
thymikee: Please revert these. Looks like you have different Prettier settings. Running `yarn lint --fix` will be helpful

pveyes: yes sorry this is from my vscode setting. Just pushed an update after running `lint --fix` but turns out it updates `findSymlinkedModules.ts` as well. Is this intended?

thymikee: If ESLint says so, then it is intended

PR #695, opened 2019-09-10
chore: convert bundle command to TS

Quadriphobs1:
Summary:
---------

Part of #683
Converted `commands/bundle/` to typescript

Test Plan:
----------

`yarn test`


packages/cli/src/commands/bundle/assetPathUtils.ts
```diff
@@ -5,14 +5,13 @@
  * LICENSE file in the root directory of this source tree.
  *
  * @format
```
thymikee: you can remove format as well

packages/cli/src/commands/bundle/buildBundle.ts
```diff
@@ -4,25 +4,45 @@
...
-
+// @ts-ignore - no typed definition for the package
 import outputBundle from 'metro/src/shared/output/bundle';
 import path from 'path';
 import chalk from 'chalk';
-import type {CommandLineArgs} from './bundleCommandLineArgs';
-import type {ConfigT} from 'types';
+import {CommandLineArgs} from './bundleCommandLineArgs';
+// @ts-ignore FIXME after converting types
+import {ConfigT} from 'types';
```
thymikee:
```suggestion
import {Config} from '@react-native-community/cli-types';
```
Remove the above ts-ignore as well

packages/cli/src/commands/bundle/buildBundle.ts
```diff
@@ -4,25 +4,45 @@
...
+  height: number | null;
+  httpServerLocation: string;
+  name: string;
+  scales: number[];
+  type: string;
+  width: number | null;
+  files: string[],
+};
+
+async function buildBundle(args: CommandLineArgs, ctx: ConfigT, output: outputBundle) {
```
thymikee:
```suggestion
async function buildBundle(args: CommandLineArgs, ctx: Config, output: outputBundle) {
```

packages/cli/src/commands/bundle/buildBundle.ts
```diff
@@ -61,7 +81,7 @@ async function buildBundle(
     sourceMapUrl,
     dev: args.dev,
     minify: args.minify !== undefined ? args.minify : !args.dev,
-    platform: args.platform,
```
thymikee: Let's revert this change, since we use `arg.platform` everywhere in this function

packages/cli/src/commands/bundle/bundle.ts
```diff
@@ -6,14 +6,16 @@
  *
  * @format
  */
-
+// @ts-ignore FIXME after converting types
+import {ConfigT} from 'types';
```
thymikee: same as marked above

packages/cli/src/commands/bundle/bundle.ts
```diff
@@ -6,14 +6,16 @@
...
+import {CommandLineArgs} from './bundleCommandLineArgs';
 import buildBundle from './buildBundle';
 import bundleCommandLineArgs from './bundleCommandLineArgs';
 
 /**
  * Builds the bundle starting to look for dependencies at the given entry path.
  */
-function bundleWithOutput(_, config, args, output) {
+function bundleWithOutput(_: any, config: ConfigT, args: CommandLineArgs, output: any) {
```
thymikee:
```suggestion
function bundleWithOutput(_: Array<string>, config: Config, args: CommandLineArgs, output: any) {
```
You can find the interface of `output` at `metro/src/shared/output/RamBundle`

packages/cli/src/commands/bundle/ramBundle.ts
```diff
@@ -6,15 +6,18 @@
...
  * @format
  */
+// @ts-ignore - no typed definition for the package
 import outputUnbundle from 'metro/src/shared/output/RamBundle';
-
+import {CommandLineArgs} from './bundleCommandLineArgs';
 import {withOutput as bundleWithOutput} from './bundle';
 import bundleCommandLineArgs from './bundleCommandLineArgs';
+// @ts-ignore FIXME after converting types
+import {ConfigT} from 'types';
```
thymikee: fix import to cli-types

packages/cli/src/commands/bundle/saveAssets.ts
```diff
@@ -43,15 +48,15 @@ function saveAssets(assets, platform, assetsDest) {
...
+function copyAll(filesToCopy: CopiedFiles) {
   const queue = Object.keys(filesToCopy);
   if (queue.length === 0) {
     return Promise.resolve();
   }
 
   logger.info(`Copying ${queue.length} asset files`);
   return new Promise((resolve, reject) => {
-    const copyNext = error => {
+    const copyNext = (error?: any) => {
```
thymikee:
```suggestion
    const copyNext = (error?: NodeJS.ErrnoException) => {
```

packages/cli/src/commands/bundle/saveAssets.ts
```diff
@@ -60,7 +65,7 @@ function copyAll(filesToCopy) {
         logger.info('Done copying assets');
         resolve();
       } else {
-        const src = queue.shift();
+        const src = queue.shift() || '';
```
thymikee: we should bail out of copying in such case, instead of making source empty

packages/cli/src/commands/bundle/buildBundle.ts
Quadriphobs1: Its has been assigned on line 48  
```
const platform: string = args.platform || '';
```

thymikee: That's what I refer to. It's only assigned there and used here, but other parts of this function rely on `args.platform` directly. No point in changing that for now :)

packages/cli/src/commands/bundle/buildBundle.ts
```diff
@@ -72,13 +94,14 @@ async function buildBundle(
...
     // Save the assets of the bundle
-    const outputAssets = await server.getAssets({
+    const outputAssets: AssetData[] = await server.getAssets({
       ...Server.DEFAULT_BUNDLE_OPTIONS,
       ...requestOpts,
       bundleType: 'todo',
     });
 
     // When we're done saving bundle output and the assets, we're done.
+    // @ts-ignore
```
thymikee: why ts ignore? can you leave a comment explaining?

packages/cli/src/commands/bundle/bundle.ts
```diff
@@ -4,16 +4,21 @@
  * This source code is licensed under the MIT license found in the
  * LICENSE file in the root directory of this source tree.
  *
- * @format
  */
-
+import {Config} from '@react-native-community/cli-types';
+import {CommandLineArgs} from './bundleCommandLineArgs';
 import buildBundle from './buildBundle';
 import bundleCommandLineArgs from './bundleCommandLineArgs';
```
thymikee: You can remove line 9 nad move the import here
```suggestion
import bundleCommandLineArgs, {CommandLineArgs} from './bundleCommandLineArgs';
```

packages/cli/src/commands/bundle/getAssetDestPathAndroid.ts
```diff
@@ -4,12 +4,10 @@
...
  *
- * @format
- * @flow strict
  */
 
 import path from 'path';
-import type {PackagerAsset} from './assetPathUtils';
+import {PackagerAsset} from './assetPathUtils';
 
 import assetPathUtils from './assetPathUtils';
```
thymikee: please consolidate imports :) 
```suggestion
import assetPathUtils, {PackagerAsset} from './assetPathUtils';
```

packages/cli/src/commands/bundle/ramBundle.ts
```diff
@@ -4,17 +4,18 @@
...
  * LICENSE file in the root directory of this source tree.
  *
- * @format
  */
+// @ts-ignore - no typed definition for the package
 import outputUnbundle from 'metro/src/shared/output/RamBundle';
-
+import {CommandLineArgs} from './bundleCommandLineArgs';
 import {withOutput as bundleWithOutput} from './bundle';
 import bundleCommandLineArgs from './bundleCommandLineArgs';
```
thymikee:
```suggestion
import bundleCommandLineArgs, {CommandLineArgs} from './bundleCommandLineArgs';
```

Quadriphobs1: My bad 🙈

packages/cli/src/commands/bundle/buildBundle.ts
Quadriphobs1: the value `args.platform` is actually seen as `string | undefined` however in the function `saveAssets()` it is required which causes type error... looking at the code it looks like the `args.platform` would be provided at all times which makes no sense to actually make it optional from the code as well which is why I suppress the warning but what do you think of it

thymikee: Definitely shouldn't be suppressed. This is possible failure point and we need to make sure args.platform is defined before passing it further

Quadriphobs1: I just checked in now the `args.platform` has a default value of `ios` which means making it options is actually not necessary

packages/cli/src/commands/bundle/bundleCommandLineArgs.ts
```diff
@@ -4,29 +4,28 @@
...
-};
+export interface CommandLineArgs {
+  assetsDest?: string;
+  entryFile: string;
+  resetCache: boolean;
+  resetGlobalCache: boolean;
+  transformer?: string;
+  minify?: boolean;
+  config?: string;
+  platform?: string;
```
thymikee: We can change it to 'platform: string' then, if it has a default value :)

packages/cli/src/commands/bundle/buildBundle.ts
```diff
@@ -4,24 +4,46 @@
...
 import path from 'path';
 import chalk from 'chalk';
-import type {CommandLineArgs} from './bundleCommandLineArgs';
-import type {ConfigT} from 'types';
+import {CommandLineArgs} from './bundleCommandLineArgs';
+import {Config} from '@react-native-community/cli-types';
 import saveAssets from './saveAssets';
-import loadMetroConfig from '../../tools/loadMetroConfig';
+// @ts-ignore FIXME after converting tools/loadMetroConfig
+import loadMetroConfig from '../../tools/loadMetroConfig'; // eslint-disable-line import/namespace, import/default
```
thymikee: you can remove the ignores after a rebase :)

thymikee:
Thank you!

PR #696, opened 2019-09-10
chore: migrated link command to TS

anikethsaha:
Summary:
---------

https://github.com/react-native-community/cli/issues/683
Converted `link` command to `Typescript`
Also added `preulink` , `postunlink` to the `cli-types` in the config hooks 

Test Plan:
----------
- `yarn test`
- `yarn lint`


packages/cli-types/src/index.ts
```diff
@@ -106,6 +106,8 @@ export interface Config {
       hooks: {
         prelink?: string;
         postlink?: string;
+        preulink?: string;
```
FLGMwt: I think this was a typo in the command itself, should probably be `preunlink`

packages/cli/src/commands/link/__tests__/link-test.js
```diff
@@ -1,7 +1,7 @@
 import {func as link} from '../link';
 import loadConfig from '../../../tools/config';
 import makeHook from '../makeHook';
-
+// $FlowFixMe - converted to TS
```
FLGMwt: We can migrate the `link-test` and `makeHook-test` to ts as well. We shouldn't need the `FlowFixMe` then

packages/cli/src/commands/link/getPlatformName.ts
```diff
@@ -0,0 +1,7 @@
+type names = {[key: string]: string};
+const names: names = {
```
FLGMwt: Having the type with the exact same name as the variable is a bit confusing. Maybe inline it instead?:

```suggestion
const names: {[key: string]: string} = {
```

packages/cli/src/commands/link/linkAssets.ts
```diff
@@ -1,15 +1,12 @@
...
 import {logger} from '@react-native-community/cli-tools';
 
 const linkAssets = (
-  platforms: PlatformsT,
-  project: ProjectConfigT,
+  platforms: Config['platforms'],
+  project: Config['project'],
   assets: Array<string>,
-) => {
+): void | null => {
```
FLGMwt: Hrm, I don't see the `null` return. can this just be void?

packages/cli-types/src/index.ts
Esemesek: Yup. It seems like this is the case.

anikethsaha: yeah my mistake. Will fix it soon !

packages/cli/src/commands/link/__tests__/link-test.js
anikethsaha: I tried migrating but I was getting some weird unrelated errors mostly with namespace... I will try once more then !

packages/cli/src/commands/link/getPlatformName.ts
anikethsaha: yeah ...this suggestion looks neat ! 👍

packages/cli/src/commands/link/linkAssets.ts
anikethsaha: https://github.com/react-native-community/cli/blob/master/packages/cli/src/commands/link/linkAssets.js#L14 
In this line its `return` thats it....so will it be covered with `void`  ?

thymikee: When a function returns nothing (either by specifcyting `return;` or, well, not specifying it), it implicitly returns `undefined`. `void` in a type def is equivalent to `undefined`, but it's also an expression in JS, that always returns undefined: `void 1 === undefined`, `void 'text' === undefined`.

packages/cli/src/commands/link/__tests__/link-test.js
anikethsaha: done !!!

packages/cli/src/commands/link/getPlatformName.ts
anikethsaha: done

packages/cli/src/commands/link/linkAssets.ts
anikethsaha: changed to void alone

packages/cli-types/src/index.ts
anikethsaha: done

anikethsaha:
Need Help with CI errors 
image

FLGMwt:
You can repro this w/ `yarn build:ts`. Looks like inquirer has a types package so you can do:

```
yarn workspace @react-native-community/cli add @types/inquirer
```

To fix the first error.

I poked into the platforms one a bit and that might have a long tail. One option would be to change `FlagsType.platforms` (in link and unlink) from `Array<'ios' | 'android'>` to `Array<'ios' | 'android'>`.

This would cause some issues with out of branch platforms (web, windows, etc) but there's a bit to unravel to support that with the types anyway.

The main issue is that we're using the `Config['platforms']` type here which includes `ios` and `android` properties. `pick` can possibly return an object that has neither. I'm not sure how to modify the main platforms type to fix this without having to export a bunch more types from the types package : /

FLGMwt:
Wow, and by "a bunch more types" I actually meant 2.

Looks like this worked out: https://github.com/react-native-community/cli/commit/2eaeb70b4d818c206546a3850f9df368df7edf57

anikethsaha:
I am getting a good number of errors when adding the type package for inquirer. 

> FlagsType.platforms (in link and unlink) from Array<'ios' | 'android'> to Array<'ios' | 'android'>

A bit confused with this one. Where do I get those ios and android type ! 

> Looks like this worked out: 2eaeb70

Will rebase it and try it

assuncaocharles:
> You can repro this w/ `yarn build:ts`. Looks like inquirer has a types package so you can do:
> 
> ```
> yarn workspace @react-native-community/cli add @types/inquirer
> ```
> ....

Yes, I think it's missing the `@types/inquirer` and after installing it some things should change. e.g:

```typescript
import {prompt, QuestionCollection} from 'inquirer';

export default (questions: QuestionCollection) =>
  new Promise<any>((resolve, reject) => {
    if (!questions) {
      resolve({});
      return;
    }

    prompt(questions).then(resolve, reject);
  });
```

anikethsaha:
> Looks like this worked out: [2eaeb70](https://github.com/react-native-community/cli/commit/2eaeb70b4d818c206546a3850f9df368df7edf57)

Rebasing will work right ! @FLGMwt

anikethsaha:
image

@assuncaocharles I tried installing the `@types/inquirer` and `@types/node` , still getting that error

Command used `yarn workspace @react-native-community/cli add @types/inquirer @types/node`

assuncaocharles:
> image
> 
> @assuncaocharles I tried installing the `@types/inquirer` and `@types/node` , still getting that error
> 
> Command used `yarn workspace @react-native-community/cli add @types/inquirer @types/node`
> 
> PS:
> I even got both `node` and `inquirer` in my `node_modules`
> 
> image

That's weird, I reproduced it locally and it worked well.. Try to run `yarn add @types/inquirer @types/node` from inside `packages/cli/src/commands`

anikethsaha:
from `packages/cli/src/commands` or from `packages/cli` ? I guess from `packages/cli/` right ?
I tried....no luck

FLGMwt:
Can you push you branch to so we can take a look? You should have `@types/node` in your `packages/cli/package.json `.

As for the `export` changes from my branch, if you're having issues with rebasing the changes, just copy over the changes to your branch. There should be changes in the following files:

- packages/cli-types/src/index.ts
- packages/platform-android/src/link/warnAboutManuallyLinkedLibs.ts
- packages/platform-ios/src/link/warnAboutManuallyLinkedLibs.ts

anikethsaha:
@FLGMwt @assuncaocharles I updated my fork, https://github.com/anikethsaha/cli/tree/migration/typescript-link-command
here it is!

anikethsaha:
> As for the export changes from my branch, if you're having issues with rebasing the changes, just copy over the changes to your branch. There should be changes in the following files:

I manually copy pasted it......IDK why the rebasing didnt work !

FLGMwt:
I played around and I don't understand why, but adding an explicit version of @types/node made it work. I think it's something with how typescript resolved which type package to use for deps? It's probably the transitive @types/node@8 throwing things off but I have no idea. Give this a shot:

```
 yarn workspace @react-native-community/cli add @types/node@^12.7.5
```

Diff should look like:

image

anikethsaha:
> I played around and I don't understand why, but adding an explicit version of @types/node made it work. I think it's something with how typescript resolved which type package to use for deps? It's probably the transitive @types/node@8 throwing things off but I have no idea. Give this a shot:

hmm...yeah this is a weird behavior.

I will give it a shot !

thymikee:
We use `@types/node@8` in root config to make sure we use APIs that are compatible with the target we compile to. Does inquirer has issues with older typings as well? cc @Esemesek halp pls

anikethsaha:
>  Does inquirer has issues with older typings as well? 

I guess typescript is not able to resolve older versions.!

FLGMwt:
> We use `@types/node@8` in root config to make sure we use APIs that are compatible with the target we compile to. Does inquirer has issues with older typings as well? cc @Esemesek halp pls

Ah, that makes sense, sorry for the confusion. Seems like it used to be called `ReadLine` instead of `Interface`: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/v8/base.d.ts#L1850.

Latest types have `Interface` re-exported as `ReadLine` so maybe `@types/inquirer` could use that instead? https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/readline.d.ts#L110

Unsure what to do in `cli` though other than mark as untyped to get this in.

anikethsaha:
Should I downgrade the version then ?

anikethsaha:
should I change the `@types/node` to `v8` ?

thymikee:
Lint is failing :)

anikethsaha:
Yeah I am not sure about the fix of those error.
Need Help with it !

thymikee:
"Unused eslint-disable directive" means you can safely remove it from that place. Also, please rebase to latest master because of recent conflicts

anikethsaha:
> "Unused eslint-disable directive" means you can safely remove it from that place.

means I can remove the import statement !?

thymikee:
You can remove the "eslint-disable-line" or similar

anikethsaha:
```
/home/circleci/react-native-cli/packages/cli/src/commands/link/__tests__/link-test.ts
  2:24  error  Parse errors in imported module '../../../tools/config': '=' expected. (16:12)  import/namespace
  2:24  error  Parse errors in imported module '../../../tools/config': '=' expected. (16:12)  import/default
```
ohh and about the lint error in the test file ? should I change it to `require` ?

thymikee:
Add `// eslint-disable-line import/namespace, import/default` as in other files.

anikethsaha:
okay !!!!

anikethsaha:
To fix the current errors , `cli-types/src/index.js` at https://github.com/react-native-community/cli/pull/696/commits/72f5a0e9411168863ce01d2c11cbe2e37e2a447a will fix it !....
But not  sure as it may break files from https://github.com/react-native-community/cli/pull/689

PR #697, opened 2019-09-10
chore: convert generator files into TS

KangYoosam:
## Summary:
Part of https://github.com/react-native-community/cli/issues/683
Converted `commands/tools/generator/*` to typescript

## Test Plan:
- yarn test
- yarn flow-check
- yarn lint

### DIFF
--- a/packages/cli/src/tools/generator/promptSync.ts
+++ b/packages/cli/src/tools/generator/promptSync.ts
@@ -43,11 +43,13 @@ function create() {
     const echo = opts.echo;
     const masked = 'echo' in opts;
 
-    let fd = fs.openSync('/dev/tty', 'rs');
-    if (process.platform === 'win32') {
-      // @ts-ignore MEMO: fd exists when win32
-      fd = process.stdin.fd;
-    }
+    /*eslint-disable prettier/prettier*/
+    const fd =
+      process.platform === 'win32'
+        // @ts-ignore
+        ? process.stdin.fd
+        : fs.openSync('/dev/tty', 'rs');
+    /*eslint-enable prettier/prettier*/
 
     const wasRaw = process.stdin.isRaw;
     if (!wasRaw && process.stdin.setRawMode) {


packages/cli/src/tools/generator/promptSync.ts
```diff
@@ -12,33 +12,45 @@
...
     const masked = 'echo' in opts;
 
-    const fd =
-      process.platform === 'win32'
-        ? process.stdin.fd
-        : fs.openSync('/dev/tty', 'rs');
+    let fd = fs.openSync('/dev/tty', 'rs');
+    if (process.platform === 'win32') {
+      // @ts-ignore MEMO: fd exists when win32
+      fd = process.stdin.fd;
```
thymikee: Can we bring back previous behavior? No need to reassign, a conditional assignment per platform is ok

packages/cli/src/tools/generator/templates.ts
```diff
@@ -109,7 +108,10 @@ async function createFromRemoteTemplate(
   }
 }
 
-async function installTemplateDependencies(templatePath, destinationRoot) {
+async function installTemplateDependencies(
+  templatePath: string,
+  _destinationRoot: string,
```
thymikee: please remove since it's not used

packages/cli/src/tools/generator/templates.ts
```diff
@@ -135,7 +137,10 @@ async function installTemplateDependencies(templatePath, destinationRoot) {
   execSync('react-native link', {stdio: 'inherit'});
 }
 
-async function installTemplateDevDependencies(templatePath, destinationRoot) {
+async function installTemplateDevDependencies(
+  templatePath: string,
+  _destinationRoot: string,
```
thymikee: please remove as well

packages/cli/src/tools/generator/promptSync.ts
```diff
@@ -12,33 +12,45 @@
...
+  function prompt(
+    ask?: string | Options,
+    value?: string | Options,
+    opts?: Options,
+  ) {
     let insert = 0;
     opts = opts || {};
 
-    if (Object(ask) === ask) {
+    if (typeof ask === 'object') {
```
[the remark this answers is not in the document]
KangYoosam: When I do so, It causes ts error 😭 
image

packages/cli/src/tools/generator/promptSync.ts
KangYoosam: image

When I use Typescript, it saids ts error occurred.
And When I `@ts-ignore` it, It also tells error since we can not use `@ts-ignore` in ternary operator. 😭 
image

packages/cli/src/tools/generator/promptSync.ts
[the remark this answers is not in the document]
thymikee: Ok, let's leave as is, since there's no `null` there

packages/cli/src/tools/generator/promptSync.ts
thymikee: What about
```ts
let fd;
if (win32) { something; } else { somethingElse; }
```

thymikee: The thing is I'd like to avoid touching the code in tools/generator too much, because it works and will be removed in a year or so, because we'll get rid of legacy init.

KangYoosam: @thymikee Thanks for comment!
I brought back previous behavior and make it work!

changes are in ↓ 👍 
8f83905

Esemesek:
Seems okay, but we should probably try legacy init flow before merging it. Will test things out tomorrow.

PR #698, opened 2019-09-10
chore: converted install command to typescript

Quadriphobs1:
Summary:
---------

Part of #683
Converted commands/install/* to typescript

Test Plan:
yarn test

### DIFF
--- a/packages/cli/src/commands/install/install.ts
+++ b/packages/cli/src/commands/install/install.ts
@@ -5,9 +5,7 @@
  * LICENSE file in the root directory of this source tree.
  *
  */
-/* eslint-disable import/namespace, import/default */
 import {logger} from '@react-native-community/cli-tools';
-// @ts-ignore FIXME after converting tools/packageManager
 import * as PackageManager from '../../tools/packageManager';
 // @ts-ignore FIXME after converting link/link
 import link from '../link/link';
--- a/packages/cli/src/commands/install/install.ts
+++ b/packages/cli/src/commands/install/install.ts
@@ -8,9 +8,9 @@
 import {logger} from '@react-native-community/cli-tools';
 import * as PackageManager from '../../tools/packageManager';
 // @ts-ignore FIXME after converting link/link
-import link from '../link/link';
+import link from '../link/link'; // eslint-disable-line import/namespace, import/default
 // @ts-ignore FIXME after converting tools/config
-import loadConfig from '../../tools/config';
+import loadConfig from '../../tools/config'; // eslint-disable-line import/namespace, import/default
 
 async function install(args: Array<string>): Promise<void> {
   const name = args[0];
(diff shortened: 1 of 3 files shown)


thymikee:
Can you rebase and make sure lint is passing? :)

packages/cli/src/commands/install/install.ts
```diff
@@ -4,16 +4,17 @@
  * This source code is licensed under the MIT license found in the
  * LICENSE file in the root directory of this source tree.
  *
- * @flow
  */
-
-import type {ConfigT} from 'types';
+/* eslint-disable import/namespace, import/default */
```
thymikee: Please don't do this for entire file, just where it's needed. We can remove these inline later when cleaning up

packages/cli/src/commands/install/install.ts
```diff
@@ -4,16 +4,17 @@
...
 import {logger} from '@react-native-community/cli-tools';
+// @ts-ignore FIXME after converting tools/packageManager
 import * as PackageManager from '../../tools/packageManager';
```
thymikee: packageManager is already converted, you need to rebase

packages/cli/src/commands/install/uninstall.ts
```diff
@@ -4,15 +4,16 @@
  * This source code is licensed under the MIT license found in the
  * LICENSE file in the root directory of this source tree.
  *
- * @flow
  */
-
-import type {ConfigT} from 'types';
+/* eslint-disable import/namespace, import/default */
```
thymikee: same as above

packages/cli/src/commands/install/uninstall.ts
```diff
@@ -4,15 +4,16 @@
...
+import {Config} from '@react-native-community/cli-types';
 import {logger} from '@react-native-community/cli-tools';
+// @ts-ignore FIXME after converting tools/packageManager
 import * as PackageManager from '../../tools/packageManager';
```
thymikee: packageManager is already converted

Quadriphobs1:
@thymikee  Can you check it now

packages/cli/src/commands/install/install.ts
thymikee: still there, please see my message above

thymikee:
The issues I brought are still there, please address them first.

packages/cli/src/commands/install/install.ts
Quadriphobs1: Removing it won't pass the test and the file `link/unlink` is not typed which is causing `import/namespace` typescript error

thymikee: All I ask for is to disable only the lines that need it. It's easy to forget about global disables, that's why we avoid them.

Quadriphobs1: +1

thymikee:
Thanks!

DanielUche:
wow some bit me to this task. I can take up any other task if any is available

thymikee:
@DanielUche sorry, I didn't notice it was assigned to somebody else. Currently we don't have any TS-related task left. But you're free to help us with regular issues and code reviews :)

Issue #694, opened 2019-09-10
Doctor command


lucasbento:
## Describe the Feature

This lists the work that we still need to do in order to get `react-native doctor` ready to ship.

⚠️ This list will be kept up-to-date with whatever issue I encounter. ⚠️

➡️ **If you want to work on any of this just comment what you want to work on to be assigned.**

<details>
  <summary>Specify why a health check is showing status as failed - #763</summary>

  Be more specific about the statuses:

  ```
	x Node.js Required version 10 is not installed (found Node.js 7)
	x ANDROID_HOME: The specified path doesn't exist
	x Android NDK: Required version 20 is not installed
  ```

   Suggestion by @satya164.
</details>

<details>
  <summary>Add tests (@thib92) - #728 #742</summary>
  
  Some stuff is pretty complicated on this flow and we should add tests before we forget how they were implemented 😅
</details>

### Done

<details>
  <summary>Change `Fix all issues` to `Try to fix issues automatically` - #763</summary>

  `Fix all issues` sounds too good to be true and sometimes a health check won't work, better to be safe on wording.

  Suggestion by @satya164.
</details>

<details>
  <summary>Agree on version ranges</summary>

  With the `doctor` command we will have to officially support specific version ranges, we need to agree on which ones those will be and how we will decide when to bump them, the current version ranges are listed [here](https://github.com/react-native-community/cli/blob/master/packages/cli/src/commands/doctor/versionRanges.ts).

  Keep in mind that I have put the version ranges that I think that should be supported based on my environment, all those versions are up for discussion.
</details>

<details>
  <summary>Errors being thrown should be consistent - #746</summary>

  Currently the errors that are being thrown are not consistent, for instance the `CocoaPods` installation error is much larger than `ios-deploy`, some of them also have indentation to match with the whole "`doctor`" template and some don't.

  An idea to fix this is to have helpers to log errors to the screen.
</details>

<details>
  <summary>Add description to each software that needs to be installed - #743</summary>

  Currently we show this:

  ![Current state](https://user-images.githubusercontent.com/6207220/64615864-a4576400-d3db-11e9-8e9b-6c07184c5553.png)

  It would be nicer to have a `description` field on each healthcheck so they can give better info about why it's needed.
</details>

<details>
  <summary>Don't show the "usage" section if everything is good - #729</summary>

  The usage section shouldn't be shown if there are no error nor warnings.
</details>

<details>
  <summary>Recognition of which package manager the user is using is not 100% - #726</summary>

[...]

PR #699, opened 2019-09-10
doctor: fix missing `loader.fail()` call before showing how to fix manually on `ios-deploy`

lucasbento:
Summary:
---------

This is part of the `react-native doctor` issue at #694.

It fixes a problem with the `ios-deploy` healthcheck throwing an error without actually putting a _final state_ in the loader, causing it to leak into the next healthcheck:

before

It also changes some stuff regarding brew installation to keep consistency of what is being shown to the user (e.g. Watchman showing as `Installing watchman` after you choose fix and the others showing without `Installing` prefix).

<details>
  <summary>Before</summary>

  before
</details>

<details>
  <summary>After</summary>

  after
</details>

### DIFF
--- a/packages/cli/src/commands/doctor/healthchecks/common.js
+++ b/packages/cli/src/commands/doctor/healthchecks/common.js
@@ -22,13 +22,13 @@ const logManualInstallation = ({
 
   if (url) {
     return logMessage(
-      `Read more about how to download ${healthcheck} at ${chalk.dim(url)}`,
+      `Read more about how to download ${healthcheck} at ${chalk.bold(url)}`,
     );
   }
 
   if (command) {
     return logMessage(
-      `Please install ${healthcheck} by running ${chalk.dim(command)}`,
+      `Please install ${healthcheck} by running ${chalk.bold(command)}`,
     );
   }
 };
--- a/packages/cli/src/commands/doctor/healthchecks/common.js
+++ b/packages/cli/src/commands/doctor/healthchecks/common.js
@@ -22,7 +22,9 @@ const logManualInstallation = ({
 
   if (url) {
     return logMessage(
-      `Read more about how to download ${healthcheck} at ${chalk.bold(url)}`,
+      `Read more about how to download ${healthcheck} at ${chalk.dim.underline(
+        url,
+      )}`,
     );
   }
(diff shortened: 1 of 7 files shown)


packages/cli/src/tools/brewInstall.ts
```diff
@@ -1,16 +1,27 @@
...
     await execa('brew', ['install', pkg]);
+
     loader.succeed();
   } catch (error) {
-    logger.log(error.stderr);
-    loader.fail(
-      `An error occured while trying to install ${pkg}. Please try again manually: brew install ${pkg}`,
+    loader.fail();
+    logger.log(`\n${error.stderr}`);
+    logger.log(
```
thymikee: shouldn't we use `logger.error` here?

lucasbento: I forgot that there's a `logger.error`, this is how it looks with it:

image

I would rather keep the `logger.log` because the `error` prefix here doesn't really add much, same goes for `Error: Command failed: brew install watchman` as it's already mentioned below.

thymikee: Ok, so let's maybe dim the output and highlight (chalk.bold) the brew install command

lucasbento: image

Like this? I like it.

thymikee: 👍

lucasbento: Cool, just pushed the change!

packages/cli/src/commands/doctor/healthchecks/common.js
```diff
@@ -22,13 +22,13 @@ const logManualInstallation = ({
 
   if (url) {
     return logMessage(
-      `Read more about how to download ${healthcheck} at ${chalk.dim(url)}`,
+      `Read more about how to download ${healthcheck} at ${chalk.bold(url)}`,
```
thymikee: We `chalk.dim.underline` the links, can we keep it consistent?

lucasbento: Good eye, fixed.

PR #708, opened 2019-09-10
chore: convert init command to TS

gvarandas:
Summary:
---------
Part of issue #683
Converted `commands/init/*` to typescript.

The conversion involved the following steps:
- [x] Renaming the files from `.js` to `.ts`;
- [x] Removing all the `@flow` annotations from the header;
- [x] Transforming the type definitions from flow syntax to typescript;
- [x] Adding explicit type definitions when needed (for most of the cases the inferred type sufficed);
- [x] Extracting some interfaces to explicit definitions to increase readability;

Test Plan:
----------
- [x] Running `yarn flow` and `yarn lint` to guarantee static analysis with no errors
- [x] Running `yarn test` to guarantee unit test coverage (this also involved making sure that the new snapshots are generated in the correct `.ts` format).
- [x] Creating a new project using the local `cli`, following the guidelines from the `CONTRIBUTING.md` file.


packages/cli/src/commands/init/__tests__/template.test.ts
```diff
@@ -22,7 +19,7 @@ afterEach(() => {
 });
 
 test('installTemplatePackage', async () => {
-  jest.spyOn(PackageManger, 'install').mockImplementationOnce(() => {});
+  jest.spyOn(PackageManger, 'install').mockImplementationOnce(() => null);
```
gvarandas: Changed the mock implementation to avoid type conflicts. There was no change in the overall outcome from the unit tests since we're not asserting the result from the mock.

packages/cli/src/commands/init/__tests__/template.test.ts
```diff
@@ -63,7 +60,7 @@ test('copyTemplate', async () => {
   const CWD = '.';
 
   jest.spyOn(path, 'resolve').mockImplementationOnce((...e) => e.join('/'));
-  jest.spyOn(copyFiles, 'default').mockImplementationOnce(() => {});
+  jest.spyOn(copyFiles, 'default').mockImplementationOnce(() => null);
```
gvarandas: Same as the comment above.

packages/cli/src/commands/init/__tests__/validate.test.ts
```diff
@@ -28,6 +27,8 @@ test.each([
     name: 'helloworld_test',
     error: HelloWorldError,
   },
-])("'%s' is invalid name", ({name, error}: {name: string, error: Error}) => {
+  // @ts-ignore-next-line FIXME extending the Error class causes weird TS validation errors
+  // https://stackoverflow.com/questions/41102060/typescript-extending-error-class
+])("'%s' is invalid name", ({name, error}: {name: string; error: Error}) => {
```
gvarandas: This is the only unsolved case from the conversion.
Apparently TS has weird issues when extending the `Error` class by not recognizing the correct properties.
Tried different solutions with no success. 😞

packages/cli/src/commands/init/init.ts
```diff
@@ -1,53 +1,63 @@
-// @flow
 import os from 'os';
 import path from 'path';
 import fs from 'fs-extra';
-import Ora from 'ora';
 import minimist from 'minimist';
+import ora from 'ora';
```
gvarandas: Changin the import name to follow the same convention as the other files.

packages/cli/src/tools/installPods.ts
```diff
@@ -103,7 +103,7 @@ async function installPods({
 }: {
   projectName: string;
   loader?: ora.Ora;
-  shouldUpdatePods: boolean;
+  shouldUpdatePods?: boolean;
```
gvarandas: Updated the type definition for this property since we're checking if it is passed or not in the method execution.
I'll be happy to revert this and explicitly pass the `false` option on `commands/init/initCompat:88`.

PR #709, opened 2019-09-11
chore: assign TS migration

radko93:
Summary:
---------
Migrated `assign.js` and `isValidPackageName.js` to typescript. Related to #683 
Is `isValidPackageName` used by external tool? Because it's not being used inside CLI so maybe we can remove it.


thymikee:
Looks like we can remove `isValidPackageName`, likely a leftover when refactoring init.

packages/cli/src/tools/assign.ts
```diff
@@ -4,14 +4,16 @@
...
- * @flow
  */
 export default function assign(target: Object, ...sources: Object[]) {
   sources.forEach(source => {
-    let descriptors = Object.keys(source).reduce((acc, key) => {
-      acc[key] = Object.getOwnPropertyDescriptor(source, key);
-      return acc;
-    }, {});
+    let descriptors = Object.keys(source).reduce(
+      (acc: {[k: string]: any}, key) => {
```
thymikee: how about moving the type to initial value in second argument of reduce? also, please use "unknown" instead of "any"

assuncaocharles: May `PropertyDescriptor` it's a better fit than `unknown`?

```
(acc: {[key: string]: PropertyDescriptor | undefined}, key: string) => {
```

radko93: New update using `PropertyDescriptorMap`. I had to add a safety check because `getOwnPropertyDescriptor` also returns undefined.

radko93: Also `uknown` does not work that great here because you cannot pass it to `Object.defineProperties` later on.

radko93:
> Looks like we can remove `isValidPackageName`, likely a leftover when refactoring init.

Removed.

packages/cli/src/tools/assign.ts
```diff
@@ -4,14 +4,19 @@
...
-    }, {});
+    let descriptors = Object.keys(source).reduce(
+      (acc, key) => {
+        const propertyDescriptor = Object.getOwnPropertyDescriptor(source, key);
+        if (propertyDescriptor !== undefined) {
+          acc[key] = propertyDescriptor;
+        }
+        return acc;
+      },
+      {} as PropertyDescriptorMap,
```
assuncaocharles: I guess it's more readable with the typing in the parameters?!?

```typescript
(acc: PropertyDescriptorMap, key: string) => {
```

So you avoid the typecasting in the initial value ```as PropertyDescriptorMap ```

But it's just my opnion 🤔

assuncaocharles:
You may want to add ```// $FlowFixMe - converted to TS``` in the ```tools/config/index```. It's the reason why the CI is not passing :)

radko93:
> You may want to add `// $FlowFixMe - converted to TS` in the `tools/config/index`. It's the reason why the CI is not passing :)

@assuncaocharles missed that from other PRs, thanks

packages/cli/src/tools/assign.ts
radko93: IMO it's a bit also more readable, but we can wait for @thymikee opinion

thymikee: I prefer `{} as PropertyDescriptorMap`. As long as TS is able to give me correct autocompletion, it's better to rely on inference and type less imho :)

PR #712, opened 2019-09-11
Migrate doctor command to TypeScript

thib92:
Summary:
---------

Migrate `packages/cli/src/commands/doctor` to TypeScript, with the exported type definitions in `cli-types`.

Related to #683 

Current progress:
-----------------

- [ ] `healthchecks/**`
- [ ] `doctor`
- [ ] `index`
- [x] `printFixOptions`
- [ ] `runAutomaticFix`
- [x] `versionRanges`

Test Plan:
----------

CI should be green

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -66,6 +66,7 @@
     "react-native": "^0.60.0"
   },
   "devDependencies": {
+    "@types/command-exists": "^1.2.0",
     "@types/graceful-fs": "^4.1.3",
     "@types/semver": "^6.0.2",
     "slash": "^3.0.0",
--- a/packages/cli/src/commands/doctor/checkInstallation.ts
+++ b/packages/cli/src/commands/doctor/checkInstallation.ts
@@ -1,4 +1,3 @@
-// @flow
 import semver from 'semver';
 import commandExists from 'command-exists';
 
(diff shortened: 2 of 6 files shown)


thib92:
I'm having an issue trying to move the exported types to `cli-types`, and I'm not sure how to resolve it:

When I try to move the `types.js` Flow file to the `cli-types` folder and export them with the `cli-types` package, I have some TypeScript errors regarding the version of the Node typings. I don't know how to fix this, and it's blocking me from moving the rest of the files to TS, since they all use these type definitions. If anyone has an idea on how to fix these, I'm all ears!

thymikee:
Which types you try to "move"? cc @lucasbento who's working on `doctor` command and may help you with some issues :)

lucasbento:
Solid work, thank you for doing this @thib92!

I would go with merging this in small chunks if possible as a lot of stuff is being changed on `doctor` as it's still WIP.

packages/cli-types/src/commands/doctor/index.ts
```diff
@@ -0,0 +1,79 @@
+// @flow
```
thymikee: remove flow

thymikee: Btw, how about putting the types into `doctor/types.ts`, similar as we have with Flow ones? In general, `index.js`/`index.ts` should only be used to actually "index" (list) some files/variables to export

thib92:
@thymikee thanks for the response. I thought that `cli-types` was a package for all exported types. My bad.

I wanted to "move" the `types.js` file which contains the Flow types for the `doctor` command.

Therefore, moving it to `types.ts` sounds like a good solution.

@lucasbento sure. Now that I have an answer to my question, I'll make a minimal passing build so that it can be merged soon to allow continuous work on the command.

thib92:
@thymikee @lucasbento I think the PR is good to review as is. I moved to the types to the `doctor` command in the `cli` package as asked.

The flow check is failing now, since some JS files are trying to import files converted to TS.

Should I remove the `// @flow-check` mark from the files for the `doctor` command so that CI is green again, just for the time to migrate to TS?

thymikee:
If the file is js, leave flow alone, just add $FlowFixMe comments to the imports of TS files. See how it's done in other modules :)

thib92:
Great! That's a lot of `$FlowFixMe`, will need to get this moved to TS fast 😄

PR #718, opened 2019-09-13
doctor: improve CocoaPods installation

lucasbento:
Summary:
---------

This PR fixes a small problem on the UI of `doctor` with the CocoaPods installation (tracked at #694), however it introduced a lot of stuff in the code in order to get it done.

The issue basically was that the question was being shown in the middle of the _fixes_ and felt a little bit out-of-place, the idea was to remove the question and do the installation together with a prefix of the method, e.g. `CocoaPods (installing with gem)`.

This turned out to be slightly bigger than expected, in order to remove the question after it's prompted I had to calculate the length of lines the question was using in the terminal.

### Before

before

### After

after

Test Plan:
----------

1. `sudo gem uninstall cocoapods`
1. `node /path/to/cli doctor`
1. Press `f`
1. Fall in love


packages/cli/src/tools/installPods.ts
```diff
@@ -43,56 +44,85 @@ function runSudo(command: string): Promise<void> {
...
-      await brewInstall({
-        pkg: 'cocoapods',
-        label: 'Installing CocoaPods',
-        loader,
-      });
-      break;
+  const shouldInstallWithGem = shouldInstallCocoaPods === installWithGem;
+
+  return {
+    installWithGem: shouldInstallWithGem,
```
thib92: `installWithGem` and `installWithHomebrew` are mutually exclusive.
Maybe you could have a `installMethod` property instead, which is equal to `homebrew` or `gem`?

You can even type this with TypeScript by doing this:

```typescript
type InstallMethod = 'homebrew' | 'gem';
```

lucasbento: I just left it like this so it doesn't need to compare with a string here: https://github.com/react-native-community/cli/blob/8beab8cd631d3942d3e35428624e8f33664e6156/packages/cli/src/tools/installPods.ts#L101.

Thoughts on this @thymikee?

thymikee: I like the `installMethod`, let's do this

lucasbento: 2x1
I'll do it 😄

packages/cli/src/commands/doctor/healthchecks/cocoaPods.ts
```diff
@@ -1,11 +1,92 @@
...
+} from '../../../tools/installPods';
+import {brewInstall} from '../../../tools/brewInstall';
 import {HealthCheckInterface} from '../types';
 
+function calculateQuestionSize(promptQuestion: string) {
+  return Math.max(
+    1,
+    Math.ceil(
+      wcwidth(promptQuestion) /
+        (process.stdout.columns! ? process.stdout.columns! : 80),
```
tido64: Not quite sure why you have to force unwrap here. Can we do something like this instead:
```suggestion
        (process.stdout.columns || 80),
```

packages/cli/src/tools/brewInstall.ts
```diff
@@ -6,15 +6,16 @@ import ora from 'ora';
 type InstallArgs = {
   pkg: string;
   label?: string;
+  succeedMessage?: string;
   loader: ora.Ora;
 };
 
-async function brewInstall({pkg, label, loader}: InstallArgs) {
+async function brewInstall({pkg, label, succeedMessage, loader}: InstallArgs) {
```
tido64: Would this be cleaner if we passed `onSuccess` and `onError` handlers instead?

packages/cli/src/tools/installPods.ts
tido64: Agree with @thib92 here. All strings are interned anyway, so it's just a cheap ref comparison. Alternatively, we could also just return the installer function. That way we need only check whether it's non-null.

packages/cli/src/commands/doctor/healthchecks/cocoaPods.ts
lucasbento: I have tried it but TypeScript gives an error, I'll check it again.

I do agree that this piece of code is a little bit strange.

packages/cli/src/tools/brewInstall.ts
lucasbento: Hmm, yes.

So, we call `onSuccess` only if it's not `undefined`, we default to showing `loader.succeed()`?

packages/cli/src/commands/doctor/healthchecks/cocoaPods.ts
lucasbento: Aaaand... I just tried this now and don't get an error 😐

Just pushed it, thanks for the hint!

lucasbento:
@thib92 @thymikee @tido64: all done, please check it out again 🙂

packages/cli/src/commands/doctor/healthchecks/cocoaPods.ts
```diff
@@ -1,11 +1,93 @@
...
+    const {
+      installMethod,
+      promptQuestion,
+    } = await promptCocoaPodsInstallationQuestion();
+
+    // Capitalise `Homebrew` when printing on the screen
+    const installMethodCapitalized =
+      installMethod === 'homebrew'
+        ? installMethod.substr(0, 1).toUpperCase() +
+          installMethod.substr(1).toLowerCase()
```
thymikee:
```suggestion
          installMethod.substr(1)
```
no need to lowercase already lowercased letters

packages/cli/src/tools/installPods.ts
```diff
@@ -43,56 +49,83 @@ function runSudo(command: string): Promise<void> {
...
-        label: 'Installing CocoaPods',
-        loader,
-      });
-      break;
+  const shouldInstallWithGem = shouldInstallCocoaPods === installWithGem;
+
+  return {
+    installMethod: shouldInstallWithGem ? 'gem' : 'homebrew',
+    // This is used for removing the message in `doctor` after it's answered
+    promptQuestion: `? ${stripAnsi(promptQuestion)} ${
```
thymikee: Seems like a hack, because we clear the question line(s), isn't it? Can we do it better, e.g. not clearing the line? I know terminals are tricky, but it would be nice to avoid hacks if possible

packages/cli/src/tools/installPods.ts
```diff
@@ -43,56 +49,83 @@ function runSudo(command: string): Promise<void> {
...
+
+  try {
+    // First attempt to install `cocoapods`
+    await execa('gem', options);
+  } catch (_error) {
+    // If that doesn't work then try with sudo
+    try {
+      await runSudo(`gem ${options.join(' ')}`);
+    } catch (error) {
+      throw error;
```
thymikee: if we re-throw without any side effects around, then there's no need to try/catch

packages/cli/src/tools/installPods.ts
lucasbento: Well, no.

In terms of clearing the message, I would definitely continue in this path for a better DX, after the question is answered you no longer need it clogging your UI.

I searched multiple ways of doing this, best way is to calculate the size of the message and remove it, unfortunately.

It does look a little bit hacky, I can _maybe_ improve this later by listening to which messages are being printed on the screen and have a helper function to remove it later, this will only get rid of the need to store what message is being shown but the calculation and all the rest will remain.

thymikee: Yea, that's what I was afraid of. Let's continue doing what we do now and try to think of a simple abstraction around it for ease of use (but don't stress it)

lucasbento:
@thymikee: all fixed 🙂

PR #719, opened 2019-09-13
chore: Migrate doctor command to TypeScript (3)

thib92:
Summary:
---------

Finish the healthchecks migration to TS, continuing the work from #712 and #716 (linked to #683):

* `androidSDK`
* `iosDeploy`
* `nodeJs`
* `packageManagers`
* `watchman`

Test Plan:
----------

- [x] CI is green ✅ 
- [ ] The changes made make sense

### DIFF
--- a/packages/cli/src/commands/doctor/healthchecks/androidSDK.ts
+++ b/packages/cli/src/commands/doctor/healthchecks/androidSDK.ts
@@ -1,22 +1,17 @@
-// @flow
 import chalk from 'chalk';
-import Ora from 'ora';
-// $FlowFixMe - converted to TS
 import {logManualInstallation} from './common';
-// $FlowFixMe - converted to TS
 import versionRanges from '../versionRanges';
-// $FlowFixMe - converted to TS
 import {doesSoftwareNeedToBeFixed} from '../checkInstallation';
 import execa from 'execa';
-import type {EnvironmentInfo, HealthCheckInterface} from '../types';
+import {HealthCheckInterface} from '../types';
 
 const installMessage = `Read more about how to update Android SDK at ${chalk.dim(
   'https://developer.android.com/studio',
 )}`;
 
-export default ({
+export default {
   label: 'Android SDK',
-  getDiagnostics: async ({SDKs}: EnvironmentInfo) => {
+  getDiagnostics: async ({SDKs}) => {
     let sdks = SDKs['Android SDK'];
 
     // This is a workaround for envinfo's Android SDK check not working on
@@ -58,13 +53,7 @@ export default ({
           })),
     };
   },
-  runAutomaticFix: async ({
-    loader,
-    environmentInfo,
-  }: {
-    loader: typeof Ora,
-    environmentInfo: EnvironmentInfo,
-  }) => {
+  runAutomaticFix: async ({loader, environmentInfo}) => {
     const version = environmentInfo.SDKs['Android SDK'][0];
     const isSDKInstalled = version !== 'Not Found';
 
@@ -81,4 +70,4 @@ export default ({
       url: 'https://facebook.github.io/react-native/docs/getting-started',
     });
   },
-}: HealthCheckInterface);
+} as HealthCheckInterface;
--- a/packages/cli/src/commands/doctor/healthchecks/androidSDK.ts
+++ b/packages/cli/src/commands/doctor/healthchecks/androidSDK.ts
@@ -38,6 +38,9 @@ export default {
         if (matches.length > 0) {
           sdks = {
             'Build Tools': matches,
+            'API Levels': 'Not Found',
+            'Android NDK': 'Not Found',
+            'System Images': 'Not Found',
           };
         }
       } catch {}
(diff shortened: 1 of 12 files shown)


thib92:
Tests are failing because of the `androidSdk` healthcheck, because here:

https://github.com/thib92/react-native-cli/blob/2ff26561138c1a7c4e71687dc0a12d4f00f20b9b/packages/cli/src/commands/doctor/healthchecks/androidSDK.ts#L20

We check for `sdks === 'Not Found'`, but it is never the case as the typings describe them as:

```typescript
'Android SDK': {
  'API Levels': string[];
  'Build Tools': string[];
  'System Images': string[];
  'Android NDK': string;
};
```

@lucasbento do you know if `'Not Found'` is actually a possible value for the `Android SDK` key?

thymikee:
@thib92 I think `envinfo` may output a string `"Not Found"` when a certain required environment info is not available (looks like it happens on Windows at least, but make sure to check the envinfo codebase or typings)

thib92:
Thanks @thymikee, I’ll take a look!

lucasbento:
> do you know if 'Not Found' is actually a possible value for the Android SDK key?

@thib92: yeah, it is, you can check it out [here](https://github.com/tabrindle/envinfo/blob/master/src/helpers/sdks.js#L62-L65).

thib92:
Thanks. I’ll update the typings then.

packages/cli/src/commands/doctor/types.ts
```diff
@@ -32,12 +32,14 @@ export type EnvironmentInfo = {
     'iOS SDK': {
       Platforms: string[];
     };
-    'Android SDK': {
-      'API Levels': string[];
-      'Build Tools': string[];
-      'System Images': string[];
-      'Android NDK': string;
-    };
+    'Android SDK':
```
thib92: Ideally, we should contribute these typings to the `envinfo` repo. There are no typings yet in there. There is [an ongoing PR](https://github.com/tabrindle/envinfo/pull/105), but the typings are quite incomplete.

thib92:
CI is failing but doesn't seem related 🤔

thib92:
@thymikee is there a way to "try again" on Azure Pipeline? I ran the E2E tests locally and they worked. It might just be an "infrastructure bug" while it ran the test.

Esemesek:
@thib92 Tests on Windows seem to fail every now and then. I wouldn't care too much C:

packages/cli/src/commands/doctor/healthchecks/androidSDK.ts
```diff
@@ -58,14 +56,8 @@ export default ({
...
   },
-  runAutomaticFix: async ({
-    loader,
-    environmentInfo,
-  }: {
-    loader: typeof Ora,
-    environmentInfo: EnvironmentInfo,
-  }) => {
-    const version = environmentInfo.SDKs['Android SDK'][0];
+  runAutomaticFix: async ({loader, environmentInfo}) => {
```
Esemesek: Can you add function parameter type?

packages/cli/src/commands/doctor/healthchecks/iosDeploy.ts
```diff
@@ -20,13 +16,13 @@ const getInstallationCommand = () => {
...
 
-export default ({
+export default {
   label: 'ios-deploy',
   isRequired: false,
   getDiagnostics: async () => ({
     needsToBeFixed: await checkSoftwareInstalled('ios-deploy'),
   }),
-  runAutomaticFix: async ({loader}: {loader: typeof Ora}) => {
+  runAutomaticFix: async ({loader}) => {
```
Esemesek: Missing function parameters type here as well.

packages/cli/src/commands/doctor/healthchecks/nodeJS.ts
```diff
@@ -1,28 +1,23 @@
...
-// $FlowFixMe - converted to TS
 import {logManualInstallation} from './common';
-import type {EnvironmentInfo, HealthCheckInterface} from '../types';
+import {HealthCheckInterface} from '../types';
 
-export default ({
+export default {
   label: 'Node.js',
-  getDiagnostics: async ({Binaries}: EnvironmentInfo) => ({
+  getDiagnostics: async ({Binaries}) => ({
```
Esemesek: Here too C:

packages/cli/src/commands/doctor/healthchecks/nodeJS.ts
```diff
@@ -1,28 +1,23 @@
...
     version: Binaries.Node.version,
     needsToBeFixed: doesSoftwareNeedToBeFixed({
       version: Binaries.Node.version,
       versionRange: versionRanges.NODE_JS,
     }),
   }),
-  runAutomaticFix: async ({loader}: {loader: typeof Ora}) => {
+  runAutomaticFix: async ({loader}) => {
```
Esemesek: And here C:

thib92:
@Esemesek I guess you figured it out after the comments, but the function types are not needed here because we already have the `HealthCheckInterface`, so everything is implicit 😄

Esemesek:
@thib92 Yup. Figured it out C:

PR #722, opened 2019-09-14
Migrate the rest of doctor command to TypeScript

thib92:
Summary:
---------

This is the last PR to move the `doctor` command to TypeScript. This one was quite a long one because I needed to unify the type definitions of all the functions manipulating the type checks, and it can be a little bit confusing (at least to me).

So I hope with this, all the functions and their roles in the healthchecks will be easier to read.

Feel free to comment if anything feels off.

* Migrate the rest of doctor command to TypeScript
* Unify type definitions for the command
* Remove Flow typings

Test Plan:
----------

* CI is green ✅ 
* Everything makes sense


packages/cli/src/commands/doctor/doctor.ts
```diff
@@ -108,23 +107,24 @@ export default (async function runDoctor(
             : undefined,
         };
       }),
-    )).filter(Boolean),
+    )).filter(healthcheck => healthcheck !== undefined) as HealthCheckResult[],
```
thib92: Thought this would be a bit simpler to understand.
The `as HealthCheckResult[]` part is unfortunately needed, otherwise TypeScript isn't smart enough to know that the objects inside the array can't be `undefined` anymore.

Could have used [type predicates](https://www.typescriptlang.org/docs/handbook/advanced-types.html#using-type-predicates) though.

packages/cli/src/commands/doctor/doctor.ts
```diff
@@ -108,23 +107,22 @@ export default (async function runDoctor(
...
-    // $FlowFixMe - bad Object.values typings
+  const iterateOverCategories = (categories: HealthcheckCategory[]) =>
     Promise.all(categories.map(iterateOverHealthChecks));
 
-  const healthchecksPerCategory = await iterateOverCategories(
-    Object.values(getHealthchecks(options)),
-  );
+  const healthchecksPerCategory = await iterateOverCategories(Object.values(
+    getHealthchecks(options),
+  ).filter(category => category !== undefined) as HealthcheckCategory[]);
```
thib92: Same here, the `as` part is necessary, but could use type predicates.

packages/cli/src/commands/doctor/doctor.ts
```diff
@@ -161,13 +161,14 @@ export default (async function runDoctor(
     });
   }
 
-  const onKeyPress = async key => {
-    // $FlowFixMe
+  const onKeyPress = async (key: string) => {
+    // @ts-ignore
     process.stdin.setRawMode(false);
```
thib92: Does `stdin.setRawMode` really exist? It's not in the typings.

packages/cli/src/commands/doctor/runAutomaticFix.ts
```diff
@@ -1,38 +1,41 @@
...
+  };
+  loader: Ora;
+  environmentInfo: EnvironmentInfo;
+}
+
+export default async function({
   healthchecks,
   automaticFixLevel,
   stats,
-  loader,
```
thib92: I saw here that this function doesn't use its `loader` argument, but uses `ora` directly from the import. Is this intended?

packages/cli/src/commands/doctor/doctor.ts
```diff
@@ -1,18 +1,19 @@
...
-// $FlowFixMe - converted to TS
 import printFixOptions, {KEYS} from './printFixOptions';
 import runAutomaticFix, {AUTOMATIC_FIX_LEVELS} from './runAutomaticFix';
-import type {ConfigT} from 'types';
-import type {HealthCheckInterface} from './types';
-
-const printCategory = ({label, key}) => {
+import {CommandFunction} from '@react-native-community/cli-types';
+import {
+  HealthcheckCategory,
```
thymikee: Let's unify how we refer to "health check"
```suggestion
  HealthCheckCategory,
```

packages/cli/src/commands/doctor/doctor.ts
thymikee: It exists in TTY, which is our primary use case, so yep: https://nodejs.org/api/tty.html#tty_readstream_setrawmode_mode
Please check if `typeof process.stdin.setRawMode === "function"` before accessing it. We should add it to every call of this method at some point.

packages/cli/src/commands/doctor/runAutomaticFix.ts
```diff
@@ -1,46 +1,47 @@
...
-  healthchecks: any,
-  automaticFixLevel: $Values<typeof AUTOMATIC_FIX_LEVELS>,
-  stats: {errors: any, warnings: any},
-  loader: typeof ora,
-  environmentInfo: EnvironmentInfo,
-}) => {
+}: RunAutomaticFixArgs) {
   // Remove the fix options from screen
-  // $FlowFixMe
+  // @ts-ignore
```
thymikee: instead of ignore, please add `if (process.stdout.isTTY) {}` check, or make sure `moveCursor` and `clearScreenDown` exist

packages/cli/src/commands/doctor/types.ts
```diff
@@ -66,15 +66,42 @@ export type EnvironmentInfo = {
   };
 };
 
+export type HealthcheckCategory = {
+  label: string;
+  healthchecks: HealthCheckInterface[];
+};
+
+export type Healthchecks = {
+  common: HealthcheckCategory;
```
thymikee: Please rename to `HealthChecks` and `HealthCheckCategory` to stay consistent consistency

packages/cli/src/commands/doctor/doctor.ts
thib92: My bad, it's optional. Thanks!

packages/cli-types/src/index.ts
```diff
@@ -13,14 +13,16 @@ import {
 
 export type InquirerPrompt = any;
 
+export type CommandFunction = (
```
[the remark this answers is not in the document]
thib92: Good call!

thib92:
All done @thymikee, thanks for the review!

packages/cli/src/commands/doctor/runAutomaticFix.ts
thymikee: if loader is unused, please remove it from the type def as well

thymikee:
cc @lucasbento

packages/cli/src/commands/doctor/doctor.ts
```diff
@@ -1,18 +1,19 @@
-// @flow
 import chalk from 'chalk';
+// @ts-ignore
```
lucasbento: Noobie question: shouldn't we type these instead of just adding it to ignore?

packages/cli/src/commands/doctor/runAutomaticFix.ts
```diff
@@ -1,46 +1,49 @@
...
 import {HEALTHCHECK_TYPES} from './healthchecks';
-import type {EnvironmentInfo} from './types';
+import {EnvironmentInfo, HealthCheckCategoryResult} from './types';
 
-const AUTOMATIC_FIX_LEVELS = {
-  ALL_ISSUES: 'ALL_ISSUES',
-  ERRORS: 'ERRORS',
-  WARNINGS: 'WARNINGS',
-};
+export enum AUTOMATIC_FIX_LEVELS {
```
lucasbento: so good

packages/cli/src/commands/doctor/runAutomaticFix.ts
```diff
@@ -1,46 +1,49 @@
...
-}) => {
+}: RunAutomaticFixArgs) {
   // Remove the fix options from screen
-  // $FlowFixMe
-  process.stdout.moveCursor(0, -6);
-  // $FlowFixMe
-  process.stdout.clearScreenDown();
+  if (process.stdout.isTTY) {
+    // @ts-ignore
+    process.stdout.moveCursor(0, -6);
```
lucasbento: Those should be part of `readline` as they are apparently not typed under `process.stdout`: DefinitelyTyped/DefinitelyTyped#31505.

I can change those later as there's no harm on ignoring this.

packages/cli/src/commands/doctor/runAutomaticFix.ts
```diff
@@ -1,46 +1,49 @@
...
-};
+export enum AUTOMATIC_FIX_LEVELS {
+  ALL_ISSUES = 'ALL_ISSUES',
+  ERRORS = 'ERRORS',
+  WARNINGS = 'WARNINGS',
+}
 
-export {AUTOMATIC_FIX_LEVELS};
-export default async ({
+interface RunAutomaticFixArgs {
```
lucasbento: @thib92: small question, when do you use `type` and `interface`?

thib92: There's a small difference between the two, but to be fair I usually use `type` because you can union them, type functions, etc. I'm not sure what the advantages of `interface` is though. I saw you used `interface` before so I used them where I could

packages/cli/src/commands/doctor/doctor.ts
thib92: Yep we should. I might give it a try, but there is a pending WIP PR on the repo. Would probably be worth being put on DefinitelyTyped

thib92:
Thanks for the merge. @thymikee @lucasbento if you need help on other stuff, please send me a message! I'd love to contribute more on the project! I'll be looking at issues that I could take 😄

thymikee:
@thib92 thank you! ❤️

packages/cli/src/commands/doctor/runAutomaticFix.ts
lucasbento: Cool, thanks for replying!

lucasbento:
@thib92: thanks for the contributions!

We have a good list of issues that we need to work on at #694 in order to get `doctor` out of experimental, if you are feeling adventurous you can take one and open a PR 🙂

thib92:
@lucasbento thanks a lot, will take a look!

PR #726, opened 2019-09-17
doctor: improve `ios-deploy` installation

lucasbento:
Summary:
---------

This PR is related to #694, it improves the experience of when the command can not "guess" which package manager to use when installing `ios-deploy`, with this change, we try to guess, if not available then we prompt a question asking which package manager to use (`yarn` or `npm`) and an option to skip, if the user chooses skip then we just print a message for manual installation. 

### Before (when it couldn't guess the package manager)

before

#### When it guesses the package manager

When it guesses the package manager

#### When it shows options

When it shows options

#### When you choose "skip"

When you choose "skip"

➡️ Skipping doesn't cancel the other health checks, it continues normally.

Test Plan:
----------

1. `yarn global remove ios-deploy` / `npm uninstall -g ios-deploy`
1. `/path/to/cli doctor`
1. Press `F`


lucasbento:
@thib92: if you have time, please take a look 🙂

packages/cli/src/commands/doctor/healthchecks/iosDeploy.ts
```diff
@@ -1,58 +1,116 @@
...
 };
 
+const installLibrary = async ({
+  installationCommand,
+  packageManagerToUse,
+  loader,
+}: {
+  installationCommand: string;
+  packageManagerToUse: 'yarn' | 'npm';
+  loader: ora.Ora;
```
thib92: You can use this if you prefer (that's what I did for the rest of the command)

```typescript
import {Ora} from 'ora';
```

```typescript
loader: Ora;
```

packages/cli/src/commands/doctor/healthchecks/iosDeploy.ts
```diff
@@ -1,58 +1,116 @@
...
 
   if (packageManager === PACKAGE_MANAGERS.NPM) {
-    return 'npm install ios-deploy --global';
+    return installationWithNpm;
   }
 
   return undefined;
 };
 
+const installLibrary = async ({
```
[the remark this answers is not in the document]
lucasbento: Do you mean in the sense that it can be shared?

I would keep it like this until it actually has the need to move to a `common` file, to be clear I would wait until we have another health check with a library from npm registry to move this so we know better what should be kept or changed.

packages/cli/src/commands/doctor/healthchecks/iosDeploy.ts
lucasbento: Done 🙂

packages/cli/src/commands/doctor/healthchecks/iosDeploy.ts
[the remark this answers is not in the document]
thib92: Sure! That sounds good. It's ready to be shared when needed 😄

packages/cli/src/commands/doctor/healthchecks/iosDeploy.ts
```diff
@@ -1,58 +1,116 @@
...
+  installationCommand: string;
+  packageManagerToUse: 'yarn' | 'npm';
+  loader: Ora;
+}) => {
+  try {
+    loader.start(`${label} (installing with ${packageManagerToUse})`);
+
+    const installationCommandArgs = installationCommand.split(' ');
+
+    await execa(installationCommandArgs[0], installationCommandArgs.splice(1));
```
thymikee: This should work:
```suggestion
    await execa(installationCommand);
```

packages/cli/src/commands/doctor/healthchecks/iosDeploy.ts
```diff
@@ -1,58 +1,116 @@
...
+
+    loader.succeed(`${label} (installed with ${packageManagerToUse})`);
+  } catch (_error) {
```
thymikee: I think we should reject with some kind of error, since this function is returned

packages/cli/src/commands/doctor/healthchecks/iosDeploy.ts
```diff
@@ -1,58 +1,116 @@
...
-      const installationCommandArgs = installationCommand.split(' ');
+      if (whichPackageManagerToUse === skipInstallation) {
+        loader.fail();
 
-      await execa(
-        installationCommandArgs[0],
-        installationCommandArgs.splice(1),
-      );
+        // Then we just print out the URL that the user can head to download the library
+        return logManualInstallation({
```
thymikee: if `logManualInstallation` doesn't return, please revert this change, I prefer being explicit with returns

packages/cli/src/commands/doctor/healthchecks/iosDeploy.ts
```diff
@@ -1,58 +1,116 @@
...
+      const skipInstallation = 'Skip installation';
 
-      // Then we just print out the URL that the user can head to download the library
-      logManualInstallation({
-        healthcheck: 'ios-deploy',
-        url: 'https://github.com/ios-control/ios-deploy#readme',
-      });
-      return;
-    }
+      const {whichPackageManagerToUse} = await inquirer.prompt([
```
thymikee: This is quite a long variable name. How about `chosenPackageManager`?

lucasbento:
@thymikee: thanks for the review, changes done!

packages/cli/src/commands/doctor/healthchecks/iosDeploy.ts
```diff
@@ -1,58 +1,117 @@
...
+}) => {
+  try {
+    loader.start(`${label} (installing with ${packageManagerToUse})`);
+
+    await execa(installationCommand);
+
+    loader.succeed(`${label} (installed with ${packageManagerToUse})`);
+  } catch (error) {
+    loader.fail();
+    logger.log(chalk.dim(`\n${error}`));
```
thymikee: Shuoldn't we just log the message? We don't need the stack trace here
```suggestion
    logger.log(chalk.dim(`\n${error.message}`));
```

lucasbento: I checked with `error` and it returns the same as `error.message`, I'll change this though.

In `cocoapods` we are also logging only `error`, shall I change that there as well?

thymikee: Please do

lucasbento: Done!

packages/cli/src/commands/doctor/healthchecks/iosDeploy.ts
lucasbento: I just checked this and it doesn't work, it only gives me `spawn npm install ios-deploy --global ENOENT`.

Shall I revert to what it was?

lucasbento: The API from execa: https://github.com/sindresorhus/execa#execafile-arguments-options.

thymikee: Oh, I thought this would work. Revert then

lucasbento: Done :D

PR #728, opened 2019-09-18
[doctor command] Add tests for androidHomeEnvVariable

thib92:
Summary:
---------

Started writing tests for the `doctor` healthchecks, starting with `androidHomeEnvVariable`.

Since I needed the `envinfo` output in the test, and I will need it in all the upcomming tests, I moved it to the `tools` folder. Maybe it's not the best place, I'm not sure.

I only wrote one test to have feedback about the way of writing them, feel free to comment.

IMHO since these are unit tests we should mock all the imports and just make sure we call them when expected.
I didn't test the hard-coded SO links because it seems like a detail that wasn't worth testing.

Related to #694

Test Plan:
----------

This is a test so... CI is green? ✅

### DIFF
--- a/packages/cli/src/commands/doctor/doctor.ts
+++ b/packages/cli/src/commands/doctor/doctor.ts
@@ -57,7 +57,7 @@ export default (async (_, __, options) => {
 
   loader.start('Running diagnostics...');
 
-  const environmentInfo = JSON.parse(await getEnvironmentInfo());
+  const environmentInfo = await getEnvironmentInfo();
 
   const iterateOverHealthChecks = async ({
     label,
@@ -117,10 +117,7 @@ export default (async (_, __, options) => {
   };
 
   healthchecksPerCategory.forEach((issueCategory, key) => {
-    printCategory({
-      ...issueCategory,
-      key,
-    });
+    printCategory({...issueCategory, key});
 
     issueCategory.healthchecks.forEach(healthcheck => {
       printIssue(healthcheck);
--- a/packages/cli/src/commands/doctor/doctor.ts
+++ b/packages/cli/src/commands/doctor/doctor.ts
@@ -1,6 +1,4 @@
 import chalk from 'chalk';
-// @ts-ignore
-import envinfo from 'envinfo';
 import {logger} from '@react-native-community/cli-tools';
 import {getHealthchecks, HEALTHCHECK_TYPES} from './healthchecks';
 import {getLoader} from '../../tools/loader';
@@ -12,6 +10,7 @@ import {
   HealthCheckCategoryResult,
   HealthCheckResult,
 } from './types';
+import getEnvironmentInfo from '../../tools/envinfo';
 
 const printCategory = ({label, key}: {label: string; key: number}) => {
   if (key > 0) {
@@ -58,18 +57,7 @@ export default (async (_, __, options) => {
 
   loader.start('Running diagnostics...');
 
-  const environmentInfo = JSON.parse(
-    await envinfo.run(
-      {
-        Binaries: ['Node', 'Yarn', 'npm', 'Watchman'],
-        IDEs: ['Xcode', 'Android Studio'],
-        SDKs: ['iOS SDK', 'Android SDK'],
-        npmPackages: ['react', 'react-native', '@react-native-community/cli'],
-        npmGlobalPackages: ['*react-native*'],
-      },
-      {json: true, showNotFound: true},
-    ),
-  );
+  const environmentInfo = JSON.parse(await getEnvironmentInfo());
 
   const iterateOverHealthChecks = async ({
     label,
@@ -106,8 +94,9 @@ export default (async (_, __, options) => {
     )).filter(healthcheck => healthcheck !== undefined) as HealthCheckResult[],
   });
 
-  // Remove all the categories that don't have any healthcheck with `needsToBeFixed`
-  // so they don't show when the user taps to fix encountered issues
+  // Remove all the categories that don't have any healthcheck with
+  // `needsToBeFixed` so they don't show when the user taps to fix encountered
+  // issues
   const removeFixedCategories = (categories: HealthCheckCategoryResult[]) =>
     categories.filter(category =>
       category.healthchecks.some(healthcheck => healthcheck.needsToBeFixed),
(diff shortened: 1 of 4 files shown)


thib92:
Looks like E2E tests failing because of Azure.

thib92:
@lucasbento please take a look if you have time 😄

packages/cli/src/commands/doctor/healthchecks/__tests__/androidHomeEnvVariable.test.ts
```diff
@@ -0,0 +1,56 @@
...
+
+jest.mock('../common');
+
+import {logManualInstallation} from '../common';
+
+describe('androidHomeEnvVariables', () => {
+  const OLD_ENV = process.env;
+
+  it('returns a message if no ANDROID_HOME is defined', async () => {
+    delete process.env.ANDROID_HOME;
```
lucasbento: I was wondering if this can cause problems but I saw the `afterEach` so it seems good to go 🙂

packages/cli/src/commands/doctor/healthchecks/__tests__/androidHomeEnvVariable.test.ts
```diff
@@ -0,0 +1,56 @@
...
+      await getEnvironmentInfo(),
+    );
+    const diagnostics = await androidHomeEnvVariables.getDiagnostics(
+      environmentInfo,
+    );
+    expect(diagnostics.needsToBeFixed).toBe(false);
+  });
+
+  it('logs manual installation steps to the screen', async () => {
+    // @ts-ignore
```
lucasbento: Why do you need this?

packages/cli/src/commands/doctor/healthchecks/__tests__/androidHomeEnvVariable.test.ts
```diff
@@ -0,0 +1,56 @@
...
+      info: jest.fn(),
+    };
+
+    const environmentInfo: EnvironmentInfo = JSON.parse(
+      await getEnvironmentInfo(),
+    );
+
+    androidHomeEnvVariables.runAutomaticFix({loader, environmentInfo});
+
+    expect(logManualInstallation).toHaveBeenCalledTimes(1);
```
lucasbento: Perhaps you can snapshot the `logManualInstallation` calls as well? 😄

packages/cli/src/commands/doctor/healthchecks/__tests__/androidHomeEnvVariable.test.ts
thib92: OLD_ENV will contain the environment before running any test, so since we restore it on afterEach, yeah it should be good to go 😉

packages/cli/src/commands/doctor/healthchecks/__tests__/androidHomeEnvVariable.test.ts
thib92: I was thinking that snapshotting the exact link wasn't really useful, since changing the link would break the test. Depends on individual testing preferences though, I'm not feeling strongly on it.

packages/cli/src/commands/doctor/healthchecks/__tests__/androidHomeEnvVariable.test.ts
thib92: Because I'm forcing `loader` to be of type `Ora`, but I only put the `info` property, not the rest.

Though I changed it to `any` to avoid the `// @ts-ignore`.

thymikee: You can use `NoopLoader` (grep for it)

thib92: I need to watch it though, so I need to use `jest.fn()`. Though using `NoopLoader`, I could use `jest.spyOn()`. What do you reckon is the best?

thymikee: Generate the mock: https://jestjs.io/docs/en/jest-object#jestgenmockfrommodulemodulename

thib92: Actually my mistake, I don't need to watch it, so I used the `NoopLoader`. Thanks @thymikee!

packages/cli/src/commands/doctor/doctor.ts
```diff
@@ -128,7 +117,10 @@ export default (async (_, __, options) => {
   };
 
   healthchecksPerCategory.forEach((issueCategory, key) => {
-    printCategory({...issueCategory, key});
+    printCategory({
+      ...issueCategory,
+      key,
+    });
```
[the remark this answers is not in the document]
thib92: Yeah sorry, my VSCode formatter settings changed for some reason. Reverted :)

thib92:
All fixed @thymikee, thanks for the review and the good tips on Jest best practices 😄

packages/cli/src/commands/doctor/healthchecks/__tests__/androidHomeEnvVariable.test.ts
```diff
@@ -3,19 +3,22 @@ import getEnvironmentInfo from '../../../../tools/envinfo';
...
+    jest.resetAllMocks();
+  });
+
   it('returns a message if no ANDROID_HOME is defined', async () => {
     delete process.env.ANDROID_HOME;
 
-    const environmentInfo: EnvironmentInfo = JSON.parse(
-      await getEnvironmentInfo(),
-    );
+    const environmentInfo: EnvironmentInfo = await getEnvironmentInfo();
```
thymikee: you likely don't need to cast the type now :)

packages/cli/src/tools/envinfo.ts
```diff
@@ -2,17 +2,19 @@
...
-      IDEs: ['Xcode', 'Android Studio'],
-      SDKs: ['iOS SDK', 'Android SDK'],
-      npmPackages: ['react', 'react-native', '@react-native-community/cli'],
-      npmGlobalPackages: ['*react-native*'],
-    },
-    {
-      json: true,
-      showNotFound: true,
-    },
+  return JSON.parse(
```
thymikee: Please cast to `EnvironmentInfo`

thib92: Good catch, thanks!

PR #730, opened 2019-09-18
[Docs][Autolinking] Fix monorepo examples relative path

Ashoat:
The example folder structure as well as the edited paths to `node_modules` reflect a path that is only one extra level up. This PR updates the relative paths passed into the scripts to match. I noticed they don't match while I was upgrading to RN 0.60 because I have a comparable folder structure, but had issues on both iOS and Android when I initially tried passing in `../..` as described in the docs.

### DIFF
--- a/docs/autolinking.md
+++ b/docs/autolinking.md
@@ -54,7 +54,7 @@ you'll need to set a custom root. Pass it as an argument to `use_native_modules!
 require_relative '../../node_modules/@react-native-community/cli-platform-ios/native_modules'
 target 'RNapp' do
   # React pods and custom pods here...
-  use_native_modules!("../..")
+  use_native_modules!("..")
 end
 ```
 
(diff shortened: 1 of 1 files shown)


thymikee:
This setup is for custom root (the default is `..`) and expects you to run `yarn react-native` from the root of the project. How do you run your project? Regardless the answer, we need to think about better story for handling monorepos and custom structures, because it causes a lot of headaches today.

Ashoat:
Interesting, I wasn’t aware that where I run `react-native` from matters!

My [project](https://github.com/ashoat/squadcal) structure is a yarn workspaces setup with a `native` folder that contains the React Native project.

In my case:

1. Running `pod install` from `native/ios` works when passing in `..` to `use_native_modules!`. If I pass in `../..` the autolinking logic fails to detect any native modules. `Podfile` is [here](https://github.com/Ashoat/squadcal/blob/master/native/ios/Podfile)
2. Running `react-native run-android` from `native` works when passing `..` to `applyNativeModulesSettingsGradle` and `applyNativeModulesAppBuildGradle`. If I pass in `../..`, Gradle doesn’t pick up the native modules.

grabbou:
@Ashoat, I think the confusion is caused by our understanding of a `project root`.

> "The project root is where node_modules with react-native is"

In your example, I think `react-native` and other modules are still under `native`, which means no additional changes are needed in order to make it work.

Now, if this was e.g. Yarn workspaces, then, you would need to apply it.

Ashoat:
It in fact _is_ a Yarn workspaces set-up, but you're correct in identifying that `native/node_modules/react-native` exists. I have tooling that symlinks dependencies from the root `node_modules` so that various other tools/libraries that expect them there can find them.

What continues to confuse me is that passing in `../..` broke both the iOS and Android set-ups. You would think that since `node_modules/react-native` exists, that parameter would work. But it doesn't - only `..` seems to work for me.

Anyways - thanks for explaining the context here. If others aren't seeing this issue then it's probably just on my end.

PR #736, opened 2019-09-20
Update run instructions to use projectDir

raldred:
Summary:
---------

Run instructions use projectName
Updated to use projectDir to display correct instruction when using the --direction option


grabbou:
That change looks good to me. I defer the final merge to @thymikee since I wasn't following the code changes in the past few weeks.

thymikee:
We intentionally displayed only the project name so the command is shorter, but now that we have `--directory` param it makes sense to display the whole path, even though it's uglier. Thanks!

PR #737, opened 2019-09-20
fix: find node_modules at relative root

ishabo:
Summary:
---------
For monorepo apps with custom folder structure (where android folder is not in the same root folder as node_modules) building fails with error: `Text must not be null or empty`.

Test Plan:
----------
To replicate the issue:
- create a new project using the react-native cli tool.
- create a new dir called `app`
- move `android` to `app`
- change all references to node_modules in android build gradle files `../../node_modules` instead of `../node_modules`
- Try building the project for Android
- Watch it fail

With this fix, the build will work no matter what the structure of folders and relative position of the `android` folder to `node_modules` is.

### DIFF
--- a/packages/platform-android/native_modules.gradle
+++ b/packages/platform-android/native_modules.gradle
@@ -183,7 +183,7 @@ class ReactNativeModules {
 
     def cmdProcess
     def root = getReactNativeProjectRoot()
-    def command = "node ./node_modules/react-native/cli.js config"
+    def command = "node ${root.toString()}/node_modules/react-native/cli.js config"
     def reactNativeConfigOutput = ""
 
     try {


packages/platform-android/native_modules.gradle
```diff
@@ -183,7 +183,7 @@ class ReactNativeModules {
 
     def cmdProcess
     def root = getReactNativeProjectRoot()
-    def command = "node ./node_modules/react-native/cli.js config"
+    def command = "node ${root.toString()}/node_modules/react-native/cli.js config"
```
thymikee: This is called on line 190 like:
```groovy
```
So I wonder, why does it change anything, since we call the command from that root directory?

ishabo: In a monorepo setup we have a shared node_modules between apps, so we call the build command from within each android app with some custom task script. This fix makes it work for those custom situations as well as the conventional situation.

thymikee: Ok, but that still doesn't address my question. I'd expect `Runtime.getRuntime` ran from the root to have `.` being a root. cc @Salakar

Salakar: Hmm I'm confused by this a bit, the third of exec is already specifying where to run the command - in the specified root dir?

ishabo: I see what you guys mean ... I just added some log and it gives the path to the `current android folder with ../` 

I tested again removing both `node` & `${root.String()}` and it didn't give the error anymore. I am not an android dev so the gradle syntax is a bit strange to me. 

Is there a way to pass special settings from the `apply` command in `settings.gradle` to let this script know where to find `node_modules`?

ishabo: Just to clarify, I thinkI got a false positive when I first changed the line to:

`def command = "${root.toString()}/node_modules/react-native/cli.js config"`
That was because I removed `node`, not because I added `${root.toString()}` 
I later added `node` by mistake when changing to make the PR without testing. Then just tested again with the last change 
`def command = "node ${root.toString()}/node_modules/react-native/cli.js config"` and I was still getting the error.

Sorry about the confusion.

ishabo:
I realised the fix is to add the custom root as a second param to applyNativeModulesAppBuildGradle in the settings.gradle. This was a dumb solution.

Issue #740, opened 2019-09-22
Android Studio synchronization fails


generalpiston:
## Environment

```
info Fetching system and libraries information...
System:
    OS: macOS 10.14.6
    CPU: (12) x64 Intel(R) Core(TM) i9-8950HK CPU @ 2.90GHz
    Memory: 4.54 GB / 32.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 12.4.0 - ~/.nodenv/versions/12.4.0/bin/node
    npm: 6.9.0 - ~/.nodenv/versions/12.4.0/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  SDKs:
    iOS SDK:
      Platforms: iOS 12.4, macOS 10.14, tvOS 12.4, watchOS 5.3
    Android SDK:
      API Levels: 23, 26, 27, 28
      Build Tools: 27.0.3, 28.0.3, 29.0.0
      System Images: android-23 | Google APIs Intel x86 Atom, android-27 | Google Play Intel x86 Atom, android-28 | Google Play Intel x86 Atom
  IDEs:
    Android Studio: 3.5 AI-191.8026.42.35.5791312
    Xcode: 10.3/10G8 - /usr/bin/xcodebuild
  npmPackages:
    react: 16.8.6 => 16.8.6
    react-native: 0.60.5 => 0.60.5
  npmGlobalPackages:
    react-native-cli: 2.0.1
```

## Description

Android Studio synchronization fails with the following error:

```
ERROR: Text must not be null or empty

org.gradle.internal.exceptions.LocationAwareException: Script '.../node_modules/@react-native-community/cli-platform-android/native_modules.gradle' line: 191
A problem occurred evaluating settings '...'.
  at org.gradle.initialization.exception.DefaultExceptionAnalyser.transform(DefaultExceptionAnalyser.java:99)
  at org.gradle.initialization.exception.DefaultExceptionAnalyser.collectFailures(DefaultExceptionAnalyser.java:65)
  at org.gradle.initialization.exception.MultipleBuildFailuresExceptionAnalyser.transform(MultipleBuildFailuresExceptionAnalyser.java:47)
  at org.gradle.initialization.exception.StackTraceSanitizingExceptionAnalyser.transform(StackTraceSanitizingExceptionAnalyser.java:29)
  at org.gradle.initialization.DefaultGradleLauncher.finishBuild(DefaultGradleLauncher.java:174)
[...]

PR #741, opened 2019-09-22
propagate react-native config error stream to logs

generalpiston:
Use output of subprocess in gradle file instead of input

Summary:
---------

See https://github.com/react-native-community/cli/issues/740.

Test Plan:
----------

Android Studio synchronizes.

### DIFF
--- a/packages/platform-android/native_modules.gradle
+++ b/packages/platform-android/native_modules.gradle
@@ -188,7 +188,7 @@ class ReactNativeModules {
 
     try {
       cmdProcess = Runtime.getRuntime().exec(command, null, root)
-      def bufferedReader = new BufferedReader(new InputStreamReader(cmdProcess.getOutputStream()))
+      def bufferedReader = new BufferedReader(new InputStreamReader(cmdProcess.getInputStream()))
       def buff = ""
       def readBuffer = new StringBuffer()
       while ((buff = bufferedReader.readLine()) != null){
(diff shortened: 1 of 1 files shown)


packages/platform-android/native_modules.gradle
```diff
@@ -188,7 +188,7 @@ class ReactNativeModules {
 
     try {
       cmdProcess = Runtime.getRuntime().exec(command, null, root)
-      def bufferedReader = new BufferedReader(new InputStreamReader(cmdProcess.getInputStream()))
+      def bufferedReader = new BufferedReader(new InputStreamReader(cmdProcess.getOutputStream()))
```
thymikee: Why not `OutputStreamReader` instead of `InputStreamReader` then?

generalpiston: AFAIK that doesn't exist.

zxxlcc:
The problem I encountered is also like this, how can I solve it? I tried to replace it with 'getOutputStream()', but I get the following error:
image

thymikee:
@zxxlcc you need to run [jetifier](https://github.com/mikehardy/jetifier). 
However, please don't spam under a PR and please [update the `@react-native-community/cli`](https://github.com/react-native-community/cli#updating-the-cli) to the latest version where it runs automatically when calling `run-android`.

packages/platform-android/native_modules.gradle
Taym95: I think you mean `OutputStreamWriter`?

Taym95: > Why not `OutputStreamReader` instead of `InputStreamReader` then?

I think you mean `OutputStreamWriter`?

generalpiston: Hmm I see the problem with this patch. getInputStream yields the output stream of the subprocess.

generalpiston:
It turns out that node was throwing an exception. The environment of the groovy process was using a version of node that I was not.

Let's pipe the contents of the error stream to the parent process instead?

PR #742, opened 2019-09-23
[doctor command] Add tests for androidSDK and androidNDK

thib92:
Summary:
---------

Add unit tests for the healthchecks for `androidSDK` and `androidNDK`

Also removed a `$FlowFixMe` that wasn't removed when the `androidSDK` file was moved to TS.

Related to #694 

Test Plan:
----------

CI is green

### DIFF
--- a/packages/cli/src/commands/doctor/healthchecks/__tests__/androidNDK.test.ts
+++ b/packages/cli/src/commands/doctor/healthchecks/__tests__/androidNDK.test.ts
@@ -1,0 +1,66 @@
+import androidNDK from '../androidNDK';
+import getEnvironmentInfo from '../../../../tools/envinfo';
+import {EnvironmentInfo} from '../../types';
+import {NoopLoader} from '../../../../tools/loader';
+
+import * as common from '../common';
+
+const logSpy = jest.spyOn(common, 'logManualInstallation');
+
+describe('androidHomeEnvVariables', () => {
+  let initialEnvironmentInfo: EnvironmentInfo;
+  let environmentInfo: EnvironmentInfo;
+
+  beforeAll(async () => {
+    initialEnvironmentInfo = await getEnvironmentInfo();
+  });
+
+  beforeEach(() => {
+    environmentInfo = initialEnvironmentInfo;
+  });
+
+  afterEach(() => {
+    jest.resetAllMocks();
+  });
+
+  it('returns a message if the Android SDK is not installed', async () => {
+    environmentInfo.SDKs['Android SDK'] = 'Not Found';
+    const diagnostics = await androidNDK.getDiagnostics(environmentInfo);
+    expect(typeof diagnostics.needsToBeFixed).toBe('string');
+  });
+
+  it('returns a message if the Android NDK is not installed', async () => {
+    // @ts-ignore
+    environmentInfo.SDKs['Android SDK'] = {
+      'Android NDK': 'Not Found',
+    };
+    const diagnostics = await androidNDK.getDiagnostics(environmentInfo);
+    expect(typeof diagnostics.needsToBeFixed).toBe('string');
+  });
+
+  it('returns a message if the NDK version is not in range', async () => {
+    // @ts-ignore
+    environmentInfo.SDKs['Android SDK'] = {
+      'Android NDK': '18',
+    };
+    const diagnostics = await androidNDK.getDiagnostics(environmentInfo);
+    expect(typeof diagnostics.needsToBeFixed).toBe('string');
+  });
+
+  it('returns false if the NDK version is in range', async () => {
+    // @ts-ignore
+    environmentInfo.SDKs['Android SDK'] = {
+      'Android NDK': '19',
+    };
+    const diagnostics = await androidNDK.getDiagnostics(environmentInfo);
+    expect(diagnostics.needsToBeFixed).toBe(false);
+  });
+
+  it('logs manual installation steps to the screen', async () => {
+    const loader = new NoopLoader();
+
+    androidNDK.runAutomaticFix({loader, environmentInfo});
+
+    expect(logSpy).toHaveBeenCalledTimes(1);
+  });
+});
--- a/packages/cli/src/commands/doctor/healthchecks/__tests__/androidNDK.test.ts
+++ b/packages/cli/src/commands/doctor/healthchecks/__tests__/androidNDK.test.ts
@@ -7,7 +7,7 @@ import * as common from '../common';
 
 const logSpy = jest.spyOn(common, 'logManualInstallation');
 
-describe('androidHomeEnvVariables', () => {
+describe('androidNDK', () => {
   let initialEnvironmentInfo: EnvironmentInfo;
   let environmentInfo: EnvironmentInfo;
 
@@ -56,7 +56,7 @@ describe('androidHomeEnvVariables', () => {
     expect(diagnostics.needsToBeFixed).toBe(false);
   });
 
-  it('logs manual installation steps to the screen', async () => {
+  it('logs manual installation steps to the screen', () => {
     const loader = new NoopLoader();
 
     androidNDK.runAutomaticFix({loader, environmentInfo});
--- a/packages/cli/src/commands/doctor/healthchecks/__tests__/androidNDK.test.ts
+++ b/packages/cli/src/commands/doctor/healthchecks/__tests__/androidNDK.test.ts
@@ -30,6 +30,7 @@ describe('androidNDK', () => {
   });
 
   it('returns a message if the Android NDK is not installed', async () => {
+    // To avoid having to provide fake versions for all the Android SDK tools
     // @ts-ignore
     environmentInfo.SDKs['Android SDK'] = {
       'Android NDK': 'Not Found',
@@ -39,6 +40,7 @@ describe('androidNDK', () => {
   });
 
   it('returns a message if the NDK version is not in range', async () => {
+    // To avoid having to provide fake versions for all the Android SDK tools
     // @ts-ignore
     environmentInfo.SDKs['Android SDK'] = {
       'Android NDK': '18',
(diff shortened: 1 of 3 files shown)


thib92:
@lucasbento tests are failing on Windows because apparently the `androidSDK` check specific to Windows is not responding within 5s.

I'm not too sure why but I'm suspecting the `execa` call here is not responding. Should we put a timeout on it?

https://github.com/react-native-community/cli/blob/ed0a6fd419dba778e5ea737a040092afdda51c47/packages/cli/src/commands/doctor/healthchecks/androidSDK.ts#L24-L29

lucasbento:
@thib92: I would say so, it seems like a command that can take quite a while to respond.

I'm not sure if 5 seconds would be sufficient though, @thymikee can you tell us about that?

I didn't actually do this piece of code so I am not sure about it.

thymikee:
I'd mock a response that we expect, no time for long timeouts.

lucasbento:
Yes, that's definitely a better approach 😄

packages/cli/src/commands/doctor/healthchecks/__tests__/androidNDK.test.ts
```diff
@@ -0,0 +1,66 @@
+import androidNDK from '../androidNDK';
+import getEnvironmentInfo from '../../../../tools/envinfo';
+import {EnvironmentInfo} from '../../types';
+import {NoopLoader} from '../../../../tools/loader';
+
+import * as common from '../common';
+
+const logSpy = jest.spyOn(common, 'logManualInstallation');
+
+describe('androidHomeEnvVariables', () => {
```
lucasbento: Need to change this to `androidNDK`.

packages/cli/src/commands/doctor/healthchecks/__tests__/androidNDK.test.ts
```diff
@@ -0,0 +1,66 @@
...
+  });
+
+  afterEach(() => {
+    jest.resetAllMocks();
+  });
+
+  it('returns a message if the Android SDK is not installed', async () => {
+    environmentInfo.SDKs['Android SDK'] = 'Not Found';
+    const diagnostics = await androidNDK.getDiagnostics(environmentInfo);
+    expect(typeof diagnostics.needsToBeFixed).toBe('string');
```
lucasbento: This has actually been changed on #743, you can check the change for `androidHomeEnvVariable` [here](https://github.com/react-native-community/cli/pull/743/files#diff-9c043f20be368d1006196f0b0cb243c2R24).

lucasbento: Just to clarify, this field was being used to hold a boolean and a string, the string was used as a `description` sorta of thing, on #743 I actually created the `description` field.

packages/cli/src/commands/doctor/healthchecks/__tests__/androidSDK.test.ts
```diff
@@ -0,0 +1,78 @@
+import androidSDK from '../androidSDK';
+import getEnvironmentInfo from '../../../../tools/envinfo';
+import {EnvironmentInfo} from '../../types';
+import {NoopLoader} from '../../../../tools/loader';
+
+import * as common from '../common';
+
+const logSpy = jest.spyOn(common, 'logManualInstallation');
+
+describe('androidHomeEnvVariables', () => {
```
lucasbento: Need to change this to `androidSDK`.

packages/cli/src/commands/doctor/healthchecks/__tests__/androidSDK.test.ts
```diff
@@ -0,0 +1,78 @@
...
+    environmentInfo.SDKs['Android SDK'] = 'Not Found';
+    const diagnostics = await androidSDK.getDiagnostics(environmentInfo);
+
+    jest.mock('execa', () => ({
+      default: () => {
+        throw new Error('No SDK found');
+      },
+    }));
+
+    expect(typeof diagnostics.needsToBeFixed).toBe('string');
```
lucasbento: Same as the other one.

packages/cli/src/commands/doctor/healthchecks/__tests__/androidSDK.test.ts
```diff
@@ -0,0 +1,78 @@
...
+
+  afterEach(() => {
+    jest.resetAllMocks();
+  });
+
+  it('returns a message if the Android SDK is not installed', async () => {
+    environmentInfo.SDKs['Android SDK'] = 'Not Found';
+    const diagnostics = await androidSDK.getDiagnostics(environmentInfo);
+
+    jest.mock('execa', () => ({
```
lucasbento: I suggest mocking this at the top of the file, you can do:

```js
jest.mock('execa', () => ({
  default: jest.fn(),
});
```

And then, on each test case:

```js
execa.mockReturnValueOnce('my value');
```

I believe this would work.

packages/cli/src/commands/doctor/healthchecks/__tests__/androidSDK.test.ts
```diff
@@ -0,0 +1,78 @@
...
+        return 'build-tools;25';
+      },
+    }));
+
+    const diagnostics = await androidSDK.getDiagnostics(environmentInfo);
+    expect(typeof diagnostics.needsToBeFixed).toBe('string');
+  });
+
+  it('returns false if the SDK version is in range', async () => {
+    // @ts-ignore
```
lucasbento: Why is this needed?

packages/cli/src/commands/doctor/healthchecks/__tests__/androidSDK.test.ts
```diff
@@ -0,0 +1,78 @@
...
+      default: () => {
+        return 'build-tools;27';
+      },
+    }));
+
+    const diagnostics = await androidSDK.getDiagnostics(environmentInfo);
+    expect(diagnostics.needsToBeFixed).toBe(false);
+  });
+
+  it('logs manual installation steps to the screen', async () => {
```
lucasbento: This doesn't have to be `async` as it's not `await`ing anything.

packages/cli/src/commands/doctor/healthchecks/__tests__/androidSDK.test.ts
thib92: Thanks! That worked 😄 
I'm not an export on Jest mocks 🙄

packages/cli/src/commands/doctor/healthchecks/__tests__/androidSDK.test.ts
thib92: Good catch!

packages/cli/src/commands/doctor/healthchecks/__tests__/androidSDK.test.ts
thib92: Because the `Android SDK` key of envinfo needs more than just `Build Tools`, but I didn't want to fake version for everything it provides, as they are not needed by the `androidSDK` healthcheck.

packages/cli/src/commands/doctor/healthchecks/__tests__/androidNDK.test.ts
thib92: Okay nice one. I like it like that!

thib92:
@lucasbento thanks for the good review. Everything should be fixed.

thib92:
Btw I'm doing a lot of copy-paste between all the healthcheck tests. Once I'm done with all of them, I'll see if I can create util functions to avoid repeating code in the tests.

lucasbento:
@thymikee: can you give a review?

packages/cli/src/commands/doctor/healthchecks/__tests__/androidSDK.test.ts
```diff
@@ -0,0 +1,66 @@
...
+import getEnvironmentInfo from '../../../../tools/envinfo';
+import {EnvironmentInfo} from '../../types';
+import {NoopLoader} from '../../../../tools/loader';
+
+import * as common from '../common';
+
+const logSpy = jest.spyOn(common, 'logManualInstallation');
+
+const mockExeca = (stdout: string) => {
+  jest.mock('execa', () => jest.fn());
```
thymikee: `jest.mock` is hosited to the top of the parent scope, it's not meant to be called mutliple times. Can you move there so it's a bit more obvious the `execa` _module_ is mocked?

packages/cli/src/commands/doctor/healthchecks/__tests__/androidSDK.test.ts
```diff
@@ -0,0 +1,66 @@
...
+import androidSDK from '../androidSDK';
+import getEnvironmentInfo from '../../../../tools/envinfo';
+import {EnvironmentInfo} from '../../types';
+import {NoopLoader} from '../../../../tools/loader';
+
+import * as common from '../common';
+
+const logSpy = jest.spyOn(common, 'logManualInstallation');
+
+const mockExeca = (stdout: string) => {
```
thymikee: Let's rename it to `mockSdkManagerResponse` or something so that it's closer to what it's actually doing, since we don't extract execa call to a function we could mock – which would make it obvious what is mocked.

packages/cli/src/commands/doctor/healthchecks/__tests__/androidSDK.test.ts
thib92: Changed it a little bit to be easier to understand and more standard. What do you think?

packages/cli/src/commands/doctor/healthchecks/__tests__/androidSDK.test.ts
thib92: See what I changed. What do you think?

PR #746, opened 2019-09-24
doctor: improve error messages

lucasbento:
Summary:
---------

This is related to #694, it improves the error messages by maintaining consistency between the health checks with one separated function to log errors.

It also fixes a problem introduced on #743 where it shows `undefined` when the `description` is not provided on a health check.

### Before

Before

### After

After with consistent error messages

Test Plan:
----------

1. Make a health check fail;
1. `/path/to/cli doctor`.

### DIFF
--- a/packages/cli/src/commands/doctor/healthchecks/androidHomeEnvVariable.ts
+++ b/packages/cli/src/commands/doctor/healthchecks/androidHomeEnvVariable.ts
@@ -18,7 +18,7 @@ const platform = process.platform as 'darwin' | 'win32' | 'linux';
 
 const message = `Read more about how to set the ${label} at ${chalk.dim(
   URLS[platform],
-)}.`;
+)}`;
 
 export default {
   label,
--- a/packages/cli/src/commands/doctor/healthchecks/androidHomeEnvVariable.ts
+++ b/packages/cli/src/commands/doctor/healthchecks/androidHomeEnvVariable.ts
@@ -26,7 +26,7 @@ export default {
     needsToBeFixed: !process.env.ANDROID_HOME,
   }),
   runAutomaticFix: async ({loader}: {loader: Ora}) => {
-    loader.info();
+    loader.fail();
 
     logManualInstallation({
       message,
--- a/packages/cli/src/commands/doctor/healthchecks/common.ts
+++ b/packages/cli/src/commands/doctor/healthchecks/common.ts
@@ -2,13 +2,28 @@ import chalk from 'chalk';
 import readline from 'readline';
 import wcwidth from 'wcwidth';
 import stripAnsi from 'strip-ansi';
+import {Ora} from 'ora';
 import {logger} from '@react-native-community/cli-tools';
 
 // Space is necessary to keep correct ordering on screen
-const logMessage = (message: string) => logger.log(`   ${message}`);
+const logMessage = (message?: string) => {
+  const indentation = '   ';
+
+  if (typeof message !== 'string') {
+    logger.log();
+
+    return;
+  }
+
+  const messageByLine = message.split('\n');
+
+  return logger.log(`${indentation}${messageByLine.join(`\n${indentation}`)}`);
+};
+
+const addBlankLine = () => logMessage();
 
 const logManualInstallation = ({
-  healthcheck = '',
+  healthcheck,
   url,
   command,
   message,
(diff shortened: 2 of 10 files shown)


packages/cli/src/commands/doctor/healthchecks/androidHomeEnvVariable.ts
```diff
@@ -18,15 +18,15 @@ const platform = process.platform as 'darwin' | 'win32' | 'linux';
...
+)}`;
 
 export default {
   label,
   getDiagnostics: async () => ({
     needsToBeFixed: !process.env.ANDROID_HOME,
   }),
   runAutomaticFix: async ({loader}: {loader: Ora}) => {
-    loader.info();
+    loader.fail();
```
thymikee: why logger.fail?

lucasbento: Consistency, we can't fix that automatically.

The other health checks do a `fail` when logging how to manually fix it when they try to fix something automatically and it doesn't succeed, UI looks really strange with two messages that have the same content but with different loaders.

thymikee: Can't we move this call up? `runAutomaticFix` is a generic interface which we run in a single place, I think it just makes more sense to call `loader.fail()` once there, instead of thinking about it in every fix fn, or I don't see something?

lucasbento: I think it was like that originally, I have changed it over time because the health check has to have full control of what is displayed on the screen once it's started IMO.

We could throw an error here and catch in on `doctor.ts` but I am not sure if that would be an improvement.

packages/cli/src/commands/doctor/healthchecks/common.ts
```diff
@@ -2,13 +2,28 @@ import chalk from 'chalk';
...
+
+    return;
+  }
+
+  const messageByLine = message.split('\n');
+
+  return logger.log(`${indentation}${messageByLine.join(`\n${indentation}`)}`);
+};
+
+const addBlankLine = () => logMessage();
```
thymikee: This is adding 2 blank lines (it shows on the screenshot a well), can we call empty `logger.log()`?

lucasbento: It does call `logger.log()` on line `13` as no `message` is provided.

thymikee: With extra newline, resulting in 2 newlines instead of one.

thymikee: Oh, I missed that, sorry!

packages/cli/src/commands/doctor/healthchecks/common.ts
```diff
@@ -23,20 +38,58 @@ const logManualInstallation = ({
...
+
+    return;
+  }
+
+  logMessage(
+    `The error above occured while trying to install ${healthcheck}. Please try again manually: ${chalk.bold(
+      command,
+    )}`,
+  );
+  addBlankLine();
```
thymikee: I think this is the one that causes extra space which is not necessary imho

lucasbento: Which space do you mean? the one before `iOS` or the one after `iOS` is done?

thymikee: image

lucasbento: That's because there's a space after each health check category (in this case `Common` and `iOS`) and also a space after each error is shown, that's why that extra space is there.

To be sincere I actually tried to get rid of it but there isn't an easy way to do at the moment, if we remove this last space then it can get to a funky UI when the previous health check succeeded but the next one failed.

lucasbento: The best way to do this is to actually check if the last health check failed, if so then print only one space instead of two.

thymikee: Let's figure this out later, this is a very minor inconvenience

lucasbento: Thanks for commenting about this though!

PR #757, opened 2019-09-29
chore: convert some server files into TypeScript

Naturalclar:
Summary:
---------

This PR is part of #683 
Converted some of server related files into TypeScript

Test Plan:
----------

- yarn
- yarn lint
- yarn flow-check


packages/cli/package.json
```diff
@@ -30,6 +30,8 @@
     "@react-native-community/cli-types": "^3.0.0-alpha.2",
     "@types/mkdirp": "^0.5.2",
     "@types/semver": "^6.0.2",
+    "@types/ws": "^6.0.3",
```
thymikee: We use `ws@1`, shouldn't we match types? Ideally, we'd like to migrate to the newest version, but I'm not sure if we're able to do it without syncing Metro `ws` version as well

packages/cli/src/commands/server/launchDebugger.ts
```diff
@@ -69,6 +68,7 @@ function getChromeAppName(): string {
 }
 
 function launchChrome(url: string) {
+  // @ts-ignore open's callback argument definition is not defined in open
   open(url, {app: [getChromeAppName()]}, err => {
```
thymikee: Maybe using ąsync/await syntax would help with the typings? Based on the readme this is what they recommend

packages/cli/src/commands/server/launchDefaultBrowser.ts
```diff
@@ -5,14 +5,14 @@
...
- * @flow
  */
 
 import open from 'open';
 import {logger} from '@react-native-community/cli-tools';
 
 function launchDefaultBrowser(url: string) {
-  open(url, err => {
+  // @ts-ignore open's second argument takes an option, not callback
+  open(url, (err: string) => {
```
thymikee: async/await maybe?

packages/cli/src/commands/server/messageSocket.ts
```diff
@@ -192,12 +216,14 @@ function attachToServer(server, path) {
 
     clients.set(clientId, clientWs);
     const onCloseHandler = () => {
+      // @ts-ignore
       clientWs.onmessage = null;
```
thymikee: likely because types version mismatch

packages/cli/src/commands/server/webSocketProxy.ts
```diff
@@ -9,18 +9,22 @@
 
 import ws from 'ws';
 import {logger} from '@react-native-community/cli-tools';
+import {Server as HttpServer} from 'http';
+import {Server as HttpsServer} from 'https';
+import WebSocket from 'ws';
```
thymikee: `ws` is imported twice, can we do it one?

thymikee:
cc @Esemesek

packages/cli/package.json
Naturalclar: It seems like the type definition for `ws@1` does not exist in DefinitelyTyped. The closest I could find was `3.0.0` or `0.0.42`
I'll look into this further to see if I can improve with typings.

Naturalclar:
Thanks for the feedback! I've made some changes to improve with typings and removed some `ts-ignore`s.

packages/cli/src/commands/server/webSocketProxy.ts
```diff
@@ -54,20 +57,27 @@ function attachToServer(server, path) {
...
-      debuggerSocket.onmessage = ({data}) => send(clientSocket, data);
+      if (debuggerSocket) {
+        debuggerSocket.onerror = debuggerSocketCloseHandler;
+        debuggerSocket.onclose = debuggerSocketCloseHandler;
+        debuggerSocket.onmessage = ({data}) => send(clientSocket, data);
+      }
     } else if (url.indexOf('role=client') > -1) {
       if (clientSocket) {
+        // @ts-ignore current typing of websocket does not expect null for onerror
         clientSocket.onerror = null;
```
grabbou: Would `() => {}` do the trick?

Naturalclar: it did indeed! thanks!

packages/cli/src/commands/server/launchDebugger.ts
```diff
@@ -46,8 +45,14 @@ function getChromeAppName(): string {
   }
 }
 
-function launchChrome(url: string) {
-  return open(url, {app: [getChromeAppName()], wait: true});
+async function launchChrome(url: string) {
```
thymikee: Please revert this whole change. It introduces a regression, since we don't want to catch this exception here

packages/cli/src/commands/server/launchEditor.ts
```diff
@@ -10,11 +10,34 @@
...
+  | 'wstorm'
+  | 'appcode'
+  | 'charm'
+  | 'idea'
+  | 'joe'
+  | 'emacs'
+  | 'emacsclient'
+  | 'rmate'
+  | 'mate'
+  | 'mine';
```
thymikee: Nah, we don't have resources to maintain this list. Please use just string

packages/cli/src/commands/server/webSocketProxy.ts
```diff
@@ -54,20 +57,24 @@ function attachToServer(server, path) {
...
+        clientSocket.onerror = () => {};
+        clientSocket.onclose = () => {};
+        clientSocket.onmessage = () => {};
         clientSocket.close(1011, 'Another client connected');
       }
       clientSocket = connection;
-      clientSocket.onerror = clientSocketCloseHandler;
-      clientSocket.onclose = clientSocketCloseHandler;
-      clientSocket.onmessage = ({data}) => send(debuggerSocket, data);
+      if (clientSocket) {
```
thymikee: we assume `connection` is defined on line 52, let's skip this check if possible

packages/cli/src/commands/server/webSocketProxy.ts
```diff
@@ -54,20 +57,24 @@ function attachToServer(server, path) {
...
+        debuggerSocket.onmessage = ({data}) => send(clientSocket, data);
+      }
     } else if (url.indexOf('role=client') > -1) {
       if (clientSocket) {
-        clientSocket.onerror = null;
-        clientSocket.onclose = null;
-        clientSocket.onmessage = null;
+        clientSocket.onerror = () => {};
+        clientSocket.onclose = () => {};
+        clientSocket.onmessage = () => {};
```
thymikee: I don't think these reassignments are necessary, because they're overwritten right after. It only makes sense to `null` them when `clientSocket.close` throws for some reason and we try to recover – the handlers could leak memory because `clientSocket` would hold them. I'm not sure if that happens, didn't have time to go through this deeply.

For now, let's revert to what was here before and ts-ignore it with a todo comment to rewrite, once we migrate to newer version of ws library.

packages/cli/src/commands/server/webSocketProxy.ts
Naturalclar: changed the type of `connection` in this function argument and removed this check

thymikee:
Thank you!

Issue #739, opened 2019-09-22
"iPhone X" simulator is disabled


Resten1497:
React Native version: 0.60.4

## Steps to reproduce

<img width="626" alt="스크린샷 2019-09-22 오후 10 17 33" src="https://user-images.githubusercontent.com/19689492/65388500-54877e00-dd87-11e9-9b41-a645a5aa5cbb.png">

The Xcode is updated version 11 and the "iPhone X" simulator is disabled and cannot be run with an existing "react-native run-ios".

ScreamZ:
F*ck apple:

Workaround solution awaiting fix : 

`yarn react-native run-ios --simulator "iPhone 11 Pro Max"`
`npx react-native run-ios --simulator "iPhone 11 Pro Max"`
`react-native run-ios --simulator "iPhone 11 Pro Max"`
`./node_modules/.bin/react-native run-ios --simulator "iPhone 11 Pro Max"`

thymikee:
The message I got from @bartolkarouza having this issue:
> The problem was that I had the iPhone X sim open _during_ the xcode upgrade. This leaves it in a weird unupgraded state and it becomes unreachable to Xcode or the sim cli.

Have you tried restarting Xcode _and_ Simulator as well?

satya164:
I just upgraded to Catalina and XCode 11 today and I have only iPhone 11 simulators. I didn't have iPhone X simulator open during the upgrade.

<img width="659" alt="Screenshot 2019-09-22 at 7 55 30 PM" src="https://user-images.githubusercontent.com/1174278/65392256-06687f80-dd73-11e9-8535-f9e94ef2f7f4.png">

thymikee:
@satya164 this seems like a bug in Simulator, likely because Catalina is still in beta

satya164:
@thymikee XCode 11 is stable though

thymikee:
Hah, this is the list I have, and surprisingly no iPhone X there :D  
![image](https://user-images.githubusercontent.com/5106466/65392431-31ec6980-dd75-11e9-9eb2-3098c3c66792.png)

I guess we'll need to change the logic to:

1. try "iPhone 11"
1. try "iPhone X"
1. try "iPhone 8"
1. bail

What do you think, makes sense?

wbadam:
Another workaround is to add back the removed iPhone X in Simulator's Hardware > Device > Manage Devices menu

sasurau4:
[...]

PR #758, opened 2019-09-29
feat: add fallback behavior when running a default simulator on iOS

Naturalclar:
Summary:
---------

This is an attempt to solve problem on #739, based on idea provided at https://github.com/react-native-community/cli/issues/739#issuecomment-533904514

This changes the behavior of runIOS when no simulator argument is provided.

When no simulator argument is provided, the cli will look for simulators in the following order
- `iPhone 11`
- `iPhone X`
- `iPhone 8`

Test Plan:
----------
yarn
yarn test 
yarn lint

### DIFF
--- a/packages/platform-ios/src/commands/runIOS/index.ts
+++ b/packages/platform-ios/src/commands/runIOS/index.ts
@@ -129,8 +129,7 @@ async function runOnSimulator(
   }
 
   /**
-   * If no simulator was chosen, try simulators in following order
-   * - iPhone 11
+   * If provided simulator does not exist, try simulators in following order
    * - iPhone X
    * - iPhone 8
    */
--- a/packages/platform-ios/src/commands/runIOS/index.ts
+++ b/packages/platform-ios/src/commands/runIOS/index.ts
@@ -128,37 +128,20 @@ async function runOnSimulator(
     throw new CLIError('Could not parse the simulator list output', error);
   }
 
-  let simulator: string = args.simulator;
-  let selectedSimulator: {
-    udid: string;
-    name: string;
-    booted: boolean;
-    version: string;
-  } | null = null;
-
   /**
    * If no simulator was chosen, try simulators in following order
    * - iPhone 11
    * - iPhone X
    * - iPhone 8
    */
-  const defaultSimulators = ['iPhone 11', 'iPhone X', 'iPhone 8'];
-  if (!simulator) {
-    defaultSimulators.some(target => {
-      selectedSimulator = findMatchingSimulator(simulators, target);
-      return selectedSimulator;
-    });
 
-    if (!selectedSimulator) {
-      throw new CLIError(
-        'Could not find any of the following simulators: "iPhone 11" "iPhone X "iPhone 8"',
-      );
-    }
-  } else {
-    selectedSimulator = findMatchingSimulator(simulators, simulator);
-    if (!selectedSimulator) {
-      throw new CLIError(`Could not find "${simulator}" simulator`);
-    }
+  const fallbackSimulators = ['iPhone X', 'iPhone 8'];
+  const selectedSimulator = fallbackSimulators.reduce((simulator, fallback) => {
+    return simulator || findMatchingSimulator(simulators, fallback);
+  }, findMatchingSimulator(simulators, args.simulator));
+
+  if (!selectedSimulator) {
+    throw new CLIError(`Could not find "${args.simulator}" simulator`);
   }
 
   /**
@@ -520,6 +503,7 @@ export default {
       description:
         'Explicitly set simulator to use. Optionally include iOS version between' +
         'parenthesis at the end to match an exact version: "iPhone 6 (10.0)"',
+      default: 'iPhone 11',
     },
     {
       name: '--configuration [string]',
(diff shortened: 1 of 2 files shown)


packages/platform-ios/src/commands/runIOS/index.ts
```diff
@@ -492,7 +507,7 @@ export default {
       description:
         'Explicitly set simulator to use. Optionally include iOS version between' +
         'parenthesis at the end to match an exact version: "iPhone 6 (10.0)"',
-      default: 'iPhone X',
+      default: '',
```
thymikee: I think we can default to undefined (or just don't provide the default) – can we change that and adjust docs as well?

packages/platform-ios/src/commands/runIOS/index.ts
```diff
@@ -128,9 +128,24 @@ async function runOnSimulator(
...
+  let simulator: string = args.simulator;
+  if (!simulator) {
+    simulator = findMatchingSimulator(simulators, 'iPhone 11')
+      ? 'iPhone 11'
+      : findMatchingSimulator(simulators, 'iPhone X')
+      ? 'iPhone X'
+      : 'iPhone 8';
+  }
+
+  const selectedSimulator = findMatchingSimulator(simulators, simulator);
```
thymikee: this is calling `findMatchingSimulator` twice for iPhone 11 and iPhone X. Can we move this call to line 143 (iPhone 8) and get rid of `selectedSimulator` variable?

Naturalclar:
Thanks for the feedback! I've made some modification to reduce the number of times `findMatchingSimulator` run

packages/platform-ios/src/commands/runIOS/index.ts
```diff
@@ -128,9 +128,40 @@ async function runOnSimulator(
...
+    version: string;
+  } | null = null;
+
+  /**
+   * If no simulator was chosen, try simulators in following order
+   * - iPhone 11
+   * - iPhone X
+   * - iPhone 8
+   */
+  if (!simulator) {
```
grabbou: I wonder if this could be replaced with a simple `while` loop or a `for` loop and `break` statement. Thoughts?

Naturalclar: thanks for the suggestion!
I modified it to use `array.some` so that loop would break when simulator is found.

packages/platform-ios/src/commands/runIOS/index.ts
```diff
@@ -128,9 +128,37 @@ async function runOnSimulator(
...
+    });
+
+    if (!selectedSimulator) {
+      throw new CLIError(
+        'Could not find any of the following simulators: "iPhone 11" "iPhone X "iPhone 8"',
+      );
+    }
+  } else {
+    selectedSimulator = findMatchingSimulator(simulators, simulator);
+    if (!selectedSimulator) {
```
grabbou: I believe we can move this and another `if` statement at line `152` outside of this block. If we would place it below line 163, we would be able to keep only one `if` statement, instead of the two.

packages/platform-ios/src/commands/runIOS/index.ts
grabbou: That is a good improvement, but still, I find doing side effects inside `some` a bit of a code smell.

How about:
```js
for (const defaultSimulator of defaultSimulators) {
  selectedSimulator = findMatchingSimulator(simulators, defaultSimulator);
  if (selectedSimulator) {
     break;
  }
}
```
or
```js
const selectedSimulator = args.simulator 
   ? findMatchingSimulator(simulators, args.simulator)
   : defaultSimulators.reduce(
       (selectedSimulator, simulator) => selectedSimulator || findMatchingSimulator(simulators, simulator),
       null
     );
```

The below (in my opinion) would be the most concise:
```js
const selectedSimulator = defaultSimulators.reduce(
   (selectedSimulator, simulator) => selectedSimulator || findMatchingSimulator(simulators, simulator),
   findMatchingSimulator(simulators, args.simulator)
);
```
but it would require we bring: `args.simulator` default value to be `iPhone 11` and rename `defaultSimulators` to be `fallbackSimulators` (in other words, try these ones in case user provided/default value does not exist on the machine).

Leaving that up to further discussion.

Naturalclar: Thank you for the feedback and suggestions!
I find doing side effects inside `some` feels iffy as well, but was not able to come up with a solution at the moment.

I like the last suggestion of providing fallback when `args.simulator` is not available. It let me remove multiple `if` s.

I've modified the default simulator to be `iPhone 11` and modified the document to match the code.

packages/platform-ios/src/commands/runIOS/index.ts
```diff
@@ -128,7 +128,17 @@ async function runOnSimulator(
...
+  /**
+   * If provided simulator does not exist, try simulators in following order
+   * - iPhone X
+   * - iPhone 8
+   */
+
+  const fallbackSimulators = ['iPhone X', 'iPhone 8'];
+  const selectedSimulator = fallbackSimulators.reduce((simulator, fallback) => {
+    return simulator || findMatchingSimulator(simulators, fallback);
+  }, findMatchingSimulator(simulators, args.simulator));
```
thymikee: This loop always run through all simulators from `fallbackSimulators`, but since the data set is small and we avoid running `findMatchingSimulator` if necessary, I'm fine with that.

PR #763, opened 2019-10-01
doctor: show required versions

lucasbento:
Summary:
---------

This PR is related to #694 and it:

1. Shows the required current and required version for a failed health check;
1. Changes the wording from `to fix all issues` to `to try to fix all issues;
1. Fixes a problem where if you type a character not recognised on the usage it just doesn't listen to the correct characters anymore;
1. Puts the `description` content below the health check name rather than on the right side.

### Preview

Preview

Test Plan:
----------

Run `doctor`.

### DIFF
--- a/packages/cli/src/commands/doctor/healthchecks/androidSDK.ts
+++ b/packages/cli/src/commands/doctor/healthchecks/androidSDK.ts
@@ -11,7 +11,7 @@ const installMessage = `Read more about how to update Android SDK at ${chalk.dim
 
 export default {
   label: 'Android SDK',
-  description: 'required for building and installing your app on Android',
+  description: 'Required for building and installing your app on Android',
   getDiagnostics: async ({SDKs}) => {
     let sdks = SDKs['Android SDK'];
 
@@ -47,11 +47,15 @@ export default {
       } catch {}
     }
 
+    const version = sdks === 'Not Found' ? sdks : sdks['Build Tools'][0];
+
     return {
+      version,
+      versionRange: versionRanges.ANDROID_SDK,
       needsToBeFixed:
         sdks === 'Not Found' ||
         doesSoftwareNeedToBeFixed({
-          version: sdks['Build Tools'][0],
+          version,
           versionRange: versionRanges.ANDROID_SDK,
         }),
     };
--- a/packages/cli/src/commands/doctor/types.ts
+++ b/packages/cli/src/commands/doctor/types.ts
@@ -96,6 +96,8 @@ export type HealthCheckInterface = {
 export type HealthCheckResult = {
   label: string;
   needsToBeFixed: boolean;
+  version?: 'Not Found' | string;
+  versionRange?: string;
   description: string | undefined;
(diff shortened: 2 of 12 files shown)


thymikee:
> Shows the required current and required version for a failed health check

I can see how that looks nice with colors, but please notice that this command will be ran in case of issues, likely by inexperienced RN users. We need to be more descriptive than showing an arrow. I'd rather like to give it a human-readable text like "We found vX, however we need vY".

> Puts the description content below the health check name rather than on the right side.

I'm not a fan of this change. Initially when I read the "Watchman" and its description below, I was confused where it belongs. Let's either keep it in the same line (why moving it? it fits nicely even with version), or indent it one level or do something else.

How about this UI:

```
Android
  v ANDROID_HOME
  x Android SDK - Required for building and installing your app on Android devices.
    - found at: <dim>/Users/user/Android/SDK
    - version:  <red>v23.0.1
    - to fix: 	<bold>upgrade to v26.0.0. Here's how: <dim link>
  x Android NDX - description
    - version: 	<red>not available
    - to fix: 	<bold>install v19.0.0. Here's how: <dim link>
```

packages/cli/src/commands/doctor/types.ts
```diff
@@ -89,13 +89,19 @@ export type HealthCheckInterface = {
   description?: string;
   getDiagnostics: (
     environmentInfo: EnvironmentInfo,
-  ) => Promise<{version?: string; needsToBeFixed: boolean | string}>;
+  ) => Promise<{
+    version?: string;
+    versionRange?: string;
```
thymikee: shouldn't we require versions and ranges from all health checks?

lucasbento:
@thymikee: cool, thanks for the review.

The approach you suggested looks good, I'll see what I can do!

packages/cli/src/commands/doctor/types.ts
lucasbento: Some of them are not based on comparing versions, just checking if they are there or not (e.g. `ios-deploy` and `cocoapods`).

lucasbento:
image

@thymikee: what do you think about this?

A few points related to your proposal:

- I'm not sure if the path to the software/lib is useful, perhaps it's just more info clogging the UI;
- We currently don't have a way to show a _to fix_ part on this step as it's shown only after the user tries to fix it all.

thymikee:
Looks good!

lucasbento:
@thymikee: pushed :)

packages/cli/src/commands/doctor/healthchecks/androidSDK.ts
```diff
@@ -47,11 +47,15 @@ export default {
       } catch {}
     }
 
+    const version = sdks === 'Not Found' ? sdks : sdks['Build Tools'][0];
```
Rainsho: What if I forgot to install `build-tools` when using `sdkmanager` manually? Here will throw an `Cannot read property '0' of undefined` error, instead of telling me to install `build-tools` through `sdymanager "build-tools;29.0.0"`.

PR #767, opened 2019-10-03
Remove hardcoded React Native configuration

grabbou:
Summary:
---------

Historically, React Native used to be tightly coupled with the CLI, support for Android and iOS was hardcoded. Since 0.60, we have made React Native ship with its own configuration, specifying Android and iOS platforms just like React Native Windows.

This PR removes "dead-code" that was only needed during the transition period. It wasn't used and called in the 0.60.

I have also removed some obsolete tests that were no longer needed.


packages/cli/src/tools/config/__tests__/index-test.ts
```diff
@@ -79,8 +105,7 @@ test('should return dependencies from package.json', () => {
 
 test('should read a config of a dependency and use it to load other settings', () => {
   writeFiles(DIR, {
-    'node_modules/react-native/package.json': '{}',
-    'node_modules/react-native-test/package.json': '{}',
```
thymikee: why this is removed?

grabbou: Accident, but I am surprised that the tests are passing.

grabbou:
Tests are failing on Windows. I will need a Windows machine to debug why the mocking logic doesn't work. 

If there's anybody with Windows that's willing to give it a go, let me know!

thymikee:
@thecodrr would you like to help us debug this?

thecodrr:
Yup. On it.

thymikee:
@grabbou can you rebase and rerun CI btw?

grabbou:
Any luck @thecodrr? We do happen to have same on #811. I have Windows machine with me as I am working on React Native Windows project right now. Will try to debug this in the coming days.

Meanwhile, help is appreciated :)

PR #768, opened 2019-10-03
feat: support `autolinking` in monorepos

### LINKED ISSUE #537: autoLinking doesn't work with monorepo

### LINKED ISSUE #633: iOS autolinking not working in monorepo
## Environment

System:
    OS: macOS 10.15
    CPU: (4) x64 Intel(R) Core(TM) i5-7360U CPU @ 2.30GHz
    Memory: 255.71 MB / 8.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 10.15.3 - ~/.nvm/versions/node/v10.15.3/bin/node
    Yarn: 1.16.0 - /usr/local/bin/yarn
    npm: 6.4.1 - ~/.nvm/versions/node/v10.15.3/bin/npm
  SDKs:
    iOS SDK:
      Platforms: iOS 13.0, DriverKit 19.0, macOS 10.15, tvOS 13.0, watchOS 6.0
    Android SDK:
      API Levels: 23, 25, 26, 27, 28
      Build Tools: 23.0.1, 25.0.1, 25.0.2, 26.0.1, 26.0.2, 27.0.3, 28.0.0, 28.0.1, 28.0.2, 28.0.3
      System Images: android-28 | Google APIs Intel x86 Atom
  IDEs:
    Android Studio: 3.4 AI-183.5429.30.34.5452501
    Xcode: 11.0/11M382q - /usr/bin/xcodebuild
  npmGlobalPackages:
    create-react-native-app: 2.0.2
    react-native-create-library: 3.1.2

## Description

I created a project with react native 0.60.4 in a monorepo (using yarn workspaces). The project has de following structure:
```
project
  |
  |- node_modules
  |- packages
     |-react-native-app
       |- ios
       |- android
       |- src
```

I followed the react native docs (https://github.com/react-native-community/cli/blob/master/docs/autolinking.md) on how to configure autolinking with custom roots, so I changed the paths in my Podfile to:

```
require_relative '../../../node_modules/@react-native-community/cli-platform-ios/native_modules'

use_native_modules!("../../..")
```

The problem is that when I run `pod install` it install only the react-native default dependencies and doesn't identify any fo the 3rd party libraries added to the app.

Is the podfile configuration right? Should I run any other command to make it work?

### LINKED ISSUE #657: use_native_modules! should take in consideration pod install `--project-directory=ios` option
## Environment

```
System:
    OS: macOS 10.15
    CPU: (8) x64 Intel(R) Core(TM) i7-6920HQ CPU @ 2.90GHz
    Memory: 975.33 MB / 16.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 10.15.1 - /var/folders/x_/wwvkzdz104n7j0m5jj2dq06m0000gn/T/fnm-shell-1102175/bin/node
    Yarn: 1.16.0 - /usr/local/bin/yarn
    npm: 6.9.0 - /var/folders/x_/wwvkzdz104n7j0m5jj2dq06m0000gn/T/fnm-shell-1102175/bin/npm
  SDKs:
    iOS SDK:
      Platforms: iOS 12.2, macOS 10.14, tvOS 12.2, watchOS 5.2
    Android SDK:
      API Levels: 26, 27, 28, 29
      Build Tools: 27.0.3, 28.0.3, 29.0.0
      System Images: android-28 | Intel x86 Atom, android-28 | Google APIs Intel x86 Atom, android-28 | Google Play Intel x86 Atom, android-29 | Google Play Intel x86 Atom
  IDEs:
    Android Studio: 3.4 AI-183.6156.11.34.5692245
    Xcode: 10.2.1/10E1001 - /usr/bin/xcodebuild
  npmPackages:
    react: ^16.9.0 => 16.9.0
    react-native: ^0.61.0-rc.0 => 0.61.0-rc.0
```

## Description

If you don't run `pod install` from `./ios` folder (eg: via `pod install --project-directory=ios` instead of `cd ios && pod install`, you can get this error

```
❯ pod install --project-directory=ios

[!] Invalid `Podfile` file: No such file or directory - ./node_modules/.bin/react-native.

 #  from /Users/moox/******/PROJECTROOT/ios/Podfile:74
 #  -------------------------------------------
 #
 >    use_native_modules!
 #  end
 #  -------------------------------------------
```

In this precise scenario, (`pod install` with `--project-directory=ios`), using `use_native_modules!(".")` instead of `use_native_modules!` does the trick (as default root value is `".."`).

That said, it's not a safe option as now, running `pod install` from `ios` folder won't work anymore.

Maybe this tool should take `--project-directory` option in consideration.

## Reproducible Demo

You can reproduce this with any project that
- have `@react-native-community/cli-platform-ios" "^3.0.0-alpha.2` or something around that 
- have `use_native_modules!` in the `ios/Podfile`

Run from the root of the project (and not `./ios`)

```console
```

### LINKED ISSUE #717: Properly configuring cli

grabbou:
Fixes #537 
Fixes #717
Fixes #657 (path to `.bin` no longer hardcoded)
Fixes #633 

The problem:
---------

There are various issues complaining about weak or broken support for mono repositories, in regards to auto-linking. The workarounds are temporary and error-prone, being executed on top of a confusing resolution system.

Surprisingly, we already have quite a good support for `mono-repo` in the CLI. For example, we check for a package in `node_modules` up the directory tree. We also do that for resolving the configuration file.

In this PR, I am reworking some of the fundamental assumptions about the paths within auto-linking in order to make it support various project structures without any extra work.

*Automatic mono repository support for automatic linking* 😂

Summary:
---------

Please look at the structure that I am working on. I believe this is the most common mono repo:

```
- /node_modules
   - /react-native
   - /react-native-firebase
- /packages
   - /mobile
      - /node_modules
         - /react-native-webview
      - /ios
```

This structure is based on the [@brunolemos repository](https://github.com/brunolemos/react-native-web-monorepo/tree/rn61) that he provided for reproduction purposes.

Please keep in mind the following two design decisions:

> My main assumption is that all configuration relative to the project you are working on (metro configuration, CLI configuration) should be placed inside `packages/mobile`. I think scoping is better and is going to support multiple apps that co-exist within a single codebase.

> Dependencies can live at different levels in the project hierarchy (hence I had to remove hardcoded `root`, which wouldn't make this work). This is important when there are multiple React Native apps living next to each other when a package cannot be hoisted.

Test:
---------

When running `pod install` on iOS and Android, both `react-native-webview` and `react-native-firebase` are linked properly with their relative paths. 

Status:
---------

- [x] iOS
- [x] Android

Let me know if this structure is similar to what you are running and what you'd expect.

Changes
----------
- feat: deprecate `root` in auto-linking and introduce automatic detection of the project location 
- refactor: refactor `PackageManager` to not rely on a `setProjectRoot` being called in different places
- refactor: limit dependency on an implicit `process.cwd()` being set by few functions
- feat: introduce `detachedFunction` (internally) that runs without configuration (useful for init and other commands that don't need project context)


packages/platform-ios/native_modules.rb
```diff
@@ -5,24 +5,23 @@
...
     json = []
 
-    # Make sure `react-native config` is ran from your project root
-    Dir.chdir(root) do
-      IO.popen("./node_modules/.bin/react-native config") do |data|
-        while line = data.gets
-          json << line
-        end
+    # @todo: Do not use Yarn here, but find path dynamically
+    IO.popen("yarn run --silent react-native config") do |data|
```
grabbou: I am most likely going to revert back to the original resolution mechanism we had, which was spawning `node` and looking for a location of a file. 

Yarn is nice, because it does that for us. If we could assume certain `npm` compatibility, I guess we would be able to use it as well.

Thoughts?

Certainly, hardcoded path to `node_modules/.bin` is not a go in a mono-repository structure (where CLI can be even installed globally and we shouldn't make any assumptions about that - it's one of the hidden architecture decisions to be run independently of the location of the project).

packages/platform-ios/native_modules.rb
```diff
@@ -5,24 +5,23 @@
 #
 require 'pathname'
 
-def use_native_modules!(root = "..", config = nil)
```
grabbou: This is a breaking change, unless I leave this argument here and print a warning that it does not take any effect and will be removed.

Otherwise, due to the presence of the next argument, `config`, we are potentially running into silent errors.

How about: `if config is string` (means that somebody is passing path to the function), print out a warning and ignore it.

We use `config` parameter for testing purposes only (AFAIK).

packages/platform-ios/native_modules.rb
thymikee: https://github.com/react-native-community/cli/pull/384

grabbou: Thanks for linking some context! We will need to find a better alternative! Will look into it to remind myself about some of the constraints.

grabbou: @thymikee 

> Since @react-native-community/cli is not declared as template dependency, there is a possibility for Node to not be able to resolve it. It may happen because package managers can put the package inside react-native/node_modules/ behind a symlink (like npm does in some scenarios) which is invisible to require.resolve.

I believe we can address this by using `node` and `require.resolve` in the following manner:

- Call `path.dirname(require.resolve('react-native/package.json'))` to get path to React Native
- Use `require.resolve('@react-native-community/cli/..../index.js', { paths: [reactNativePath] })` to get path to the CLI from context of React Native where **we have it in package.json**

We use this mechanism already inside `resolveNodeModuleDir` and it works quite well. I believe we didn't have it at the time of that PR.

tbergq:
I remember looking into `brunolemos repository`, but if I understand correctly, this setup works because of the `react-native.config.js` in the root pointing to the android project path?

And this would only work with one `react-native` app in your monorepo right? What if I want to have more react-native apps in the same monreopo?

packages/platform-ios/native_modules.rb
thymikee: That may work, definitely worth exploring.

grabbou: Works flawlessly and resolves to an absolute path to the CLI, including supporting symlinks (while in development).

I consider this "resolved"!

grabbou:
Speaking to @thymikee, we have decided to pursue using `yarn` and `npx` again and checking which versions are supported. Using them lets us make even fewer assumptions about the paths, which is making it better for `monorepo`.

Will resume later.

packages/platform-ios/native_modules.rb
```diff
@@ -5,24 +5,22 @@
...
```
grabbou: Not needed, because `react-native config` is now `cwd` agnostic and will automatically resolve from the correct directory (project root).

packages/platform-ios/native_modules.rb
```diff
@@ -5,24 +5,22 @@
...
+    IO.popen("npx react-native config") do |data|
+      while line = data.gets
+        json << line
       end
     end
 
     config = JSON.parse(json.join("\n"))
   end
 
+  project_root = Pathname.new(config["project"]["ios"]["sourceDir"])
```
thymikee: oh yea, I wanted to set that a long time ago

packages/platform-ios/native_modules.rb
```diff
@@ -5,24 +5,22 @@
...
+
   packages = config["dependencies"]
-  config_root = config["root"]
```
thymikee: can you make sure this doesn't break setups like Slider? https://github.com/react-native-community/react-native-slider/blob/master/react-native.config.js

grabbou: I guess I need more context on this setup because the configuration looks a bit unfamiliar to me. 

@michalchudziak

grabbou: Ah, I see. Looks like we shouldn't break it because we are only reading `config.root` and `project`, not a `dependency` namespace, but will save it down to further debug.

grabbou:
Added Android support. Feedback is welcome :)

packages/cli/src/tools/config/findProjectRoot.ts
```diff
@@ -0,0 +1,26 @@
...
+
+  /**
+   * It is possible that `package.json` doesn't exist
+   * in the tree. In that case, we want to throw an error.
+   *
+   * When executing via `npx`, this will never happen as `npm`
+   * requires that file to be present in order to run.
+   */
+  if (!packageLocation) {
+    throw new CLIError(`
```
thymikee: this fails init, which is a special case we need to account for

packages/platform-android/native_modules.gradle
```diff
@@ -245,12 +211,12 @@ class ReactNativeModules {
 
 def autoModules = new ReactNativeModules(logger)
 
-ext.applyNativeModulesSettingsGradle = { DefaultSettings defaultSettings, String root = ".." ->
-  autoModules.applySettingsGradle(defaultSettings, root)
+ext.applyNativeModulesSettingsGradle = { DefaultSettings defaultSettings ->
+  autoModules.addReactNativeModuleProjects(defaultSettings)
 }
 
-ext.applyNativeModulesAppBuildGradle = { Project project, String root = ".." ->
```
thymikee: let's keep the root (unused) in gradle methods for backwards compatibility, output a warning that it's deprecated and remove it in v4.

packages/cli/src/cliEntry.js
```diff
@@ -174,13 +176,17 @@ async function setupAndRun() {
...
-  const ctx = loadConfig();
-
   logger.enable();
 
-  setProjectDir(ctx.root);
+  detachedCommands.forEach(addCommand);
 
-  [...commands, ...ctx.commands].forEach(command => addCommand(command, ctx));
+  try {
+    const ctx = loadConfig();
```
grabbou: Probably needs a better try/catch startegy

packages/cli/src/cliEntry.js
```diff
@@ -174,13 +176,17 @@ async function setupAndRun() {
...
+
+    [...projectCommands, ...ctx.commands].forEach(command =>
+      addCommand(command, ctx),
+    );
+  } catch (e) {}
```
thymikee: add a logger.debug maybe?

docs/autolinking.md
```diff
@@ -142,3 +95,14 @@ module.exports = {
...
+## How can I use autolinking in a monorepo?
+
+There is nothing extra you need to do - monorepos are supported by default.
+
+Please note that in certain scenarios, such as when using Yarn workspaces, your packages might be hoisted to the root of the repository. If that is the case, please make sure that the following paths are pointing to the
+correct location and update them accordingly:
+
+- path to `native_modules.rb` in your `ios/Podfile`
+- path to `native_modules.gradle` in your `android/settings.gradle`
+- path to `native_modules.gradle` in your `android/app/build.gradle`
```
Titozzz: Great addition

packages/platform-ios/native_modules.rb
```diff
@@ -57,10 +63,10 @@ def use_native_modules!(root = "..", config = nil)
     end
 
     podspec_dir_path = Pathname.new(File.dirname(podspec_path))
-    project_root = Pathname.new(config_root)
+    
     relative_path = podspec_dir_path.relative_path_from project_root
```
thymikee: Now that we don't have custom root, I think instead of using `project_root` we need to use cwd, because we look for – a relative path from cwd (ios dir) to our native module (in node_modules or custom path). This fixes the Slider repo example for me:
```suggestion
    relative_path = podspec_dir_path.relative_path_from Dir.pwd
```

Salakar:
It might be good to test this on the React Native Firebase monorepo's test project also (https://github.com/invertase/react-native-firebase/tree/master/tests), though I have no availability until probably end of next week 😫

grabbou:
Will test React Native Slider and React Native Firebase before we ship it, just to be sure.

grabbou:
React Native Slider isn't properly configured right now and this PR leveraged its issues. Running `react-native config` reveals that `project.ios.sourceDir` points to `ios` instead of `example/ios`.

The following configuration makes everything run smoothly:
```js
const root = __dirname;

module.exports = {
  project: {
    android: {
      sourceDir: './example/android',
    },
    ios: {
      project: './example/ios/example.xcodeproj',
    },
  },
  dependencies: {
    'react-native-slider': {
      root,
    },
  },
};
```

Note that in the future, once we add support to `runAndroid` and `runIOS` to use `config` (right now, they don't), it will make it easier. 

For example, no `--root` flag will be needed to `runAndroid` to point it to the correct directory.

Personal note: I am not a fan of `root` being set under `dependencies`. I would prefer to keep it just like for `project` - you specify explicit paths to the Android and iOS projects in case the detection mechanism failed.

grabbou:
Note: I wouldn't recommend doing React Native Slider setup in the future. Ideally, every project goes React Native Firebase way, having a separate `package.json` and independent package for an example project.

thymikee:
Doesn't this configuration leak to the users if config is published? 

I think it's about time to end this hack (because it clearly was one). The now seemingly correct approach there would be to include a real package.json with `"react-native": "*"` and `"@react-native-community/slider": ".."` and setup workspaces for less stuff to download

grabbou:
@thymikee `project` doesn't leak outside. `dependency` would "leak".

I actually just wrote this down - I believe [this](https://github.com/invertase/react-native-firebase/blob/master/tests/package.json) is the best way to set it up and doesn't include this hack.

grabbou:
@thymikee we should have a generator for initing such a structure like React Native Firebase to work on. Once it's supported here by default, we can make it official and the only use-case supported.

thymikee:
@grabbou sounds like something that Bob could provide

grabbou:
React Native Firebase works w/o issues with this PR @Salakar. Great project set up by the way!

Salakar:
> React Native Firebase works w/o issues with this PR @Salakar. Great project set up by the way!

Thanks for checking!

vomchik:
@grabbou Hi. Could you help me? Should both dependencies be added to package.json?
Currently, I'm using Lerna and run this command `npx lerna booststap --hoist` so all my deps hoist to root node_modules folder.
Thanks.

vongohren:
Yeah some examples of this would be nice, because Im still battling to no-hoist or not, to make sure that all is packaged right. Im following a couple of leads: https://engineering.brigad.co/react-native-monorepos-code-sharing-f6c08172b417

But keep getting missing in the project on some files and even getting past that I get firebase issues: https://github.com/invertase/react-native-firebase/issues/2122

So as the commentes in the medium says, is it still too early to do react native monorepo?

Issue #655, opened 2019-09-01
Cannot get property 'packageName' on null object


brunolemos:
> Moving from #537

## Environment

System:
    OS: macOS 10.15
    CPU: (4) x64 Intel(R) Core(TM) i5-6267U CPU @ 2.90GHz
    Memory: 2.70 GB / 16.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 12.8.0 - /var/folders/xh/gcxkb1qx2rd3fmqz5vxrs2lh0000gn/T/yarn--1567321836628-0.386672585258681/node
    Yarn: 1.17.3 - /var/folders/xh/gcxkb1qx2rd3fmqz5vxrs2lh0000gn/T/yarn--1567321836628-0.386672585258681/yarn
    npm: 6.10.3 - /usr/local/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  SDKs:
    iOS SDK:
      Platforms: iOS 12.4, macOS 10.14, tvOS 12.4, watchOS 5.3
    Android SDK:
      API Levels: 28
      Build Tools: 28.0.3
  IDEs:
    Android Studio: 3.5 AI-191.8026.42.35.5791312
    Xcode: 10.3/10G8 - /usr/bin/xcodebuild

## Description

Can't get react-native 0.61 to work on a monorepo on Android. Build fails with the error: `Cannot get property 'packageName' on null object`. iOS works as expected.

Running `yarn react-native config` returns this result:

```
{
  "root": "/Users/user/Projects/react-native-web-monorepo",
  "reactNativePath": "/Users/user/Projects/react-native-web-monorepo/node_modules/react-native",
  "dependencies": {},
  "commands": [],
  "assets": [],
  "platforms": {},
  "haste": {
    "providesModuleNodeModules": [],
    "platforms": []
  },
  "project": {}
}
```

Using `nohoist` or installing `react-native` inside `packages/mobile` instead of the root is not an option because that causes a range of other issues. 

## Reproducible Demo

Branch `rn61` with error: https://github.com/brunolemos/react-native-web-monorepo/tree/rn61
Branch `master` working, with [workarounds](https://github.com/react-native-community/cli/issues/655#issuecomment-531899969): https://github.com/brunolemos/react-native-web-monorepo

subashtimalsina:
I am also facing the same problem though my react-native version is 0.59, I couldn't fix this issue. If you have any optional idea please share me.

[...]

PR #769, opened 2019-10-03
fix: do not read configuration outside of the root

grabbou:
Summary:
---------

Reading configuration outside of the current working directory (in other words, project root), is not supported and can lead to dangerous errors. [I was pretty sure we never supported that](https://github.com/react-native-community/cli/pull/749#issuecomment-537931760), until I've found this out today.

Since I was able to remove this functionality without changing any tests, my recommendation is to not include it until: a) there is such a requirement (validated) from the user-side and that b) we have thoroughly tested this behavior.

Context:
---------

Fixes https://github.com/react-native-community/cli/issues/655).

As per [a comment from the above issue](https://github.com/react-native-community/cli/issues/655#issuecomment-538009106), the configuration is placed at the root of the repository whereas the CLI is run from the root of the project (which is different and resolves to `./packages/mobile`). That tiny difference is crucial and because paths in the configuration are always resolved relative to the root of the project, it creates silent issues that are hard to debug.

I believe it's a good practice to require running the CLI from the root of the mobile project (where `react-native` is defined in the `package.json`, just like Metro already does).

Few workarounds relied on this mechanism to provide support for mono-repos (it was broken anyway). However, the PR #768 is going to address these cases by rolling out a new mechanism for this.

### DIFF
--- a/packages/cli/src/tools/config/readConfigFromDisk.ts
+++ b/packages/cli/src/tools/config/readConfigFromDisk.ts
@@ -92,7 +92,10 @@ function readLegacyConfigFromDisk(rootFolder: string): UserConfig | void {
  * workspace.
  */
 export function readConfigFromDisk(rootFolder: string): UserConfig {
-  const explorer = cosmiconfig('react-native', {searchPlaces});
+  const explorer = cosmiconfig('react-native', {
+    searchPlaces,
+    stopDir: rootFolder,
+  });
 
   const {config} = explorer.searchSync(rootFolder) || {
     config: readLegacyConfigFromDisk(rootFolder),


grabbou:
I believe that the test failures are unrelated.

tbergq:
> I believe that the test failures are unrelated.

On #749 this test also failed randomly.

tbergq:
Ok, this solves half the problem. It crashes because of `def command = "node ./node_modules/react-native/cli.js config"` <- this line in `node_modules.gradle`. 
This is simply not where that module is installed, after changing it to `def command = "node ../../node_modules/react-native/cli.js config"` it all works as expected 🙂 


If you want to see my setup, I have a [monorepo here](https://github.com/tbergq/monorepo).

Still, the bundler cannot find the react-navigation back image.

grabbou:
@tbergq exactly, this is what I am tackling in the other PR - #768. On iOS, I have resolved that issue. Working on Android now.

tbergq:
Ok, thank you @grabbou 🙂 I will follow both PR's closely

grabbou:
Up, @thymikee, can I count on your review & merge? This is pretty important for the other PR to ensure there's no confusion while setting things up.

This is a small breaking change. It alters the undocumented behaviour, but changes setup for monorepositories which I believe could've been quite popular with this workaround.

PR #770, opened 2019-10-04
Fix buggy behaviour in launchDebugger on macOS

### LINKED ISSUE #762: Buggy behaviour in launchDebugger.js on macOS
## Environment

```
System:
    OS: macOS Mojave 10.14.6
    CPU: (8) x64 Intel(R) Core(TM) i7-6820HQ CPU @ 2.70GHz
    Memory: 1.77 GB / 16.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 10.16.3 - /usr/local/bin/node
    Yarn: 1.19.0 - /usr/local/bin/yarn
    npm: 6.11.3 - /usr/local/bin/npm
  SDKs:
    iOS SDK:
      Platforms: iOS 13.0, DriverKit 19.0, macOS 10.15, tvOS 13.0, watchOS 6.0
    Android SDK:
      API Levels: 27, 28, 29
      Build Tools: 27.0.3, 28.0.3, 29.0.2
      System Images: android-28 | Google APIs Intel x86 Atom, android-29 | Google APIs Intel x86 Atom
      Android NDK: 20.0.5594570
  IDEs:
    Android Studio: 3.5 AI-191.8026.42.35.5791312
    Xcode: 11.0/11A420a - /usr/bin/xcodebuild
  npmPackages:
    react: 16.9.0 => 16.9.0 
    react-native: 0.61.1 => 0.61.1

```

## Description

I was wondering why do I still get the message "For a better debugging experience please install Google Chrome" while I do have it installed, and the debugger was opening in Safari before I changed the default browser to Chrome...

So I found that the file
`node_modules/@react-native-community/cli/build/commands/server/launchDebugger.js`
was responsible for that message and I did some debugging.

I found out that you use `open` package to open urls in Chrome, and you correctly have `'google chrome'` as the app name parameter for that package (on a mac). **BUT!** You check for the existence of command with such name before actually launching Chrome. And such a command (`'google chrome'`) obviously does not exist, that's why `please install Google Chrome` message is displayed no matter what . So I suggest dropping the check for command existence altogether and rewriting the code similar to this:

```
  try {
    launchChrome(url);
  } catch {
    _cliTools().logger.info(`For a better debugging experience please install Google Chrome from: ${_chalk().default.underline.dim('https://www.google.com/chrome/')}`);
    (0, _launchDefaultBrowser.default)(url);
  } 
```

P.S.: Looks like nothing happens if I pass a wrong app name to 'open', it just suppresses any errors, so the fix might be not that easy.

P.P.S.: Changing from `'google chrome'` to `'/Applications/Google Chrome.app/Contents/MacOS/Google Chrome'` in `getChromeAppName` function and also adding double quotes around `${commandName}` in `commandExistsUnixSync` function seem to be enough to fix this bug.

Taym95:
Fix #762

### DIFF
--- a/packages/cli/src/commands/server/launchDebugger.js
+++ b/packages/cli/src/commands/server/launchDebugger.js
@@ -26,28 +26,6 @@ function commandExistsUnixSync(commandName) {
   }
 }
 
-function commandExistsWindowsSync(commandName) {
-  try {
-    const stdout = execSync('where ' + commandName, {stdio: []});
-    return !!stdout;
-  } catch (error) {
-    return false;
-  }
-}
-
-function commandExists(commandName) {
-  switch (process.platform) {
-    case 'win32':
-      return commandExistsWindowsSync(commandName);
-    case 'linux':
-    case 'darwin':
-      return commandExistsUnixSync(commandName);
-    default:
-      // assume it doesn't exist, just to be safe.
-      return false;
-  }
-}
-
 function getChromeAppName(): string {
   switch (process.platform) {
     case 'darwin':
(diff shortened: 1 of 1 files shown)


packages/cli/src/commands/server/launchDebugger.js
```diff
@@ -77,7 +97,7 @@ function launchChrome(url: string) {
 }
 
 function launchDebugger(url: string) {
-  if (!commandExists(getChromeAppName())) {
+  if (!commandExists(getChromeCommandName())) {
```
thymikee: I think we could get rid of `commandExists`, `getChromeCommandName` and `getChromeAppName` if we try/catch the `launchChrome(url)` and then fall back to `launchDefaultBrowser(url)` with the info log. What do you think?

StasD: @thymikee That is what I thought could be done here, but, for some reason, when I played with the code, I could not manage the `open` library to get to error handling code when I passed some garbage command to it. It seemed it just swallowed any errors... Maybe I will give it another try.

thymikee: Use async/await interface of open, that should help

Taym95: Do you mean something like that?

```
try {
    launchChrome(url);
  } catch (error) {
    launchDefaultBrowser(url);
	  logger.info(
      `For a better debugging experience please install Google Chrome from: 	 
       ${chalk.underline.dim('https://www.google.com/chrome/',
      )}`,);
  }
```

thymikee: You're missing the key part: `await launchChrome` and converting launchChrome into an `async` function

Taym95: Yes sure, but we will need `getChromeAppName` anyways because `launchChrome` used it, right?

thymikee: if it uses it then sure, let it stay

StasD: looking at `open`'s code gave me some clue. You also need to pass wait: true, for it to pass any errors...
So, it becomes like this:

```
function launchChrome(url) {
  return (0, _open().default)(url, {
    app: [getChromeAppName()],
    wait: true,
  });
}

async function launchDebugger(url) {
  try {
    await launchChrome(url);
  } catch (err) {
    _cliTools().logger.info(`For a better debugging experience please install Google Chrome from: ${_chalk().default.underline.dim('https://www.google.com/chrome/')}`);
    (0, _launchDefaultBrowser.default)(url);
  }
}
```
also need to delete commandExists and commandExistsWindowsSync functions.

StasD: Submitted an alternative PR here: https://github.com/react-native-community/cli/pull/771 :)

PR #771, opened 2019-10-04
fix: buggy behaviour when launching debugger on macOS

StasD:
Summary:
---------

Fixes https://github.com/react-native-community/cli/issues/762

Test Plan:
----------

### DIFF
--- a/packages/cli/src/commands/server/launchDebugger.js
+++ b/packages/cli/src/commands/server/launchDebugger.js
@@ -26,28 +26,6 @@ function commandExistsUnixSync(commandName) {
   }
 }
 
-function commandExistsWindowsSync(commandName) {
-  try {
-    const stdout = execSync('where ' + commandName, {stdio: []});
-    return !!stdout;
-  } catch (error) {
-    return false;
-  }
-}
-
-function commandExists(commandName) {
-  switch (process.platform) {
-    case 'win32':
-      return commandExistsWindowsSync(commandName);
-    case 'linux':
-    case 'darwin':
-      return commandExistsUnixSync(commandName);
-    default:
-      // assume it doesn't exist, just to be safe.
-      return false;
-  }
-}
-
 function getChromeAppName(): string {
   switch (process.platform) {
     case 'darwin':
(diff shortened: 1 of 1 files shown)


packages/cli/src/commands/server/launchDebugger.js
```diff
@@ -69,24 +47,20 @@ function getChromeAppName(): string {
 }
 
 function launchChrome(url: string) {
-  open(url, {app: [getChromeAppName()]}, err => {
-    if (err) {
-      logger.error('Google Chrome exited with error:', err);
-    }
-  });
+  return open(url, {app: [getChromeAppName()], wait: true});
```
Taym95: Test it, It's not working properly, still getting `For a better debugging experience please install Google Chrome from` and I have Chrome installed and my default browser is `Brave`.

I don't think we need to use `wait: true` here, from `open` documentation:
```
wait
Type: boolean

Wait for the opened app to exit before fulfilling the promise. If false it's fulfilled immediately when opening the app.
```

You need to use async/await with open, something like this:

```
async function launchChrome(url: string) {
  const {err} = await open(url, {app: getChromeAppName()});
  if (err) {
    logger.error('Google Chrome exited with error:', err);
  }
}
```

StasD: Without `wait: true`, it will not work as we need it to: no errors will be reported to the calling code and try/catch will not make any difference. Here is an extract from `open`'s code:

```
	const subprocess = childProcess.spawn(command, cliArguments, childProcessOptions);

	if (options.wait) {
		return new Promise((resolve, reject) => {
			subprocess.once('error', reject);

			subprocess.once('close', exitCode => {
				if (exitCode > 0) {
					reject(new Error(`Exited with code ${exitCode}`));
					return;
				}

				resolve(subprocess);
			});
		});
	}

	subprocess.unref();

	return subprocess;
```

It seems to be working well with my modifications for me . :)

Taym95: Ah sorry my bad, works great!

PR #787, opened 2019-10-14
chore: rewrite main CLI entry point to TypeScript

grabbou:
This pull requests converts main entry point of the CLI to use TypeScript.


packages/cli/tsconfig.json
```diff
@@ -1,13 +1,14 @@
 {
   "extends": "../../tsconfig.json",
   "compilerOptions": {
-    "rootDir": "src",
+    "rootDir": "",
```
grabbou: @thymikee can you help me understand what this file is doing? Had to change it so that `../package.json` require works. 

This doesn't seem to be affecting the directory structure inside `outDir`, as it's still flat (`build/index.js` is present, it's not `build/src/index.js` or something).

thymikee: It's an indication that we keep all of our source files inside src/ dir. Likely makes the compilation a bit faster. For the package.json include, I'd go with `"include": ["../package.json"]`

grabbou:
Thanks for late-night merge! One more PR and I can go to sleep! haha

PR #789, opened 2019-10-14
feat: use config from `run-ios`

### LINKED ISSUE #12: `run-ios` should use config

grabbou:
Unifies `run-ios` to use config.

Fixes #12


packages/platform-ios/src/commands/runIOS/index.ts
```diff
@@ -510,13 +504,6 @@ export default {
       name: '--scheme [string]',
       description: 'Explicitly set Xcode scheme to use',
     },
-    {
-      name: '--project-path [string]',
```
thymikee: remove from the docs as well

grabbou: Done!

grabbou:
This PR is on hold when we remove `legacy link`. It requires a change to config I don't want to make right now.

PR #790, opened 2019-10-14
chore: remove "rnpm"

grabbou:
This PR removes all `rnpm` configuration files and related code. Everything was already deprecated and messages were being displayed.

This will ship with the next React Native.

Removed:
- `link`
- `unlink`
- `install`
- `uninstall`
- `hooks`, `params` and `assets` (now taken care by `autolinking` or not supported)
- `linkConfig` + extraneous config properties
- RNPM migration guides from documentation

Updated:
- Documentation to use TypeScript instead of Flow in the interface breakdown

This implements https://github.com/react-native-community/discussions-and-proposals/issues/96


grabbou:
e2e tests are also working now because `install` and `uninstall` were flaky (now removed).

grabbou:
On that note, we should add "e2e" tests for autolinking too.

grabbou:
@thymikee I don't think I have removed anything that would make #791 not work. In that case, we will re-add them once you rebase your PR.

docs/configuration.md
```diff
@@ -14,107 +12,3 @@ Check the documentation for
 - [plugins](./plugins.md)
 
 to learn more about different types of configuration and features available.
-
-## Migration guide
```
thymikee: I think it would be nice of ours to leave this section and link to the certain github tag, so it's still easy to find. I'd assume there's still a lot of folks using RN 0.59 out there.

docs/dependencies.md
```diff
@@ -54,18 +31,20 @@ In most cases, as a library author, you should not need to define any of these.
 The following settings are available on iOS and Android:
 
 ```ts
-type DependencyParamsIOST = {
+interface IOSDependencyParams {
   project?: string;
   podspecPath?: string;
-  sharedLibraries?: string[];
-};
+  scriptPhases?: Array<any>;
```
thymikee: any? would be nice to describe it somehow

grabbou: Yeah, it's been there like this in the code, mainly because it's not used inside JS but inside Ruby. Good call!

docs/configuration.md
grabbou: I love this idea! I'll just link to the version tag with a note that maybe `rnpm` config is only supported within React Native <= 0.61 - in that case, check the migration guides under ...?

docs/projects.md
```diff
@@ -4,36 +4,37 @@ A project is an app that contains React code and has a dependency on `react-nati
...
+      IOSProjectConfig,
+      IOSProjectParams,
+      IOSDependencyConfig,
+      IOSDependencyParams
+    >;
+    [name: string]: PlatformConfig<any, any, any, any>;
+  };
+  commands: Command[];
+}
 type ProjectConfigT = {
```
thymikee: This type is outdated. From the source:
```ts
export type ProjectConfig = {
  android?: AndroidProjectConfig;
  ios?: IOSProjectConfig;
  [key: string]: any;
};
```

docs/platforms.md
```diff
@@ -78,35 +73,19 @@ module.exports = {
...
-  projectPath: string;
-  projectName: string;
-  libraryFolder: string;
-  sharedLibraries: Array<any>;
-  plist: Array<any>;
-};
+  scriptPhases: Array<any>;
 
-type ProjectConfigAndroidT = {
+  podfile?: string;
```
thymikee: Looks like podfile is always there in iOS project config, but may be undefined. This type would make it more accurate:
```suggestion
  podfile: string | undefined;
```

docs/projects.md
```diff
@@ -4,36 +4,37 @@ A project is an app that contains React code and has a dependency on `react-nati
...
-For example, below configuration informs CLI of the additional assets to link and about a custom project location.
-
-```js
-module.exports = {
-  project: {
-    ios: {
-      project: './CustomProject.xcodeproj',
-    },
-  },
-  assets: ['./assets'],
```
thymikee: I think it's time to figure out a better story for assets, which are not autolinked and currently require running `react-native link`, which is now removed

TheSavior:
Awesome!

When you say that these commands are removed now, do you mean that they will throw an exception when called?

It might be nice if the commands still existed but give a nice message and link to something. With all the older libraries with readmes that say to run link I can imagine people getting confused if those start erroring.

FWIW, I haven’t read the code. I’m on mobile and it’s too big to scroll and look for those related changes. So hopefully this isn’t a stupid question :-)

docs/projects.md
grabbou: It's a mistake I have left it at all there. Was replaced by just "Config" to get rid of this confusion!

grabbou:
@TheSavior, I like this idea! I will bring these commands back and print an appropriate message

docs/projects.md
grabbou: I have already started appropriate discussions on the Discord. I am hoping to write a follow-up PR with proper, built-in support for assets (in case it doesn't work right now).

grabbou:
I will send a follow-up PR with better assets support. Removing them from documentation now because we don't support it and that the documentation will change.

This will be a non-breaking change.

packages/cli/src/tools/config/index.ts
```diff
@@ -156,13 +143,7 @@ function loadConfig(projectRoot: string = findProjectRoot()): Config {
...
-    /**
-     * Legacy `rnpm` config required `haste` to be defined. With new config,
-     * we do it automatically.
-     *
-     * @todo: Remove this once `rnpm` config is deprecated and all major RN libs are converted.
-     */
-    const haste = config.haste || {
+    const haste = {
       providesModuleNodeModules: isPlatform ? [dependencyName] : [],
       platforms: Object.keys(config.platforms),
```
thymikee: I think we don't need to provide haste anymore since 0.61?

packages/platform-ios/src/config/findPodspec.ts
```diff
@@ -1,11 +1,11 @@
 import glob from 'glob';
 import path from 'path';
 
-export default function findPodspec(folder: string): string | null {
+export default function findPodspec(folder: string): string | undefined {
   const podspecs = glob.sync('*.podspec', {cwd: folder});
 
   if (podspecs.length === 0) {
-    return null;
+    return;
```
thymikee: Just FYI: I don't see a reason for this change, but I don't have strong opinions here

grabbou: Originally, I did it because I wanted to use `poodfile?: string` and that was incompatible with null being returned here.

Not an issue now that we have `string | undefined` vs `string | null`, I think I can revert this change anyway.

packages/cli/src/tools/config/index.ts
grabbou: Writing it down and will send a follow-up PR to remove it.

packages/platform-android/src/config/index.ts
```diff
@@ -70,7 +70,7 @@ export function dependencyConfig(
   if (!manifestPath) {
     return null;
   }
-
+  console.log(manifestPath);
```
thymikee: remove log

packages/platform-android/src/config/findManifest.ts
```diff
@@ -1,25 +1,9 @@
...
-    ],
-  })[0];
-
-  return manifestPath ? path.join(folder, manifestPath) : null;
-}
+module.exports = {
+  dependency: {
+    platforms: {
+      android: {
+        packageImportPath: 'com.lufinkey.react.spotify.RNSpotifyPackage',
```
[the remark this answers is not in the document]
grabbou: Haha! It's not, my mistake.

PR #791, opened 2019-10-14
feat: use config inside run-android

### LINKED ISSUE #781: `run-android` should use config
Follow up to https://github.com/react-native-community/cli/issues/12, we should make `run-android` use config and deprecate custom `--root` flag.

It is confusing and unnecessary.

thymikee:
Summary:
---------

Defer reading as much as possible from the shared config.
Details:
- removed `root` option in favor of deriving it from config. 
- (**possibly breaking**) changed `sourceDir` to only look for `android/` directory, which is a Gradle root project – this aligns `sourceDir` to iOS and allows for greater flexibility, because `android/app` was a hardcoded name for a default `appFolder`, which would likely break stuff when somebody started using `appFolder` flag
- remove `--appFolder` and `--appId` flags
- add `android.project.appName` to compensate for `appFolder`
- fix running app with `appIdSuffix`, when app uses flavors

Fixes #781. 

Test Plan:
----------

Updated tests.

### DIFF
--- a/packages/platform-android/src/commands/runAndroid/index.ts
+++ b/packages/platform-android/src/commands/runAndroid/index.ts
@@ -114,14 +114,30 @@ function getPackageNameWithSuffix(
 // Builds the app and runs it on a connected emulator / device.
 function buildAndRun(args: Flags, sourceDir: string) {
   const gradlew = process.platform === 'win32' ? 'gradlew.bat' : './gradlew';
+  const manifestPath = `${sourceDir}/${
+    args.appFolder
+  }/src/main/AndroidManifest.xml`;
+  let packageName;
+  let manifestContent;
 
-  // @ts-ignore
-  const packageName = fs
-    .readFileSync(
-      `${sourceDir}/${args.appFolder}/src/main/AndroidManifest.xml`,
-      'utf8',
-    )
-    .match(/package="(.+?)"/)[1];
+  try {
+    manifestContent = fs.readFileSync(manifestPath, 'utf8');
+  } catch (error) {
+    throw new CLIError(
+      "Failed to read AndroidManifest.xml. Maybe it's in a different Gradle project? If your project uses custom Android app directory, try `--appFolder` flag.",
+      error,
+    );
+  }
+
+  const packageMatch = manifestContent.match(/package="(.+?)"/);
+
+  if (packageMatch) {
+    packageName = packageMatch[1];
+  } else {
+    throw new CLIError(
+      "Couldn't determine the package name of your Android project. Make sure it's present in your app's AndroidManifest.xml",
+    );
+  }
 
   const packageNameWithSuffix = getPackageNameWithSuffix(
     args.appId,
@@ -351,7 +367,7 @@ function startServerInNewWindow(
       return execa.sync('sh', [launchPackagerScript], procConfig);
     }
   }
-  if (/^win/.test(process.platform)) {
+  if (process.platform === 'win32') {
     //Temporary fix for #484. See comment on line 254
     fs.writeFileSync(launchPackagerScript, launchPackagerScriptContent, {
       encoding: 'utf8',
--- a/packages/platform-android/src/commands/runAndroid/index.ts
+++ b/packages/platform-android/src/commands/runAndroid/index.ts
@@ -40,7 +40,8 @@ export interface Flags {
  * Starts the app on a connected Android emulator or device.
  */
 async function runAndroid(_argv: Array<string>, config: Config, args: Flags) {
-  if (!config.project.android) {
+  const androidProject = config.project.android;
+  if (!androidProject) {
     logger.error(
       'Android project not found. Are you sure this is a React Native project?',
     );
@@ -64,7 +65,7 @@ async function runAndroid(_argv: Array<string>, config: Config, args: Flags) {
   }
 
   if (!args.packager) {
-    return buildAndRun(args, config);
+    return buildAndRun(args, androidProject);
   }
 
   return isPackagerRunning(args.port).then(result => {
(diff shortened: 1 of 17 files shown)


packages/platform-android/src/commands/runAndroid/index.ts
```diff
@@ -115,15 +109,18 @@ function getPackageNameWithSuffix(
 }
 
 // Builds the app and runs it on a connected emulator / device.
-function buildAndRun(args: Flags) {
-  process.chdir(path.join(args.root, 'android'));
-  const cmd = process.platform.startsWith('win') ? 'gradlew.bat' : './gradlew';
+function buildAndRun(args: Flags, config: Config) {
+  const androidConfig = config.project.android;
```
grabbou: Didn't we check for existence already? If it's guaranteed to be there, maybe we can pass "config.project.android" directly or use some special Typescript utility type?

packages/platform-android/src/commands/runAndroid/index.ts
```diff
@@ -115,15 +109,18 @@ function getPackageNameWithSuffix(
...
+  if (!androidConfig) {
+    return;
+  }
+  process.chdir(androidConfig.sourceDir);
```
grabbou: While we are at it, I am wondering if we should remove this and prefer setting `cwd` explicitly while running a command or fixing paths in other functions? Would likely help with testing (similar what we did with PackageManager).

packages/platform-android/src/commands/runAndroid/index.ts
```diff
@@ -115,15 +109,18 @@ function getPackageNameWithSuffix(
...
+
+  const gradlew = process.platform === 'win32' ? 'gradlew.bat' : './gradlew';
```
grabbou: If we remove `process.cwd`, I guess this should be `./android/gradlew`?

packages/platform-android/src/commands/runAndroid/index.ts
grabbou: If not, maybe we can make this a first-line, on line 42? That way we can keep a great balance of making it explicit and working inside "android" folder if it makes things easier.

packages/platform-android/src/commands/runAndroid/index.ts
thymikee: Yea not quite, because Gradle will think that cwd is `./` and will try to apply tasks to the root project instead of app project. We could, however, pass cwd to `execa` and avoid process.chdir if you like. I left it, because we have similar logic for iOS

packages/platform-android/src/commands/runAndroid/index.ts
```diff
@@ -46,7 +40,8 @@ export interface Flags {
  * Starts the app on a connected Android emulator or device.
  */
 async function runAndroid(_argv: Array<string>, config: Config, args: Flags) {
-  if (!checkAndroid(args.root)) {
+  const androidProject = config.project.android;
+  if (!androidProject) {
     logger.error(
       'Android project not found. Are you sure this is a React Native project?',
```
grabbou: Not directly related to the PR, but if we did:
```
throw new CLIError(`
  Android project not found. Are you sure this is a React Native project?

  If your Android files are located in a non-standard location (e.g. not inside \'android\' folder), consider setting
  \`project.android.sourceDir\` option to point to a new location.
`)
```
it would be easier to style it later, when we do better error handling.

I would assume that running `run-android` w/o an Android package is most likely an error.

packages/platform-android/src/commands/runAndroid/index.ts
```diff
@@ -359,6 +366,7 @@ function startServerInNewWindow(
   logger.error(
     `Cannot start the packager. Unknown platform ${process.platform}`,
```
grabbou: Shouldn't we throw here? 

TIP: This codeebase hasn't been touched in a while, I am pretty sure it has some wild `logger.error` instead of `CLIError` in few places.

Good opportunity to update.

grabbou: It's not related to the PR, we can take it separately under "improving error messags and handling inside `run-android`". Whatever makes more sense!

packages/platform-android/src/commands/runAndroid/index.ts
```diff
@@ -114,14 +114,30 @@ function getPackageNameWithSuffix(
...
+    throw new CLIError(
+      "Failed to read AndroidManifest.xml. Maybe it's in a different Gradle project? If your project uses custom Android app directory, try `--appFolder` flag.",
+      error,
+    );
+  }
+
+  const packageMatch = manifestContent.match(/package="(.+?)"/);
+
+  if (packageMatch) {
+    packageName = packageMatch[1];
```
grabbou: One more thing, don't we already have a [`packageName`](https://github.com/react-native-community/cli/blob/next/packages/platform-android/src/config/index.ts#L43) in the configuration?

I would also consider returning `manifestPath` from the configuration as well, as [we already use it to derive `packageName`](https://github.com/react-native-community/cli/blob/next/packages/platform-android/src/config/index.ts#L41).

Otherwise, we risk using two different AndroidManifests which might be dangerous.

thymikee: Yes we have, but the tricky thing is the `appFolder`. Gradle has a concept of root project (where root build.gradle is) and sub projects (e.g. `app`, `core`, `ui`). In a typical RN app, in `settings.gradle` we only include `app` project. But technically, in an Android-first project, there could be many gradle sub projects – and hence `appFolder`. Our configuration doesn't have this as an option. We could add it, but even when overwritten through a flag, the `npx react-native config` won't know about it. This is likely a super edge-case or even not happening, but we still support it.

That's why I can't really use the packageName or even appId from the configuration – because they only refer to a single project.

grabbou:
Any update on this PR? The tests are failing, not reviewing at this point.

thymikee:
Didn't have time to revisit yet

thymikee:
Superseded by  #1029

Issue #765, opened 2019-10-02
Handle Work Profile / Several Android Users


adrienthiery:
## Describe the Feature

We are using [Android for Work](https://www.android.com/enterprise/employees/) in our company, which makes a nice "sandbox" and makes sense in the context of our jobs.

A great feature would be to be able to install  the project ONLY in that profile and run it in that profile, instead of in the "first" profile.

In addition to a feature request, it is actually a feature proposition, as I would like to implement it.

## Possible Implementations

The idea would be that when running on Android (`react-native run-android`) : 
* a check on the user profiles (`adb shell pm list users`) could be run
* if several users are detected, propose a choice to the user : 
   * Choose a user workspace to install in (with an option `all`)
   * Choose a user to launch it in
* Propose to save that as a default for that project.

When running the app subsequently, use the default.

This could come with a setting in `react-native.config.js` and/or a flag to clear the default when running `run-android`

## Related Issues

None that I know of

## Relevant links

- https://stackoverflow.com/questions/43344576/appium-how-to-launch-app-from-android-for-work-work-profile
- https://stackoverflow.com/questions/20746422/how-to-use-adb-to-install-development-apps-for-one-user-only

thymikee:
Sounds like a good idea. We can start with an interactive behavior + a flag for non-interactive mode and see how that works. Not yet sure what would be the most ergonomic way of configuring commands through the configuration. Let's leave that for a separate discussion.

One more thing, we'll likely make choosing an emulator interactive as well (on master there's a feature that it picks the first available). Please take that into consideration when picking the best DX for this – answering 2 or more questions for every `run-android` may be a chore, but maybe not.

adrienthiery:
Hi @thymikee, I started to work on that but I can't seem to find how to test my changes.

I initialized a new RN App, and in it I tried :
* To `yarn add <local-path-of-cli-repo>/packages/cli` or `yarn link <local-path-of-cli-repo>/packages/cli` and then in my project `yarn link @react-native-community/cli`
* To `yarn add <local-path-of-cli-repo>/packages/platform-android` or `yarn link <local-path-of-cli-repo>/packages/platform-android` and then in my project `yarn link @react-native-community/platform-android`

I also tried to modify the `node_modules/.bin/react-native` with no luck.

How can I run my local version of the CLI packages? 🤔

homersimpsons:
Is there any news on this one ? Adding a cli options should be the first step, I don't think having an interactive behavior is desirable at first.

[...]

PR #807, opened 2019-10-17
Added interactivity for launching emulators and app on devices

adrienthiery:
Summary:
---------

Motivation is explained in https://github.com/react-native-community/cli/issues/765

Test Plan:
----------

Environment: 
* MacOS Mojave (10.14.6)
* node v8.16.0
* Android Debug Bridge version 1.0.41 / Version 29.0.2-5738569
* With a project started with RN 0.61.2

For each configuration hereafter, I ran :
  * `npx react-native run-android`
  * `npx react-native run-android --no-interactive`

Tried the change with several configurations :
* No device connected, no emulator started  (2 exist on my machine)
* No device, two emulators started
* One device (Galaxy s8 - Android 9), no emulator started
* One device (Galaxy s8 - Android 9), two emulators started
* Two devices  (Galaxy s8 - Android 9/One Plus 6 - Android 9), two emulators started

Both devices have a work profile and a normal profile.

For starting the emulators, when prompted which one to choose (case where no emulator started & no device connected) : 
* Tried starting the first one
* Tried starting the second one
* Tried starting all

When launching the application when several devices/emulators are connected/running:
* Tried launching on each one individually
* Tried launching on all

When choosing the profile to launch the app on:
* Tried the normal profile
* And the work profile

With one phone being in french and the other in english.

### DIFF
--- a/packages/platform-android/src/commands/runAndroid/index.ts
+++ b/packages/platform-android/src/commands/runAndroid/index.ts
@@ -41,6 +41,7 @@ export interface Flags {
   port: number;
   terminal: string;
   jetifier: boolean;
+  interactive: boolean;
 }
 
 /**
@@ -263,6 +264,7 @@ function installAndLaunchOnDevice(
     packageName,
     adbPath,
     args.mainActivity,
+    args.interactive,
   );
 }
 
@@ -435,6 +437,12 @@ export default {
       name: '--no-jetifier',
       description:
         'Do not run "jetifier" – the AndroidX transition tool. By default it runs before Gradle to ease working with libraries that don\'t support AndroidX yet. See more at: https://www.npmjs.com/package/jetifier.',
+      default: false,
+    },
+    {
+      name: '--no-interactive',
+      description:
+        'Do not prompt the user for choices if applicable, choose the default value.',
       default: false,
     },
   ],
--- a/packages/platform-android/src/commands/runAndroid/index.ts
+++ b/packages/platform-android/src/commands/runAndroid/index.ts
@@ -89,15 +89,11 @@ async function runAndroid(_argv: Array<string>, config: Config, args: Flags) {
     // result == 'not_running'
     logger.info('Starting JS server...');
     try {
-      startServerInNewWindow(
-        args.port,
-        args.terminal,
-        config.reactNativePath,
-      );
+      startServerInNewWindow(args.port, args.terminal, config.reactNativePath);
     } catch (error) {
       logger.warn(
         `Failed to automatically start the packager server. Please run "react-native start" manually. Error details: ${
-        error.message
+          error.message
         }`,
       );
     }
@@ -145,11 +141,16 @@ async function buildAndRun(args: Flags) {
     devices = adb.getDevices(adbPath);
   }
 
-  if (devices && devices.length > 1 && devices.filter(Boolean).length > 1 && args.interactive) {
+  if (
+    devices &&
+    devices.length > 1 &&
+    devices.filter(Boolean).length > 1 &&
+    args.interactive
+  ) {
     devices = await chooseDevice(devices);
   }
 
-  if (args.deviceId || devices && devices.length === 1) {
+  if (args.deviceId || (devices && devices.length === 1)) {
     return runOnSpecificDevice(
       args,
       cmd,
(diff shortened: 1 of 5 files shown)


adrienthiery:
Whoopsy, didn't check the tests. Sorry about that.

adrienthiery:
Hi guys,
I'm wondering if we should make `interactive` as `false` by default to avoid breaking all CI builds, or is that okay if we mention it in the changelog?
What do you think ?

Also, isn't it strange that the e2e tests pass on circle CI but not on Azure ? 🤔

packages/platform-android/src/commands/runAndroid/index.ts
```diff
@@ -135,34 +135,74 @@ function buildAndRun(args: Flags) {
...
   packageNameWithSuffix: string,
   packageName: string,
   adbPath: string,
+  devices: Array<string> | undefined,
 ) {
-  const devices = adb.getDevices(adbPath);
   const {deviceId} = args;
+
+  if (!devices) {
+    devices = [];
```
grabbou: nit: consider using default function argument?

packages/platform-android/src/commands/runAndroid/index.ts
```diff
@@ -179,6 +219,15 @@ function runOnSpecificDevice(
         ...devices,
       );
     }
+  } else if (devices.length === 1) {
+    buildApk(gradlew);
+    installAndLaunchOnDevice(
```
grabbou: This is super similar to https://github.com/react-native-community/cli/pull/807/files#diff-2302ca84a9ef71bfb8384b1d669d2aaaR208-R216. Can we simplify it to call `installAndLaunchOnDevice` only once?

packages/platform-android/src/commands/runAndroid/index.ts
```diff
@@ -135,34 +135,74 @@ function buildAndRun(args: Flags) {
...
+  }
+
+  if (args.deviceId || (devices && devices.length === 1)) {
     return runOnSpecificDevice(
       args,
       cmd,
       packageNameWithSuffix,
       packageName,
       adbPath,
+      devices,
```
grabbou: If the function is `runOnSpecificDevice`, wouldn't it make more sense to pass just `device`? (not the entire array)? We could reduce the logic inside the function.

packages/platform-android/src/commands/runAndroid/index.ts
```diff
@@ -135,34 +135,74 @@ function buildAndRun(args: Flags) {
...
+  let devices;
+
+  if (!args.deviceId) {
+    devices = adb.getDevices(adbPath);
+  }
+
+  if (
+    devices &&
+    devices.length > 1 &&
+    devices.filter(Boolean).length > 1 &&
```
grabbou: I would personally prefer a bit more verbose check here. I don't understand what's the purpose of `filter(Boolean)`, so maybe a simple inline comment would do?

Also, I think that `devices.length > 1` is not needed if we have the `devices.filter(Boolean).length > 1` already in place.

packages/platform-android/src/commands/runAndroid/index.ts
```diff
@@ -135,34 +135,74 @@ function buildAndRun(args: Flags) {
     packageName,
   );
   const adbPath = getAdbPath();
-  if (args.deviceId) {
+  let devices;
```
grabbou: How about we ALWAYS get `const devices = adb.getDevices(adbPath)` and... `const device = args.deviceId ? selectSimulator(devices, args.deviceId) : undefined`

If `device` is undefined and `args.interactive` is `true`, we would proceed with the interactive logic.

packages/platform-android/src/commands/runAndroid/index.ts
```diff
@@ -135,34 +135,74 @@ function buildAndRun(args: Flags) {
...
       args,
       cmd,
       packageNameWithSuffix,
       packageName,
       adbPath,
+      devices,
     );
-  } else {
+  } else if (devices && devices.length > 1) {
     return runOnAllDevices(
```
grabbou: While we are at it, why we can't just `devices.forEach(d => runOnSpecificDevice(....))` but need a separate function? Could be potentially a good area for refactor.

packages/platform-android/src/commands/runAndroid/index.ts
grabbou: It's also a bit confusing that we ignore `deviceId` passed to the function if the `devices.lenght` is `1`. Something to keep in mind.

grabbou:
Closing due to inactivity and merge issues. Let's reopen once there's more time to revisit this! Thank you for submitting this PR!

PR #811, opened 2019-10-18
feat: remove haste

grabbou:
Summary:
---------

Haste has been removed in 0.61 of React Native and we should remove it too. Otherwise, we are setting `hasteImplPath` pointing to a file that has been removed from React Native 0.61 via https://github.com/facebook/react-native/commit/7a2463e1f396ffcfbd86e68170f624d31a1be4e1


grabbou:
Note: I haven't tested this yet, will do later today. For now, I am just sending out a PR for a review.

grabbou:
@thymikee, could you CC somebody from the Metro team on this PR? I want to point out that we always explicitly set "platforms" and "sourceExts" and that Metro doesn't need to explicitly set these platforms here: https://github.com/facebook/metro/blob/master/packages/metro-config/src/defaults/defaults.js#L58

grabbou:
I wonder how we can debug the Windows failure?

grabbou:
I have rebased this against `next` branch.

grabbou:
Merging since tests are passing and I had former approval of @thymikee

PR #812, opened 2019-10-18
feat: cleaned-up `start` command

grabbou:
Summary:
---------

This PR cleans up "start" command by:
- removing extraneous options that were never actually read (e.g. "transformer" was specified as a flag, but never passed to any line of code. "platforms" was defined, but never read. "providesNodeModules" as well). I am not entirely sure how we ended up in such state, but removing them should help with reducing potential confusion when developers are listing down flags that a command accepts
- removing options that are possible to be defined in a regular Metro configuration and there is no need to repeat support for them ("projectRoots", "assetsExts" and more, I have consulted official Metro website for that)
- renamed "config" to "metroConfig"

For use cases when someone wants to customize "react-native start" on per execution basis, they can either edit `metro.config.js` or create a new one and use `--metroConfig` option.

In the follow-up PR, I will be working on improving "MiddlewareManager" to have a smaller API surface and take advantage of the recent developments in Metro.

The goal is to reduce the amount of Metro related code and make Metro versions independent of the CLI. 

Test Plan:
----------

Building and running an app should work.

### DIFF
--- a/packages/cli/src/commands/server/runServer.ts
+++ b/packages/cli/src/commands/server/runServer.ts
@@ -28,7 +28,7 @@ export type Args = {
   port?: number;
   resetCache?: boolean;
   verbose?: boolean;
-  config?: string;
+  metroConfig?: string;
 };
 
 async function runServer(_argv: Array<string>, ctx: Config, args: Args) {
@@ -37,7 +37,8 @@ async function runServer(_argv: Array<string>, ctx: Config, args: Args) {
   const reporter = new ReporterImpl(terminal);
 
   const metroConfig = await loadMetroConfig(ctx, {
-    config: args.config,
+    config: args.metroConfig,
+    port: args.port,
     resetCache: args.resetCache,
     reporter,
   });
--- a/packages/cli/src/commands/server/runServer.ts
+++ b/packages/cli/src/commands/server/runServer.ts
@@ -22,12 +22,12 @@ import releaseChecker from '../../tools/releaseChecker';
 export type Args = {
   cert?: string;
   customLogReporterPath?: string;
-  key?: string;
-  config?: string;
   host?: string;
   https?: boolean;
+  key?: string;
   resetCache?: boolean;
   verbose?: boolean;
+  config?: string;
 };
 
 async function runServer(_argv: Array<string>, ctx: Config, args: Args) {
--- a/packages/cli/src/commands/server/runServer.ts
+++ b/packages/cli/src/commands/server/runServer.ts
@@ -20,15 +20,13 @@ import loadMetroConfig from '../../tools/loadMetroConfig';
 import releaseChecker from '../../tools/releaseChecker';
 
 export type Args = {
-  customLogReporterPath?: string;
-  config?: string;
-
   cert?: string;
+  customLogReporterPath?: string;
   key?: string;
+  config?: string;
   host?: string;
   https?: boolean;
   resetCache?: boolean;
-
   verbose?: boolean;
 };
 
@@ -38,8 +36,8 @@ async function runServer(_argv: Array<string>, ctx: Config, args: Args) {
   const reporter = new ReporterImpl(terminal);
 
   const metroConfig = await loadMetroConfig(ctx, {
-    resetCache: args.resetCache,
     config: args.config,
+    resetCache: args.resetCache,
     reporter,
   });
 
--- a/packages/cli/src/commands/server/runServer.ts
+++ b/packages/cli/src/commands/server/runServer.ts
@@ -20,22 +20,16 @@ import loadMetroConfig from '../../tools/loadMetroConfig';
 import releaseChecker from '../../tools/releaseChecker';
 
 export type Args = {
-  assetPlugins?: string[];
-  cert?: string;
   customLogReporterPath?: string;
+  config?: string;
+
+  cert?: string;
+  key?: string;
   host?: string;
   https?: boolean;
-  maxWorkers?: number;
-  key?: string;
-  platforms?: string[];
-  port?: number;
   resetCache?: boolean;
-  sourceExts?: string[];
-  transformer?: string;
+
   verbose?: boolean;
-  watchFolders?: string[];
-  config?: string;
-  projectRoot?: string;
 };
 
 async function runServer(_argv: Array<string>, ctx: Config, args: Args) {
@@ -44,21 +38,10 @@ async function runServer(_argv: Array<string>, ctx: Config, args: Args) {
   const reporter = new ReporterImpl(terminal);
 
   const metroConfig = await loadMetroConfig(ctx, {
-    config: args.config,
-    maxWorkers: args.maxWorkers,
-    port: args.port,
     resetCache: args.resetCache,
-    watchFolders: args.watchFolders,
-    projectRoot: args.projectRoot,
-    sourceExts: args.sourceExts,
+    config: args.config,
     reporter,
   });
-
-  if (args.assetPlugins) {
-    metroConfig.transformer.assetPlugins = args.assetPlugins.map(plugin =>
-      require.resolve(plugin),
-    );
-  }
 
   const middlewareManager = new MiddlewareManager({
     host: args.host,
--- a/packages/cli/src/commands/server/runServer.ts
+++ b/packages/cli/src/commands/server/runServer.ts
@@ -25,6 +25,7 @@ export type Args = {
   host?: string;
   https?: boolean;
   key?: string;
+  port?: number;
   resetCache?: boolean;
   verbose?: boolean;
   config?: string;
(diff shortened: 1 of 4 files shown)


grabbou:
Tests are failing, this needs to be wrapped up. Opening for initial feedback.

grabbou:
My goal is to make "start" command more of a generic command and keep Metro implementation separate.

In order to be able to do it, we need to make "start" command Metro-agnostic in terms of the flags it accepts.

packages/cli/src/commands/server/runServer.ts
```diff
@@ -20,22 +20,14 @@ import loadMetroConfig from '../../tools/loadMetroConfig';
...
 export type Args = {
-  assetPlugins?: string[];
   cert?: string;
   customLogReporterPath?: string;
   host?: string;
   https?: boolean;
-  maxWorkers?: number;
   key?: string;
-  platforms?: string[];
-  port?: number;
```
grabbou: A bit controversial, but `--port` can be set in the configuration of Metro and I don't think it's something users would like to change on per run basis. If it's non-standard, they would probably prefer to persist it inside a custom configuration.

packages/cli/src/commands/server/runServer.ts
```diff
@@ -20,22 +20,14 @@ import loadMetroConfig from '../../tools/loadMetroConfig';
...
   resetCache?: boolean;
-  sourceExts?: string[];
-  transformer?: string;
```
grabbou: This is a dead option and was never accessed from code.

packages/cli/src/commands/server/runServer.ts
```diff
@@ -20,22 +20,14 @@ import loadMetroConfig from '../../tools/loadMetroConfig';
...
 
 export type Args = {
-  assetPlugins?: string[];
   cert?: string;
   customLogReporterPath?: string;
   host?: string;
   https?: boolean;
-  maxWorkers?: number;
   key?: string;
-  platforms?: string[];
```
grabbou: This is a dead option and was never accessed from code.

packages/cli/src/commands/server/server.ts
```diff
@@ -22,60 +21,6 @@ export default {
...
-      parse: (val: string) => val.split(','),
-    },
-    {
-      name: '--platforms [list]',
-      description:
-        'Specify any additional platforms to be used by the packager',
-      parse: (val: string) => val.split(','),
-    },
-    {
-      name: '--providesModuleNodeModules [list]',
```
grabbou: This is a dead option and was never accessed from code. It's also removed from Metro anyway in the recent release.

packages/cli/src/commands/server/runServer.ts
```diff
@@ -20,22 +20,14 @@ import loadMetroConfig from '../../tools/loadMetroConfig';
...
-  port?: number;
   resetCache?: boolean;
-  sourceExts?: string[];
-  transformer?: string;
   verbose?: boolean;
-  watchFolders?: string[];
   config?: string;
```
thymikee: we should rename it to `metroConfig`, what do you think?

packages/cli/src/commands/server/runServer.ts
thymikee: Let's leave it as is. `port` is important and I use it frequently. I'd use it more often, if iOS worked with it (Android doesn't have issues) – we need to fix that separately

packages/cli/src/commands/server/server.ts
```diff
@@ -103,7 +48,7 @@ export default {
     },
     {
       name: '--config [string]',
```
thymikee: rename 
```suggestion
      name: '--metroConfig [string]',
```

packages/cli/src/tools/loadMetroConfig.ts
```diff
@@ -110,32 +119,22 @@ export const getDefaultConfig = (ctx: Config): MetroConfig => {
...
  * This allows the CLI to always overwrite the file settings.
  */
 export default function load(
   ctx: Config,
-  options?: ConfigOptionsT,
+  opts: ConfigOptions = {},
 ): Promise<MetroConfig> {
-  const defaultConfig = getDefaultConfig(ctx);
-  if (options && options.reporter) {
-    defaultConfig.reporter = options.reporter;
```
thymikee: If `options.reporter` is not there, we're going to pass `reporter: undefined` to Metro, and if I remember correctly, it resulted in no reporters at all, instead of using a default reporter, because Metro checked for presence of keys, instead of their values. But maybe it's not the case anymore. Please test thoroughly

grabbou:
Thanks @thymikee, I'll get working on this PR either today or tomorrow.

packages/cli/src/commands/server/runServer.ts
grabbou: Is this something you change as you run it? E.g. is the `--port` set "always" to same value or you change it each time you run the command?

I am wondering whether being able to set port in `metro.config.js` (just like with Webpack) wouldn't be enough.

packages/cli/src/commands/server/runServer.ts
grabbou: Sounds like a plan!

packages/cli/src/tools/loadMetroConfig.ts
grabbou: Ok, thanks, will double check that one tomorrow!

grabbou:
Thanks @thymikee, these sound like valid concerns / questions to me! I've left some answers and will get coding on them tomorrow!

packages/cli/src/commands/server/runServer.ts
thymikee: The typical scenario is having multiple RN apps and being able to run them on different Metro instances. Passing `--port` is convenient to do from the command line. I'd see removing this as controversial, especially that it doesn't bring too much complexity

grabbou: > The typical scenario is having multiple RN apps and being able to run them on different Metro instances.

Do you mean having multiple separate React Native projects running at the same time? Wouldn't the ability to set `port: XXX` in their respective configuration files be equally easy? 

All I am saying is that if you already change a port, you may have a reason to do it. And in that case, you may prefer to store it permanently in the configuration file and forget about it.

grabbou: I am not really opinionated here - just wanted to kick start the discussion and challenge how do we use this while running our apps these days.

Before finally deciding on leaving it as is, I'd like to get some extra thoughts on that from the community.

thymikee: I may work on N projects with N different teams – for each of which their project is the only one they work with. How would I justify changing configuration permanently for my own workflow to them? What if one of them start to have similar setup, but want different ports? With `--port` flag I don't have to change any files myself all the time and still keep the workflow.

I understand your urge to remove as much code as possible, but specifying a port is universal across many different bundlers, so it doesn't lock us in with Metro or any other.

satya164: Maybe it's better to leave `port` in CLI option as well.

Personally I think having ability to specify separate config files is more useful (e.g. `--config metro.prod.config.js`). Not sure if CLI supports it.

grabbou: Yeah, CLI supports that @satya164 and that was the reason why I considered `--port` redundant (an unnecessary complication of configuration layers).

grabbou: > I understand your urge to remove as much code as possible,

@thymikee, it's not the urge to remove code. I am simply challenging the use cases and configurations that we have established 3 to 4 years ago, while not having some of the current Metro options and features.

The intent here is to not make the UX more complex or to remove the code "for the sake of it", but to check whether there are better alternatives right now and how do they compare against the preferred developer workflow.

grabbou: > I may work on N projects with N different teams – for each of which their project is the only one they work with. How would I justify changing configuration permanently for my own workflow to them? What if one of them start to have similar setup, but want different ports? With --port flag I don't have to change any files myself all the time and still keep the workflow.

This makes sense. It's still probably a very edge-case to have multiple React Native apps running at the same time, but we can leave it as-is for now and revisit in the future when Metro configuration establishes its presence a bit more.

thymikee: I agree this is edge-case. I don't bother too much personally, but I know people feel strongly about their workflows 🤷‍♂. It may make sense to start a discussion on that in a new GH issue and share it on Twitter or somewhere to see how users feel about it.

grabbou:
I believe this is now ready to go! I have addressed the necessary changes.

grabbou:
Rebased on top of `master`. Ready for another review.

Issue #823, opened 2019-10-24
cli-platform-ios pod install failure


cormander:
## Environment

```
$ react-native info
info Fetching system and libraries information...
System:
    OS: macOS 10.15
    CPU: (12) x64 Intel(R) Core(TM) i7-8700B CPU @ 3.20GHz
    Memory: 7.31 GB / 32.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 10.16.3 - /usr/local/bin/node
    npm: 6.12.0 - /usr/local/bin/npm
  SDKs:
    iOS SDK:
      Platforms: iOS 13.1, DriverKit 19.0, macOS 10.15, tvOS 13.0, watchOS 6.0
    Android SDK:
      API Levels: 23, 25, 27, 28, 29
      Build Tools: 28.0.3, 29.0.1, 29.0.2
      System Images: android-28 | Google Play Intel x86 Atom, android-29 | Google Play Intel x86 Atom
  IDEs:
    Android Studio: 3.5 AI-191.8026.42.35.5900203
    Xcode: 11.1/11A1027 - /usr/bin/xcodebuild
  npmPackages:
    react: 16.11.0 => 16.11.0 
    react-native: 0.61.2 => 0.61.2
```

## Description

On a current project, after removing the `package-lock.json` and re-running `npm install`, a `pod install` in the `ios` directory fails:

```
$ pod install
Command `config` unrecognized. Make sure that you have run `npm install` and that you are inside a react-native project.

[!] Invalid `Podfile` file: 767: unexpected token at ''.

 #  from /Users/ourvoiceusa/git/HelloVoter/mobile/ios/Podfile:44
 #  -------------------------------------------
 #  
 >    use_native_modules!
 #  end
 #  -------------------------------------------
```

## Reproducible Demo

```
git clone https://github.com/OurVoiceUSA/HelloVoter.git
cd HelloVoter/mobile
npm install
cd ios
pod install
```
Then if you remove `package-lock.json` and re-install and try the `npx react-native config` from the `ios` directory again:
```
rm -f package-lock.json
npm install
cd ios
pod install
```
I believe this is due to `npx react-native config` failing to work inside the `ios` directory on the newer version of `@react-native-community/cli-platform-ios`
```
$ cd ios
$ npx react-native config
Command `config` unrecognized. Make sure that you have run `npm install` and that you are inside a react-native project.
```
Here's the `package-lock.json` change that broke it:
```
[...]

PR #824, opened 2019-10-24
feat: run npx from JS app dir

### LINKED ISSUE #793: Gradle sync fail - ERROR: Text must not be null or empty
## Env. Info
```
System:
    OS: macOS Mojave 10.14.4
    CPU: (8) x64 Intel(R) Core(TM) i7-4870HQ CPU @ 2.50GHz
    Memory: 1.20 GB / 16.00 GB
    Shell: 5.3 - /bin/zsh
  Binaries:
    Yarn: 1.12.3 - /usr/local/bin/yarn
    npm: 6.10.2 - ~/.nvm/versions/node/v10.13.0/bin/npm
  SDKs:
    iOS SDK:
      Platforms: iOS 12.4, macOS 10.14, tvOS 12.4, watchOS 5.3
    Android SDK:
      API Levels: 26, 28, 29
      Build Tools: 28.0.3, 29.0.2
      System Images: android-29 | Google APIs Intel x86 Atom
  IDEs:
    Android Studio: 3.5 AI-191.8026.42.35.5900203
    Xcode: 10.3/10G8 - /usr/bin/xcodebuild
  npmPackages:
    react: 16.9.0 => 16.9.0 
    react-native: 0.61.2 => 0.61.2
```

## Current cli version.

 Andrei  ~  desktop  ReactConf   native_navigation ✚ 1 … 1  yarn list @react-native-community/cli 
yarn list v1.12.3
warning Filtering by arguments is deprecated. Please use the pattern option instead.
└─ @react-native-community/cli@3.0.0-alpha.2
✨  Done in 0.57s.

## Description

I'm able to build and run on device using `npx react-native run-android`, however, when I open Android Studio I'm unable to sync gradle. 

## Reproducible Demo
Clean init with `npx react-native init FreshApp`

```  
Caused by: org.gradle.internal.exceptions.LocationAwareException: Script '/Users/Andrei/desktop/ReactConf/node_modules/@react-native-community/cli-platform-android/native_modules.gradle' line: 191
A problem occurred evaluating settings 'ReactConf'.

```

Full log file from Gradle failure:

[idea.log](https://github.com/react-native-community/cli/files/3729600/idea.log)

## Note

I noticed some other issues related to this, I tried to going over them and to find a fix, but I failed so far. Most seem related to a Monorepo structure, which is not my case. 

Any clues how I can fix this?

### LINKED ISSUE #804: Android: latest release broke support custom react-native Root
## Environment

System:
    OS: Windows 10
    CPU: (8) x64 Intel(R) Core(TM) i
    Memory: 15.44 GB 
  Binaries:
    Node: 12.10.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.19.1 - C:\Users\v\AppData\Roaming\npm\yarn.CMD
    npm: 6.12.0 - C:\Program Files\nodejs\npm.CMD
  npmPackages:
    react: 16.9.0 => 16.9.0
    react-native: ^0.61 => 0.61.2

## Description
my directory structure is
andrapp\ <--  has android ./app  ,   gradlew  and so on
rn.common\   <-- has package.json  index.android.js   and ./node_modules

in react-native 60.6   I was able to specify the React native root (in app/build.gradle)
```

project.ext.react = [

        entryFile: "index.android.js",
        // whether to bundle JS and assets in debug mode
        bundleInDebug: false,
        // whether to bundle JS and assets in release mode
        bundleInRelease: true,
        bundleCommand: "ram-bundle",
        enableHermes: false,  // clean and rebuild if changing
        **root:** "${myRnAppRootDir_abs}"
]

```

And  cli was able to correctly invoke   
npx.cmd --quiet react-native config   
from within rn.common directory

However since this change
https://github.com/react-native-community/cli/commit/5724d29dc979a4b645c620109348a1e5c8c74d03#diff-375026a0607eb034a6fc70cca7d74689

**native_modules.gradle**
Now invokes   npx  in 'default' directly, not in React's project root.  Which causes
`npx.cmd --quiet react-native config
`
to complain with 
```
error Unrecognized command "config".

```

Which, then, causes JSON parsing error in native_modules.js

Here is code change in native_modules.js that causes the error on Android

// old code, that was working

// new code that does not work 
cmdProcess = Runtime.getRuntime().exec(command)

From the design/documentation perspective, it seems that the whole infrastructure for getReactNativeProjectRoot()  have been removed, and no alternative/migration documentation has been suggested.

Note also, that react-native 0.61 project template, continues to assert that configuring custom root (where package.json) resides, by specifying root in project.ext.react (in app/build.gradle) -- is still a valid and supported option

https://github.com/facebook/react-native/blob/master/template/android/app/build.gradle

thymikee:
Summary:
---------

Gradle may be run from a driectory that's outside of JS project, with a passed project dir (through `-p` flag) like `path/to/app/android/gradlew bundleRelease -p ../app/android/app`. 

```
project/
  app/ -> js, android, ios files
  scripts/ -> running gradlew from here, e.g. through Fastlane
```

Gradle CWD is usually (maybe always, but they don't give guarantees) set to the directory where it's ran from. This tricks `npx`, which doesn't know where to find `react-native` binary. To account for it, we can spawn `npx` with a CWD passed to it. We don't need an exact path to android project directory, path to the project which holds `@react-native-community/cli` in node_modules is just fine with current config search mechanism.

I'm not 100% sure about this fix, open for discussion.

Fixes #793 
Fixes #804

Test Plan:
----------

I don't have an automated test for this, you'll need to verify on repro provided in #793.

### DIFF
--- a/packages/platform-android/native_modules.gradle
+++ b/packages/platform-android/native_modules.gradle
@@ -2,6 +2,7 @@ import groovy.json.JsonSlurper
 import org.gradle.initialization.DefaultSettings
 import org.apache.tools.ant.taskdefs.condition.Os
 
+def jsAppDir = buildscript.getProperties().get("sourceFile").toString().split("node_modules/@react-native-community/cli-platform-android/native_modules.gradle")[0]
 def generatedFileName = "PackageList.java"
 def generatedFilePackage = "com.facebook.react"
 def generatedFileContentsTemplate = """
@@ -71,12 +72,14 @@ public class PackageList {
 class ReactNativeModules {
   private Logger logger
   private String packageName
+  private String jsAppDir
   private ArrayList<HashMap<String, String>> reactNativeModules
 
   private static String LOG_PREFIX = ":ReactNative:"
 
-  ReactNativeModules(Logger logger) {
+  ReactNativeModules(Logger logger, String jsAppDir) {
     this.logger = logger
+    this.jsAppDir = jsAppDir
 
     def (nativeModules, packageName) = this.getReactNativeConfig()
     this.reactNativeModules = nativeModules
@@ -150,13 +153,13 @@ class ReactNativeModules {
     if (this.reactNativeModules != null) return this.reactNativeModules
     ArrayList<HashMap<String, String>> reactNativeModules = new ArrayList<HashMap<String, String>>()
 
-    def cmdProcess
     def npx = Os.isFamily(Os.FAMILY_WINDOWS) ? "npx.cmd" : "npx"
-    def command = "${npx} --quiet react-native config"
+    def command = "${npx} --quiet --no-install react-native config"
     def reactNativeConfigOutput = ""
 
     try {
-      cmdProcess = Runtime.getRuntime().exec(command)
+      // When running Gradle from CWD that's outside of JS project, it's necessary to pass a CWD that's close to JS app,// so that npx can resolve to correct local CLI, instead of possibly installed global, corrupting config output
+      def cmdProcess = Runtime.getRuntime().exec(command, null, new File(this.jsAppDir))
       def bufferedReader = new BufferedReader(new InputStreamReader(cmdProcess.getInputStream()))
       def buff = ""
       def readBuffer = new StringBuffer()
@@ -164,23 +167,28 @@ class ReactNativeModules {
           readBuffer.append(buff)
       }
       reactNativeConfigOutput = readBuffer.toString()
-    } catch (Exception exception) {
-      this.logger.warn("${LOG_PREFIX}${exception.message}")
-      this.logger.warn("${LOG_PREFIX}Automatic import of native modules failed.")
 
-      def bufferedErrorReader = new BufferedReader(new InputStreamReader(cmdProcess.getErrorStream()))
-      def buff = ""
-      def readBuffer = new StringBuffer()
-      while ((buff = bufferedErrorReader.readLine()) != null){
-          readBuffer.append(buff)
+      if (!reactNativeConfigOutput) {
+        def bufferedErrorReader = new BufferedReader(new InputStreamReader(cmdProcess.getErrorStream()))
+        def errBuff = ""
+        def readErrorBuffer = new StringBuffer()
+        while ((errBuff = bufferedErrorReader.readLine()) != null){
+            readErrorBuffer.append(errBuff)
+        }
+        throw new Exception(readErrorBuffer.toString())
       }
-      this.logger.warn("${LOG_PREFIX}${readBuffer.toString()}")
-
-      return reactNativeModules
+    } catch (Exception exception) {
+      this.logger.error("${LOG_PREFIX}Automatic import of native modules failed.")
+      throw exception
     }
 
     def json = new JsonSlurper().parseText(reactNativeConfigOutput)
     def dependencies = json["dependencies"]
+    def project = json["project"]["android"]
+
+    if (project == null) {
+      throw new Exception("React Native CLI failed to determine Android project configuration. This is likely due to misconfiguration. Config output:\n${json.toMapString()}")
+    }
 
     dependencies.each { name, value ->
       def platformsConfig = value["platforms"];
@@ -211,7 +219,7 @@ class ReactNativeModules {
  *    Exported Extensions
  * ------------------------ */
 
-def autoModules = new ReactNativeModules(logger)
+def autoModules = new ReactNativeModules(logger, jsAppDir)
 
 ext.applyNativeModulesSettingsGradle = { DefaultSettings defaultSettings, String root = null ->
   if (root != null) {
--- a/packages/platform-android/native_modules.gradle
+++ b/packages/platform-android/native_modules.gradle
@@ -158,7 +158,9 @@ class ReactNativeModules {
     def reactNativeConfigOutput = ""
 
     try {
-      // When running Gradle from CWD that's outside of JS project, it's necessary to pass a CWD that's close to JS app,// so that npx can resolve to correct local CLI, instead of possibly installed global, corrupting config output
+      // Running npx from the directory of the JS app which holds this script in its node_modules.
+      // We do so, because Gradle may be ran from a different directory, that's outside of JS project,
+      // in which case npx wouldn't resolve correct `react-native` binary
       def cmdProcess = Runtime.getRuntime().exec(command, null, new File(this.jsAppDir))
       def bufferedReader = new BufferedReader(new InputStreamReader(cmdProcess.getInputStream()))
       def buff = ""
--- a/packages/platform-android/native_modules.gradle
+++ b/packages/platform-android/native_modules.gradle
@@ -160,7 +160,7 @@ class ReactNativeModules {
       }
       output = readBuffer.toString()
       if (!output) {
-        this.logger.error("${LOG_PREFIX}Unexpected empty result of running '${command}' command.")
+        this.logger.error("${LOG_PREFIX}Unexpected empty result of running '${command}' command from '${directory}' directory.")
         def bufferedErrorReader = new BufferedReader(new InputStreamReader(cmdProcess.getErrorStream()))
         def errBuff = ""
         def readErrorBuffer = new StringBuffer()
@@ -171,7 +171,7 @@ class ReactNativeModules {
       }
       return output
     } catch (Exception exception) {
-      this.logger.error("${LOG_PREFIX}Running '${command}' command failed.")
+      this.logger.error("${LOG_PREFIX}Running '${command}' command from '${directory}' directory failed.")
       throw exception
     }
   }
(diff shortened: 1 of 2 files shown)


packages/platform-android/native_modules.gradle
```diff
@@ -2,6 +2,7 @@ import groovy.json.JsonSlurper
 import org.gradle.initialization.DefaultSettings
 import org.apache.tools.ant.taskdefs.condition.Os
 
+def jsAppDir = buildscript.getProperties().get("sourceFile").toString().split("node_modules/@react-native-community/cli-platform-android/native_modules.gradle")[0]
```
thymikee: Thinking about it again, we don't need the app folder. Npx should be fine with running from sourceFile dir as well, which should simplify this a bit. Will update later

thymikee: Wrong trait, packages in node_modules also have `package.json`s, so it won't work with current project root discovery.

packages/platform-android/native_modules.gradle
```diff
@@ -150,37 +153,44 @@ class ReactNativeModules {
     if (this.reactNativeModules != null) return this.reactNativeModules
     ArrayList<HashMap<String, String>> reactNativeModules = new ArrayList<HashMap<String, String>>()
 
-    def cmdProcess
     def npx = Os.isFamily(Os.FAMILY_WINDOWS) ? "npx.cmd" : "npx"
-    def command = "${npx} --quiet react-native config"
+    def command = "${npx} --quiet --no-install react-native config"
```
thymikee: Adding `--no-install` so we don't accidentally install `react-native` into local npm cache. Verified that it's already present in npx shipped with Node 8.

ifyapishore: Is it related to https://github.com/react-native-community/cli/issues/823?

packages/platform-android/native_modules.gradle
```diff
@@ -150,37 +153,42 @@ class ReactNativeModules {
...
     def reactNativeConfigOutput = ""
 
     try {
-      cmdProcess = Runtime.getRuntime().exec(command)
+      // When running Gradle from CWD that's outside of JS project, it's necessary to pass a CWD that's close to JS app,// so that npx can resolve to correct local CLI, instead of possibly installed global, corrupting config output
+      def cmdProcess = Runtime.getRuntime().exec(command, null, new File(this.jsAppDir))
```
ifyapishore: What about PATH variable?

thymikee: > If envp is null, the subprocess inherits the environment settings of the current process.

https://docs.oracle.com/javase/8/docs/api/java/lang/Runtime.html#exec-java.lang.String:A-java.lang.String:A-java.io.File-

AndreiCalazans:
I applied these changes inside node_modules and I"m still seeing an issue, it's no longer the same one pointed out in #793. 

Here is a gist of the error outputted by Gradle 
https://gist.github.com/AndreiCalazans/808217902369be04e0ab5c591afcccc7

thymikee:
@AndreiCalazans it's a normal project structure, in terms of running from Android Studio. I'll need you to debug this a little deeper, as Gradle sync and build works on my machine. This is likely an environmental issue I cannot reproduce.

packages/platform-android/native_modules.gradle
thymikee: Not really sure, but it's preferable to run with `--no-install` anyway. Feel free to send a PR, as I don't want to mix iOS and Android in this one :)

grabbou:
How about simple heuristics here: if `-p` flag is present, spawn `(cd $pValue && react-native config)`, otherwise just `react-native config`?

thymikee:
I'm not sure how to access the -p value. Cc @Salakar

Salakar:
> I'm not sure how to access the -p value. Cc @Salakar

Afaik it's not possible to read this. Only properties passed via `-P` (e.g. `-PcustomProp=foo) or `--args` can be read in the build script.

codestersunion:
In my testing, a `cd <path> && npx...` does not solve the issue.

packages/platform-android/native_modules.gradle
ifyapishore: You are correct. But parent java process (eclipse, gradle, AndroidStudio) do not always have same PATH like you have in shell.

https://stackoverflow.com/questions/30118905/read-path-variable-using-java-in-eclipse-on-mac-os-x-differs-from-shell/36282645

For instance my Eclipse have only PATH=/usr/bin:/bin:/usr/sbin:/sbin and can't find npx.
But in shell PATH=/Users/ify/.jenv/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/usr/local/MacGPG2/bin:/opt/X11/bin:/Users/ify/Library/Android/sdk/tools:/Users/ify/Library/Android/sdk/platform-tools

Workaround is to launch npx via bash under OSX. This way same PATH like in shell will be used.
```
Runtime.getRuntime().exec("bash npx", null, new File("/"));
```

thymikee: If you could come up with a way to run it like this for Windows as well, that would be great!

ifyapishore: I have no Window box to test, but I expect something like this will work
```
Runtime.getRuntime().exec("cmd /C npx", null, new File("/"));
```
https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/cmd

thymikee: Now that I think about it, `bash` environment and `PATH` it holds may differ from a default shell (like `zsh` on latest macOS or `fish`). Doesn't Java has a method to execute the command in a default shell? That would make it easier

ifyapishore: echo $SHELL

grabbou:
@thymikee @Salakar that sounds to me like a serious issue not being able to read args. Being able to do that or at least have the equivalent of `__dirname` would seriously help with reducing the amount of code needed to resolve this.

Let's discuss this tomorrow and figure how to move forward!

ifyapishore:
As far as I know, module scope is better because Gradle can run task in
separate thread with different env variables and current dir.

packages/platform-android/native_modules.gradle
thymikee: I didn't find a reliable way to get the `$SHELL` from Java's Runtime exec, so I'll pass on that. This is also not the scope of this PR.

The issue with not inheriting PATH seems related to the environment, and there are people discussing this in SO:
- https://stackoverflow.com/questions/40168112/path-is-different-when-i-run-bash-from-java-processbuilder
- https://stackoverflow.com/questions/603785/environment-variables-in-mac-os-x

For example, I don't have troubles running this script through Android Studio on macOS Catalina (10.15). Would appreciate any additional help in this field though :)

packages/platform-android/native_modules.gradle
```diff
@@ -71,12 +72,14 @@ public class PackageList {
...
   private String packageName
+  private String jsAppDir
   private ArrayList<HashMap<String, String>> reactNativeModules
 
   private static String LOG_PREFIX = ":ReactNative:"
 
-  ReactNativeModules(Logger logger) {
+  ReactNativeModules(Logger logger, String jsAppDir) {
     this.logger = logger
+    this.jsAppDir = jsAppDir
```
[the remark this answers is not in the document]
thymikee: The module scope in Gradle is not inherited by class instances, so I need to inject it somehow. Found it easier to do through the constructor, similar to how `logger` is passed.

lucasbento:
Bump
What's left to be done for this PR?

Looking forward to v3

thymikee:
It's done, waiting for @grabbou and @Esemesek to review. And maybe @Salakar :)

grabbou:
Do we know why the Windows CI is failing? It's blocking few important PRs that are good to be merged otherwise. I'd ship them all if that's not the case.

packages/platform-android/native_modules.gradle
```diff
@@ -2,6 +2,7 @@ import groovy.json.JsonSlurper
 import org.gradle.initialization.DefaultSettings
 import org.apache.tools.ant.taskdefs.condition.Os
 
+def jsAppDir = buildscript.sourceFile.toString().split("node_modules/@react-native-community/cli-platform-android/native_modules.gradle")[0]
```
grabbou: nit: I think we can just split by `node_modules/@react-native-community/cli-platform-android`, which is a path to the module (the gradle file at the end is not needed and may break if we refactor files).

I was thinking to go with `node_modules`, but that would break in case dependencies were not hoisted.

packages/platform-android/native_modules.gradle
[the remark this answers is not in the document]
grabbou: Interesting, thanks!

packages/platform-android/native_modules.gradle
```diff
@@ -142,45 +145,64 @@ class ReactNativeModules {
...
+   * Runs a process to call the React Native CLI Config command and parses the output
+   */
+  ArrayList<HashMap<String, String>> getReactNativeConfig() {
+    if (this.reactNativeModules != null) return this.reactNativeModules
 
-      return reactNativeModules
+    ArrayList<HashMap<String, String>> reactNativeModules = new ArrayList<HashMap<String, String>>()
+    def npx = Os.isFamily(Os.FAMILY_WINDOWS) ? "npx.cmd" : "npx"
+    def command = "${npx} --quiet --no-install react-native config"
+    // Running npx from the directory of the JS app which holds this script in its node_modules.
```
grabbou: nit: blank space before comment and prefer `/* */`?
nit: I'd rephrase the comment a bit to explain when such a case happens, in other words, when it's run with a `-p` flag. We can also link to this PR for further ref.

mike-oanda:
I believe this has broken projects running in a monorepo as the `react-native config` needs to be run in the package directory, however is now running in the root directory as the node_module dependencies are hoisted.

mbios:
I believe this has broken projects running in a monorepo as the react-native config needs to be run in the package directory, however is now running in the root directory as the node_module dependencies are hoisted.

thymikee:
@mbios could you provide a repro we could investigate?

mbios:
```
git clone git@github.com:brunolemos/react-native-web-monorepo.git
cd react-native-web-monorepo
rm yarn.lock
yarn
yarn android```

but works without removing the yarn.lock as that points to 3.0.0 alpha

thymikee:
Yep, it's a regression, thanks for noticing!

PR #829, opened 2019-10-26
fix: Properly handle --host param for remote debugging

asklar:
Summary:
---------
When running metro on one machine and a react-native app on another, /launch-js-devtools causes the metro process to launch localhost:8081. However, the hostname can be overridden by the --host parameter to react-native start, which because of the hardcoding of localhost, causes the web debugger to not start, breaking the react-native app.

Test Plan:
----------

Manually verified that the RN app can run when running react-native start with both: no params as well as --host <mymachinename>


asklar:
The snapshot test failure seems unrelated to my change?

PR #831, opened 2019-10-30
✨Ensure we run config 'collection' from root

builtbyproxy:
Summary:
---------

When you run `react-native config` it gives a super handy breakdown of all packages in your config. From there the `use_native_modules!` call will parse them all in. However, in my project, this would not run from inside of the ios folder.

In order to call this, I would add `use_native_modules!` to my podfile. I would then call `pod install` and it would throw the error: 
```
Command `config` unrecognized. Make sure that you have run `npm install` and that you are inside a react-native project.
```

Config was undefined because I was calling `react-native config` inside of an 'XCode' project instead of native. Therefore this simple PR is to ensure we go back a directory in order to run `react-native config`

I had no issues from here and would love to discuss to ensure this isn't a project-specific fix. If it is I'm happy to close and upload a patch-package for other googlers coming across this!

Test Plan:
----------
1. Ensure you add the following line to the top of your podfile: 
image

2. **Optional** Ensure you are running least React-Native `0.61.2` (any `0.60` version should suffice)

3. Add `use_native_modules!` to the end of your podfile, much like this:
image

4. ensure you are in your projects `ios` directory. `cd yourProject/ios`
5. run `pod install`

You will encounter the error 
```
Command `config` unrecognized. Make sure that you have run `npm install` and that you are inside a react-native project.
```

If you run this change, it will cd back a folder into your root project where you can collect your config to continue on.

### DIFF
--- a/packages/platform-ios/native_modules.rb
+++ b/packages/platform-ios/native_modules.rb
@@ -17,7 +17,7 @@ def use_native_modules!(config = nil)
   if (!config)
     json = []
 
-    IO.popen("npx --quiet react-native config") do |data|
+    IO.popen('cd "../" ; npx --quiet react-native config') do |data|
       while line = data.gets
         json << line
       end


thymikee:
Thanks for sending a PR! 

Running `npx react-native` from any nested directory should resolve to the `node_modules/react-native` that it's closest to (in your case, the `<projectRoot>/node_modules/react-native`. If it doesn't it may mean you have a global `react-native` or/and `react-native-cli` installed (hence the message "Command `config` unrecognized" – it comes from the legacy init living in these packages). 

Please remove global versions and run again without this patch, it should work.

builtbyproxy:
Awesome! Thanks a lot for explaining that
Out of curiocity what is the harm in cd'ing back to root to run this to account for both use cases?

thymikee:
You have it somewhere, this error message is only present in `react-native-cli` package: https://github.com/react-native-community/cli/blob/master/packages/global-cli/index.js#L154

> Out of curiocity what is the harm in cd'ing back to root to run this to account for both use cases?

Supporting monorepos

builtbyproxy:
Awesome, thanks a lot for the clarification @thymikee

builtbyproxy:
Hey @thymikee 
Sorry to dig up an old issue/PR

Im having this same error now in android. I can confirm I do not have any form of `react-native-cli` installed globally. 

if I run `yarn run react-native config` it will work in my android directory. However if I run `npx --quiet react-native config` it won't..

My error message exactly is: 
```
Command `config` unrecognized. Make sure that you have run `npm install` and that you are inside a react-native project.
```

if i run in the same android directory `react-native -v` it tells me `zsh: command not found: react-native`

Any help would be HUGELY appreciated!!

thymikee:
Uninstall global `react-native-cli` and `react-native` packages, they're no longer necessary (you can run `npx react-native` or `yarn react-native` locally) and happen to conflict in some situations.

builtbyproxy:
I can confirm I do not have any react related packages installed globally.
`yarn global list` and `npm list -g --depth 0` do not show any react packages

builtbyproxy:
Additionally, if I cd into my android folder and run `npx react-native -v` it prints out: 
image

thymikee:
Please uninstall `react-native-cli`, wherever it is. It often conflicts with local CLI

builtbyproxy:
ooof progress!! Weird, sorry I didnt think those would clash, im not entirely across the node ecosystem and it's package management. Ill check in as I progress thankyou!

builtbyproxy:
I can confirm this has fixed the problem :) Thanks a lot @thymikee 
Ive got more errors but they seem to be unrelated to the config collection

builtbyproxy:
As a final update on this issue. It turns out I had `react-native` packages installed globally for a different `nvm` version. This was then being used in my project causing issues. Anyone reading this later please ensure you check your `nvm` versions as well.

SaeedZhiany:
@builtbyproxy 

I even haven't installed `nvm` on my OS
image

also running `npx react-native -v` results in error message `error: unknown option '-v'`

I also uninstalled react-native and react-native-cli globally using `npm uninstall -g react-native react-native-cli`

Where else I should check to ensure no other cli globally installed.

PR #832, opened 2019-10-30
imp: Extract debugger-ui

Esemesek:
Summary:
---------

Extract `debugger-ui` module out of `cli`.

TODO:
- [x] - Cleanup `html` file (extract styles, scripts and assets)
- [x] - Improve dev scripts

Next steps:
- [ ] - Rewrite to Typescript
- [ ] - Use React

Test Plan:
----------

- [x] - Debugger UI working


packages/cli/package.json
```diff
@@ -31,6 +31,7 @@
     "@react-native-community/cli-platform-ios": "^3.0.0-alpha.7",
     "@react-native-community/cli-tools": "^3.0.0-alpha.7",
     "@react-native-community/cli-types": "^3.0.0-alpha.7",
+    "@react-native-community/debugger-ui": "^3.0.0-alpha.7",
```
thymikee: need to add debugger-ui to tsconfig references

packages/debugger-ui/src/ui/index.js
```diff
@@ -0,0 +1,200 @@
+/* eslint-env browser */
+import './index.css';
+import blueIcon from './assets/blue-icon.png';
+import grayIcon from './assets/gray-icon.png';
+import orangeIcon from './assets/orange-icon.png';
+
+('use strict');
+(function() {
```
thymikee: Do we need this IIFE? Isn't parcel wrapping the bundled module?

Esemesek: It is unnecessary, left it because this was present in the original script. Will change that in a sec.

grabbou:
Thanks for sending this over @Esemesek and happy to see you back with PRs. Just wondering if you have applied the things we have discussed in the office already? I believe we have talked about extracting out the entire debugger middleware out to a separate package.

Esemesek:
@grabbou I extracted the debugger UI as a middleware (https://github.com/react-native-community/cli/blob/debugger-ui/packages/cli/src/commands/server/middleware/MiddlewareManager.ts#L48). Or did you mean to extract the whole `MiddlewareManager` to a separate package?

grabbou:
Is this PR ready to ship / be reviewed? Also, shall we target `next` or `master` with this one? I think there's a little confusion (at least on my end) which branch is the preferred one for breaking changes from now on.

packages/debugger-ui/src/middleware/index.ts
```diff
@@ -0,0 +1,6 @@
+import serveStatic from 'serve-static';
+import path from 'path';
+
+export function debuggerUIMiddleware() {
+  return serveStatic(path.join(__dirname, '..', 'ui'));
```
grabbou: nit: `../ui` is just fine, `path` normalises for Windows by default.

grabbou:
Let's merge it to `master` since it's non breaking.

PR #844, opened 2019-11-14
fix: debugger-ui name and reload button

thymikee:
Summary:
---------

- rename module to `@react-native-community/cli-debugger-ui`
- fix "Reload app" 

Test Plan:
----------

Reload button works

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -8,7 +8,6 @@
   "editor.formatOnSave": true,
   "flow.useNPMPackagedFlow": true,
   "javascript.validate.enable": false,
-  "prettier.eslintIntegration": true,
   "eslint.validate": [
     "javascript",
     "javascriptreact",
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
     "prebuild": "yarn build:ts",
     "build": "node ./scripts/build.js && yarn build:debugger",
     "build:ts": "node ./scripts/buildTs.js",
-    "build:debugger": "yarn workspace @react-native-community/debugger-ui build",
+    "build:debugger": "yarn workspace @react-native-community/cli-debugger-ui build",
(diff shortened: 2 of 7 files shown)


slorber:
Hey,

Like this button.

Wonder if it would be possible to customize this view more and let users add custom buttons?

For example the options in the dev shake menu, or other contextual things like a button to skip current form in DEV?

thymikee:
Sure. I think we could include some of those by default. But making the UI extensible sounds like terrific idea as well. 

We plan to rewrite the debugger UI to React, which will make it possible to compose the whole experience from reusable components. By providing a custom entry-point and exporting some common functionalities (e.g. logging, priority override, connected devices, controls) from the `@react-native-community/cli-debugger-ui` package, we could empower users to build their own UIs and eventually ship them as defaults if they prove helpful.

However, there's also an ongoing effort from [Flipper](https://fbflipper.com/) core team to make it easier to integrate with React Native and write plugins. I think the near future where I'd like to be is making Flipper a go-to development tool that would embed other tools (like React DevTools, Reactotron, [perf logging](https://github.com/oblador/flipper-plugin-react-native-performance), and many more).

cc @rickhanlonii @axe-fb

slorber:
That would be great :) 
My current solution is to include a dev toolbar at the top of all my apps with contextual actions (like "fill this form"). But it's a bit annoying to have to use some screen space for that while it could fit in another UI.

rickhanlonii:
@slorber for the dev shake menu we added a way to add new options, does this work for you? https://github.com/facebook/react-native/pull/25848

slorber:
@rickhanlonii not exactly, I find it particularly annoying to have to take my device from the desk and shake it :)

If there would be a devtools option to toggle the dev menu without shaking, that would remove a part of that pain and is probably more easy to support, as it would be static ;)

PR #845, opened 2019-11-14
chore: remove platforms from cli dependencies

thymikee:
Summary:
---------

Followup PR to https://github.com/react-native-community/cli/pull/767.
If you're curious, iOS and Android platforms are already shipping with `react-native` as default platforms since 0.60: https://github.com/facebook/react-native/blob/v0.60.0/package.json#L83-L85.

Test Plan:
----------

Green CI


grabbou:
Awesome! The core is now decoupled from the RN specific code.

PR #852, opened 2019-11-15
fix: temporary fix for a regression in monorepos

thymikee:
Summary:
---------

Regression introduced in #824. Still looking for a better fix, because otherwise we'll need to reopen #793 and #804.

cc @mbios

Test Plan:
----------

### DIFF
--- a/packages/platform-android/native_modules.gradle
+++ b/packages/platform-android/native_modules.gradle
@@ -191,7 +191,7 @@ class ReactNativeModules {
      * (e.g. when running with -p flag), in which case npx wouldn't resolve correct `react-native` binary.
      */
     def dir = new File(this.jsAppDir)
-    def reactNativeConfigOutput = this.getCommandOutput(command, dir)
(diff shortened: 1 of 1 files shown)


patrickschmelter:
https://github.com/react-native-community/cli/commit/1799e4440123821db1085517695c51ad7c2dacb8#commitcomment-36010896

grabbou:
Since this is causing issues, I am wondering whether we shouldn't revert #824 entirely and just leave it as it was before?

I believe the amount of regressions is larger than a small number of developers facing issues with a non-standard Gradle project setup (e.g. https://github.com/react-native-community/cli/issues/804).

We should probably think about it once again and ship a better fix in the next release.

mrinx:
Any news on this one? Currently if I start a fresh project with `npx react-native init AwesomeProject` on MacOS, `npx react-native run-android` fails unless I uncomment the dir.

This may cause some frustration to people who are just starting with the framework and have no idea where to look to fix this.

PR #856, opened 2019-11-19
fix: update Xcode version range to >= 10; change wording of supported version

lucasbento:
Summary:
---------

This PR fixes an issue with Xcode being shown as `Minimum required version: 10.0.0` even though we officially support `11.x`.

This issue has been reported through https://twitter.com/chakrihacker/status/1196750533224091648.

### DIFF
--- a/packages/cli/src/commands/doctor/doctor.ts
+++ b/packages/cli/src/commands/doctor/doctor.ts
@@ -1,6 +1,5 @@
 import chalk from 'chalk';
 import {logger} from '@react-native-community/cli-tools';
-import semver from 'semver';
 import {getHealthchecks, HEALTHCHECK_TYPES} from './healthchecks';
 import {getLoader} from '../../tools/loader';
 import printFixOptions, {KEYS} from './printFixOptions';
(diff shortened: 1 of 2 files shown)


packages/cli/src/commands/doctor/versionRanges.ts
```diff
@@ -8,5 +8,5 @@ export default {
   ANDROID_SDK: '>= 26.x',
   ANDROID_NDK: '>= 19.x',
   // iOS
-  XCODE: '10.x',
+  XCODE: '11.x',
```
thymikee: um, but we support both 10.x and 11.x?

packages/cli/src/commands/doctor/doctor.ts
```diff
@@ -36,22 +35,17 @@ const printIssue = ({
...
     const versionToShow = version && version !== 'Not Found' ? version : 'N/A';
-    const cleanedVersionRange = semver.valid(semver.coerce(versionRange)!);
 
-    if (cleanedVersionRange) {
-      logMessage(`- Version found: ${chalk.red(versionToShow)}`);
-      logMessage(
-        `- Minimum version required: ${chalk.green(cleanedVersionRange)}`,
-      );
+    logMessage(`- Version found: ${chalk.red(versionToShow)}`);
+    logMessage(`- Version supported: ${chalk.green(versionRange)}`);
```
thymikee: We could show the version range instead of semver coerced version. Here's a sample output how it looks now: 
image

PR #867, opened 2019-11-26
fix: add cocoapods requirement

Gray-Wind:
As file is using `Pod`, cocoapods should be imported.

Summary:
---------

This PR might solve this issue: https://github.com/CocoaPods/Core/issues/601

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -36,6 +36,13 @@ commands:
           root: .
           paths:
             - .
+
+  install-cocoapods:
+    steps:
+      - attach_workspace:
+           at: ~/react-native-cli
+      - run: gem install cocoapods
+
   run-lint:
     steps:
       - attach_workspace:
@@ -72,6 +79,7 @@ jobs:
   cocoa-pods:
     executor: noderuby
     steps:
+      - install-cocoapods
       - run-cocoa-pods-tests
   unit-tests:
     executor: node8
--- a/packages/platform-ios/native_modules.rb
+++ b/packages/platform-ios/native_modules.rb
@@ -218,7 +218,7 @@ if $0 == __FILE__
       })
       @printed_messages.must_equal [
         "Detected React Native module pod for ios-dep",
-        "Detected React Native module pods for ios-dep, and ios-dep"
+        "Detected React Native module pods for ios-dep and ios-dep"
       ]
     end
 
--- a/packages/platform-ios/native_modules.rb
+++ b/packages/platform-ios/native_modules.rb
@@ -4,6 +4,7 @@
 # imports those into your current target.
 #
 require 'pathname'
+require 'cocoapods'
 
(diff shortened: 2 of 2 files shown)


thymikee:
@alloy @orta can you advise here?

orta:
I think this should be fine, normally when running for `pod install` CocoaPods has to have been included at some point. If you're calling this code another way, then it's possible to not be set up.

To get the CI green, I think it will need a `gem install cocoapods` before running the tests

Gray-Wind:
I'm not sure where to do it. In CircleCI config? Just add another step to install this dependency?

orta:
Hrm, these other failing tests look pretty worrying - unsure what's going on there though. I don't feel like they should be affected by these changes

Gray-Wind:
I noticed, that tests are failing for a long time already: https://github.com/react-native-community/cli/commits/master

alloy:
The require wasn’t there previously to keep the test suite light-weight. However, it’s probably better to not mock and actually require the env 👍

Gray-Wind:
I've dropped mocking cocoapods, as there is no more need.

But test is failing because of

```
--- expected
+++ actual
@@ -1 +1 @@
-["Detected React Native module pod for ios-dep", "Detected React Native module pods for ios-dep, and ios-dep"]
+["Detected React Native module pod for ios-dep", "Detected React Native module pods for ios-dep and ios-dep"]
```

This error shouldn't be related to changes in this branch.

Gray-Wind:
Should I do something else?

alloy:
Yeah update that test, it’s probably using the real ActiveSupport `to_sentence` implementation now.

Gray-Wind:
Done.

alloy:
Nice, thanks! This looks good to me 👍 the one thing is those failing CI tests…

It’s true that those have been failing previously as well, but for the last commit the results were:

```
Test Suites: 1 failed, 67 passed, 68 total
Tests:       5 failed, 2 skipped, 2 todo, 252 passed, 261 total
Snapshots:   6 failed, 19 passed, 25 total
```

Whereas for this PR the results are:

```
Test Suites: 2 failed, 66 passed, 68 total
Tests:       6 failed, 2 skipped, 2 todo, 251 passed, 261 total
Snapshots:   6 failed, 19 passed, 25 total
```

So it looks like there may have been additional regression?

Gray-Wind:
I think it is flakiness of tests themselves. You could check not the last, but previous one and there are 2 failed tests:
https://github.com/react-native-community/cli/runs/305041507

alloy:
Gotcha, works for me 👍 

@thymikee What’s the process of this repo? Should I squash & merge and add `fix` to the commit message to cut a patch release?

thymikee:
We don't have an automated release yet, but I'll edit the message and merge myself then :)

PR #435, opened 2019-06-20
Add import for `applicationId.R` class in generated `PackageList.java`

robertying:
Summary:
---------

- Fix https://github.com/react-native-community/cli/issues/434
- Also remove the redundant semicolon after package imports

Test Plan:
----------

It fixes #434.

Try to modify `node_modules/@react-native-community/cli-platform-android/native_modules.gradle` in https://github.com/robertying/autolink-issue-example and it can build succesfully now!

Issue #434, opened 2019-06-20
Autolink failed to get `react-native-code-push` deployment key


robertying:
## Environment

System:
    OS: macOS 10.14.5
    CPU: (4) x64 Intel(R) Core(TM) i3-8100B CPU @ 3.60GHz
    Memory: 224.80 MB / 8.00 GB
    Shell: 5.7.1 - /usr/local/bin/zsh
  Binaries:
    Node: 12.4.0 - /usr/local/Cellar/node/12.4.0/bin/node
    Yarn: 1.16.0 - /usr/local/Cellar/node/12.4.0/bin/yarn
    npm: 6.9.0 - /usr/local/Cellar/node/12.4.0/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  SDKs:
    iOS SDK:
      Platforms: iOS 12.2, macOS 10.14, tvOS 12.2, watchOS 5.2
  IDEs:
    Android Studio: 3.4 AI-183.6156.11.34.5522156
    Xcode: 10.2.1/10E1001 - /usr/bin/xcodebuild
  npmPackages:
    react: 16.8.6 => 16.8.6 
    react-native: 0.60.0-rc.1 => 0.60.0-rc.1

## Description

I put `reactNativeCodePush_androidDeploymentKey` in `res/values/codepush.xml` with:

```xml
<resources>
    <string moduleConfig="true" name="reactNativeCodePush_androidDeploymentKey">example</string>
</resources>
```

But got error in `/android/build/generated/rn/src/main/java/com/facebook/react/PackageList.java`

```
error: cannot find symbol
      new AsyncStoragePackage(),new BlurViewPackage(),new RNBackgroundFetchPackage(),new CalendarEventsPackage(),new CodePush(getResources().getString(R.string.reactNativeCodePush_androidDeploymentKey), getApplicationContext(), BuildConfig.DEBUG),new RNDeviceInfo(),new ExtraDimensionsPackage(),new RNGestureHandlerPackage(),new Interactable(),new LinearGradientPackage(),new ReanimatedPackage(),new RNSharePackage(),new VectorIconsPackage(),new RNCWebViewPackage(),new RNFetchBlobPackage()
                                                                                                                                                               ^
  symbol:   variable reactNativeCodePush_androidDeploymentKey
  location: class string
```

## Reproducible Demo

https://github.com/robertying/autolink-issue-example

Run `cd android && ./gradlew assembleDebug` and you can get the error.

[...]

PR #871, opened 2019-11-29
feat: replace {{ packageName }} in deps' configuration

sjchmiela:
Summary:
---------

Hard importing `${packageName}.R` and `${packageName}.BuildConfig` helps to install React Native modules which use `R` and `BuildConfig` without package prefix (eg. `react-native-code-push`, see https://github.com/react-native-community/cli/pull/435 fixing  https://github.com/react-native-community/cli/issues/434.) `BuildConfig` import has been introduced in https://github.com/react-native-community/cli/pull/258, so right from the beginning (specific commit: https://github.com/react-native-community/cli/commit/64f3f5a3db137c5cee4bbe26b18b26a7c55a205e).

This pull request:
- adds support for interpolating `{{ packageName }}` in `packageImportPath` and `packageInstance`
- adds support for `generatePackageList` task to libraries (imagine a project where its the submodule (`:app` depends on ➡️ `:library`) which depends on autolinked React Native modules)

Test Plan:
----------

I introduced these changes to my project, added to `react-native-maps`' `react-native.config.js`

```diff
 module.exports = {
   dependency: {
     platforms: {
       android: {
         sourceDir: './lib/android',
+        packageInstance: 'new MapsPackage({{ packageName }}.BuildConfig)',
+        packageImportPath: 'import {{ packageName }}.R;\nimport com.maps;'
       },
     },
   },
 };
```

started the app, and `PackageList.java` contained

```java
// react-native-maps
import host.exp.exponent.R;
import com.maps;
```

and

```java
new MapsPackage(host.exp.exponent.BuildConfig),
```

### DIFF
--- a/packages/platform-android/native_modules.gradle
+++ b/packages/platform-android/native_modules.gradle
@@ -126,9 +126,11 @@ class ReactNativeModules {
     String packageClassInstances = ""
 
     if (packages.size() > 0) {
+      def interpolateDynamicValues = {
+        it.replace("{{ packageName }}", packageName)
+      }
       packageImports = packageImports + packages.collect {
-        String interpolatedPackageImportPath = it.packageImportPath.replace("{{ packageName }}", packageName)
-        "// ${it.name}\n${interpolatedPackageImportPath}"
+        "// ${it.name}\n${interpolateDynamicValues(it.packageImportPath)}"
       }.join('\n')
       packageClassInstances = ",\n      " + packages.collect { it.packageInstance }.join(",\n      ")
     }
--- a/packages/platform-android/native_modules.gradle
+++ b/packages/platform-android/native_modules.gradle
@@ -134,22 +134,22 @@ class ReactNativeModules {
                 // let's prefix all non-FQDN references to those classes with
                 // the package name.
                 //
-                // We want to match:
+                // We want to match "R" or "BuildConfig":
                 //  - new Package(R.string…),
                 //  - Module.configure(BuildConfig);
                 //    ^ hence including (BuildConfig|R)
-                // but we don't want to match:
+                // but we don't want to match "R":
                 //  - new Package(getResources…),
                 //  - new PackageR…,
                 //  - new Royal…,
                 //    ^ hence excluding \w before and after matches
-                //  - Module.configure({{ packageName }}.BuildConfig);
+                // and "BuildConfig" that has FQDN reference:
+                //  - Module.configure(com.acme.BuildConfig);
                 //    ^ hence excluding . before the match.
                 .replaceAll(~/([^.\w])(BuildConfig|R)([^\w])/, {
                   wholeString, prefix, className, suffix ->
-                    "${prefix}{{ packageName }}.${className}${suffix}"
+                    "${prefix}packageName.${className}${suffix}"
                 })
-                .replace("{{ packageName }}", packageName)
       }
       packageImports = packages.collect {
         "// ${it.name}\n${interpolateDynamicValues(it.packageImportPath)}"
--- a/packages/platform-android/native_modules.gradle
+++ b/packages/platform-android/native_modules.gradle
@@ -128,11 +128,11 @@ class ReactNativeModules {
     if (packages.size() > 0) {
       def interpolateDynamicValues = {
         it
-                // Before adding the {{ packageName }} replacement mechanism,
+                // Before adding the package replacement mechanism,
                 // BuildConfig and R classes were imported automatically
-                // into the scope of the file. In order to be backwards-compatible
-                // let's prefix all non-FQDN references to those classes with
-                // the package name.
+                // into the scope of the file. We want to replace all
+                // non-FQDN references to those classes with the package name 
+                // of the MainApplication.
                 //
                 // We want to match "R" or "BuildConfig":
                 //  - new Package(R.string…),
@@ -148,7 +148,7 @@ class ReactNativeModules {
                 //    ^ hence excluding . before the match.
                 .replaceAll(~/([^.\w])(BuildConfig|R)([^\w])/, {
                   wholeString, prefix, className, suffix ->
-                    "${prefix}packageName.${className}${suffix}"
+                    "${prefix}${packageName}.${className}${suffix}"
                 })
       }
       packageImports = packages.collect {
--- a/packages/platform-android/native_modules.gradle
+++ b/packages/platform-android/native_modules.gradle
@@ -127,7 +127,29 @@ class ReactNativeModules {
 
     if (packages.size() > 0) {
       def interpolateDynamicValues = {
-        it.replace("{{ packageName }}", packageName)
+        it
+                // Before adding the {{ packageName }} replacement mechanism,
+                // BuildConfig and R classes were imported automatically
+                // into the scope of the file. In order to be backwards-compatible
+                // let's prefix all non-FQDN references to those classes with
+                // the package name.
+                //
+                // We want to match:
+                //  - new Package(R.string…),
+                //  - Module.configure(BuildConfig);
+                //    ^ hence including (BuildConfig|R)
+                // but we don't want to match:
+                //  - new Package(getResources…),
+                //  - new PackageR…,
+                //  - new Royal…,
+                //    ^ hence excluding \w before and after matches
+                //  - Module.configure({{ packageName }}.BuildConfig);
+                //    ^ hence excluding . before the match.
+                .replaceAll(~/([^.\w])(BuildConfig|R)([^\w])/, {
+                  wholeString, prefix, className, suffix ->
+                    "${prefix}{{ packageName }}.${className}${suffix}"
+                })
+                .replace("{{ packageName }}", packageName)
       }
       packageImports = packages.collect {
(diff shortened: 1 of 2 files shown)


sjchmiela:
I should note this will unfortunately break `react-native-code-push` installations until it adds those value placeholders to the config (or users are provided with a `patch-package` solution 😕).

packages/platform-android/native_modules.gradle
```diff
@@ -126,11 +126,15 @@ class ReactNativeModules {
     String packageClassInstances = ""
 
     if (packages.size() > 0) {
-      packageImports = "import ${packageName}.BuildConfig;\nimport ${packageName}.R;\n\n"
+      def interpolateDynamicValues = {
+        it.replace("{{ packageName }}", packageName)
+      }
       packageImports = packageImports + packages.collect {
```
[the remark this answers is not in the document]
sjchmiela: Sure! I left the `+=` specifically in case someone would add something to that variable before the `if` or as the initial value — this would be more failsafe, but I guess we can remove this safely.
```suggestion
      packageImports = packages.collect {
```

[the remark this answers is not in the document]
sjchmiela: Huh, interesting thing is happening with that suggestion, it removes the initial space too…
```suggestion
       packageImports = packages.collect {
```

packages/platform-android/native_modules.gradle
```diff
@@ -126,11 +126,15 @@ class ReactNativeModules {
     String packageClassInstances = ""
 
     if (packages.size() > 0) {
-      packageImports = "import ${packageName}.BuildConfig;\nimport ${packageName}.R;\n\n"
-      packageImports = packageImports + packages.collect {
-        "// ${it.name}\n${it.packageImportPath}"
+      def interpolateDynamicValues = {
+        it.replace("{{ packageName }}", packageName)
+      }
+       packageImports = packages.collect {
```
sjchmiela:
```suggestion
      packageImports = packages.collect {
```

sjchmiela:
`packageName` gets resolved to `packageName` of `:app` (let's say `com.app`) via `react-native config`, whereas `:library` has another package, eg. `com.library`. Adding `import com.app.R` in `:library` wouldn't work (without adding a dependency on `:app`, which is something I wouldn't need/shouldn't be required to do).

grabbou:
This is great - @sjchmiela - what is the status of it? Is this good to go? One of the previous comments mentions that this is potentially going to break CodePush. Was this already addressed?

sjchmiela:
No, it wasn't up until now! I have brought back support for eg.
```java
new CodePush(
    getResources().getString(R.string.CodePushDeploymentKey),
    getApplicationContext(),
    BuildConfig.DEBUG
)
```
by a middle step of replacing `R` with `{{ packageName }}.R` and `BuildConfig` with `{{ packageName }}.BuildConfig`.

I have tested the change by manually copying the `interpolateDynamicValues` function to a Gradle file and executing it on
```java
// 1
new CodePush(getResources().getString(R.string.CodePushDeploymentKey), getApplicationContext(), BuildConfig.DEBUG)
// 2
new PackageR(), new Royal(), Module.configure(other.package.BuildConfig)
```
which printed
```java
// 1
new CodePush(getResources().getString(com.test.sjchmiela.R.string.CodePushDeploymentKey), getApplicationContext(), com.test.sjchmiela.BuildConfig.DEBUG)
// 2
new PackageR(), new Royal(), Module.configure(other.package.BuildConfig)
```

Let me know if this will be the proper way to go in your opinion.

PS. I think the test failures weren't caused by this change. 🤔

Salakar:
Thanks for making this backwards compat @sjchmiela - LGTM

packages/platform-android/native_modules.gradle
```diff
@@ -126,11 +126,37 @@ class ReactNativeModules {
...
+                //  - new Package(R.string…),
+                //  - Module.configure(BuildConfig);
+                //    ^ hence including (BuildConfig|R)
+                // but we don't want to match:
+                //  - new Package(getResources…),
+                //  - new PackageR…,
+                //  - new Royal…,
+                //    ^ hence excluding \w before and after matches
+                //  - Module.configure({{ packageName }}.BuildConfig);
+                //    ^ hence excluding . before the match.
```
grabbou: This is a nice explanation, thanks!

packages/platform-android/native_modules.gradle
```diff
@@ -126,11 +126,37 @@ class ReactNativeModules {
...
+      def interpolateDynamicValues = {
+        it
+                // Before adding the {{ packageName }} replacement mechanism,
+                // BuildConfig and R classes were imported automatically
+                // into the scope of the file. In order to be backwards-compatible
+                // let's prefix all non-FQDN references to those classes with
+                // the package name.
+                //
+                // We want to match:
+                //  - new Package(R.string…),
```
grabbou: So I guess the essence of this PR is that: "if automatic prefixing doesn't work - produces an error, please use overrides and set {{packageName}} yourself"? 

If we consider this deprecated, we should probably print a warning that this will stop working in the future.

If we consider this backward-compatible support something that will stay for longer, then I would say we should document the exact use-case in the documentation - I've seen a comment from you to @Salakar, but being not that proficient with Android, I have trouble wrapping my head around it fully. In what cases this is going to break?

sjchmiela: I'm not sure if I have understood your comment correctly, so let me rephrase the essence:

> if the library doesn't use `{{ packageName }}` template, but it looks like it expects `BuildConfig` and `R` to be imported globally, prefix their usages with the template

This way we:
- keep the solution backwards-compatible,
- if for some reason a developer would expect to be able to access some other `packageName`-dependant class, they will be able to use `{{ packageName }}` instead of making a pull request to add another `import {{ packageName }}.AnotherClass` to the top of the file.

I will explain the exact use case by the exact use case that resulted in this PR 😃 

In Expo repository our Android project currently looks like this:

- `android/app` — where `host.exp.exponent.MainApplication` lives
- `android/expoview` — where the `host.exp.expoview` library lives, a monolith native Expo library containing vendored native modules (eg. `react-native-screens`)
- `packages/…` — unimodules projects

`app` depends on `expoview`, `expoview` depends on unimodules. It doesn't need to depend on React Native modules, since all of them are copied into `expoview`.

What I wanted to do was to replace copied RN modules with autolinking (from `/node_modules`).

In order to do this I added the autolinking script to the `expoview` project. The generated `PackageList` contained `import host.exp.exponent.R;`. Notice the `host.exp.exponent` — it has been inferred from the package of the `MainApplication`, so a different Android project. 

On one hand it makes sense, given that `MainApplication` is in `host.exp.exponent` package; on the other hand, this `PackageList` was being used in `host.exp.expoview` package, so the appropriate `R` class would be `host.exp.expoview.R` package. This mismatch resulted in a compilation error, since the mispackaged class could not have been found.

This particular PR won't solve this use case fully (if we wanted to depend on CodePush in `expoview`, it would still import `host.exp.exponent.BuildConfig`, so the proper fix would also include a different strategy for fetching `packageName`), but it's good enough for us and other multi-library Android projects (+ I suspect the `packageName` strategy fix would a major refactor).

To sum up, this will allow using React Native autolinking in Android workspaces where one or more applications-projects depend on an internal Android library-project which autolinks React Native modules. 🙂

grabbou: Before I dive deep, let's clarify one thing (I need to ask fewer questions at once, haha). Happy to hop on a call as well in case it's easier.

> if the library doesn't use {{ packageName }} template, but it looks like it expects BuildConfig and R to be imported globally, prefix their usages with the template

What is the benefit of doing a custom property on a CLI config like this: `new AwesomePackage({{ packageName }}.R.string.key)` if I can just leave do `new AwesomePackage(R.string.key)` and the backward-compatible mechanism will take care of prefixes?

sjchmiela: None! 😅  Maybe the PR’s essence should be rephrased to something along the lines of

> only use `{{ packageName }}.{R,BuildConfig}` if they are expected

which can also be read as

> do not import `{{ packageName }}.{R,BuildConfig}` if they aren’t used

and it could have been done like:

1. see if any of the packages uses R or BuildConfig class
2. if they do, add respective imports to the imports section of the file

I thought adding a placeholder would make it easier to maintain (if for whatever reason any package would need any other packageName-dependant class, it wouldn’t require any CLI change).

grabbou:
I am not sure if this PR will fix #938. The issue that #938 is having is due to multiple apps - CLI currently doesn't support an array of apps and will always use the first one.

grabbou:
The purpose of this PR was to enable a scenario where we don't import the `import {{packageName}}.R;` and `import {{packageName}}.BuildConfig` to the scope of the file.

There are scenarios where this would break the build.

What this PR does is... rather than importing the `BuildConfig` to the scope, it will replace its reference from `BuildConfig` to `{{packageName}}.BuildConfig` directly at the call site.

As a result, when library author uses FQDN reference themselves, e.g. `com.acme.BuildConfig` rather than `BuildConfig`, we will not refer to MainApplication `packageName` from that file, making it work.

PR #873, opened 2019-11-30
Check java keywords when init new project

sunnylqm:
Summary:
---------
Use java keywords as a project name would get error like `Package 'com.xxxx' from AndroidManifest.xml is not a valid Java package name as 'xxxx' is a Java keyword.`

Test Plan:
----------
Use reserved words to init a new project will fail


grabbou:
Thanks for sending this. Can you run ESLint locally and fix the issues? I believe that this code, while looking good from the technical perspective, needs to break the line after each element - otherwise it's not passing the tests.

To make it more readable, maybe we could move the reserved keywords to a separate file.

grabbou:
Okay, to speed things up, I have updated the PR in place. Will merge once the CI passes.

grabbou:
e2e tests seem to be flaky (I have seen this error before). Going to merge as is as this is unlikely to be causing issues.

PR #875, opened 2019-12-03
feat(yarn): prefer local react-native insteda of global one using npx

sibelius:
Summary:
---------

run local react-native cli version, so it does not have some kind of mismatch of react-native version

Test Plan:
----------

Run against a monorepo using yarn workspaces without nohoist option

### DIFF
--- a/packages/platform-android/native_modules.gradle
+++ b/packages/platform-android/native_modules.gradle
@@ -183,8 +183,7 @@ class ReactNativeModules {
     if (this.reactNativeModules != null) return this.reactNativeModules
 
     ArrayList<HashMap<String, String>> reactNativeModules = new ArrayList<HashMap<String, String>>()
-    def npx = Os.isFamily(Os.FAMILY_WINDOWS) ? "npx.cmd" : "npx"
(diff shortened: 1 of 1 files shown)


sibelius:
before this fix the error was like this one

```jsx
> React Native CLI failed to determine Android project configuration. This is likely due to misconfiguration. Config output:
  [root:/app/packages/app/android, reactNativePath:/app/node_modules/react-native, dependencies:[:], commands:[], assets:[], platforms:[:], haste:[providesModuleNodeModules:[], platforms:[]], project:[:]]
```

sibelius:
can we use `npm start react-native config` ?

thymikee:
No, we can't. "start" usually (and likely not always, as projects differ) runs metro and won't work with extra "config" argument. Can you share a repro where the problem happens so we can investigate?

sibelius:
like this https://github.com/entria/entria-fullstack

but we need to remove nohoist on that

sibelius:
worked well here https://github.com/entria/entria-fullstack with npx

gonna need more time to figure it out why our internal repo is failing

I'm gonna reopen if I figure it out

PR #879, opened 2019-12-05
[hold] fix: pass watch: false to metro for offline bundles

cdsanchez:
Summary:
---------

**Note:** This PR cannot land until the corresponding PR in metro is released. This PR will be updated once that happens.

This PR goes along with a corresponding change in https://github.com/facebook/metro/pull/491, in which I'm adding a new flag to metro config which allows me to configure the `watch` flag that it passes to `JestHasteMap`. I need this capability because having `watch` always set to `true` causes failures on some internal build machines, due to the high number of file watchers required.

More specifically, this sets the `watch` flag to false on the metro config when invoking metro via `react-native bundle`.

Test Plan:
----------

Using the corresponding local change in `metro` (via npm link):

Ran `react-native start`
- Observed that the `watch` flag in `metro` was true

Ran `react-native bundle`
- Observed that `watch` in `metro` was set to `false`

Observations were captured via `console.log` statements in metro.

If there are any other test cases you would like me to run, please let me know!

### DIFF
--- a/packages/cli/src/commands/bundle/buildBundle.ts
+++ b/packages/cli/src/commands/bundle/buildBundle.ts
@@ -50,6 +50,9 @@ async function buildBundle(
     config: args.config,
   });
 
+  // Don't need file watcher for building offline bundle
+  config.watch = false;
+
   if (config.resolver.platforms.indexOf(args.platform) === -1) {
     logger.error(
       `Invalid platform ${
--- a/packages/cli/src/tools/loadMetroConfig.ts
+++ b/packages/cli/src/tools/loadMetroConfig.ts
@@ -59,6 +59,7 @@ export interface MetroConfig {
     assetPlugins?: Array<string>;
   };
   watchFolders: string[];
+  watch: boolean;
   reporter?: any;
 }
 
@@ -110,6 +111,7 @@ export const getDefaultConfig = (ctx: Config): MetroConfig => {
       ),
     },
     watchFolders: getWatchFolders(),
+    watch: true,
   };
 };
 
@@ -122,6 +124,7 @@ export interface ConfigOptionsT {
   sourceExts?: string[];
   reporter?: any;
   config?: string;
+  watch?: boolean;
 }
 
 /**


packages/cli/src/commands/bundle/buildBundle.ts
```diff
@@ -50,6 +50,9 @@ async function buildBundle(
     config: args.config,
   });
 
+  // Don't need file watcher for building offline bundle
+  config.watch = false;
```
thymikee: Please don't mutate the config object like this. Also, isn't this failing validation built into `metro-config`? We'll likely need to wait for a new version and include it after Metro upgrade

cdsanchez: Thanks, I'll update that.

Regarding validation - I didn't see any validation errors in my local workspace with the corresponding changes to metro.

I updated the description to make it clear that this PR depends on a release of https://github.com/facebook/metro/pull/491 before it can land.

grabbou:
Looks like they didn't implement this parameter, and provided an environmental workaround. Please look here: https://github.com/facebook/metro/commit/c90ea3abda60c2b4cfa0e995b3781c466ad45c9c#diff-7b182d5ba2fea830f66aca8de9366067R63

With that in mind, I am going to close this PR. Thank you for your time and digging deeper into it!

cdsanchez:
Hey @grabbou long story short, it was actually implemented, but in a slightly different way than I envisioned it. 

You can see the commit here:
https://github.com/facebook/metro/commit/1a38e0097eedf61fb01780a215fe446336872ef0

I haven't had time to update my PR for cli just yet though.

The gist of it is that the cli needs to be updated to use the `runBuild` API, or to pass a new `watch: false` parameter to `Server` (not part of config). In either case, I believe my metro changes need to be patched into a previous metro release, or the cli needs to update to the latest version of metro.

I probably won't get to it for another few weeks, so I'm okay with leaving this PR closed. However, I will create an issue to make sure the remaining work is tracked somewhere for posterity.

PR #864, opened 2019-11-25
Bumping metro packages to v0.57.0, which fixes a Windows CLI bug

tad3j:
Windows CLI bug solved with https://github.com/facebook/metro/commit/7c6f30b592b2fb23cee55b54f1aa4b7a522dec18 and released under v0.57.0.

Summary:
---------

Updating all metro related packages to v0.57.0

Fixes #900
Fixes #853

Test Plan:
----------

I ran test command bug seems like there are some issues:
Test Suites: 1 failed, 71 passed, 72 total
Tests:       5 failed, 2 skipped, 2 todo, 263 passed, 272 total
Snapshots:   6 failed, 20 passed, 26 total

Not sure if it's due to new metro version...

PR #893, opened 2019-12-18
Fix invalid regular expression on Windows

t0rr3sp3dr0:
Summary:
---------

Metro had a bug that crashed React Native builds on Windows facebook/metro#468 that was solved on release v0.57.0.

[React Native for Windows](https://github.com/microsoft/react-native-windows) still needs to use React Native CLI 2.x as it is only compatible with React Native 0.60.6. I'm updating Metro version on this branch to solve this problem until React Native for Windows becomes compatible with the latest version of React Native.


thymikee:
We usually need to keep Metro in CLI and RN in sync and, as far as I remember, using Metro 0.56+ with RN 0.60 won't work. 
For now you can use yarn `resolutions` to overcome this, if that works at all.

thymikee:
Duplicate of #864

Issue #889, opened 2019-12-13
react-native init does not work with react-native-tvos repo


dlowder-salesforce:
## Environment

Mac desktop
@react-native-community/cli version: 3.0.4

## Description and Reproducible Demo

When using the `react-native-cli` package, the following works correctly:

```
react-native init TestApp --version=react-native@npm:react-native-tvos@latest
```

When this command is used with `@react-native-community/cli`, I get this error:

```
✖ Downloading template
error Error: Command failed: yarn add react-native@react-native@npm:react-native-tvos@0.61.5-0
```

Modifying the command as follows:

```
react-native init TestApp --version=npm:react-native-tvos@0.61.5-0
```

successfully downloads the template, but then has this error:

```
✖ Copying template
error Error: Cannot find module '/var/folders/wx/d0rry04n0qj6yl6ltg77llf9lnm333/T/rncli-init-template-O0xLE3/node_modules/react-native@npm:react-native-tvos/template.config'
```

dlowder-salesforce:
I also tried using `--template` flag, but had similar problems.

dlowder-salesforce:
Interestingly

```
react-native init TestApp --template=react-native@https://github.com/react-native-community/react-native-tvos#tvos-v0.61.2
```

does seem to work.  Investigating.

dlowder-salesforce:
Initial hacky fix for the issue at https://github.com/dlowder-salesforce/cli/commit/ad529c01264c15f5188cfa671ee6af7153b15e68 -- will keep working on this

thymikee:
@dlowder-salesforce `react-native init` (legacy init) and `npx react-native init` (new init) are 2 different commands. TVOS needs to implement `template.config.js` file to be compatible with the new format. Check https://github.com/react-native-community/cli/blob/master/docs/init.md for details :)

dlowder-salesforce:
@thymikee thanks for pointing me to the correct doc.  In fact, the `react-native-tvos` repo is a fork of RN core, and does have the required `template.config.js`.

Using the correct command (`npx react-native init`) still leads to an error unless my fix is applied.  Reopening this and will continue to work on this issue.

dlowder-salesforce:
Thanks @thymikee for pointing me to the correct command:

```
npx react-native init TestApp --template=react-native-tvos@latest
```

No fix needed, just need to document this :)

PR #898, opened 2019-12-18
Allow the init command to use a forked RN NPM package

dlowder-salesforce:
Summary:
---------

Add an additional protocol and handler in `templateNames.ts` so that this CLI can be used to init a new project with the `react-native-tvos` package (forked from RN core).  This fixes issue #889.  After this change, the following will work:

```
npx react-native init TestApp --template=react-native@npm:react-native-tvos@latest
```

Test Plan:
----------

- Manual testing to init a new react-native-tvos project and verify that everything works
- Added a unit test for the new handler

### DIFF
--- a/packages/cli/src/commands/init/__tests__/templateName.test.ts
+++ b/packages/cli/src/commands/init/__tests__/templateName.test.ts
@@ -2,6 +2,7 @@ import {processTemplateName} from '../templateName';
 
 const RN_NPM_PACKAGE = 'react-native';
 const ABS_RN_PATH = '/path/to/react-native';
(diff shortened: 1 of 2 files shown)


thymikee:
Oh, I think we failed to communicate how new init works. The `--template` argument accepts an npm module name by default. So in your case, using `--template react-native-tvos` is perfectly fine.

image

Do you think we can close this and the corresponding issue now?

dlowder-salesforce:
Amazing -- I guess I tried every possible command except that one :)

Yes let's close the PR and the issue, and I'll document this in the `react-native-tvos` README.

PR #908, opened 2020-01-05
Use default derived data path for runIOS command

janicduplessis:
Summary:
---------

Currently when building using the react-native-cli the derived data path is in the `ios/build` folder while in xcode it is in `~/Library/Developer/Xcode` by default. This means builds started from xcode and the cli don't share the same cache. This results in longer builds and more disk space usage (xcode build cache size is no joke 😄) when using both xcode and the cli.

While it is possible to have xcode build in a different directory I think it is better to use the derived data path specified in the project instead of forcing one.

Before we relied on hardcoding the path to launch the app in the simulator or on device. Now we use `xcodebuild` with `-showBuildSettings` to print the build configuration and get the `TARGET_BUILD_DIR` env variable which is the path where the app will be.

Note this can be a breaking change if a script assumes the build directory location since now the default build directory won't be `ios/build`.

Test Plan:
----------

You can see that the app is being installed from the xcode shared derived data directory from the cli install logs.

Before:

```
info Installing "build/th3rdwave/Build/Products/Debug-iphonesimulator/th3rdwave.app"
```

After:

```
info Installing "/Users/janic/Library/Developer/Xcode/DerivedData/th3rdwave-ebjnyfazbruzovcnoejrwfwcgpqv/Build/Products/Debug-iphonesimulator/th3rdwave.app"
```

I also tested that the build cache is shared by clearing all build caches and building from xcode then building with the cli to confirm that a full rebuild is not needed.

### DIFF
--- a/packages/platform-ios/src/commands/runIOS/index.ts
+++ b/packages/platform-ios/src/commands/runIOS/index.ts
@@ -175,7 +175,13 @@ async function runOnSimulator(
     args,
   );
 
-  const appPath = getBuildPath(args.configuration, appName, false, scheme);
+  const appPath = getBuildPath(
+    xcodeProject,
+    args.configuration,
+    appName,
+    false,
+    scheme,
+  );
 
   logger.info(`Installing "${chalk.bold(appPath)}"`);
 
@@ -238,7 +244,7 @@ async function runOnDevice(
 
   const iosDeployInstallArgs = [
     '--bundle',
-    getBuildPath(args.configuration, appName, true, scheme),
+    getBuildPath(xcodeProject, args.configuration, appName, true, scheme),
     '--id',
     selectedDevice.udid,
     '--justlaunch',
@@ -279,8 +285,6 @@ function buildProject(
       scheme,
       '-destination',
       `id=${udid}`,
-      '-derivedDataPath',
-      `build/${scheme}`,
     ];
     logger.info(
       `Building ${chalk.dim(
@@ -359,7 +363,15 @@ function bootSimulator(selectedSimulator: Device) {
   }
 }
 
+function getTargetBuildDir(buildSettings: string) {
+  const targetBuildMatch = /TARGET_BUILD_DIR = (.+)$/m.exec(buildSettings);
+  return targetBuildMatch && targetBuildMatch[1]
+    ? targetBuildMatch[1].trim()
+    : null;
+}
+
 function getBuildPath(
+  xcodeProject: ProjectInfo,
   configuration: string,
   appName: string,
   isDevice: boolean,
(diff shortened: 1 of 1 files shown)


grabbou:
Oh my... @janicduplessis this is brilliant! I think this is what's been bothering me (and many others) for so long! 

I am so excited to ship it soon!

Thank you for the great PR, description and clear code changes. A pleasure to work with :)

packages/platform-ios/src/commands/runIOS/index.ts
```diff
@@ -359,7 +363,15 @@ function bootSimulator(selectedSimulator: Device) {
   }
 }
 
+function getTargetBuildDir(buildSettings: string) {
+  const targetBuildMatch = /TARGET_BUILD_DIR = (.+)$/m.exec(buildSettings);
+  return targetBuildMatch && targetBuildMatch[1]
+    ? targetBuildMatch[1].trim()
+    : null;
```
leethree: this broke the build for me because we have multiple targets in the project, and this line matched the incorrect target.

thymikee: cc @janicduplessis

janicduplessis: Looks like xcodebuild has a -target flag, maybe we can try setting that.

thymikee: Yep. Looking quickly, we'd need to introduce it as a new flag to `run-ios`

janicduplessis: Yea not sure how that worked before with the hardcoded path. @leethree How did you handle building with multiple targets before this change? Which target gets selected now and which got selected before this change?

leethree: I think Xcode just put the resulted app in `${appName}.app` and hardcoded path worked fine before.

We have targets for the iOS app itself, and some watchkit and widget stuff in our build scheme, which means there are multiple build settings in whatever `xcodebuild -showBuildSetting` outputs.

If I understand correctly, this script will match the first `TARGET_BUILD_DIR` in the output, which is not necessarily the app target we want.

janicduplessis: Okay, I'm able to get the output for a specific target using `xcodebuild -project projectName.xcodeproj -sdk iphonesimulator -configuration Debug -showBuildSettings -target targetName`

I wonder if it would be safe to assume the target name is the same name as scheme, seems like a scheme with the same name as the target is created when you create a new target. We could use that as the default and allow customizing the target with a new option maybe in case the name is different.

janicduplessis: @leethree In your case does your scheme and target have the same name?

leethree: no we don't have the same name for scheme and target. I guess a more reliable way is to match all `TARGET_BUILD_DIR` and find the one that actually contains `${appName}.app`.

PR #911, opened 2020-01-08
Upgrade morgan to 1.9.1

safaiyeh:
Summary:
---------

morgan@1.9.0 triggers security warnings (https://nvd.nist.gov/vuln/detail/CVE-2019-5413)

First I tried using package resolutions to override the version; however, the existence of 1.9.0 in the `yarn.lock` file under this package continues to trigger the warning.

Test Plan:
----------

Ran CLI locally, worked as expected.

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -52,7 +52,7 @@
     "metro-react-native-babel-transformer": "^0.56.0",
     "minimist": "^1.2.0",
     "mkdirp": "^0.5.1",
-    "morgan": "^1.9.0",
+    "morgan": "^1.9.1",
     "node-notifier": "^5.2.1",
     "open": "^6.2.0",
     "ora": "^3.4.0",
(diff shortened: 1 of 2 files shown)


thymikee:
It's in a semver range, there's no point doing that.

safaiyeh:
@thymikee I used yarn resolutions to use 1.9.1; however, the existence of 1.9.0 in the yarn file still triggers the security alert.

If cli does not use 1.9.1 our current work around is to manually edit our yarn.lock file which is not optimal.

thymikee:
After this is merged, and new version is released, you'll still need to bump the version. I'd say invalidate lock file and move on, I don't see it as "not optimal".

I appreciate your care about this though. It's just unnecessary

safaiyeh:
Makes sense. I’ll invalidate our lock file.

Thanks for being responsive! @thymikee

PR #913, opened 2020-01-13
feat(chore): Recommend matching commands

### LINKED ISSUE #905: Suggest matching commands if the user makes a typo
## Describe the Feature

Suggest matching commands if the user mistypes. For instance, `react-native-cli ini` should show up the help message along with a friendly suggestion such as `Did you mean init?`

## Possible Implementations

We can make use of the [didyoumean.js](https://github.com/dcporter/didyoumean.js/) matching engine.

jamesgeorge007:
Summary:
---------
Suggest matching commands if the user makes a typo.
closes #905 

This is how it goes:
----------
Screenshot from 2020-01-13 20-33-23

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -37,6 +37,7 @@
     "connect": "^3.6.5",
     "cosmiconfig": "^5.1.0",
     "deepmerge": "^3.2.0",
+    "didyoumean": "^1.2.1",
     "envinfo": "^7.1.0",
     "errorhandler": "^1.5.0",
     "execa": "^1.0.0",
--- a/packages/cli/src/index.ts
+++ b/packages/cli/src/index.ts
@@ -103,7 +103,7 @@ function printUnknownCommand(cmdName: string) {
 function suggestCommands(cmd: string) {
   const suggestion = didYouMean(cmd, commander.commands.map(cmd => cmd._name));
     if (suggestion) {
-  	  logger.warn(`\n\nDid you mean ${suggestion}?`)
+  	  logger.warn(`Did you mean ${suggestion}?`)
     }
 }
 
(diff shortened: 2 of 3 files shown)


jamesgeorge007:
@thymikee it seems `leven` just returns the actual difference between 2 given input strings. What required in the respective case is a check against some group of values which isn't possible. I would say, `didyoumean.js` would be the candidate here. If you're concerned about it's popularity, `Vue-CLI` makes use `didyoumean` for the purpose.

thymikee:
> it seems leven just returns the actual difference between 2 given input strings

Yup, that's all you need, the Levenshtein distance. I'm familiar with `leven` because we use it for Jest CLI and it serves use well. But if Vue is happy with `didyoumean` then sure, why not, it definitely is more specialized in what we need :)

packages/cli/package.json
```diff
@@ -37,6 +37,7 @@
     "connect": "^3.6.5",
     "cosmiconfig": "^5.1.0",
     "deepmerge": "^3.2.0",
+    "didyoumean": "^1.2.1",
```
thymikee: Did you run `yarn` so that `yarn.lock` is updated?

jamesgeorge007: I use `npm` and always miss it out!

thymikee:
CI is still failing, please make sure the linter passes.

jamesgeorge007:
\cc @thymikee

packages/cli/src/index.ts
```diff
@@ -1,6 +1,7 @@
 import chalk from 'chalk';
 import childProcess from 'child_process';
 import commander from 'commander';
+import didYouMean from 'didyoumean';
```
Taym95: eslint giving this error `Unable to resolve path to module 'didyoumean'  import/no-unresolved`!

grabbou: Linter passing upstream, so going to ignore this and ship it. Will be fixed later if needed, thank you!

grabbou:
Great feature, thank you for working on it and @thymikee for spending time on reviewing it! Let me merge it!

PR #916, opened 2020-01-16
Fix log-ios on Xcode 11.2

laurent22:
Summary:
---------

The way Xcode 11.2 reports device availability has changed. This change adds support for this while maintaining backward compatibility with older Xcode versions.

Also changed how it detects the absence of a device. Previously it would do strict equality check with `undefined` but the problem is that the device was `null`. I think it's fine to simply do `!device` as that will handle both cases.

### DIFF
--- a/packages/platform-ios/src/commands/logIOS/index.ts
+++ b/packages/platform-ios/src/commands/logIOS/index.ts
@@ -15,7 +15,7 @@ import {Device} from '../../types';
 function findAvailableDevice(devices: {[index: string]: Array<Device>}) {
   for (const key of Object.keys(devices)) {
     for (const device of devices[key]) {
-      if (device.availability === '(available)' && device.state === 'Booted') {
+      if ((device.availability === '(available)' || device.isAvailable) && device.state === 'Booted') {
         return device;
       }
     }
@@ -38,7 +38,7 @@ async function logIOS() {
   };
 
   const device = findAvailableDevice(devices);
-  if (device === null) {
+  if (!device) {
     logger.error('No active iOS device found');
     return;
   }


packages/platform-ios/src/commands/logIOS/index.ts
```diff
@@ -38,7 +38,7 @@ async function logIOS() {
   };
 
   const device = findAvailableDevice(devices);
-  if (device === null) {
+  if (!device) {
```
thymikee: This can be either `null` or an object (see line 18 where you assert on `device` being able to access keys like `availability`). Can we please leave the check as it was?

packages/platform-ios/src/commands/logIOS/index.ts
```diff
@@ -15,7 +15,7 @@ import {Device} from '../../types';
 function findAvailableDevice(devices: {[index: string]: Array<Device>}) {
   for (const key of Object.keys(devices)) {
     for (const device of devices[key]) {
-      if (device.availability === '(available)' && device.state === 'Booted') {
+      if ((device.availability === '(available)' || device.isAvailable) && device.state === 'Booted') {
```
thymikee: We have similar logic in the `findMatchingSimulator.ts` file. I just noticed that we even call  the same command (`simctl list devices --json`). Can you think of a way to reuse availability check? Can be done in a followup, no worries :)

grabbou:
@laurent22, do you plan on working on this issue anytime soon? If not, that's fine! In that case, someone from our team will take it over and ship within next React Native release.

Thank you!

laurent22:
Hi, sorry for the lack of feedback. Unfortunately I don't think I'll have time to look into this PR any time soon, so feel free to take it over.

grabbou:
Sounds good. I am going to close this pull request then and open another one where I continue the work.

PR #919, opened 2020-01-20
Don't use deprecated createBlacklist metro api

janicduplessis:
Summary:
---------

Some legacy apis were removed in the latest metro release. The only one that seem to be affecting here is createBlacklist. we import it from the metro-config package instead. See https://github.com/facebook/metro/commit/871cafbd567cdb3bf231270cdbb6c28f6535c2fd#diff-82e93cde29a5080883ceae93070af8feL13

Test Plan:
----------

Tested that running the metro server works with metro 0.58.0

### DIFF
--- a/packages/cli/src/tools/loadMetroConfig.ts
+++ b/packages/cli/src/tools/loadMetroConfig.ts
@@ -3,9 +3,9 @@
  */
 import path from 'path';
 // @ts-ignore - no typed definition for the package
-import {createBlacklist} from 'metro';
-// @ts-ignore - no typed definition for the package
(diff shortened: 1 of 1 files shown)


thymikee:
We have a PR open currently that bumps Metro to 0.57, see https://github.com/react-native-community/cli/pull/864. When updating it I remember I verified the default blacklist and there was no need for us to actually use it. Can you please have a look?

janicduplessis:
@thymikee This is only required for metro 0.58.0 that was released a few days ago. I think there is no harm in updating now since the blacklist file exists even in older metro versions.

PR #928, opened 2020-01-29
Ditch `npx` and call CLI scripts directly

### LINKED ISSUE #886: Is npx needed?
The server used to build our app doesn't have npx installed and we would prefer not installing it. Is there any benefit to using npx instead of just calling the react-native binary directly? The one in node_modules/.bin/react-native? Seems to me it accomplishes the same thing without the need for npx.

grabbou:
Summary:
---------

Remove `npx` from auto-linking to reduce regressions and issues around `npx`.

As a next step, I am going to resolve some regressions aroudn Gradle / root that people were submitting.

Fixes #886


packages/tools/src/printCLIPath.ts
```diff
@@ -0,0 +1,13 @@
+import path from 'path';
+import pkg from '@react-native-community/cli/package.json';
```
thymikee: This assumes `cli-tools` has a dependency on `cli`, which is not true. Moving `printCLIPath` to `cli` would solve this problem

thymikee: I forgot to mention this creates a circular dep, because cli relies on cli-tools

grabbou: We can move it to the CLI, no problem. It will be actually even a bit more simple.

grabbou: Implemented alternative solution that doesn't rely on package.json.

packages/platform-android/native_modules.gradle
```diff
@@ -160,7 +160,7 @@ class ReactNativeModules {
       }
       output = readBuffer.toString()
       if (!output) {
-        this.logger.error("${LOG_PREFIX}Unexpected empty result of running '${command}' command from '${directory}' directory.")
```
grabbou: This is `null` right now, so I have removed the extra logging for better error messages without "null" directory in the output.

packages/platform-android/native_modules.gradle
```diff
@@ -183,15 +183,20 @@ class ReactNativeModules {
...
     /**
-     * Running npx from the directory of the JS app which holds this script in its node_modules.
-     * We do so, because Gradle may be ran with a different directory as CWD, that's outside of JS project
-     * (e.g. when running with -p flag), in which case npx wouldn't resolve correct `react-native` binary.
+     * Resolve the CLI location from Gradle file
+     *
+     * @todo: Sometimes Gradle can be called outside of the JavaScript hierarchy (-p flag) which
+     * will fail to resolve the script and the dependencies. We should resolve this soon.
+     *
+     * @todo: `fastlane` has been reported to not work too.
```
grabbou: I have managed to confirm one additional place where it fails and I am hoping this will let us better debug this.

packages/platform-android/native_modules.gradle
```diff
@@ -183,21 +183,27 @@ class ReactNativeModules {
...
      */
-    def dir = new File(this.jsAppDir)
-    def reactNativeConfigOutput = this.getCommandOutput(command, /* dir */) // Temporary disable changing dir, as it introduces regressions
+    def cliResolveScript = "console.log(require('@react-native-community/cli').bin);"
+    def cliPath = this.getCommandOutput("node -e ${cliResolveScript}")
```
thymikee: Can you make sure `node -e` works on Windows?

grabbou: Well, `bin` comes from the Node.js `require.resolve` so it's coming from the API layer. The only question is whether `node` can be accessed on Windows this way or not

grabbou: I have confirmed on a Virtual Machine that running `node -e script` works in both PowerShell and CMD. 

I think we're good to go and ship it as "next"

thymikee: Why not shippining it to current 3.x?

grabbou:
Note: e2e tests are passing locally. I believe it's a CircleCI issue with the dependencies (fails on `npm install`) but I seem to not have permissions to reset the cache.

Going to continue as is and we will see how's master post-merge.

PR #935, opened 2020-01-30
chore: remove REACT_NATIVE_APP_ROOT, use --watchFolders

grabbou:
Summary:
---------

I don't think this is needed. Users can pass `--watchFolders` flag with the CLI (if they really need). They can also create their own `metro.config.js` where they define that value.

In my opinion, this is really confusing. I haven't found any places where it's referred to.

In other words, setting this:
```
REACT_NATIVE_APP_ROOT=path/to/cli node path/to/cli/packages/cli/build/index.js start
```
is equivallent to
```
node path/to/cli/packages/cli/build/index.js start --watchFolders path/to/cli 
```

(whether this makes sense or not is another topic)


grabbou:
CC: @lithin I have a hard time understanding the use-case of that - can you please help me by bringing the issue you've run into and when this was a fix? That would help!

My understanding is that the location of the CLI doesn't matter. The `projectRoot` of the project is always the CWD. 

And according to the Metro documentation:
> Specify any additional (to projectRoot) watch folders, this is used to know which files to watch

Source: https://facebook.github.io/metro/docs/en/configuration#watchfolders

grabbou:
Okay, looks like when this environmental variable is present, we will scan for symlinks by calling `findSymlinkedModules` - this performs a search inside `node_modules`.

Now, I haven't seen this referred to anywhere and I am not really sure what's the intent of this functionality. 

Is there anyone here that has more knowledge of this particular feature?

CONTRIBUTING.md
```diff
@@ -89,14 +89,6 @@ react-native init --version ${RN_VERSION}
 npm config set registry https://registry.npmjs.org/
 ```
 
-## Running `start` command
-
-In order for symlinks to work correctly when running `start` locally, set REACT_NATIVE_APP_ROOT as the root folder of your cli project:
-
-```
-REACT_NATIVE_APP_ROOT=path/to/cli node path/to/cli/packages/cli/build/index.js start
```
thymikee: I wonder how you test the `start` or any command that calls Metro now? This was a hack for Metro to "see" files that are symlinked alongside the cwd you're in. In Metro pre <0.57 you could also see these paths displayed in the UI (normally it's your CWD, but with this env passed it's CWD + this specified dir and its subdirs)

grabbou: Hm, isn't this what `projectRoots` and `watchFolders` stands for? 

Also, following up with this example:
```
REACT_NATIVE_APP_ROOT=path/to/cli node path/to/cli/packages/cli/build/index.js start
```

why Metro would care about symlinks that are in `path/to/cli`, if it's working in `cwd` where your React Native application business code is? Metro is not calling into `path/to/cli`.

grabbou: Also, the flag name is `REACT_NATIVE_APP_ROOT` which suggests you're setting application root. In the example, it's the CLI root, not the application root. It's also confusing since you can set a project root in a different way.

grabbou: FWIW, I have never used this way to debug `react-native start` command locally and it worked flawlessly.

However, as I stated in my comment [here](https://github.com/react-native-community/cli/pull/935#issuecomment-580274130), I am trying to understand the original intent and use case for that code. Looks like we brought it over from React Native.

Was it prior to `--projectRoots`? Or `--watchFolders`? We scan for symlinks when this environmental variable is present and that functionality seems useful. However, in 99.9% cases we do not do it.

grabbou: There you go https://github.com/facebook/react-native/commit/a1e105eea51ec7b8962bd43eec8be8aa15d94c5b

thymikee: > why Metro would care about symlinks that are in path/to/cli, if it's working in cwd where your React Native application business code is? Metro is not calling into path/to/cli.

It does for me locally, when I link local build of CLI into a RN project with `yarn link <cli_packages>` – Metro tries to resolve some files, like some babel plugin I don't remember now, from my local CLI dir.

So with this PR, you're removing that possibility, and essentially `resolveSymlinksForRoots` and `findSymlinkedModules` become dead code (should be removed if we want to go that path). 

I just don't want to end up in a position where we can't test local changes to CLI commands invoking Metro on a real project.

packages/cli/src/tools/loadMetroConfig.ts
```diff
@@ -15,11 +15,6 @@ function resolveSymlinksForRoots(roots: string[]): string[] {
   );
 }
 
-function getWatchFolders(): string[] {
-  const root = process.env.REACT_NATIVE_APP_ROOT;
-  return root ? resolveSymlinksForRoots([path.resolve(root)]) : [];
-}
```
grabbou: @thymikee also note that back in the day, we used to call `resovleSymlinksForRoots` in all cases - https://github.com/facebook/react-native/blob/0.55-stable/local-cli/util/Config.js#L54

Looks like this has changed and I don't know when. But it sounds like a regression.

grabbou: Okay, this has been done in this commit https://github.com/facebook/react-native/commit/893c0274d6c071c5fa99dee097e189441af74148 as a part of 0.57 release.

I am really confused right now because this change doesn't make any sense to me, unless Metro performs this functionally itself.

CC: @rafeca, I am hoping that you will be able to shed some light into this.

grabbou: Following up on this even further, this functionality has been introduced by @skevy in 2016 (lol) via https://github.com/facebook/react-native/commit/39f83c4a12b8350e331c3a6c575a2cab64950d7a#diff-d939fe299a50129184bf6b4208f6d497

Apologies for pinging both of you on such historical piece of code, but... does this mean symlinks used to work with React Native and could've been accidentally removed via https://github.com/facebook/react-native/commit/893c0274d6c071c5fa99dee097e189441af74148 commit?

I always believed that symlinks never worked and seeing symlinks related code always made me confused.

CONTRIBUTING.md
grabbou: > It does for me locally, when I link local build of CLI into a RN project with yarn link <cli_packages> – Metro tries to resolve some files, like some babel plugin I don't remember now, from my local CLI dir.

Can you try just setting `--watchFolders` flag when running `start`? If you can reproduce it of course. 

My guess is this: it's project roots / watch folders that were making this work, not `findSymlinkedModules`. I don't feel the urge of doing anything with this code, but I just want to understand the intents behind it, because it's really confusing.

> So with this PR, you're removing that possibility, and essentially resolveSymlinksForRoots and findSymlinkedModules become dead code (should be removed if we want to go that path).

Keep in mind, as per my different thread, that we used to call this no matter what. I have started off a different thread to understand the purpose of that file.

thymikee: Will try to find time to test it as well. I just remember I spent quite a bit of time to solve testing local changes to CLI `start` in a test project. I would like you to actually test it as well and feel the experience of changing anything in Metro-related code of CLI and `run-ios`/`run-android`/`start` a sample project seeing if Metro version local to this project picks it up or crashes because it can't find some module

packages/cli/src/tools/loadMetroConfig.ts
rafeca: Hey @grabbou ! 👋

It's hard to remember the intention of my commit there LOL (don't know if there's more context in D9444982 - I can't access diffs anymore), but it's quite probably that https://github.com/facebook/react-native/commit/893c0274d6c071c5fa99dee097e189441af74148 removed the `resolveSymlinksForRoots` call by mistake 😅

grabbou: CC: @rickhanlonii would you mind helping out and checking some context on the D9444982? Thanks @rafeca for answering!

rickhanlonii: The only thing in the diff is this comment you left @rafeca:

> From what I see, we could actually just not set watchFolders, since right now it's only set to the value of REACT_NATIVE_APP_ROOT if it exists (which is an env variable that I cannot find in any RN docs).

I think we could safely revert the change as long as we confirm that https://github.com/facebook/react-native/issues/20712 doesn't regress 👍

PR #936, opened 2020-01-30
chore: fix e2e tests

grabbou:
Summary:
---------

e2e tests are broken at the `npm install` step for a while now. This is an attempt to fix them.


grabbou:
Closing as the tests are passing locally. Don't know what's the problem.

thymikee:
Yea it bugs me why it fails. We should see if there's a way for getting better debugging info when an e2e test fails

Issue #922, opened 2020-01-25
Upgrade metro to v0.58.0 (fix inherited mem vulnerability)


sethahrenbach:
Current dependency path contains os-locale v2.1.0, which contains mem v1.1.0

Dependency path:
│ Path  @react-native-community/cli > metro > yargs > os-locale > mem  

Results in:

https://npmjs.com/advisories/1084 

> Overview

> Versions of mem prior to 4.0.0 are vulnerable to Denial of Service (DoS). The package fails to remove old values from the cache even after a value passes its maxAge property. This may allow attackers to exhaust the system's memory if they are able to abuse the application logging.

Solution:

metro v0.58.0 upgrades yargs from ^9.0.0 to ^14.2.0:
[
Diff metro v0.57.0 | metro v0.58.0](https://github.com/facebook/metro/commit/56e2c5713f5d5c75e6ddb78deeec904359982cf0#diff-223e98b09ed76095facb1ca8c90afd87L69)

yargs 14.2.0 removes os-locale from package.json:

[Diff yargs v9.0.1 | yargs v14.2.0](https://github.com/yargs/yargs/compare/v9.0.1...v14.2.0#diff-b9cfc7f2cdf78a7f4b91a753d10865a2):

itsgauravjain22:
Please resolve this. Following is output of ***npm audit***:  

```                         
                       === npm audit security report ===


                                 Manual Review
             Some vulnerabilities require your attention to resolve

          Visit https://go.npm.me/audit-guide for additional guidance


  Low             Denial of Service

  Package         mem

  Patched in      >=4.0.0

  Dependency of   react-native

  Path            react-native > @react-native-community/cli > metro >
                  metro-inspector-proxy > yargs > os-locale > mem

  More info       https://npmjs.com/advisories/1084


  Low             Denial of Service

  Package         mem

  Patched in      >=4.0.0

  Dependency of   react-native

  Path            react-native > @react-native-community/cli > metro-config >
                  metro > metro-inspector-proxy > yargs > os-locale > mem

  More info       https://npmjs.com/advisories/1084


  Low             Denial of Service

  Package         mem

  Patched in      >=4.0.0

  Dependency of   react-native

  Path            react-native > @react-native-community/cli > metro > yargs >
                  os-locale > mem

  More info       https://npmjs.com/advisories/1084


  Low             Denial of Service

  Package         mem

  Patched in      >=4.0.0

  Dependency of   react-native

[...]

PR #937, opened 2020-01-30
Upgrade metro to 0.58.0

rickhanlonii:
Summary:
---------

Upgrades to Metro 0.58.0 which includes code frames for LogBox.

Also closes https://github.com/react-native-community/cli/issues/922

Test Plan:
----------

Tested in RNTester and confirmed that codeframes work in LogBox:

Screen Shot 2020-01-30 at 4 06 39 PM


thymikee:
@grabbou can you test if this works with RN 0.61 as well? It would be amazing if it did, because we could then use cli@3.x for both 0.61 and 0.62

rickhanlonii:
Yeah it seems to work in 0.61 as well

grabbou:
Awesome! Let's ship it to 0.61 and 0.62 users via 3.x.

PR #939, opened 2020-02-02
Fix bundling assets with a path outside of the project root

janicduplessis:
Summary:
---------

When an asset is outside of the metro project root, it can lead to relative paths like `/assets/../../node_modules/coolpackage/image.png` as the `httpServerLocation`. This can happen for example when using yarn workspaces with hoisted node_modules.

This causes issues when bundling on iOS since we use this path in the filesystem. To avoid this we replace `../` with `_` to preserve the uniqueness of the path while avoiding these kind of problematic relative paths. The same logic is used when bundling assets in the rn-cli.

RN part of this PR: https://github.com/facebook/react-native/pull/27932

Test Plan:
----------

Tested that an asset in a hoisted node_modules package doesn't show up before this patch and does after in a release build.

### DIFF
--- a/packages/cli/src/commands/bundle/getAssetDestPathIOS.ts
+++ b/packages/cli/src/commands/bundle/getAssetDestPathIOS.ts
@@ -12,7 +12,13 @@ import {PackagerAsset} from './assetPathUtils';
 function getAssetDestPathIOS(asset: PackagerAsset, scale: number): string {
   const suffix = scale === 1 ? '' : `@${scale}x`;
   const fileName = `${asset.name + suffix}.${asset.type}`;
-  return path.join(asset.httpServerLocation.substr(1), fileName);
+  return path.join(
+    // Assets can have relative paths outside of the project root.
+    // Replace `../` with `_` to make sure they don't end up outside of
+    // the expected assets directory.
+    asset.httpServerLocation.substr(1).replace(/\.\.\//g, '_'),
+    fileName,
+  );
 }
 
 export default getAssetDestPathIOS;
--- a/packages/cli/src/commands/bundle/__tests__/getAssetDestPathAndroid-test.ts
+++ b/packages/cli/src/commands/bundle/__tests__/getAssetDestPathAndroid-test.ts
@@ -70,4 +70,16 @@ describe('getAssetDestPathAndroid', () => {
       path.normalize('raw/app_test_video.mp4'),
     );
   });
+
+  it('should handle assets with a relative path outside of root', () => {
+    const asset = {
+      name: 'icon',
+      type: 'png',
+      httpServerLocation: '/assets/../../test',
+    };
+
+    expect(getAssetDestPathAndroid(asset, 1)).toBe(
+      path.normalize('drawable-mdpi/__test_icon.png'),
+    );
+  });
 });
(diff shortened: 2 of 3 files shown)


grabbou:
Sounds good, thanks! I'll wait with this one before I release 0.62 of the CLI off master. This is going to go for next React Native release to sync with that diff.

janicduplessis:
RN part of the PR was merged on Feb 3rd

thymikee:
Is this backwards compatible? I can see it wasn't picked for 0.62 release (should it be?), so we could merge it to `next` branch once it's sorted out, or if it doesn't break anybody include in 4.x and have it available for people using 0.62

janicduplessis:
Yes, using paths outside the root is currently broken and it doesn't change the behaviour of paths inside the project root.

packages/cli/src/commands/bundle/getAssetDestPathIOS.ts
```diff
@@ -12,7 +12,13 @@ import {PackagerAsset} from './assetPathUtils';
 function getAssetDestPathIOS(asset: PackagerAsset, scale: number): string {
   const suffix = scale === 1 ? '' : `@${scale}x`;
   const fileName = `${asset.name + suffix}.${asset.type}`;
-  return path.join(asset.httpServerLocation.substr(1), fileName);
+  return path.join(
+    // Assets can have relative paths outside of the project root.
+    // Replace `../` with `_` to make sure they don't end up outside of
```
thymikee: Is there a possibility where this could clash with existing assets that happen to be `_`-prefixed?

janicduplessis: There is, however if I remember correctly it would already cause issues on Android which has a similar replace function for paths.

janicduplessis: Yea, see the test I added for android without any code change to the logic.

https://github.com/react-native-community/cli/pull/939/files#diff-e43a39ec280faab0d111d719851dcc60R82

janicduplessis: Actually it is probably less likely to happen on Android since paths are flattened and all `/` are replaced with `_`. I think it is fine and if someone hits an issue we can figure out a fix. I don't think a lot of people are using this anyway since it's been broken forever.

thymikee: The part that concerns is that the fix would require coordination between RN and CLI releases. But I agree this should almost never happen (maybe we can warn somehow that "assets name clash, please rename"?)

janicduplessis:
Done, tests passed!

packages/cli/src/commands/bundle/getAssetDestPathIOS.ts
janicduplessis: Yea I avoided changing the logic too much. I think ideally iOS should flatten the directory structure like android does. This would make it less likely (almost impossible) to happen. I think this would be a better long term solution but as you said would require coordination with RN.

janicduplessis: Since this is already kind of an edge case I'm more leaning towards wait and see if it's an issue before investing time in fixing it.

PR #942, opened 2020-02-04
docs: add example of dependency-level autolinking customisation

paramaggarwal:
Summary:
---------

There is currently no documentation of how can a library author customise the auto-linking behaviour. This adds an example for the same.

As per schema mentioned here:
https://github.com/react-native-community/cli/blob/e496c2a89c06ad5969608e10a749da17edee6375/packages/cli/src/tools/config/schema.ts#L41-L47

Test Plan:
----------

This was tested by looking and validating the schema reference.


thymikee:
Thank you!

PR #943, opened 2020-02-05
chore: update release cycle

grabbou:
Updated release cycle.

[3.x](https://github.com/react-native-community/cli/tree/3.x) will stay for React Native 0.61 and will not contain: 
a) Metro bump to 0.57 and 0.58 https://github.com/react-native-community/cli/commit/e496c2a89c06ad5969608e10a749da17edee6375 https://github.com/react-native-community/cli/commit/4f6b28e3e4d8276a0aac4e5a4bd5b08b953da4d5
b) fixed `run-android` with flavor which deprecates `appId` and `applicationId` parameters. This is minor change, but we're sticking to semver which requires us to bump major https://github.com/react-native-community/cli/commit/ada3e497325b5b3d0441f620d6ad276e4d773bd5 https://github.com/react-native-community/cli/commit/c199ecfdd1dcf8602cb10c3cbc55bddc5fc62435

Rather than that, it contains everything that has been released to `master` since then (including `watch` mode).

4.x will stay for React Native 0.62, includes everything that is currently on `master`

Our `next` branch will roll over to 0.63

Thoughts? I am not sure if we should ship Metro 0.58 to 0.61 users too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,13 +26,13 @@ _Note: CLI has been extracted from core `react-native` as a part of "[Lean Core]
 
 Our release cycle is independent of `react-native`. We follow semver and here is the compatibility table:
 
-| `@react-native-community/cli`                                    | `react-native`   |
-| ---------------------------------------------------------------- | ---------------- |
-| ^5.0.0 (`next`)                                                  | master           |
-| ^4.0.0                                                           | ^0.62.0          |
-| ^3.0.0                                                           | ^0.61.0          |
-| [^2.0.0](https://github.com/react-native-community/cli/tree/2.x) | ^0.60.0          |
-| [^1.0.0](https://github.com/react-native-community/cli/tree/1.x) | ^0.59.0          |
+| `@react-native-community/cli`                                               | `react-native`   |
+| --------------------------------------------------------------------------- | ---------------- |
+| [^5.0.0 (`next`)](https://github.com/react-native-community/cli/tree/next)  | master           |
+| [^4.0.0 (`master`)](https://github.com/react-native-community/cli/)         | ^0.62.0          |
+| [^3.0.0](https://github.com/react-native-community/cli/tree/2.x)            | ^0.61.0          |
+| [^2.0.0](https://github.com/react-native-community/cli/tree/2.x)            | ^0.60.0          |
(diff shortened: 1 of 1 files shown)


grabbou:
Note that right now, there's no way to say whether Metro is compatible with React Native. This makes me really nervous to release backward-compatible Metro release updates because we are only testing "latest" while releasing React Native.

rickhanlonii:
@grabbou good point, to be safe what do you think about:

- 3.x -> RN 0.61 -> Metro 0.57
- 4.x -> RN 0.62 -> Metro 0.58

thymikee:
IIRC there's no clear benefit of Metro 0.57 for RN 0.61, right (other than the UI change)? The critical Windows regex bugfix was back-ported to Metro 0.56.

grabbou:
Actually, I would think about moving Metro to React Native so that the updates can happen there. Most of the time, there's nothing to do from the CLI perspective. We just bump versions. 

This has the benefit of always testing React Native after bumping Metro and being able to see the results. Bumping Metro in the CLI will ALWAYS make the tests green, because there's nothing that can fail on the CLI side.

README.md
```diff
@@ -28,8 +28,9 @@ Our release cycle is independent of `react-native`. We follow semver and here is
 
 | `@react-native-community/cli`                                    | `react-native`   |
 | ---------------------------------------------------------------- | ---------------- |
-| ^4.0.0 (`next`)                                                  | ^0.63.0          |
-| ^3.0.0                                                           | ^0.61.0, ^0.62.0 |
+| ^5.0.0 (`next`)                                                  | master           |
```
thymikee: this should link to `next` branch

grabbou: Updated and added branches to few other releases too to make it more explicit.

grabbou:
@rickhanlonii let me know if you're okay with `3.x` being on Metro 0.56 as per @thymikee message. I am going to release `4.x` in the meantime.

rickhanlonii:
Yeah that makes sense to me, and so does moving the metro dep to RN 👍

grabbou:
Here's list of commits that land in 4.x https://github.com/react-native-community/cli/compare/v3.0.5-alpha.1...master but not in 3.x.

PR #951, opened 2020-02-11
fix: Fix install process for development versions of React Native

empyrical:
Summary:
---------

This small fix allows for development versions of React Native to be installed from the React Native CLI.

In an older version of the RNCLI, you used to be able to provide a link to a Git URL to the `--version` argument of `init`. This feature was unintentionally supported, because NPM and Yarn support installing packages with the syntax: `my-package@https://github.com/my/package.git`

This mostly works in the current revision, but it ran into a problem during the install process, because the version of development versions of React Native have the version `1000.0.0` in their `package.json`, and Yarn would error out because that is not a real version of React Native.

To fix this, a new build step is added to `createFromTemplate` that updates the "version" of `react-native` in the template's `package.json` to the URL of the development version of React Native, if provided.

Test Plan:
----------

New tests for `init --version` were added to `__e2e__/init.test.ts`.

The following syntax options for `--version` work properly:

```sh
npx react-native init MyApp --version https://github.com/facebook/react-native\#7bd1abec35f0aff0dddf0c1a68f79da29e00acdb
npx react-native init MyApp --version github:microsoft/react-native
```

### DIFF
--- a/packages/cli/src/commands/init/init.ts
+++ b/packages/cli/src/commands/init/init.ts
@@ -17,7 +17,10 @@ import {
   copyTemplate,
   executePostInitScript,
 } from './template';
-import {changePlaceholderInTemplate} from './editTemplate';
+import {
+  changePlaceholderInTemplate,
+  changeReactNativeVersionInTemplate,
+} from './editTemplate';
 import * as PackageManager from '../../tools/packageManager';
 import installPods from '../../tools/installPods';
 import {processTemplateName} from './templateName';
@@ -25,6 +28,7 @@ import banner from './banner';
 import {getLoader} from '../../tools/loader';
 
 const DEFAULT_VERSION = 'latest';
+const DEVELOPMENT_VERSION = /^(file|http|https|git|git\+ssh|git\+file|github|bitbucket)\:/;
 
 type Options = {
   template?: string;
@@ -40,6 +44,7 @@ interface TemplateOptions {
   npm?: boolean;
   directory: string;
   projectTitle?: string;
+  developmentVersion?: string;
 }
 
 function doesDirectoryExist(dir: string) {
@@ -99,6 +104,7 @@ async function createFromTemplate({
   npm,
   directory,
   projectTitle,
+  developmentVersion,
 }: TemplateOptions) {
   logger.debug('Initializing new project');
   logger.log(banner);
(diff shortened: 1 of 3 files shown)


empyrical:
Test failure appears to be the same thing affecting `test('init --template')` on `master`, not sure why it can't see "Run instructions" in the output 😕

packages/cli/src/commands/init/init.ts
```diff
@@ -17,14 +17,18 @@ import {
...
+  changeReactNativeVersionInTemplate,
+} from './editTemplate';
 import * as PackageManager from '../../tools/packageManager';
 import installPods from '../../tools/installPods';
 import {processTemplateName} from './templateName';
 import banner from './banner';
 import {getLoader} from '../../tools/loader';
 
 const DEFAULT_VERSION = 'latest';
+const DEVELOPMENT_VERSION = /^(file|http|https|git|git\+ssh|git\+file|github|bitbucket)\:/;
```
thymikee: The `--template` arg of CLI now supports whatever Yarn/npm supports and it's perfectly valid to install template from a local directory or a repository. It's an oversimplification to assume that `DEVELOPMENT_VERSION` is always gonna be `react-native`/RN fork repo. I think we'd need to introduce another flag for this specific use case, if we want to support it automatically.

Have you seen https://github.com/react-native-community/cli/blob/master/CONTRIBUTING.md#running-cli-with-react-native-from-the-source?

empyrical:
I have seen that; this is sorta automating the steps of that instruction, including the step of updating `react-native/template/package.json`. I can continue to use that method as well if this change to `--version` is undesired, but I thought I'd see if something like this could be implemented. Would make hacking away at out of tree platforms a bit nicer. Alternatively, like you suggested, a different command line flag option for a version of RN installed from a URL (e.g. a tarball)/Repo could be used.

packages/cli/src/commands/init/init.ts
empyrical: That regex is only being used for the value of `--version` to differentiate between whether or not you're passing a version or a tag (like `latest`). If you just use `--template` and don't provide a value to `--version` at all, this regex won't be used

thymikee:
I'm definitely open to make it easier and I think making it a separate flag would cause less confusion, so you could do:

```
npx react-native init MyApp --template-dev github:microsoft/react-native#hash
```

I get that putting it into `--version` is likely easier to introduce, but I'm not convinced we should teach users about it. Extra argument is also easier to discover, and putting something like `dev` inside indicates the intention better.

I'd like to get @Esemesek view on this as well.

empyrical:
Going to close this and remake this pull request as a separate command line flag. Thanks for the feedback! 💯

PR #953, opened 2020-02-11
feat: introduce /events endpoint for devtools like Flipper

mweststrate:
Summary:
---------

This PR introduce a new endpoint, `/events` to be used by developer tooling, that services two goals:
1. Stream the applications JavaScript logs and metro events
2. Support receiving commands and forward them to Metro. The setup is generic, but the two use cases for now are "reload" and "open dev menu"

This change reflects changes we made to the CLI at FB as well. This enables Flipper to have better first class support for React Native. Although some Flipper plugins could already be used on RN through the generic ADB support, this change makes it possible to see the Metro server as a real device and interact with it. 

Preview:

Screen Shot 2020-02-10 at 17 51 05

https://youtu.be/HIau7zjJ3gg

Note that currently errors are not send by RN to the Metro logging, so in the current RN version those won't be visible, but that will change in a next minor or patch release in RN.

Test Plan:
----------

1. Build the repo
2. Take a test project and update the CLI according to the contributing guidelines. 
3. Start Flipper (the supporter Flipper version didn't have an OSS release yet so need to clone from source and `yarn && yarn start`)
4. Start the RN app with `yarn android` (or `ios`)
4. React Native should show up as device, logging, open devtools and reload should work


mweststrate:
N.B.: during testing I encountered errors like `ReferenceError: SHA-1 for file /Users/mweststrate/Desktop/cli/node_modules/metro/src/lib/polyfills/require.js (/Users/mweststrate/Desktop/cli/node_modules/metro/src/lib/polyfills/require.js) is not computed`. 

However those seem to be unrelated, and the setup works despite that. But to be able to see the fully working version; using relative-deps to link the cli, rather than `yarn link` seems to solve the issue.

mweststrate:
Tests failures seem to be unrelated to this PR, let me know if they are related, or this branch needs rebasing

thymikee:
Lovely! 

Whenever you run Metro (e.g. through `start`) for local CLI development, you need to pass it extra root (it sucks, but we're working on removing this indirection): https://github.com/react-native-community/cli/blob/master/CONTRIBUTING.md#running-start-command. That should get rid of `ReferenceError: SHA-1`

Some e2e tests are flaky, we need to fix or at least disable them, so unrelated.

mweststrate:
Tried and that failed, but fiddled a bit more, and added changes to this PR describing what I needed instead :)

grabbou:
I am going to look at this later today or tomorrow, thanks for this contribution!

lucasbento:
I just tested this on iPhone simulator (iOS 13.3), reload works fine but opening the dev menu does not.

Also, I was only able to get the packager to work by running:

```sh
REACT_NATIVE_APP_ROOT=path/to/cli node path/to/cli/packages/cli/build/bin.js start
```

mweststrate:
@lucasbento against which RN version was that? I tested against a project in 0.62.rc-0

lucasbento:
I tested on 0.61.5 😅 forgot that the CLI has been bumped for 0.62, nevermind then!

packages/cli/src/commands/server/runServer.ts
```diff
@@ -108,6 +120,12 @@ async function runServer(_argv: Array<string>, ctx: Config, args: Args) {
     '/debugger-proxy',
   );
   const ms = messageSocket.attachToServer(serverInstance, '/message');
+  eventsSocket = eventsSocketModule.attachToServer(
+    serverInstance,
+    '/events',
+    ms,
+  );
+  logger.log('Created events socket');
```
grabbou: Do we need this console log here?

.watchmanconfig
```diff
@@ -0,0 +1 @@
+{}
```
grabbou: I guess we can gitignore this file?

packages/cli/src/commands/server/eventsSocket.ts
```diff
@@ -0,0 +1,161 @@
+/**
+ * (c) Facebook, Inc. and its affiliates. Confidential and proprietary.
```
grabbou: Is this header appropriate here? We're not adding them to any new files and this project is MIT based. Not a lawyer, but "confidential" is a bit concerning :D

packages/cli/src/commands/server/eventsSocket.ts
```diff
@@ -0,0 +1,161 @@
...
+type Server = HttpServer | HttpsServer;
+
+type Command = {
+  version: number;
+  type: 'command';
+  command: string;
+  params?: any;
+};
+
+const PROTOCOL_VERSION = 2;
```
grabbou: What is a `PROTOCOL_VERSION`? Something Flipper-related? Would be great to have a comment.

packages/cli/src/commands/server/eventsSocket.ts
```diff
@@ -0,0 +1,161 @@
...
+
+function parseMessage(data: any): any {
```
grabbou: Wouldn't it be better to have `parseMessage<T>(data: Object): T | undefined`?

packages/cli/src/commands/server/eventsSocket.ts
```diff
@@ -0,0 +1,161 @@
...
+    clients.set(clientId, clientWs);
+
+    clientWs.onclose = clientWs.onerror = () => {
+      clients.delete(clientId);
+    };
+
+    clientWs.onmessage = event => {
+      const message: Command = parseMessage(event.data);
+      if (message == null) {
+        logger.error('Received message not matching protocol');
```
grabbou: Just a note that we already log this information inside `parseMessage` I'd suggest to leave it just once.

packages/cli/src/commands/server/eventsSocket.ts
```diff
@@ -0,0 +1,161 @@
...
+      // This exposes the full JS logs and enables issuing commands like reload
+      // so let's make sure only locally running stuff can connect to it
+      return origin.startsWith('http://localhost:');
+    },
+  });
+
+  const clients = new Map();
+  let nextClientId = 0;
+
+  function broadCastEvent(message: any) {
```
grabbou: Took me a while to realize that this is used to actually send logs from JavaScript to all connected clients that support it (such as Flipper). Again, a little description here or at the top of the module would be great!

packages/cli/src/commands/server/eventsSocket.ts
mweststrate: I think I copied that just from some of the other CLI files, will double check😅

packages/cli/src/commands/server/runServer.ts
mweststrate: The way Flipper works, is that it scans the system for all open terminals, and parses their output stream. To find out which one belongs to Metro, it need to be able to recognize it, so this logging statement is secretly used as a marker so that we know that we are looking at the metro output stream.

mweststrate: Kidding, will remove.

packages/cli/src/commands/server/eventsSocket.ts
mweststrate: Just dropped the header, like in most files. Some files do have still this header though (e.g. messageSocket)

.watchmanconfig
mweststrate: Done

mweststrate:
Processed all review comments! Documented the general setup better, hopefully clarifies the bi-directional communication setup between the websockets and reporter better :)

packages/cli/src/commands/server/eventsSocket.ts
```diff
@@ -0,0 +1,201 @@
+import {Server as WebSocketServer} from 'ws';
+import {logger} from '@react-native-community/cli-tools';
+import prettyFormat from 'pretty-format';
+import {Server as HttpServer} from 'http';
+import {Server as HttpsServer} from 'https';
+import messageSocketModule from './messageSocket';
+
+/**
+ * The eventsSocket websocket listens a the 'events/` for websocket
```
thymikee:
```suggestion
 * The eventsSocket websocket listens at the 'events/` for websocket
```

packages/cli/src/commands/server/eventsSocket.ts
```diff
@@ -0,0 +1,201 @@
...
+ * connections, on which all metro reports will be emitted.
```
thymikee:
```suggestion
 * connections, on which all Metro reports will be emitted.
```

PR #958, opened 2020-02-12
[HOLD] [iOS] Add `--enabler-flipper` flag

alloy:
Summary:
---------

As mentioned [here](https://github.com/facebook/react-native/issues/27565#issuecomment-580950480), this makes it so there’s a clearer way to control whether or not to compile Flipper client in the user’s application.

This will set the `FB_FLIPPER_ENABLED` and `FB_SONARKIT_ENABLED` CPP definitions to `1`, which, with the default RN template will compile the flipper bits in.

Depends on this change: https://github.com/alloy/react-native/commit/ae4f936753e1ff2cfd134e220924597166db5ee4

Test Plan:
----------

The right `xcodebuild` command is invoked:

```bash
react-native run-ios --enable-flipper
```

```
info Found Xcode workspace "TestFlipper.xcworkspace"
info Building (using 'xcodebuild -workspace TestFlipper.xcworkspace -configuration Debug -scheme TestFlipper -destination id="60E7F277-DC13-46B8-ADFB-7C5C9BC01274" GCC_PREPROCESSOR_DEFINITIONS="\$(inherited) FB_SONARKIT_ENABLED=1 FB_FLIPPER_ENABLED=1"')
```

…and indeed the application only shows up in Flipper desktop when it’s enabled.

### DIFF
--- a/packages/platform-ios/src/commands/runIOS/index.ts
+++ b/packages/platform-ios/src/commands/runIOS/index.ts
@@ -37,6 +37,7 @@ type FlagsT = {
   verbose: boolean;
   port: number;
   terminal: string | undefined;
+  enableFlipper: boolean;
 };
 
 function runIOS(_: Array<string>, ctx: Config, args: FlagsT) {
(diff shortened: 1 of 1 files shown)


alloy:
Having written all this, I’m not so sure we should do it like this after all... Will follow-up on https://github.com/facebook/react-native/issues/27565.

alloy:
Superseded by https://github.com/facebook/react-native/pull/28044

thymikee:
Thanks! :)

PR #960, opened 2020-02-13
chore: migrate to leven to find the closest match

### LINKED ISSUE #959: Migrate to leven to find the closest match for an unknown command
## Describe the Feature

As of now [didyoumean.js](https://github.com/dcporter/didyoumean.js/) matching engine is used to suggest matching commands if the user mistypes. But it is not longer actively maintained. Instead I would suggest switching over to [leven](https://github.com/sindresorhus/leven) which is the fastest JS implementation of the Levenshtein distance algorithm.

## Possible Implementations
`leven` returns the distance between 2 strings which can be made use of to find the exact match.

## Related Issues
N/A

jamesgeorge007:
Summary:
---------
closes #959 

`didyoumean.js` is not actively maintained anymore. Henceforth it is a good idea migrating to `leven`.


packages/cli/src/index.ts
```diff
@@ -85,10 +85,10 @@ function printHelpInformation(
 }
 
 function printUnknownCommand(cmdName: string) {
-  const suggestion = didYouMean(
-    cmdName,
-    commander.commands.map(cmd => cmd._name),
-  );
+  const availableCommands = commander.commands.map(cmd => cmd._name);
+  const suggestion = availableCommands.find(cmd => {
+    return leven(cmd, cmdName) < 3;
```
grabbou: I don't have any preference here whether `3` is a good number or not. `didyoumean` has [threshold of 40% by default](https://github.com/dcporter/didyoumean.js#threshold), so in order to not change the behaviour, we would need to use a different number.

packages/cli/src/index.ts
```diff
@@ -85,10 +85,10 @@ function printHelpInformation(
 }
 
 function printUnknownCommand(cmdName: string) {
-  const suggestion = didYouMean(
-    cmdName,
-    commander.commands.map(cmd => cmd._name),
-  );
+  const availableCommands = commander.commands.map(cmd => cmd._name);
```
grabbou: We don't need this call. We can simply do `const suggestion = commander.commands.find(cmd => level(cmd._name, cmdName) < 3);`

packages/cli/src/index.ts
grabbou: I guess replacing this with `< cmd * 0.4` would be equivalent. It's a good UX question.

packages/cli/src/index.ts
jamesgeorge007:
```js
const suggestion = commander.commands.find(cmd => leven(cmd._name, cmdName) < 3)._name;
```
That would be hard to read. I would say grabbing command names from the entire `commander.commands` array would be the way to go.

packages/cli/src/index.ts
jamesgeorge007: That would result in `NaN`. FYI `jests-cli` makes use of the same threshold: https://github.com/facebook/jest/blob/5d1e98beda19dcfff3f10557396098524efbb639/packages/jest-validate/src/utils.ts#L55

thymikee: The intention here was `cmd.length * 0.4` and it makes sense. I'm gonna merge that, because I don't care too much. But feel free to improve this further :)

PR #968, opened 2020-02-14
fix: running CLI from directory with spaces

grabbou:
Autolinking files will not work when directory has space in the name.


grabbou:
@thymikee will do a follow-up

Issue #406, opened 2019-05-28
Cannot find module '@react-native-community/cli'


ibrahimahmed-io:
## Environment

Running 'react-native info' throws that error as well.

internal/modules/cjs/loader.js:584
    throw err;
    ^

Error: Cannot find module '@react-native-community/cli'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:582:15)
    at Function.Module._load (internal/modules/cjs/loader.js:508:25)
    at Module.require (internal/modules/cjs/loader.js:637:17)
    at require (internal/modules/cjs/helpers.js:22:18)
    at Object.<anonymous> (E:\Projects\Hady\ProcoorCommunication\node_modules\react-native\cli.js:12:11)
    at Module._compile (internal/modules/cjs/loader.js:701:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:712:10)
    at Module.load (internal/modules/cjs/loader.js:600:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:539:12)
    at Function.Module._load (internal/modules/cjs/loader.js:531:3)

## Description

Running 'react-native init AppName' results in this error (Cannot find module '@react-native-community/cli')

internal/modules/cjs/loader.js:584
    throw err;
    ^

Error: Cannot find module '@react-native-community/cli'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:582:15)
    at Function.Module._load (internal/modules/cjs/loader.js:508:25)
    at Module.require (internal/modules/cjs/loader.js:637:17)
    at require (internal/modules/cjs/helpers.js:22:18)
    at Object.<anonymous> (E:\Projects\Hady\ProcoorCommunication\node_modules\react-native\cli.js:12:11)
    at Module._compile (internal/modules/cjs/loader.js:701:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:712:10)
    at Module.load (internal/modules/cjs/loader.js:600:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:539:12)
    at Function.Module._load (internal/modules/cjs/loader.js:531:3)

## Reproducible Demo

Run 'react-native init AppName' after it finishes it will try to run cli.js at 'node_modules\react-native\local-cli\cli.js' but it fails and throws that error 'Error: Cannot find module '@react-native-community/cli''

thymikee:
[...]

PR #973, opened 2020-02-14
fix: resolve CLI through react-native to avoid hoisting issues

thymikee:
Summary:
---------

See https://github.com/react-native-community/cli/issues/406#issuecomment-586220725 for explanation what happens.

This PR makes use of a fact, that CLI is available under https://github.com/facebook/react-native/blob/master/cli.js, which works regardless of package manager hoisting strategy. It doesn't make us closer to PnP support because we're making assumption we have `react-native` in the node_modules tree without declaring a dependency, but we can fix that later.

Fixes https://github.com/react-native-community/cli/issues/406

Test Plan:
----------

CLI is properly resolved under `npm`.

### DIFF
--- a/packages/platform-ios/native_modules.rb
+++ b/packages/platform-ios/native_modules.rb
@@ -15,7 +15,8 @@ def use_native_modules!(config = nil)
     config = nil;
   end
 
-  cli_resolve_script = "console.log(require('react-native/cli').bin);"
+  # Resolving the path the RN CLI. The `@react-native-community/cli` module may not be there for certain package managers, so we fall back to resolving it through `react-native` package, that's always present in RN projects
+  cli_resolve_script = "try {console.log(require('@react-native-community/cli').bin);} catch (e) {console.log(require('react-native/cli').bin);}"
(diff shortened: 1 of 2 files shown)


thymikee:
Cocoapods tests are failing, because there's no `react-native` in our deps.

packages/platform-ios/native_modules.rb
```diff
@@ -15,7 +15,8 @@ def use_native_modules!(config = nil)
     config = nil;
   end
 
-  cli_resolve_script = "console.log(require('@react-native-community/cli').bin);"
+  # Resolving the path the RN CLI. The `@react-native-community/cli` module may not be there for certain package managers, so we fall back to resolving it through `react-native` package, that's always present in RN projects
+  cli_resolve_script = "try {console.log(require('@react-native-community/cli').bin);} catch (e) {console.log(require('react-native/cli').bin);}"
```
thymikee: cc @alloy in reality this try/catch is only necessary for testing environment. Do you think it's better to leave this or somehow detect we're in "test" environment and change path accordingly? (usually a bad idea in my experience)

alloy: Might be better to mock `execute_command` instead?

PR #975, opened 2020-02-14
Fix: If versionsToShow is not an array, return N/A

ph0ph0:
Summary:
---------

I tried using the `doctor` command in my machine but it crashed like this: 

image

Turned out to be a problem with Android SDK not being identified and throwing an error because it returns `versions` in the `androidSDK.ts`


thymikee:
Hey! Thanks for this PR. There seems to be a mismatch between your Prettier version and what we use here. Could you please run `yarn lint --fix` to resolve styling issues?

packages/cli/src/commands/doctor/doctor.ts
```diff
@@ -23,7 +23,9 @@ const printCategory = ({label, key}: {label: string; key: number}) => {
 
 const printVersions = ({version, versions, versionRange}) => {
   if (versions) {
-    const versionsToShow = versions.join(', ');
+    const versionsToShow = Array.isArray(versions)
+      ? versions.join(', ')
+      : 'N/A';
```
thymikee: Please update the `HealthCheckResult` types to account for this type, so we can avoid similar issues in the future

ph0ph0: Hey @thymikee, thanks very much for the pointers, I have made the changes :) Thanks!

PR #982, opened 2020-02-16
fix(ios): skip script phases on abstract targets

Salakar:
Summary:
---------

Abstract targets [don't support script phases](https://github.com/CocoaPods/Core/blob/master/lib/cocoapods-core/podfile/dsl.rb#L432), this change skips adding script phases if the current target is abstract.

**Fixes** https://github.com/invertase/react-native-firebase/issues/2800

cc @orta / @alloy for review please, thanks!

Test Plan:
----------

@gabrielbull would you be able to patch package this change in and confirm it now works? https://github.com/invertase/react-native-firebase/issues/2800

---

**script phases continue to work on non abstract targets**:
image

### DIFF
--- a/packages/platform-ios/native_modules.rb
+++ b/packages/platform-ios/native_modules.rb
@@ -184,6 +184,10 @@ if $0 == __FILE__
       target_definition.singleton_class.send(:define_method, :dependencies) do
         current_target_definition_dependencies
       end
+      
+      target_definition.singleton_class.send(:define_method, :abstract?) do
+        false
+      end
 
       podfile.singleton_class.send(:define_method, :current_target_definition) do
         target_definition
--- a/packages/platform-ios/native_modules.rb
+++ b/packages/platform-ios/native_modules.rb
@@ -73,7 +73,7 @@ def use_native_modules!(config = nil)
 
     pod spec.name, :path => relative_path.to_path
 
-    if package_config["scriptPhases"]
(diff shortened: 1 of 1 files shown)


thymikee:
Thanks @alloy, @Salakar! This is released in v3.2.

rivaros:
@Salakar hello, what about real targets inside abstract targets ? Pls also check link above.


```
abstract_target 'SomeAbstractTarget' do
  config = use_native_modules!

  # Flags change depending on the env values.
  flags = get_default_flags()

  pod 'FirebaseCore', :modular_headers => true  
  pod 'FirebaseMessaging', :modular_headers => true  
  pod 'GoogleUtilities', :modular_headers => true  

  use_react_native!(
    :path => config[:reactNativePath],
    # to enable hermes on iOS, change `false` to `true` and then install pods
    :production => production,
    :hermes_enabled => flags[:hermes_enabled],
    :fabric_enabled => flags[:fabric_enabled],
    :flipper_configuration => FlipperConfiguration.enabled,
    # An absolute path to your application root.
    :app_path => "#{Pod::Config.instance.installation_root}/.."
  )

  target 'RealTarget' do
    # Production-specific pods

    target 'RealTargetTests' do
      inherit! :complete
      # Pods for testing
    end
  end

end



```

AuroPick:
How was this even merged. This is not even a fix. This just comments the bug.

I want to use abstract target for multiple environments in my projects. but scripts are not working.

PR #987, opened 2020-02-18
chore: use Node 12 on Azure

thymikee:
Summary:
---------

Installing packages takes roughly 2mins, that's slow. Tried adding Yarn cache based on https://docs.microsoft.com/en-us/azure/devops/pipelines/caching/?view=azure-devops#example but it resulted in even slower times, as we constantly hit cache misses. Will leave that for later.
Upgrading Node version, which should contribute to slightly faster test runs.

Test Plan:
----------

Azure

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -4,10 +4,22 @@ trigger:
 pool:
   vmImage: windows-2019
 
+variables:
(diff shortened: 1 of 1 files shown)


grabbou:
Now, the big question is - can we finally make our Windows CI work? :D

thymikee:
That's what I wanted to play with, but I ended up trying to fix the cache. Will try again later

PR #992, opened 2020-02-19
chore: simplify config test

thymikee:
Summary:
---------

Found some dead code while looking at config test. Why not removing it.

Test Plan:
----------

CI


thymikee:
LOL, looks like it's not so dead after all

PR #993, opened 2020-02-19
chore: fix windows config tests

thymikee:
Summary:
---------

Windows expects paths to be forward-slashed, so using `slash` in our test to achieve that

Test Plan:
----------

Windows config test no longer failing


thymikee:
Yay it worked!

PR #994, opened 2020-02-20
chore: fix windows sdk test

thymikee:
Summary:
---------

Windows CI sets `$CWD` to `D:\`, which is not where the project exists. To not alter the source code too much, I mocked the `findProjectRoot` so that it returns a local path. A proper fix would be reading the root (and, in turn, gradle.build path) from the config passed to the command. See #791 for reference

Test Plan:
----------

Green windows CI


thymikee:
image
🎉

PR #996, opened 2020-02-20
Create sourcemaps for easier debugging

molant:
Summary:
---------

I really need all the help I can get when coding. Having breakpoints in TypeScript instead of JavaScript is really high on my list. Otherwise I always end up modifying the JS and then erasing all changes in the next build because I was modifying the wrong place 😅

The only way I've found to make this work in this project is by creating sourcemaps.
Here's an example of it running on VS Code:

TS breakpoints working on the project

What it does:

1. Create a valid `.map` for each file in the build process
1. Append `//# sourceMappingURL=FILE.map` to each JS generated file
1. Ignore all `.map` files during the publishing process so the package doesn't increase in size. There were some packages that didn't have a `files` property in `package.json`. I have not ignored in those cases.

To make this work from VS Code you just need a regular node configuration like in the image above or the following:

```json
{
  "version": "0.2.0",
  "configurations": [  
    {
      "type": "node",
      "request": "launch",
      "name": "CLI",
      "skipFiles": ["<node_internals>/**"],
      "program": "${workspaceFolder}/packages/cli/build/bin.js",
      "args": ["doctor"],
      "console": "integratedTerminal",
      "cwd": "${workspaceFolder}"
    }
  ]
}
```

It does not require setting up the `sourceMaps` or `outFiles` properties.

In this case I'm running the CLI package with the `doctor` parameter (`args`). This file is currently ignored by `.gitignore` even though there are some files from `.vscode` that are committed.

Happy to contribute this `launch.json` or try to come up with something more generic if wanted.

Test Plan:
----------

N/A

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ build/
 *.tsbuildinfo
 .cache
 .watchmanconfig
+coverage
--- a/babel.config.js
+++ b/babel.config.js
@@ -28,4 +28,5 @@ module.exports = {
       }),
     ],
   ],
+  sourceMaps: true,
 };
--- a/jest.config.js
+++ b/jest.config.js
@@ -15,4 +15,5 @@ module.exports = {
       testMatch: ['<rootDir>/**/__tests__/*{.,-}test.[jt]s'],
     },
   ],
+  collectCoverageFrom: ['packages/**/src/**/*.ts'],
 };
--- a/jest.config.js
+++ b/jest.config.js
@@ -16,7 +16,9 @@ module.exports = {
     },
   ],
   collectCoverageFrom: [
-    '<rootDir>/packages/**/src/**/*.ts',
-    '<rootDir>/!packages/cli-types/src/**/*.ts',
+    '**/packages/*/**/*.ts',
+    '!**/__mocks__/**',
+    '!**/__tests__/**',
+    '!**/build/**',
   ],
 };
--- a/jest.config.js
+++ b/jest.config.js
@@ -15,5 +15,8 @@ module.exports = {
       testMatch: ['<rootDir>/**/__tests__/*{.,-}test.[jt]s'],
     },
   ],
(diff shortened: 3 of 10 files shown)


thymikee:
Cool! This broke tests. Can you have a look?

molant:
Yes, I'm trying to figure out why and which ones to look at 😅
By looking at other PRs it looks like the new failing ones are `lts-tests` and `unit-tests`. Is it safe to ignore the other ones?

thymikee:
We've recently fixed Windows tests on Azure, so only e2e-tests fail now

molant:
@thymikee I'm looking into it what I'm seeing is really similar to [this comment](https://github.com/facebook/jest/issues/7921#issuecomment-466204098):

If I change `jest.config.js` and add the following:

```js
collectCoverageFrom: [
  'packages/**/src/**/*.ts',
  '!packages/cli-types/src/**/*.ts',
],
``` 

everything passes but the output will get coverage for even for the files that have not been run in the tests (quite a few).

From removing files it looks like the issue is under `/packages/cli/src/tools/config`. If I delete those tests everything passes, but if I add even one things start failing.

Any idea of what could be going on?

molant:
Looks like windows failed because of [`ENOTEMPTY`](https://dev.azure.com/react-native-community/cli/_build/results?buildId=924&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9&t=5caf77c8-9b10-50ef-b5c7-ca89c63e1c86&l=367) and unrelated to the above.

I don't think it's related to the previous error and this one could maybe be avoid by increasing maxBusyTries` to something greater than the default 3.

thymikee:
Can we rebase this PR to get green CI?

molant:
> Can we rebase this PR to get green CI?

@thymikee sure!

molant:
Everything is ✅ now 🥳

Issue #985, opened 2020-02-18
[0.62.0-rc.2] - Cannot run android: No such file or directory, open 'app\build\intermediates\merged_manifests


owinter86:
Just so it doesn't get lost, but the closed issue described here is still an issue on the new RN release candidate 0.62.0-rc.2

https://github.com/react-native-community/cli/issues/967

It can be replicated on new init projects when the build folder is empty or cleared.

```
➜ yarn android
yarn run v1.22.0
$ react-native run-android
info Running jetifier to migrate libraries to AndroidX. You can disable it using "--no-jetifier" flag.
Jetifier found 960 file(s) to forward-jetify. Using 4 workers...
info JS server already running.
error ENOENT: no such file or directory, open 'app/build/intermediates/merged_manifests/debug/AndroidManifest.xml'. Run CLI with --verbose flag for more details.
Error: ENOENT: no such file or directory, open 'app/build/intermediates/merged_manifests/debug/AndroidManifest.xml'
```

```
├─ @react-native-community/cli-debugger-ui@3.0.0
├─ @react-native-community/cli-platform-android@4.0.2
├─ @react-native-community/cli-platform-ios@4.1.0
├─ @react-native-community/cli-tools@3.0.0
├─ @react-native-community/cli-types@3.0.0
└─ @react-native-community/cli@4.1.0
```

```
➜ npx react-native info
info Fetching system and libraries information...
System:
    OS: macOS 10.15.3
    CPU: (4) x64 Intel(R) Core(TM) i5-7267U CPU @ 3.10GHz
    Memory: 85.88 MB / 16.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 10.19.0 - ~/.nvm/versions/node/v10.19.0/bin/node
    Yarn: 1.22.0 - /usr/local/bin/yarn
    npm: 6.13.4 - ~/.nvm/versions/node/v10.19.0/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  SDKs:
    iOS SDK:
      Platforms: iOS 13.2, DriverKit 19.0, macOS 10.15, tvOS 13.2, watchOS 6.1
    Android SDK:
      API Levels: 23, 26, 27, 28, 29
      Build Tools: 26.0.3, 27.0.3, 28.0.2, 28.0.3, 29.0.2
[...]

PR #934, opened 2020-01-30
feat: run-android respects applicationId from build.gradle when running app 

grabbou:
Summary:
---------

Follow up to #255 with merge fixes, TypeScript rewrites, and less aggressive modifications.

Fixes #921 

This PR has a small breaking change - it removes `appId` and `appIdSuffix` args from the CLI. You should now use a `variant` argument that works perfectly and without issues (it will read appId and appIdSuffix from Gradle).

Gradle 101: 
> Configuration can have build types and flavors. Gradle matches flavors (like demo and full) with build types (like `debug` and `release`) and generates variants, e.g `demoDebug` and `demoRelease`. 
> 
> Each flavor can include different `appId` and `appIdSuffix` and that's where we read them from. 
>
> Allowing both to exist is a configuration error and we should've removed these two back when we added `variant` support in first place.

We will release it for 0.62 exclusively by bumping major with few other commits.

PR #999, opened 2020-02-21
Revert "feat: run-android respects applicationId from build.gradle when running app "

thymikee:
Reverts react-native-community/cli#934. It breaks usage in new apps for RN 0.62-rc.x and also break in monorepo setup for android. 

Fixes https://github.com/react-native-community/cli/issues/985

cc @AnimaSA please have a look at https://github.com/react-native-community/cli/pull/791 because I _think_ it solves similar problem, but in a different way.


thymikee:
cc @janicduplessis

PR #1001, opened 2020-02-22
chore: use a better threshold factor to find the closest match

jamesgeorge007:
Summary:
---------
Follow up of #960 

```bash
react-native inf
error Unrecognized command "inf". Did you mean "init"?
``` 

Set the threshold to 40% such that `info` shows up instead in the above scenario.


jamesgeorge007:
Ig we can land this one for now, and I can shoot a follow up PR as you suggested.

PR #1004, opened 2020-02-22
chore: test case for command suggestion feature

jamesgeorge007:
Summary:
---------
Follow up of #1001 

Test Plan:
----------
Covers the test case for #960

### DIFF
--- a/__e2e__/validations.test.ts
+++ b/__e2e__/validations.test.ts
@@ -1,5 +1,3 @@
-import fs from 'fs';
-import path from 'path';
 import {run, getTempDirectory, cleanup, writeFiles} from '../jest/helpers';
 
 const DIR = getTempDirectory('command-validations');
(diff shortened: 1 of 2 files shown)


jamesgeorge007:
@thymikee having an additional `lint:fix` script would be a good idea.

thymikee:
I don't think so. You can run `yarn lint --fix`

jamesgeorge007:
@thymikee hope we're good to land :rocket:

__e2e__/validations.test.ts
```diff
@@ -0,0 +1,18 @@
+import {run, getTempDirectory, cleanup, writeFiles} from '../jest/helpers';
+
+const DIR = getTempDirectory('command-validations');
```
thymikee: Can you rename this to `test_didyoumean` and the file to `didYouMean.test.ts`?

PR #1005, opened 2020-02-22
fix: building TS definitions broken by include package.json

thymikee:
Summary:
---------

Noticed that type defs were not built properly for `packages/cli`. The reason was `"include": "../package.json"` in TS config, which resulted in TS happily passing everything. Seems like a bug in TS, present on both 3.7 and 3.8.

Test Plan:
----------

Type defs are there in `build/` dirs

### DIFF
--- a/packages/cli/src/index.ts
+++ b/packages/cli/src/index.ts
@@ -12,7 +12,7 @@ import init from './commands/init/initCompat';
 import assertRequiredOptions from './tools/assertRequiredOptions';
 import loadConfig from './tools/config';
 
-import pkgJson from '../package.json';
+const pkgJson = require('../package.json');
 
 commander
   .option('--version', 'Print CLI version')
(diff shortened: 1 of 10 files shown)


packages/cli/src/index.ts
```diff
@@ -12,7 +12,7 @@ import init from './commands/init/initCompat';
 import assertRequiredOptions from './tools/assertRequiredOptions';
 import loadConfig from './tools/config';
 
-import pkgJson from '../package.json';
+const pkgJson = require('../package.json');
```
thymikee: TS is able to resolve package.json, but with a require. This is the actual fix, rest is errors that accumulated since the bug started happening

packages/cli/tsconfig.json
```diff
@@ -8,6 +8,5 @@
     {"path": "../tools"},
     {"path": "../cli-types"},
     {"path": "../debugger-ui"}
-  ],
-  "include": ["../package.json"]
```
thymikee: this broke type defs

PR #1007, opened 2020-02-22
chore: test case for default behavior

jamesgeorge007:
Summary:
---------
Test case for default behavior, shows up help information if no arguments were passed.

Test Plan:
----------
Match snapshot.


thymikee:
The test is failing on CI, mind looking at it?

PR #1008, opened 2020-02-23
Change iOS default simulator to iPhone 11 Pro Max

codler:
Summary:
---------
Changing from iPhone 11 to  iPhone 11 Pro Max due to screen resolution. When uploading screenshots to app store you are required to open iPhone 11 Pro Max and take screenshots. So it is better to use iPhone 11 Pro Max from beginning.

Test Plan:
----------

Run `react-native run-ios` and it should start iPhone 11 Pro Max as simulator

### DIFF
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -370,12 +370,13 @@ Builds your app and starts it on iOS simulator.
 
 #### `--simulator [simulator_name]`
 
-> default: iPhone 11
+> default: iPhone 11 Pro Max
 
 Explicitly set the simulator to use. Optionally include iOS version between parenthesis at the end to match an exact version, e.g. `"iPhone 6 (10.0)"`.
 
 Notes: If selected simulator does not exist, cli will try to run fallback simulators in following order:
 
+- `iPhone 11`
 - `iPhone X`
 - `iPhone 8`
 
--- a/packages/platform-ios/src/commands/runIOS/index.ts
+++ b/packages/platform-ios/src/commands/runIOS/index.ts
@@ -132,11 +132,12 @@ async function runOnSimulator(
 
   /**
    * If provided simulator does not exist, try simulators in following order
+   * - iPhone 11
    * - iPhone X
    * - iPhone 8
    */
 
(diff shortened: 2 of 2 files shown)


thymikee:
Where did you get this requirement? AppStore guidelines doc doesn't mention this https://developer.apple.com/app-store/review/guidelines/

codler:
You will get error when you are going to upload a screenshot in submit for review in app store connect.

liamjones:
@thymikee It's from here: https://help.apple.com/app-store-connect/#/devd274dd925 - this is the page you get pointed to when you try and upload an incorrectly sized screenshot.

thymikee:
> Device size or platform
> 6.5 inch (iPhone 11 Pro Max, **iPhone 11**, iPhone XS Max, iPhone XR)

codler:
> > Device size or platform
> > 6.5 inch (iPhone 11 Pro Max, **iPhone 11**, iPhone XS Max, iPhone XR)

Try take a screenshot from iphone 11 simulator and then upload and you will see.

thymikee:
I get you. However, I haven't seen any other request like yours and you can run with `--device iPhone 11 Pro Max` flag and fix your issue. I don't think changing the default for everyone is worth it, as it will invalidate lots of people caches for no apparent reason (not every dev submits to appstore). I'm more willing to do so when a new iPhone series comes out, or when more users recognize this as an actual problem.

codler:
okey, I dont see any big issue with the cache.
This issue have been around ever since this was fixed https://github.com/react-native-community/cli/pull/758 , before you never needed to think about what iphone simulator you need to open. you just run `npm run ios`, now you need to add --simulator everytime which is very annoying.

jeffreyffs:
You should always run your application on multiple simulators to get a feeling for how your app works on different screen sizes (iPhone SE is still a valid screen size!). I think defaulting to a larger screen size is potentially dangerous for people that never run their app on a smaller form factor, IMO.

thymikee:
Thanks for your input @jeffreyffs. Closing this. Still appreciate your pro-activeness @codler, looking forward to future contributions 😊

PR #1018, opened 2020-02-25
chore: bump fs-extra

thymikee:
Summary:
---------

Our other deps use fs-extra@8 already, so time to move on.

Test Plan:
----------

CI


molant:
FYI, this is done also in #1020

Issue #976, opened 2020-02-14
☂ Umbrella: Install development dependencies when missing


molant:
## Describe the Feature

Getting an environment ready for React Native development can take some time and is error prone. Developers have also problems setting up all the Android environment and emulators and on Windows it's even worse with the mixed messages about Hyper-V and HAXM.

After talking with a few RN developers from different projects and companies, it seems like the average to get a new machine up and running is between 1/2-1 day 😓

`doctor` is great to help with this, and I think we can take it even further by automatically installing and configuring the required dependencies if missing.

## Tasks

* [x] Check python version and install if missing #979
* [x] Check for JDK and install if missing #980

## Possible Implementations

My tests were mainly on Windows because I believe it's going to be the most complicated platform. My goal is to not require admin privileges so the developer doesn't see the UAC prompt. I think it's doable.

The list of dependencies from the docs are:

**Node 10+**

Script runs on node so not a lot to check other than the version via `process.version` for completeness?

**Python 2**

We can use the same approach as [`windows-build-tools`](https://github.com/felixrieseberg/windows-build-tools/blob/676db3e840b61fd83b785e3669988fa66be9486d/src/utils/get-is-python-installed.ts#L10).
The installation on Windows should be under `%LOCALAPPDATA%/python2`

- Download link: https://www.python.org/ftp/python/2.7.9/python-2.7.9.amd64.msi
- Command: `msiexec.exe /i "${pythonInstaller}" TARGETDIR="${targetDir}" /qn /L*P "python-log.txt"`

**JDK**

Version 8 or later is recommended in the docs. I'd rather use a later one as the [installation without admin privileges will be a lot easier](https://zemian.github.io/blog/2018/03/install-windows-jdk-without-admin-rights.html).

- Download link: `https://javadl.oracle.com/webapps/download/GetFile/${JAVA_VERSION}-b${BUILD}/${ID}/windows-i586/jdk-${JDK_VERSION}-windows-x64.exe` (need to add some extra stuff in that URL...)
- Command:  `"${jdkInstaller}" /s INSTALLDIR="${targetDir}" ADDLOCAL="ToolsFeature,SourceFeature"`;

**Android packages**

My initial tests where without Android Studio and going directly with the Android SDK Tools (https://dl.google.com/android/repository/sdk-tools-windows-4333796.zip). It looks like a lot of developers still need AS for native debugging and other scenarios so I think we should start with it and then maybe add an option to only download the SDK tools.

[...]

PR #1022, opened 2020-02-26
fix: avoid crashing if build.gradle is missing

molant:
Summary:
---------

Working on #976 I had some issues making the CLI crash when  `build.gradle` is missing.

Test Plan:
----------

### DIFF
--- a/packages/cli/src/commands/doctor/healthchecks/androidSDK.ts
+++ b/packages/cli/src/commands/doctor/healthchecks/androidSDK.ts
@@ -12,6 +12,10 @@ const getBuildToolsVersion = (): string => {
 
   const buildToolsVersionEntry = 'buildToolsVersion';
 
+  if (!fs.existsSync(gradleBuildFilePath)) {
+    return 'Not Found';
+  }
+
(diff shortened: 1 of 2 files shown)


packages/cli/src/commands/doctor/healthchecks/androidSDK.ts
```diff
@@ -12,6 +12,10 @@ const getBuildToolsVersion = (): string => {
 
   const buildToolsVersionEntry = 'buildToolsVersion';
 
+  if (!fs.existsSync(gradleBuildFilePath)) {
+    return '';
```
thymikee: Can we return `Not Found` instead? I don't feel too good about returning an empty string, but this way we could be consistent. It also seems like this function can return either non-empty string or `Not Found`. Which makes code on line 52 non-reachable. Can you investigate and remove stuff if possible?

molant:
My first reaction was to use "Not found" but then I saw that the caller had a case for empty/no string and I assumed it was for this particular case.

I also prefer having less cases 😊 I will take a look as soon as I can and update the PR.

Thanks!

________________________________

PR #1023, opened 2020-02-27
Feat: New HealthCheck. Checks if react-native or react-native-cli are installed globally

ph0ph0:
Summary:

A new health check that checks if react-native or react-native-cli packages are installed in npm or yarn global directories. If they are installed globally, then they are removed. This is a rough implementation :)


lucasbento:
@thymikee: I'm not sure if `which` is the way to go here, sometimes the package is installed by `yarn` but there's nothing on `which`, for some reason it's not linked properly, happens to me all the time with `ios-deploy`, also, not sure if `which` works in a Windows machine?

Shouldn't it also check for `react-native`, `react-native-cli` and `@react-native-community/cli` as you pointed out instead of just `react-native`?

thymikee:
> Shouldn't it also check for react-native, react-native-cli and @react-native-community/cli as you pointed out instead of just react-native?

No, because they all expose a `react-native` binary and that's what's troublesome.

> `which` works in a Windows machine?

https://github.com/npm/node-which supposedly works on Windows, but I'm not an expert.

> sometimes the package is installed by yarn but there's nothing on which

If it's not there for `which`, it's also not visible by your shell, making it impossible to run `$ react-native` directly, isn't it?

lucasbento:
Makes sense but if we are cleaning the global installation of `react-native`-cli-related libraries it would be nice to have it for all of them, no?

> If it's not there for which, it's also not visible by your shell, making it impossible to run $ react-native directly, isn't it?

You can still get it with `npm`/`yarn` `ls`, it's not something that would cause the issues that we have now but it's good to prevent further problems that could be caused by it.

thymikee:
Using `npm list -g --depth=0` and `yarn global list` sounds like an option too 👍

lucasbento:
> Using `npm list -g --depth=0` and `yarn global list` sounds like an option too 👍 

Unfortunately `yarn global list` accepts no `--pattern` (I'm not even sure it has `--depth`) which makes it very slow for this use-case.

packages/cli/package.json
```diff
@@ -44,8 +44,11 @@
     "find-up": "^4.1.0",
     "fs-extra": "^7.0.1",
     "glob": "^7.1.1",
+    "global-dirs": "^2.0.1",
     "graceful-fs": "^4.1.3",
     "inquirer": "^3.0.6",
+    "is-installed-globally": "^0.3.1",
```
lucasbento: This can be removed.

packages/cli/package.json
```diff
@@ -57,6 +60,7 @@
     "node-notifier": "^5.2.1",
     "open": "^6.2.0",
     "ora": "^3.4.0",
+    "path": "^0.12.7",
```
lucasbento: I don't think this is needed, you can use `path` from node itself.

packages/cli/src/commands/doctor/healthchecks/reactNativeGlobal.ts
```diff
@@ -0,0 +1,155 @@
...
+import fs from 'fs';
+import execa from 'execa';
+
+import globalDirectories from 'global-dirs';
+import isPathInside from 'is-path-inside';
+import {logger} from '@react-native-community/cli-tools';
+import {logError} from './common';
+import path from 'path';
+
+const label = 'react-native/react-native-cli globally installed';
```
lucasbento: Can be renamed to `react-native global installation`, perhaps.

packages/cli/src/commands/doctor/healthchecks/reactNativeGlobal.ts
```diff
@@ -0,0 +1,155 @@
...
+const resolveGlobalNpmPackage = moduleId => {
+  try {
+    const modulePath = require.resolve(
+      path.join(globalDirectories.npm.packages, moduleId),
+    );
+    return modulePath;
+  } catch (error) {
+    return '';
+  }
+};
```
lucasbento: You can probably reduce this to something like

```typescript
// This probably exists somewhere, can use it from there
const packageManagers = {
  yarn: 'yarn',
  npm: 'npm',
}

const findGlobalPackage = (packageManager, module) => {
  try {
    const modulePath = require.resolve(
      path.join(globalDirectories[packageManager]packages, module),
    );
    return modulePath;
  } catch (error) {
    return '';
  }
};

// And then use it as:
findGlobalPackage(packageManagers.yarn, module);
// or
findGlobalPackage(packageManagers.npm, module);
```

packages/cli/src/commands/doctor/healthchecks/reactNativeGlobal.ts
```diff
@@ -0,0 +1,155 @@
...
+
+const checkGlobalPaths = moduleId => {
+  return !!(
+    (resolveGlobalYarnPackage(moduleId) || resolveGlobalNpmPackage(moduleId)) &&
+    true
```
lucasbento: I guess this can be removed?

packages/cli/src/commands/doctor/healthchecks/reactNativeGlobal.ts
```diff
@@ -0,0 +1,155 @@
...
+  } catch (error) {
+    return null;
+  }
+
+  try {
+    reactNativeCLIGlobal = checkGlobalPaths('react-native-cli');
+    logger.log('******Found r-n-cli: ' + reactNativeCLIGlobal);
+  } catch (error) {
+    return null;
+  }
```
lucasbento: This can be reduced to something like:

```typescript
// Can be at the top of the file
const librariesToCheck = ['react-native', 'react-native-cli'];

const checkGlobalInstallations = () => {
  const librariesGloballyInstalled = librariesToCheck.map(library =>
    checkGlobalPaths(library),
  );

  // This way `librariesGloballyInstalled` will hold an array of paths
  // then you can loop through it to remove the libraries
  return librariesGloballyInstalled;
}
```

packages/cli/src/commands/doctor/healthchecks/reactNativeGlobal.ts
```diff
@@ -15,6 +14,11 @@ const packageManagers = {
   npm: 'npm',
 };
 
+const packages = {
```
lucasbento: Is there a reason to use an object instead of array?

You are not using the entries anywhere so you can just turn this into an array and remove the need of `Object.keys` in line 35 and 37.

packages/cli/src/commands/doctor/healthchecks/reactNativeGlobal.ts
```diff
@@ -26,42 +30,48 @@ const findGlobalPackage = (packageManager, moduleId) => {
...
+        findGlobalPackage(
+          packageManagers[packageManagerKey],
+          packages[packageKey],
+        ),
+      ),
+    )
+    .filter(e => e != null);
+
+  //Cant's use flat() in ts...? This is an alternative
+  const flattenedArray = [].concat.apply([], librariesGloballyInstalled);
```
lucasbento: Why is this needed?

packages/cli/src/commands/doctor/healthchecks/reactNativeGlobal.ts
```diff
@@ -26,42 +30,48 @@ const findGlobalPackage = (packageManager, moduleId) => {
...
+  const librariesGloballyInstalled = Object.keys(packages)
+    .map(packageKey =>
+      Object.keys(packageManagers).map(packageManagerKey =>
+        findGlobalPackage(
+          packageManagers[packageManagerKey],
+          packages[packageKey],
+        ),
+      ),
+    )
+    .filter(e => e != null);
```
lucasbento: You can replace this with:

```suggestion
    .filter(Boolean);
```

packages/cli/src/commands/doctor/healthchecks/reactNativeGlobal.ts
```diff
@@ -26,42 +30,48 @@ const findGlobalPackage = (packageManager, moduleId) => {
...
-  }
-
-  try {
-    reactNativeCLIGlobal = checkGlobalPaths('react-native-cli');
-    logger.log('******Found r-n-cli: ' + reactNativeCLIGlobal);
-  } catch (error) {
-    return null;
-  }
+  const packageManager = packagePath.includes('yarn') ? 'yarn' : 'npm';
+  const packageName = packagePath.includes('react-native/')
```
lucasbento: Instead of doing this I suggest that you return the `packageName` in the loop on line 35.

packages/cli/src/commands/doctor/healthchecks/reactNativeGlobal.ts
```diff
@@ -78,82 +88,17 @@ const removeNodePackage = async (packageName, packageManager, loader) => {
...
-      )
-    ) {
-      await removeNodePackage(reactNativeCLI, 'npm', loader);
-    }
+    const globalPaths = checkGlobalInstallalations();
+
+    console.log('Global paths: ' + globalPaths);
+
+    await globalPaths.map(
+      async packagePath => await removeNodePackage(packagePath, loader),
```
lucasbento: I believe you can omit `async/await` here.

```suggestion
      packagePath => removeNodePackage(packagePath, loader),
```

packages/cli/src/commands/doctor/healthchecks/reactNativeGlobal.ts
```diff
@@ -78,82 +88,17 @@ const removeNodePackage = async (packageName, packageManager, loader) => {
 export default {
   label: label,
   getDiagnostics: async () => ({
-    needsToBeFixed: checkGlobalInstall(),
+    needsToBeFixed: !!checkGlobalInstallalations(),
```
lucasbento: Since I suggested to turn the response of `checkGlobalInstallalations()` in an array, you can do this:

```suggestion
    needsToBeFixed: checkGlobalInstallalations().length > 0,
```

Also, you can put this check at the top and assign it to a variable so you don't need to do it twice.

packages/cli/src/commands/doctor/healthchecks/reactNativeGlobal.ts
```diff
@@ -0,0 +1,104 @@
...
+    ? 'react-native'
+    : 'react-native-cli';
+
+  console.log('******* pM: ' + packageManager + ' pN: ' + packageName);
+
+  try {
+    // For some reason the execa statement is not being executed, but no error is thrown?
+    packageManager === 'yarn'
+      ? await execa('yarn', ['global', 'remove', packageName])
+      : await execa('npm', ['uninstall', '--global', packageName]);
```
lucasbento: I suggest doing something like this:

```ts
const packageRemovalArguments = {
  [packageManagers.yarn]: ['global', 'remove'],
  [packageManagers.npm]: ['uninstall', '--global'],
};

// ...
// Then you can do:

await execa(packageManager, [...packageRemovalArguments[packageManager], packageName]);
```

packages/cli/src/commands/doctor/healthchecks/reactNativeGlobal.ts
ph0ph0: I included this because mapping over the packages/packageManagers returns a nested array that needs to be flattened? When you console.log this nested array you can't see it for some reason in the terminal (not sure why?), but look in the console on this playground (you also can't see it in the playground display area):

image

packages/cli/src/commands/doctor/healthchecks/reactNativeGlobal.ts
ph0ph0: Sorry not sure I understand how this will work, are you suggesting that librariesGloballyInstalled (line 29) returns an array of packageNames only that are globally installed, and as such checkGlobalInstallations returns these? 

As far as I'm aware, we need to know both which packages are globally installed and with which package managers? Returning only the packageName would mean that we lose the package manager information, but both of these are contained in the current return value which is the path of the globally installed package? Sorry if I have misunderstood

packages/cli/src/commands/doctor/healthchecks/reactNativeGlobal.ts
lucasbento: It's because the terminal does not expand further.

To log everything you can use `JSON.stringify(variable);`.

packages/cli/src/commands/doctor/healthchecks/reactNativeGlobal.ts
```diff
@@ -0,0 +1,95 @@
...
+    const modulePath = require.resolve(
+      path.join(globalDirectories[packageManager].packages, moduleId),
+    );
+    return modulePath;
+  } catch (error) {
+    return '';
+  }
+};
+
+const checkGlobalInstallalations = () => {
```
friederbluemle: Minor typo:

```suggestion
const checkGlobalInstallations = () => {
```

grabbou:
Closing due to lack of updates. Please re-open when more time! Thank you for your efforts so far.

PR #1025, opened 2020-02-28
chore: refactor envinfo

molant:
Summary:
---------

While working on #976 I saw that I needed to modify the properties requested to `envinfo` and that it was being accessed in a couple different places with different configurations.

This PR changes that to use a single point (and single configuration) to get the info.

### DIFF
--- a/packages/cli/src/tools/envinfo.ts
+++ b/packages/cli/src/tools/envinfo.ts
@@ -2,20 +2,37 @@
 import envinfo from 'envinfo';
 import {EnvironmentInfo} from '../commands/doctor/types';
 
-export default async function getEnvironmentInfo() {
-  return JSON.parse(
-    await envinfo.run(
-      {
-        Binaries: ['Node', 'Yarn', 'npm', 'Watchman'],
-        IDEs: ['Xcode', 'Android Studio'],
-        SDKs: ['iOS SDK', 'Android SDK'],
-        npmPackages: ['react', 'react-native', '@react-native-community/cli'],
-        npmGlobalPackages: ['*react-native*'],
-      },
-      {
-        json: true,
-        showNotFound: true,
-      },
-    ),
-  ) as EnvironmentInfo;
+/**
+ * Returns information about the running system.
+ * If `json === true`, or no options are passed,
+ * the return type will be an `EnvironmentInfo`.
+ * If set to `false`, it will be a `string`.
+ */
+async function getEnvironmentInfo(): Promise<EnvironmentInfo>;
+async function getEnvironmentInfo(json: true): Promise<EnvironmentInfo>;
+async function getEnvironmentInfo(json: false): Promise<string>;
+async function getEnvironmentInfo(
+  json = true,
+): Promise<string | EnvironmentInfo> {
+  const options = {json, showNotFound: true};
+
+  const info = (await envinfo.run(
+    {
+      System: ['OS', 'CPU', 'Memory', 'Shell'],
+      Binaries: ['Node', 'Yarn', 'npm', 'Watchman'],
+      IDEs: ['Xcode', 'Android Studio'],
+      SDKs: ['iOS SDK', 'Android SDK'],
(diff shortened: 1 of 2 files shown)


packages/cli/src/commands/doctor/healthchecks/__tests__/androidSDK.test.ts
```diff
@@ -34,7 +34,7 @@ describe('androidSDK', () => {
   let environmentInfo: EnvironmentInfo;
 
   beforeAll(async () => {
-    environmentInfo = await getEnvironmentInfo();
+    environmentInfo = (await getEnvironmentInfo()) as EnvironmentInfo;
```
thymikee: This is not good. Let's type `getEnvironmentInfo` properly so it returns correct type based on the arguments

packages/cli/src/tools/envinfo.ts
```diff
@@ -2,20 +2,38 @@
...
-    ),
-  ) as EnvironmentInfo;
+/**
+ * Returns information about the running system.
+ * If `json === true`, or no options are passed,
+ * the return type will be an `EnvironmentInfo`.
+ * If set to `false`, it will be a `string`.
+ */
+async function getEnvironmentInfo(): Promise<EnvironmentInfo>;
+async function getEnvironmentInfo(json: boolean): Promise<string>;
```
molant: I'm not super thrilled with this but haven't come up with a better idea yet.

* `getEnvironmentInfo()` return type will be `EnvironmentInfo` ✔
* `getEnvironmentInfo(false)` return type will be a `string` ✔
* `getEnvironmentInfo(true)` return type will be a `string` ❌

use `getEnvironmentInfo(true)`, the real return type would be `EnvironmentInfo` but TS will incorrectly say it's a `string`.

molant: Got it finally right 😊

packages/cli/src/tools/envinfo.ts
```diff
@@ -2,20 +2,39 @@
...
+ */
+async function getEnvironmentInfo(): Promise<EnvironmentInfo>;
+async function getEnvironmentInfo(json: true): Promise<EnvironmentInfo>;
+async function getEnvironmentInfo(json: false): Promise<string>;
+async function getEnvironmentInfo(
+  json = true,
+): Promise<string | EnvironmentInfo> {
+  const options = json
+    ? {json: true, showNotFound: true}
+    : {json: false, showNotFound: true};
```
thymikee: We can have a single object and use json value because it's already boolean

molant: Done!

packages/cli/src/tools/envinfo.ts
```diff
@@ -2,20 +2,37 @@
...
+ * the return type will be an `EnvironmentInfo`.
+ * If set to `false`, it will be a `string`.
+ */
+async function getEnvironmentInfo(): Promise<EnvironmentInfo>;
+async function getEnvironmentInfo(json: true): Promise<EnvironmentInfo>;
+async function getEnvironmentInfo(json: false): Promise<string>;
+async function getEnvironmentInfo(
+  json = true,
+): Promise<string | EnvironmentInfo> {
+  const options = {json, showNotFound: json};
```
thymikee: I thought we always want to show Not Found?

molant: I'm all for always showing Not Found. Is just that in one place it was explicit and in the other not (and I think the default is `false`).

I'll update the PR

molant: Updated

thymikee:
Thanks!

grabbou:
> and that it was being accessed in a couple of different places with different configurations.

I only see a single call site being updated as a part of this PR. Are there any other places we haven't refactored?

thymikee:
This was the only one and it also got out of sync with options, so it's an improvement already.

molant:
The other call used `tools/envinfo` already. I meant a couple using the npm package 😊

PR #1035, opened 2020-03-04
chore: set correct react-native peer dependency

thymikee:
Summary:
---------

This will still result in warnings for RC versions, but that's fine
cc @alloy 

Test Plan:
----------

None

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -67,7 +67,7 @@
     "ws": "^1.1.0"
   },
   "peerDependencies": {
-    "react-native": "^0.61.0"
+    "react-native": "^0.62.0"
   },
   "devDependencies": {
     "@types/command-exists": "^1.2.0",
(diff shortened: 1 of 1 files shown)


packages/cli/package.json
```diff
@@ -67,7 +67,7 @@
     "ws": "^1.1.0"
   },
   "peerDependencies": {
-    "react-native": "^0.61.0"
+    "react-native": "^0.62.0"
```
grabbou: Why don't we do `|| 0.62.0-rc.3` or similar to remove warnings for RC versions as well? I don't think this would hurt.

grabbou: Simple answer would be that we just don't care, which is fine as well

thymikee: Ok, went with including RC as well because we support it after all

PR #1036, opened 2020-03-04
feat: bail out when initializing new project to pre-existing directories

Esemesek:
Summary:
---------

Initializing to the pre-existing directory was all about removing all the files in that directory. This approach doesn't make much sense from user perspective as it might lead to unexpected/accidental removals (even though we ask). From now on, we just fail when the directory exists and we leave it up to the user what to do with that.

### DIFF
--- a/packages/cli/src/commands/init/init.ts
+++ b/packages/cli/src/commands/init/init.ts
@@ -44,14 +44,14 @@ function doesDirectoryExist(dir: string) {
   return fs.existsSync(dir);
 }
 
-async function setProjectDirectory(projectPath: string) {
-  if (doesDirectoryExist(projectPath)) {
-    throw new DirectoryAlreadyExistsError(projectPath);
+async function setProjectDirectory(directory: string) {
+  if (doesDirectoryExist(directory)) {
+    throw new DirectoryAlreadyExistsError(directory);
   }
 
   try {
-    mkdirp.sync(projectPath);
-    process.chdir(projectPath);
+    mkdirp.sync(directory);
+    process.chdir(directory);
   } catch (error) {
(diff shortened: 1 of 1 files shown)


packages/cli/src/commands/init/init.ts
```diff
@@ -42,37 +42,21 @@ interface TemplateOptions {
...
-function doesDirectoryExist(dir: string) {
-  return fs.existsSync(dir);
+function doesDirectoryExist(path: string) {
+  return fs.existsSync(path);
 }
 
-async function setProjectDirectory(directory: string) {
-  const directoryExists = doesDirectoryExist(directory);
-  if (directoryExists) {
-    const {shouldReplaceprojectDirectory} = await inquirer.prompt([
```
thymikee: you can remove the inquirer import now

packages/cli/src/commands/init/init.ts
```diff
@@ -224,13 +206,12 @@ export default (async function initialize(
...
   const version: string = minimist(process.argv).version || DEFAULT_VERSION;
 
-  const directoryName = path.relative(root, options.directory || projectName);
+  const projectPath = path.resolve(root, options.directory || projectName);
 
   try {
-    await createProject(projectName, directoryName, version, options);
+    await createProject(projectName, projectPath, version, options);
 
-    const projectFolder = path.join(root, directoryName);
```
thymikee: This results in "Run instructions" having absolute paths. I don't mind, but remember we had a little back and forth in this area :D

Esemesek: I'll just go back to what was here before to avoid any unnecessary discussions again. XD

PR #1037, opened 2020-03-04
chore: add deprecation messages for manual linking

Esemesek:
Summary:
---------

Let's finally deprecate linking commands C:

Test Plan:
----------

Deprecation warning prints correctly.

### DIFF
--- a/packages/cli/src/commands/link/link.ts
+++ b/packages/cli/src/commands/link/link.ts
@@ -33,7 +33,6 @@ async function link(
   ctx: Config,
   opts: FlagsType,
 ) {
-  printDeprecationWarning('link');
   let platforms = ctx.platforms;
   let project = ctx.project;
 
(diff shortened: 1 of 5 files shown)


packages/cli/src/commands/link/link.ts
```diff
@@ -32,6 +33,7 @@ async function link(
   ctx: Config,
   opts: FlagsType,
 ) {
+  printDeprecationWarning('link');
```
thymikee: `link` command itself links assets by default, which is still supported – maybe we should rename it to `link-assets`? Not sure what's best here. The thing is that `link [packageName]` and `link --all` get deprecated and we need a clear path forward

Esemesek: Oh righty. Forgot about the assets.

Esemesek:
And this is roughly how it looks like:
image

PR #1040, opened 2020-03-06
feat: install python if missing on Windows

### LINKED ISSUE #979: Check python version and install if missing
Part of #976 

# Validation

~~Get the output of running `python -V` and parse it~~

Use `envinfo` and look for `Languages`

# Installation

## Windows

The Windows binary that needs to be installed is:
* x64: https://www.python.org/ftp/python/2.7.9/python-2.7.9.amd64.msi
* x86: https://www.python.org/ftp/python/2.7.9/python-2.7.9.msi

The command to run it is 
`msiexec.exe /i "${pythonInstaller}" TARGETDIR="${targetDir}" /qn /L*P "python-log.txt"`

Where `pythonInstaller` is the path to the downloaded file and `targetDir` is
`process.env.LOCALAPPDATA/python2`

## macOS

Already installed

## Linux

Already installed

molant:
Summary:
---------

Fix #979

I've started working on the automatic installation of Python under Windows.
You can see a video in action here:

[RN CLI installing Python on Windows](https://youtu.be/XjUefsUc0pA) 

I haven't done any test yet, I wanted to get your initial thoughts about this before going any further (filenames, code organization, etc.).
New healthchecks should touch less files than this one.

Test Plan:
----------
TBD


packages/cli/package.json
```diff
@@ -45,6 +45,7 @@
     "find-up": "^4.1.0",
     "fs-extra": "^8.1.0",
     "glob": "^7.1.3",
+    "got": "^10.5.7",
```
molant: I'll test the other package I mentioned that was more lightweight and hopefully will not cause to fail in CI

packages/cli/src/commands/doctor/healthchecks/python.ts
```diff
@@ -0,0 +1,55 @@
...
+import {join} from 'path';
+import {Ora} from 'ora';
+import {executeCommand} from '../../../tools/executeWinCommand';
+
+export default {
+  label: 'Python',
+  getDiagnostics: async ({Languages}) => ({
+    needsToBeFixed: doesSoftwareNeedToBeFixed({
+      version:
+        (Languages && Languages.Python && Languages.Python.version) || '',
```
molant: I need to recheck that `Languages && Languages.Python` is still needed now that envinfo always shows the missing ones.

packages/cli/src/tools/executeWinCommand.ts
```diff
@@ -0,0 +1,10 @@
+import execa from 'execa';
+
+const executeCommand = async (command: string) => {
```
molant: I'll add some `jsdoc` comments where missing if things look ok

packages/cli/src/commands/doctor/healthchecks/python.ts
```diff
@@ -0,0 +1,55 @@
...
+      versionRange: versionRanges.PYTHON,
+    }),
+
+    version: (Languages && Languages.Python && Languages.Python.version) || '',
+    versionRange: versionRanges.PYTHON,
+  }),
+  win32AutomaticFix: async ({loader}: {loader: Ora}) => {
```
lucasbento: I like this but I don't really understand why it's needed.

I assume that Python is not required on other operational systems so, wouldn't this be the same case of `ios-deploy`/`CocoaPods` health check? If the OS running is not the one targeted by the health check then it doesn't run it at all.

packages/cli/src/commands/doctor/healthchecks/python.ts
```diff
@@ -0,0 +1,55 @@
...
+      const installPath = join(process.env.LOCALAPPDATA!, 'python2');
+
+      loader.start(`Downloading Python installer from "${installerUrl}"`);
+
+      const installer = await download(installerUrl);
+
+      loader.text = `Installing Python in "${installPath}"`;
+      const command = `msiexec.exe /i "${installer}" TARGETDIR="${installPath}" /qn /L*P "python-log.txt"`;
+
+      await executeCommand(command);
```
lucasbento: I like this :D

packages/cli/src/commands/doctor/healthchecks/python.ts
molant: You are right that it should be the same as `CocoaPods` for this particular case. I put it in the common health checks because I was testing the fallback (i.e.: on Linux the `runAutomaticFix` should be executed instead of `win32AutomaticFix`) and the forgot to change it.

The idea is to make it easy to implement "installers" for those dependencies that are shared across different OS. Does it make more sense now? In https://github.com/react-native-community/cli/issues/976#issuecomment-589882936 there's a bit more information

Also I'll move Python to the Windows only checks. Thanks for catching it up!

molant: @lucasbento now Python is only added if the platform is win32

packages/cli/src/commands/doctor/healthchecks/python.ts
molant: I've removed those checks as they aren't needed anymore 🎉

packages/cli/package.json
molant: Migrated to `node-fetch`  and now CI is ✅

packages/cli/package.json
```diff
@@ -55,6 +56,7 @@
     "metro-react-native-babel-transformer": "^0.58.0",
     "minimist": "^1.2.0",
     "mkdirp": "^0.5.1",
+    "node-fetch": "^2.6.0",
```
thymikee: we have a util for fetching that uses `node-fetch` under the hood: https://github.com/react-native-community/cli/blob/master/packages/tools/src/fetch.ts. You can import it as: 
```js
import {fetch} from '@react-native-community/cli-tools';
```

packages/cli/src/tools/executeWinCommand.ts
thymikee: sounds more like `executeShellCommand` isn't it? we normally don't use `{shell: true}` so we should make it explicit

packages/cli/src/tools/variables.ts
```diff
@@ -0,0 +1,23 @@
...
+ */
+const updateEnvironment = (variable: string, value: string) => {
+  // https://superuser.com/a/601034
+  const command = `for /f "skip=2 tokens=3*" %a in ('reg query HKCU\\Environment /v ${variable}') do @if [%b]==[] ( @setx ${variable} "${value};%~a" ) else ( @setx ${variable} "${value};%~a %~b" )
+  `;
+
+  return executeCommand(command);
+};
+
+export {setEnvironment, updateEnvironment};
```
thymikee: I don't understand the name of this file, why `variables.ts`?

packages/cli/src/tools/downloader.ts
```diff
@@ -0,0 +1,34 @@
...
+import * as path from 'path';
+import * as fs from 'fs';
+
+import fetch from 'node-fetch';
+
+/**
+ * Downloads the contents of the given `url`
+ * into the temp folder of the OS.
+ */
+const downloader = (url: string): Promise<string> => {
```
thymikee: `downloader` is not a good name for this utility. I'd go with something like `downloadToTmpdir`. Also, maybe it's worth to remove the downloader once we're done or failed to do stuff?

packages/cli/src/commands/doctor/doctor.ts
```diff
@@ -118,6 +118,10 @@ export default (async (_, __, options) => {
         // Assume that it's required unless specified otherwise
         const isRequired = healthcheck.isRequired !== false;
         const isWarning = needsToBeFixed && !isRequired;
+        const platformFixMethodName = `${process.platform}AutomaticFix`;
```
thymikee: This fails the grep test. Please be explicit and create a function that does a switch/case:
```js
function getAutomaticFixForPlatform(platform: string) {
  switch (platform) {
    case 'win32':
      return healthcheck.win32AutomaticFix;
    // ...
    default:
      return healthcheck.runAutomaticFix;
  }
}
// ...
runAutomaticFix: getAutomaticFixForPlatform(process.platform)
```
This is more code, but easier to find, modify and type

packages/cli/src/tools/variables.ts
molant: Does `environmentVariables.ts` sound better?

packages/cli/package.json
molant: @thymikee I think I'm going to combine my `downloader.ts` and `fetch.ts`. The way `fetch.ts` works right now will not work well with larger files that need to be streamed instead of hold in memory.

Also is there any guidance into what should go under `packages/tools` vs `packages/cli/src/tools`?

thymikee: That's fine. You can also refactor it if it makes sense. We move stuff to `packages/tools` when other packages share the same logic.

molant: I've created a new method in `fetch.ts` called `fetchToTemp`. Hopefully it will also address your concern about the name.

packages/cli/src/tools/variables.ts
molant: Renamed

packages/cli/src/tools/downloader.ts
molant: Moved to `fetch.ts`. Not sure what you mean by remove the `downloader` 🤔

packages/cli/src/commands/doctor/doctor.ts
molant: This has been refactored

packages/cli/src/tools/executeWinCommand.ts
molant: Done

molant:
@thymikee I think I've addressed all the feedback. Let me know if it looks better now. Also I'll change the commit history and make it cleaner once I have tests and things are ready to get merged (probably one commit for `tools` and another for `cli`).

packages/cli/src/commands/doctor/types.ts
```diff
@@ -83,6 +89,7 @@ export type RunAutomaticFix = (args: {
 }) => Promise<void> | void;
 
 export type HealthCheckInterface = {
+  [key: string]: any;
```
molant: Need to remove this now that is no longer needed.

packages/tools/package.json
```diff
@@ -10,12 +10,12 @@
     "chalk": "^3.0.0",
     "lodash": "^4.17.15",
     "mime": "^2.4.1",
-    "node-fetch": "^2.5.0"
+    "node-fetch": "^2.6.0"
```
molant: I've bumped it because of https://github.com/node-fetch/node-fetch/blob/master/CHANGELOG.md#v260 :

> Fix: incorrect Content-Length was returned for stream body in 2.5.0 release; note that node-fetch doesn't calculate content length for stream body.

packages/cli/src/commands/doctor/doctor.ts
```diff
@@ -89,6 +90,27 @@ type FlagsT = {
...
+) => {
+  switch (platform) {
+    case 'win32':
+      return healthcheck.win32AutomaticFix || healthcheck.runAutomaticFix;
+    case 'darwin':
+      return healthcheck.darwinAutomaticFix || healthcheck.runAutomaticFix;
+    case 'linux':
+      return healthcheck.linuxAutomaticFix || healthcheck.runAutomaticFix;
+    default:
+    case 'win32':
```
thymikee: I don't think we need this? This case is already handled

packages/cli/src/commands/doctor/types.ts
thymikee: So... can we remove it now?

packages/cli/src/commands/doctor/doctor.ts
molant: Copy pasting 🤦‍♂️

packages/cli/src/commands/doctor/types.ts
molant: Yes, was a note for me to not forget when I started addressing more feedback

packages/tools/src/fetch.ts
```diff
@@ -16,10 +20,41 @@ async function unwrapFetchResult(response: Response) {
...
+
+        result.body.on('end', () => {
+          resolve(tmpDir);
+        });
+
+        result.body.on('error', reject);
+      });
+    });
+  } catch (e) {
+    console.error(e);
```
thymikee: use `logger.error` from `cli-tools`

packages/cli/src/commands/doctor/versionRanges.ts
```diff
@@ -4,6 +4,7 @@ export default {
   YARN: '>= 1.10.x',
   NPM: '>= 4.x',
   WATCHMAN: '4.x',
+  PYTHON: '>= 2.x', // Is there a way to force it <= 3 ?
```
thymikee: Yes, you can use `>= 2.x <= 3` for example. This is passed straight to `semver.satisfies` so anything semver-valid is good

packages/cli/src/commands/doctor/healthchecks/python.ts
```diff
@@ -0,0 +1,59 @@
...
+      versionRange: versionRanges.PYTHON,
+    }),
+
+    version: Languages.Python.version,
+    versionRange: versionRanges.PYTHON,
+  }),
+  win32AutomaticFix: async ({loader}: {loader: Ora}) => {
+    try {
+      const arch = process.arch === 'x64' ? 'amd64.' : '';
+      const installerUrl = `https://www.python.org/ftp/python/2.7.9/python-2.7.9.${arch}msi`;
```
thymikee: Can we detect if Chocolatey is available and provide user with a choice:
- install with Chocolatey
- install using official distribution (or whatever works here)

Asking, because it's a recommended way to install Node and Python2 according to the RN getting started guide

molant: I believe Chocolatey is the recommended because it will take care of setting environment variables, which I'm already doing. 

I debated using it underneath because it would have been easier, but there are some cons I couldn't ignore. Users need to:

* Have it installed (it's definitely not as popular as homebrew on mac, I don't have it for example)
* Run the script in powershell instead of regular prompt
* Run in an elevated prompt as a lot of packages will not get installed correctly unless they are run as an admin directly

What I'm doing is what chocolatey would do but without less limitations so not sure about the ROI.

packages/cli/src/commands/doctor/versionRanges.ts
```diff
@@ -4,6 +4,7 @@ export default {
   YARN: '>= 1.10.x',
   NPM: '>= 4.x',
   WATCHMAN: '4.x',
+  PYTHON: '>= 2.x <= 3',
```
thymikee: I believe we want to make sure people are on Python2, right? I'm not sure if Python3 is supported.
```suggestion
  PYTHON: '>= 2.x < 3',
```

packages/cli/src/commands/doctor/healthchecks/python.ts
lucasbento: Sorry I didn't see #976 before, this is good stuff ❤️

molant:
@thymikee @lucasbento looks like everyone is happy so I'll work on adding some tests and rewriting the commit history next (and being strict with the python version). Hopefully I'll have something in the next few days.

Thanks!

thymikee:
Don't worry about git history, we squash on merge

molant:
I've added some tests to this PR an addressed final feedback. If there's anything else that I should do please let me know!
I'll start working on the JDK one once this gets merged 😊

grabbou:
Looks good to me!

packages/cli/src/commands/doctor/healthchecks/__tests__/python.test.ts
```diff
@@ -0,0 +1,69 @@
...
+    // @ts-ignore
+    environmentInfo.Languages.Python = {
+      version: '2.7.17',
+    };
+
+    const diagnostics = await python.getDiagnostics(environmentInfo);
+    expect(diagnostics.needsToBeFixed).toBe(false);
+  });
+
+  // envinfo has a special field for reporting Python 3 so no need to check for python 3
```
lucasbento: Should this be removed?

molant: I left it for future reference in case someone wondered why we weren't testing out of range. I'm OK either way 😊

lucasbento: I see!

PR #1042, opened 2020-03-09
fix(android): Incorrect mapping of svg files

msand:
https://github.com/react-native-community/react-native-svg/issues/1306
https://github.com/facebook/react-native/pull/28266
https://github.com/facebook/metro/pull/529

Summary:
---------

Fix mapping of svg files. They're currently bundled to the drawable folder incorrectly.

Test Plan:
----------

LocalSvg.js
```jsx
import React, {useState, useEffect} from 'react';
import {SvgCss} from 'react-native-svg';
import loadLocalResource from 'react-native-local-resource';

export function LocalSvg({asset, ...rest}) {
  const [xml, setXml] = useState(null);
  useEffect(() => {
    loadLocalResource(asset).then(setXml);
  }, [asset]);
  return <SvgCss xml={xml} {...rest} />;
}

```

App.js
```jsx
import React from 'react';
import {LocalSvg} from './LocalSvg';
export default () => <LocalSvg asset={require('./test.svg')} />;
```

test.svg: any svg content

### DIFF
--- a/packages/cli/src/commands/bundle/assetPathUtils.ts
+++ b/packages/cli/src/commands/bundle/assetPathUtils.ts
@@ -41,7 +41,6 @@ const drawableFileTypes = new Set<string>([
   'jpeg',
   'jpg',
   'png',
-  'svg',
   'webp',
   'xml',
 ]);


msand:
Merging and releasing this enables using .svg files with react-native-svg without any further configuration. @grabbou @thymikee @Esemesek everyone ok with me merging this? 

I would be very surprised if this would break anything, as the current outcome leads to android not even building, unless one adds 
```
android.disableResourceValidation=true
```
to gradle.properties, at which point the files aren't fetchable, nor usable in Image components. I suspect this file extension has been added by mistake.

msand:
Android drawable do not support svg, and never have, so I'm quite certain this has been an unintentional addition. I'll go ahead and merge this. Could we cut a new release soon? Any schedule?

thymikee:
Based on the comment it looks like svg shouldn't be there indeed, but maybe it's some kind of a hack? Maybe it's not necessary now? Hard to tell at the moment. We can include it in 4.x, so it would be available in RN 0.62, and see if it breaks anywhere.

Next time, please don't just merge without approval. 

cc @Krizzu @dulmandakh

msand:
Ah great, sounds good.

Oh sorry if I stepped on anyones toes. Was quite certain it'd be alright to merge, and if not, revert it.

msand:
Did this while half asleep at 4am, quite an ass move in retrospect, sorry for any inconvenience.

thymikee:
Cool. I hope we can introduce this change as it seems to make sense. I'd just like more eyes on this :). Also, we've improved merging access to code owners only, so we're safer now 👍

msand:
Yeah, I was actually a bit surprised to have merge rights, as I haven't contributed to this package before, in my sleepy state somehow assumed it was normal react-native-community practise.

PR #1043, opened 2020-03-09
chore: remove haste and rnpm

thymikee:
Summary:
---------

- Haste is no longer used at React Native core, so removing anything Haste-related
- `rnpm` is deprecated since v2, so it's about time to remove it.
- remove `--platform` flag from `start` as it was not used anyway

This forces Windows platform to adapt to new config format.

There's no rush merging that, but I think we should be pretty safe by now.

Test Plan:
----------

Updated tests, but please make sure to test it thoroughly

### DIFF
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -461,14 +461,6 @@ Specify any additional asset extensions to be used by the packager
 
 Specify any additional source extensions to be used by the packager
 
-#### `--platforms [list]`
-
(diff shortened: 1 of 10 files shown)


thymikee:
cc @brentvatne @acoates-ms as Expo and RNWindows may be affected

grabbou:
Nice! I think this is similar to some of the work I've done on `next` branch. Let's keep merging that to `master`.

PR #1044, opened 2020-03-10
fix(cli-types): publish declaration files in npm package

Ehesp:
Summary:
---------

Currently external plugin authors using TS cannot access the `cli-types` declaration files as they're not published to NPM. This change removes the file exclusion allowing them to be published.

cc @thymikee as discussed on Discord

### DIFF
--- a/packages/cli-types/package.json
+++ b/packages/cli-types/package.json
@@ -7,7 +7,6 @@
   },
   "files": [
     "build",
-    "!*.d.ts",
     "!*.map"
   ],
   "types": "build/index.d.ts",


thymikee:
I'm ok with this (the tradeoff is adding 13.6kB to the download size), but maybe we can make it better with https://github.com/Microsoft/dts-gen, what do you think?

Ehesp:
Looks like that tool only generates declaration files from JS, whereas we already have it all types out.

thymikee:
cc @Esemesek if you have better ideas :)

Ehesp:
Thanks for merging - is it possible to get a release? :)

thymikee:
Yup, will followup as soon as I can

thymikee:
Published v4.3

PR #1057, opened 2020-03-18
fix: custom root or calling Gradle outside of project directory

### LINKED ISSUE #405: v1.9.6 breaks NPM install
## Environment
```
info
    System:
      OS: macOS 10.14.5
      CPU: (4) x64 Intel(R) Core(TM) i5-7600K CPU @ 3.80GHz
      Memory: 1.32 GB / 24.00 GB
      Shell: 5.3 - /bin/zsh
    Binaries:
      Node: 10.14.1 - ~/.nvm/versions/node/v10.14.1/bin/node
      Yarn: 1.12.3 - /usr/local/bin/yarn
      npm: 6.9.0 - ~/.nvm/versions/node/v10.14.1/bin/npm
    SDKs:
      iOS SDK:
        Platforms: iOS 12.2, macOS 10.14, tvOS 12.2, watchOS 5.2
      Android SDK:
        API Levels: 28
        Build Tools: 28.0.3
    IDEs:
      Android Studio: 3.3 AI-182.5107.16.33.5314842
      Xcode: 10.2.1/10E1001 - /usr/bin/xcodebuild
    npmPackages:
      react-native: 0.59.8 => 0.59.8
    npmGlobalPackages:
      react-native-cli: 2.0.1
```
## Description

Cannot install library since it fails during NPM install. v1.9.5 works fine.

## Reproducible Demo

v1.9.6 failure:
```
npm i @react-native-community/cli@v1.9.6
npm WARN react-native@0.59.8 requires a peer of react@16.8.3 but none is installed. You must install peer dependencies yourself.

npm ERR! path /Users/dkozma/Projects/AwesomeProject/node_modules/@react-native-community/cli/build/bin.js
npm ERR! code ENOENT
npm ERR! errno -2
npm ERR! syscall chmod
npm ERR! enoent ENOENT: no such file or directory, chmod '/Users/dkozma/Projects/AwesomeProject/node_modules/@react-native-community/cli/build/bin.js'
npm ERR! enoent This is related to npm not being able to find a file.
npm ERR! enoent

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/dkozma/.npm/_logs/2019-05-27T13_08_43_889Z-debug.log
```

compare to v1.9.5 output:
```
npm i @react-native-community/cli@v1.9.5
npm notice created a lockfile as package-lock.json. You should commit this file.
npm WARN react-native@0.59.8 requires a peer of react@16.8.3 but none is installed. You must install peer dependencies yourself.

+ @react-native-community/cli@1.9.5
added 18 packages from 21 contributors and audited 59869 packages in 2.515s
found 22 low severity vulnerabilities
  run `npm audit fix` to fix them, or `npm audit` for details
```

### LINKED ISSUE #793: Gradle sync fail - ERROR: Text must not be null or empty

### LINKED ISSUE #804: Android: latest release broke support custom react-native Root

### LINKED ISSUE #995: Cannot find module 'react-native/cli'
## Environment

<details>
<summary>react-native info output</summary>

```
λ yarn exec react-native info
yarn exec v1.22.0
System:
    OS: macOS Mojave 10.14.6
    CPU: (4) x64 Intel(R) Core(TM) i5-4308U CPU @ 2.80GHz
    Memory: 52.27 MB / 16.00 GB
    Shell: 3.0.2 - /usr/local/bin/fish
  Binaries:
    Node: 12.16.0 - /var/folders/z8/xq1r3qcd7szbpx1_77xsgpqc0000gn/T/yarn--1582222260147-0.32330923283802093/node
    Yarn: 1.22.0 - /var/folders/z8/xq1r3qcd7szbpx1_77xsgpqc0000gn/T/yarn--1582222260147-0.32330923283802093/yarn
    npm: 6.13.4 - /usr/local/bin/npm
  SDKs:
    iOS SDK:
      Platforms: iOS 12.4, macOS 10.14, tvOS 12.4, watchOS 5.3
    Android SDK:
      API Levels: 23, 25, 26, 27, 28
      Build Tools: 23.0.1, 25.0.2, 25.0.3, 26.0.1, 26.0.2, 27.0.2, 27.0.3, 28.0.3
      System Images: android-27 | Google APIs Intel x86 Atom
  IDEs:
    Android Studio: 3.1 AI-173.4720617
    Xcode: 10.3/10G8 - /usr/bin/xcodebuild
  npmPackages:
    react: ~16.9.0 => 16.9.0 
    react-native: ~0.61.5 => 0.61.5
✨  Done in 4.79s.
```

</details>

## Description

Hi folks, this issue is very closely related to #406, but I decided to open a new issue for it since it may require a different solution altogether, and that issue is closed.

In short, `@react-native-community/cli-platform-android` is breaking our builds on Appcenter (cc @kinigitbyday https://github.com/react-native-community/cli/issues/406#issuecomment-588406396 ), using the following versions:

```
λ yarn list --pattern @react-native-community/cli
yarn list v1.22.0
├─ @react-native-community/cli-debugger-ui@3.0.0
├─ @react-native-community/cli-platform-android@3.1.4
├─ @react-native-community/cli-platform-ios@3.2.0
├─ @react-native-community/cli-tools@3.0.0
├─ @react-native-community/cli-types@3.0.0
└─ @react-native-community/cli@3.2.0
```

The exact failure logged by Appcenter (aka Azure DevOps) is the following:

```
[command]/Users/runner/runners/2.165.0/work/1/s/app/android/gradlew -DAPPCENTER_KEYSTORE_PASSWORD=*** -DMOBILECENTER_KEYSTORE_PASSWORD=*** -DAPPCENTER_KEY_ALIAS=*** -DMOBILECENTER_KEY_ALIAS=*** -DAPPCENTER_KEY_PASSWORD=*** -DMOBILECENTER_KEY_PASSWORD=*** -DAPPCENTER_BUILD_VERSION=14 -DMOBILECENTER_BUILD_VERSION=14 -p app/android clean :app:assembleRelease :app:bundleRelease
Starting a Gradle Daemon, 1 incompatible Daemon could not be reused, use --status for details
Gradle 5.5
:ReactNative:Unexpected empty result of running '[node, -e, console.log(require('react-native/cli').bin);]' command.
:ReactNative:Running '[node, -e, console.log(require('react-native/cli').bin);]' command failed.

* Where:
Script '/Users/runner/runners/2.165.0/work/1/s/app/node_modules/@react-native-community/cli-platform-android/native_modules.gradle' line: 170

* What went wrong:
> internal/modules/cjs/loader.js:984  throw err;  ^Error: Cannot find module 'react-native/cli'Require stack:- /Users/runner/runners/2.165.0/work/1/s/[eval]    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:981:15)    at Function.Module._load (internal/modules/cjs/loader.js:863:27)    at Module.require (internal/modules/cjs/loader.js:1043:19)    at require (internal/modules/cjs/helpers.js:77:18)    at [eval]:1:13    at Script.runInThisContext (vm.js:120:20)    at Object.runInThisContext (vm.js:311:38)    at Object.<anonymous> ([eval]-wrapper:10:26)    at Module._compile (internal/modules/cjs/loader.js:1157:30)    at evalScript (internal/process/execution.js:94:25) {  code: 'MODULE_NOT_FOUND',  requireStack: [ '/Users/runner/runners/2.165.0/work/1/s/[eval]' ]}
```

There is a comment in the affected Gradle file `Sometimes Gradle can be called outside of the JavaScript hierarchy (-p flag)` which seems related. Appcenter calls this with `-p app/android`.

We have our RN app in a directory called `app/` in our repo, and you can see from the stacktrace that the eval is executed from the root directory of the repo instead of within `app/` or `app/android` as I would expect.

I have tried various versions of `@react-native-community/cli-platform-android`, and it seems that the issue was introduced by this commit which removes `npx`, which was released in `3.1.0`: https://github.com/react-native-community/cli/commit/b4b08545d3ccec3d2d4a545f029496e0358de49a

We hotfixed this for now by pinning `@react-native-community/cli-platform-android` to the latest version before `3.1.0` by adding the following to our `package.json`:

```
  "resolutions": {
    "@react-native-community/cli-platform-android": "~3.0.0"
  }
```

Hope this helps, let me know if I can help test any solution to this :)
Thank you!

grabbou:
TL;DR:
---------

Fixes #995 
Fixes #405
Fixes #793
Fixes #804

Improves #824 that was reverted via #852 because it caused regressions (see case 2)

Context:
---------

React Native CLI is flexible and makes no assumptions where and how you structure your project folders (at least that's our ideal scenario). However, Gradle and CocoaPods need to call the React Native CLI to read the configuration, list of all dependencies to link and get some relative paths. In order to resolve where the React Native CLI currently is, Gradle and CocoaPods ask `node` and its `require.resolve` strategy to locate where `react-native` is located.

In 99% cases, Gradle is called from the project root directory or a folder within the project root, such as `android` (when you run `./android/gradlew`). That makes `require.resolve` to work correctly, by locating the `react-native` in the closest `node_modules` traversing up the tree.

However, there's that 1% of use cases when this is not true. For example, Gradle can be sometimes given an absolute path to a project to build (via `-p` flag), which will result in a `cwd` that is different than the project root and outside of JavaScript hierarchy. This is very common for tools such as `fastlane` or `app center` and happens more often when doing a release of the apk rather than installing on a simulator.

**Case 1**

For example, calling: `<path_to_gradle>/.gradlew -p Repositories/HelloWorld/android` from `/home/grabbou` will result in the #995 error, because `cwd` is `/home/grabbou` and there's no `/home/grabbou/node_modules/react-native` to find.

**Case 2**

The previous `jsAppDir` workaround contributed by @thymikee wasn't optimal either, as it was inferring `root` of a project from the directory where `node_modules` are. In a monorepo, the dependency will be hoisted, which will result in the `root` being set incorrectly.

At the time of reverting this feature in #852, we didn't know why it was causing us troubles. However, recently, thanks to additional reproductions provided by the community, we were able to understand why that particular issue was happening.

For example, calling `<path_to_gradle>/.gradlew -p Repositories/HelloWorld/packages/app-a/android` with the aforementioned workaround would set `root` to `/users/grabbou/Repositories/HelloWorld` because React Native dependency has been hoisted to `/users/grabbou/Repositories/HelloWorld/node_modules/`. In reality, we want the actual `/users/grabbou/Repositories/HelloWorld/packages/app-a`.

Solution:
---------

The solution turned out to be simple and as I initially expected, was a matter of finding the right Gradle variable to read from.

Reading `rootProject.projectDir` and using that as a `root` always points out correctly to the folder that is either inferred automatically or set via `-p` flag.

Test Plan:
----------

Run `gradle` with `-p` flag outside of JavaScript hierarchy.


packages/platform-android/native_modules.gradle
```diff
@@ -237,12 +236,19 @@ class ReactNativeModules {
   }
 }
 
+
+/*
+ * Sometimes Gradle can be called outside of JavaScript hierarchy. Detect the directory
+ * where build files of an active project are located.
+ */
+def projectRoot = rootProject.projectDir
```
grabbou: For context: `rootProject` is what you give a name in `settings.gradle` - it's the umbrella in case you have different smaller sub-projects

https://github.com/facebook/react-native/blob/master/template/android/settings.gradle#L1

For `autolinking` and CLI, it doesn't matter - we always want the main project. But you will notice that `applyNativeModulesBuildGradle` expects `project` argument to handle such case.

grabbou:
After this PR, I'll provide a similar workaround for iOS. Just in case somebody ends up calling `pod` with `--project-directory`.

grabbou:
Added `e2e` tests to make sure this particular use casework. It's a good starting point for the future work around CLI and interactions on a "living" project.

I would say let's merge it and it if ends up being flaky, I am volunteering to fix it ASAP. Don't want to do premature optimisation tho.

__e2e__/root.test.ts
```diff
@@ -0,0 +1,64 @@
...
+  }
+});
+
+beforeEach(() => {
+  // Clean up folder and re-create a new project
+  cleanup(cwd);
+  writeFiles(cwd, {});
+
+  // Initialise React Native project
+  runCli(cwd, ['init', 'TestProject']);
```
thymikee: Let's move init to beforeAll, so it get downloaded only once. We don't really modify files, so there's no need to cleanup between tests.

jest/helpers.ts
```diff
@@ -11,39 +11,43 @@ import {Writable} from 'readable-stream';
...
-  dir: string;
-  args: string[] | undefined;
-  options: RunOptions;
-};
-
-export function run(
+/**
+ * Helper function to run CLI command in a given folder
+ */
+export function runCli(
```
thymikee: can we rename to `runCLI`?

jest/helpers.ts
```diff
@@ -11,39 +11,43 @@ import {Writable} from 'readable-stream';
...
   dir: string,
   args?: string[],
-  options: RunOptions = {
+  options: RunCliOptions = {
```
thymikee: can we rename to `RunCLIOptions`?

__e2e__/root.test.ts
grabbou: Done

jest/helpers.ts
```diff
@@ -156,25 +172,24 @@ function getCliArguments({dir, args, options}: CLIArgs) {
...
+  cmd: string,
+  options: SpawnOptions,
   result: {[key: string]: any},
-  dir: string,
   args: string[] | undefined,
 ) {
   if (!options.expectedFailure && result.code !== 0) {
-    console.log(`Running CLI failed for unexpected reason. Here's more info:
-${chalk.bold('dir:')}    ${dir}
+    console.log(`Running failed for unexpected reason. Here's more info:
```
thymikee:
```suggestion
    console.log(`Running ${cmd} command failed for unexpected reason. Here's more info:
```

jest/helpers.ts
```diff
@@ -156,25 +172,24 @@ function getCliArguments({dir, args, options}: CLIArgs) {
...
+${chalk.bold('cmd:')}    ${cmd}    
+${chalk.bold('options:')}${JSON.stringify(options)}
```
thymikee: some space wouldn't be nice here, no?

jest/helpers.ts
grabbou: No strong feelings - I wanted to keep it aligned with `runCli` command. What you think? Do we have any kind of naming policy in place to help here?

grabbou: Guess it can be just `RunOptions` as before.

__e2e__/root.test.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  // Link CLI to the project
+  for (const pkg of [
+    '@react-native-community/cli-platform-ios',
+    '@react-native-community/cli-platform-android',
+  ]) {
+    spawnScript('yarn', ['link', pkg], {
```
thymikee: you can run `yarn link` with multiple packages, no need for a for-loop here

__e2e__/root.test.ts
```diff
@@ -0,0 +1,62 @@
...
+  cleanup,
+  writeFiles,
+} from '../jest/helpers';
+
+const cwd = getTempDirectory('test_different_roots');
+
+beforeAll(() => {
+  // Register all packages to be linked
+  for (const pkg of ['platform-ios', 'platform-android']) {
+    spawnScript('yarn', ['link'], {
```
thymikee: same as in line 33, no need for a for-loop

__e2e__/root.test.ts
```diff
@@ -0,0 +1,62 @@
...
+  /*
+   * Grab absolute path to Gradle wrapper. The fact that we are using
+   * a wrapper from the project is just a convinience to avoid installing
+   * Gradle globally
+   */
+  const gradleWrapper = path.join(androidProjectRoot, 'gradlew');
+
+  // Execute `gradle` with `-p` flag and `cwd` outside of project hierarchy
+  const {stdout} = spawnScript(gradleWrapper, ['-p', androidProjectRoot], {
+    cwd: '/',
```
thymikee: Do we need to set cwd here? I'd expect a default value would be sufficient – this way we could make the `options` argument of `spawnScript` optional with `{}` as default value, as it used to be

grabbou: Truth is - we never want to rely on `default` value in our tests and that's why I decided to go with explicit `cwd`. 

In this case, I just want to make sure and be 100% explicit that `cwd` is somewhere else than the project - for example, at the root.

__e2e__/root.test.ts
grabbou: In this case, you need a `for-loop` because you call `yarn link` in package you want to link (see `cwd` changing)

__e2e__/root.test.ts
grabbou: > Truth is - we never want to rely on default value in our tests and that's why I decided to go with explicit cwd.

The reason why I believe it's better to require `cwd` is simple - `cwd` defaults to the location where the React Native CLI source files are located (or where the tests are being run from).

Conceptually, I think of `e2e` tests as something run in a totally separate environment that is close to what users are doing in real life. I feel like setting `cwd` in `spawnScript` (which I consider advanced use case, where `runCLI` is not enough) is just a way to tell "watch out, you're going into advanced territory - think about what you are doing".

__e2e__/root.test.ts
grabbou: nice, didn't notice `...` in docs!

lylest:
can you guys give me much simpler explanations please I beg you not everyone is Gradle guru I am really stuck I can't start my react native project and it just saying build fails

vladp:
@thymikee   Thank  you. Is this fixed in 4.6.3 release ?  (asking because I am moving from rn 0.60 to 0.62.2 and running into https://github.com/react-native-community/cli/issues/995 )

PR #1058, opened 2020-03-18
Update to mkdirp

scottwio:
** Fix Prototype pollution vulnerability**

Summary:
---------
A small PR that forces mkdrip to a release that contains a fix for the prototype pollution venerability can read more about it here  https://github.com/HoLyVieR/prototype-pollution-nsec18/blob/master/paper/JavaScript_prototype_pollution_attack_in_NodeJS.pdf

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,13 +50,13 @@
     "lerna": "^3.18.4",
     "metro-memory-fs": "^0.57.0",
     "micromatch": "^3.1.10",
-    "mkdirp": "^0.5.1",
+    "mkdirp": "^0.5.3",
     "rimraf": "^3.0.2",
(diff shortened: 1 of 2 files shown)


Esemesek:
I think this was already fixed in https://github.com/react-native-community/cli/pull/1050

Esemesek:
The vulnerable version was still present in the `yarn.lock` :O

scottwio:
I don't think 1.2.2 is actually high enough needs to be 1.2.3 onwards

Screen Shot 2020-03-18 at 11 40 07

thymikee:
This is internal dependency, so it doesn't surface to users.

Issue #528, opened 2019-07-12
warn and exit when `xcrun` is not installed


sw-yx:
## Description

i have cocoapods installed
![image](https://user-images.githubusercontent.com/6764957/61161346-6b7d3a80-a4d1-11e9-8c51-519f32a277b0.png)

but the cli thinks i dont
![image](https://user-images.githubusercontent.com/6764957/61161361-7932c000-a4d1-11e9-9651-8f8b0945a851.png)

i think this is because i have an implicit dependency on xcode and i didnt know i needed it

![image](https://user-images.githubusercontent.com/6764957/61161401-92d40780-a4d1-11e9-92a4-5dfbb555eda3.png)

we should detect this first and prompt

sw-yx:
furthermore, i've been using xcode 11 beta, and it looks like i need xcode 10 so i am downgrading.

sw-yx:
output of `git -v`:

![image](https://user-images.githubusercontent.com/6764957/61161871-eb0c0900-a4d3-11e9-9c6c-fa670f6f2d6b.png)

thymikee:
We can print the error output when it's failing. Would that help?

sw-yx:
yes! would be better than what i suggested.

grabbou:
We should also add this to `doctor`.

PR #1061, opened 2020-03-18
feat: run `pod repo update` when `pod install` failed

### LINKED ISSUE #1060: Automatically handle CocoaPods being out of date while creating a new project
When initing a new project and having out-dated CocoaPods repository, React Native `init` can fail with the following output:

```
✔ Downloading template
✔ Copying template
✔ Processing template
⠇ Installing CocoaPods dependencies (this may take a few minutes)Analyzing dependencies
Fetching podspec for `Folly` from `../node_modules/react-native/third-party-podspecs/Folly.podspec`
Fetching podspec for `glog` from `../node_modules/react-native/third-party-podspecs/glog.podspec`
[!] CocoaPods could not find compatible versions for pod "FlipperKit":
  In Podfile:
    FlipperKit (~> 0.33.1)

None of your spec sources contain a spec satisfying the dependency: `FlipperKit (~> 0.33.1)`.

You have either:
 * out-of-date source repos which you can update with `pod repo update` or with `pod install --repo-update`.
 * mistyped the name or version.
 * not added the source repo that hosts the Podspec to your Podfile.

✖ Installing CocoaPods dependencies (this may take a few minutes)
error Error: Failed to install CocoaPods dependencies for iOS project, which is required by this template.
Please try again manually: "cd ./RNTestProject/ios && pod install".
```

We should try/catch this and handle as recommended, that is run `pod repo update` and repeat.

grabbou:
Summary:
---------

Fixes #1060 

Test Plan:
----------

Run `pod install` on outdated repo. Also includes a mechanism to prevent loops in case `pod repo update` didn't fix the error (e.g. invalid version specified).

### DIFF
--- a/packages/cli/src/tools/installPods.ts
+++ b/packages/cli/src/tools/installPods.ts
@@ -15,7 +15,44 @@ type PromptCocoaPodsInstallation = {
   promptQuestion: string;
 };
 
-async function updatePods(loader: ora.Ora) {
+async function runPodInstall(
+  loader: ora.Ora,
+  projectName: string,
+  shouldHandleRepoUpdate: boolean = true,
+) {
+  try {
+    loader.start(
+      `Installing CocoaPods dependencies ${chalk.dim(
+        '(this may take a few minutes)',
+      )}`,
+    );
+    await execa('pod', ['install']);
+  } catch (error) {
+    // "pod" command outputs errors to stdout (at least some of them)
+    const stderr = error.stderr || error.stdout;
+
+    /**
+     * If CocoaPods failed due to repo being out of date, it will
+     * include the update command in the error message.
+     *
+     * `shouldHandleRepoUpdate` will be set to `false` to
+     * prevent infinite loop (unlikely scenario)
+     */
+    if (stderr.includes('pod repo update') && shouldHandleRepoUpdate) {
+      await runPodUpdate(loader);
+      await runPodInstall(loader, projectName, false);
+    } else {
+      loader.fail();
+      throw new Error(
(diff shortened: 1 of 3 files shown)


grabbou:
With this PR applied, I was able to successfully continue manual tests of the upcoming release.

packages/cli/src/tools/installPods.ts
```diff
@@ -15,7 +15,44 @@ type PromptCocoaPodsInstallation = {
...
+     *
+     * `shouldHandleRepoUpdate` will be set to `false` to
+     * prevent infinite loop (unlikely scenario)
+     */
+    if (stderr.includes('pod repo update') && shouldHandleRepoUpdate) {
+      await runPodUpdate(loader);
+      await runPodInstall(loader, projectName, false);
+    } else {
+      loader.fail();
+      throw new Error(
```
thymikee: Let's use `CLIError` and pass the original `error` so that [issues like this one](https://github.com/react-native-community/cli/issues/528) are easier to deal with. Would be cool to do the same for `runPodUpdate` function (line 68)

grabbou: Was just copying and pasting the existing code and actually started thinking why we no longer use CLIError.

grabbou: Turns out this will take more than should be in scope of this PR and I guess we should track it separately.

CLIError will inline the entire message into a single line + we already `logger.error` here (in other CLI commands, we don't do that as we have global try-catch). 

I would be a bit nervous about changing it here due to potential side effects. What you think of setting up a follow up issue?

thymikee: I'm good with a followup

PR #1065, opened 2020-03-23
feat: install JDK automatically if missing

### LINKED ISSUE #980: Check for JDK and install if missing
Part of #976 

# Validation

~~Get the output of running `javac -version` and parse it~~

Use `envinfo` and look for `Languages`.

# Installation

## Windows

Version 8 or later is recommended in the docs. I'd rather use OpenJDK 9 or later if it works as it doesn't require admin privileges to get installed:

> Staring JDK9, the [OpenJDK](http://jdk.java.net/9) provides binary tar ball (zipped) executable JDK package that can unpack in Windows in any directory. But if you still want Oracle Java SE JDK, then you still use above instruction to get around installing without admin rights.

Installation should go under `%LOCALAPPDATA%/jdkVERSION` or something like that.

We should also set the right [environment variables](https://stackoverflow.com/a/52531093/414145):

* `PATH`: Add bin folder of the JDK
* `JAVA_HOME`: New one pointing to the JDK installation path

To update/add environment variables see #976

# macOS

TBD

# Linux

TBD

molant:
Summary:
---------

Install JDK@11 on Windows if missing. I went with 11 because from my tests it was the latest one that was working with Gradle. After it they were all giving me problems.
The documentation mentions that is has to be 8 or later and with that version local installation for the user is painful as there is not a .zip available.

I tried to install it on mac as well but all the options I found in homebrew prompted for a password 😔

You can see it in action (+ python) in this video:

[YouTube vide to JDK installation](https://youtu.be/tRtBRMxLh6s)

Fix #980

### DIFF
--- a/packages/cli-types/src/node-stream-zip.d.ts
+++ b/packages/cli-types/src/node-stream-zip.d.ts
@@ -1,0 +1,66 @@
+// https://github.com/antelle/node-stream-zip/issues/36#issuecomment-596249657
+declare module 'node-stream-zip' {
+  import {Stream} from 'stream';
+
+  interface StreamZipOptions {
+    /**
+     * File to read
+     */
+    file: string;
+    /**
+     * You will be able to work with entries inside zip archive, otherwise the only way to access them is entry event
+     *
+     * default: true
+     */
+    storeEntries?: boolean;
+    /**
+     * By default, entry name is checked for malicious characters, like ../ or c:\123, pass this flag to disable validation errors
+     *
+     * default: false
+     */
+    skipEntryNameValidation?: boolean;
+    /**
+     * Undocumented adjustment of chunk size
+     *
+     * default: automatic
+     */
+    chunkSize?: number;
+  }
+
+  class ZipEntry {
+    name: string;
+    isDirectory: boolean;
+    isFile: boolean;
+    comment: string;
+    size: number;
+  }
+
+  class StreamZip {
+    constructor(config: StreamZipOptions);
+
+    on(event: 'ready', handler: () => void): void;
+
+    entry(entry: string): ZipEntry;
+    entries(): ZipEntry[];
+
+    entriesCount: number;
+
+    stream(
+      entry: string,
+      callback: (err: any | null, stream?: Stream) => void,
+    ): void;
+    entryDataSync(entry: string): Buffer;
+    openEntry(
+      entry: string,
+      callback: (err: any | null, entry?: ZipEntry) => void,
+      sync: boolean,
+    ): void;
+    extract(
+      entry: string | null,
+      outPath: string,
+      callback: (err?: any) => void,
+    ): void;
+    close(callback?: (err?: any) => void): void;
+  }
+  export = StreamZip;
+}
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -63,6 +63,7 @@
     "shell-quote": "1.6.1",
     "strip-ansi": "^5.2.0",
     "sudo-prompt": "^9.0.0",
+    "unzipper": "^0.10.10",
     "wcwidth": "^1.0.1",
     "ws": "^1.1.0"
   },
@@ -83,6 +84,7 @@
     "@types/mkdirp": "^0.5.2",
     "@types/pretty-format": "^24.3.0",
     "@types/semver": "^6.0.2",
+    "@types/unzipper": "^0.10.2",
     "@types/wcwidth": "^1.0.0",
     "@types/ws": "^6.0.3",
     "slash": "^3.0.0",
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -84,7 +84,6 @@
     "@types/mkdirp": "^0.5.2",
     "@types/pretty-format": "^24.3.0",
     "@types/semver": "^6.0.2",
-    "@types/unzipper": "^0.10.2",
     "@types/wcwidth": "^1.0.0",
     "@types/ws": "^6.0.3",
     "slash": "^3.0.0",
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -55,6 +55,7 @@
     "metro-react-native-babel-transformer": "^0.58.0",
     "minimist": "^1.2.0",
     "mkdirp": "^0.5.1",
+    "node-stream-zip": "^1.9.1",
     "open": "^6.2.0",
     "ora": "^3.4.0",
     "pretty-format": "^25.2.0",
@@ -63,7 +64,6 @@
     "shell-quote": "1.6.1",
     "strip-ansi": "^5.2.0",
     "sudo-prompt": "^9.0.0",
-    "unzipper": "^0.10.10",
     "wcwidth": "^1.0.1",
     "ws": "^1.1.0"
   },
--- a/packages/cli/src/commands/doctor/healthchecks/__tests__/jdk.test.ts
+++ b/packages/cli/src/commands/doctor/healthchecks/__tests__/jdk.test.ts
@@ -1,0 +1,81 @@
+import execa from 'execa';
+import jdk from '../jdk';
+import getEnvironmentInfo from '../../../../tools/envinfo';
+import {EnvironmentInfo} from '../../types';
+import {NoopLoader} from '../../../../tools/loader';
+import * as common from '../common';
+import * as u from '../../../../tools/unzip';
+
+jest.mock('execa', () => jest.fn());
+
+const mockFetchToTemp = jest.fn();
+jest.mock('@react-native-community/cli-tools', () => {
+  return {
+    fetchToTemp: mockFetchToTemp,
+  };
+});
+
+const logSpy = jest.spyOn(common, 'logManualInstallation');
+
+describe('jdk', () => {
+  let environmentInfo: EnvironmentInfo;
+
+  beforeAll(async () => {
+    environmentInfo = await getEnvironmentInfo();
+  }, 15000);
+
+  afterEach(() => {
+    jest.resetAllMocks();
+  });
+
+  it('returns a message if JDK is not installed', async () => {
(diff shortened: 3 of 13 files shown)


packages/cli/package.json
```diff
@@ -63,6 +63,7 @@
     "shell-quote": "1.6.1",
     "strip-ansi": "^5.2.0",
     "sudo-prompt": "^9.0.0",
+    "unzipper": "^0.10.10",
```
thymikee: Can we unzip using Node's `zlib` https://nodejs.org/api/zlib.html? `unzipper` brings extra 1.4MB to node_modules, I'd like to avoid that

packages/cli/src/commands/doctor/healthchecks/__tests__/jdk.test.ts
```diff
@@ -0,0 +1,81 @@
+import execa from 'execa';
+import jdk from '../jdk';
+import getEnvironmentInfo from '../../../../tools/envinfo';
+import {EnvironmentInfo} from '../../types';
+import {NoopLoader} from '../../../../tools/loader';
+import * as common from '../common';
+import * as u from '../../../../tools/unzip';
```
thymikee: Please rename to `unzip`

packages/cli/src/commands/doctor/healthchecks/jdk.ts
```diff
@@ -0,0 +1,69 @@
...
+        'JDK installed successfully. Please restart your shell to see the changes',
+      );
+    } catch (e) {
+      loader.fail(e);
+    }
+  },
+  runAutomaticFix: async () => {
+    logManualInstallation({
+      healthcheck: 'JDK',
+      url: 'http://jdk.java.net/11/',
```
thymikee: official docs instruct to use OpenJDK, can we link there? https://openjdk.java.net/

packages/cli/src/commands/doctor/healthchecks/index.ts
```diff
@@ -1,5 +1,6 @@
 import nodeJS from './nodeJS';
 import {yarn, npm} from './packageManagers';
+import java from './jdk';
```
thymikee: let's stick with `jdk`, that's what we install :) 
```suggestion
import jdk from './jdk';
```

packages/cli/src/commands/doctor/healthchecks/index.ts
```diff
@@ -34,6 +35,7 @@ export const getHealthchecks = ({contributor}: Options): Healthchecks => ({
   android: {
     label: 'Android',
     healthchecks: [
+      java,
```
thymikee:
```suggestion
      jdk,
```

packages/cli/src/commands/doctor/healthchecks/jdk.ts
```diff
@@ -0,0 +1,69 @@
...
+      // Installing JDK 11 because later versions seem to cause issues with gradle at the moment
+      const installerUrl =
+        'https://download.java.net/java/GA/jdk11/9/GPL/openjdk-11.0.2_windows-x64_bin.zip';
+      const installPath = process.env.LOCALAPPDATA || ''; // The zip is in a folder `jdk-11.02` so it can be unzipped directly there
+
+      loader.start(
+        `Downloading JDK 11 from "${installerUrl}" (please be patient)`,
+      );
+
+      const installer = await fetchToTemp(installerUrl);
```
thymikee: Can we make sure to clean up after download and unzip?

packages/cli/package.json
molant: I'll give it a try but I do worry about the size of some of those zips and making things correctly.

thymikee: Maybe there's a lighter alternative?

molant:
@thymikee I've address your comments minus the `unzip` package. I'll look into that now.

packages/cli/package.json
molant: How did you get the computed size? npm only reports 51.6kB which I imagine does not take into account dependencies:
image

I have to support node 8, right? It looks like with node 10 I could use [`stream.pipeline`](https://nodejs.org/api/stream.html#stream_stream_pipeline_source_transforms_destination_callback) and have a clean implementation but I remember seeing a comment saying that node 8 will get drop at the same time as RN.

thymikee: It has some heavy deps. Check this site: https://packagephobia.now.sh/result?p=unzipper

thymikee: Node 8 will be dropped in the next RN version, since Metro 0.59 dropped it

molant: I've replaced it with [`node-stream-zip`](https://packagephobia.now.sh/result?p=node-stream-zip) that does not have any [external dependency](https://packagephobia.now.sh/result?p=node-stream-zip).

Thanks for the site!

packages/cli/package.json
```diff
@@ -83,6 +84,7 @@
     "@types/mkdirp": "^0.5.2",
     "@types/pretty-format": "^24.3.0",
     "@types/semver": "^6.0.2",
+    "@types/unzipper": "^0.10.2",
```
thymikee, line 87: this is unused now :)

packages/cli/src/commands/doctor/healthchecks/jdk.ts
```diff
@@ -0,0 +1,72 @@
...
+        `Downloading JDK 11 from "${installerUrl}" (this may take a few minutes)`,
+      );
+
+      const installer = await fetchToTemp(installerUrl);
+
+      loader.text = `Installing JDK in "${installPath}"`;
+
+      await unzip(installer, installPath);
+
+      await deleteFile(installer);
```
thymikee, line 50: Can `unzip` throw an error? In such case it won't delete the file. Instead, we could add `finally` block to this try/catch and call delete there

packages/cli/src/tools/deleteFile.ts
```diff
@@ -0,0 +1,4 @@
+import {unlink} from 'fs';
+import {promisify} from 'util';
+
+export const deleteFile = promisify(unlink);
```
thymikee, line 4: We can use `fs.unlinkSync` and be done, without creating new files, why making it async?

molant: It's a lot easier to mock if it's in a separate file. Also I'll be using this again for the android dependencies.

thymikee: Fair call. Still doesn't need to be async, right? :)

packages/cli/src/commands/doctor/healthchecks/jdk.ts
molant: I'm going to pull some of the changes I'm doing in `install-android` here because I'm moving this around and  might as well take a look now.

packages/cli/src/tools/deleteFile.ts
molant: Doesn't have to, but I think it's better if it is.
This is a blocking call and we don't know how long it will take. We have a spinner in the UI and if it stops for whatever reason it will provide a bad experience to the user.

Another reason (and this is a potential) is if fixes start running in parallel to speed up the process this will be blocking everything.

The difference for any "consumer" of this method is just adding `await` in front of it which I don't think it's a big deal. IMO the pros outweigh the cons, especially with not blocking the UI 😊

packages/cli/package.json
molant: Really need to bring the script to check dependencies 🤦‍♂️

packages/cli/src/commands/doctor/healthchecks/jdk.ts
molant: @thymikee I've moved things a bit around. The context is that I'm reusing this method to install Android Studio and the SDK as well. Can you please take another look?

packages/cli/src/tools/downloadAndUnzip.ts
```diff
@@ -0,0 +1,32 @@
...
+  loader.start(
+    `Downloading ${component} from "${downloadUrl}" (this may take a few minutes)`,
+  );
+
+  const installer = await fetchToTemp(downloadUrl);
+
+  loader.text = `Installing ${component} in "${installPath}"`;
+  await unzip(installer, installPath);
+
+  await deleteFile(installer);
```
thymikee, line 31: Can we wrap it in try/finally? I'd like to make sure we don't leave garbage after unzipping. Hope that makes sense
```suggestion
  loader.text = `Installing ${component} in "${installPath}"`;
  try {
    await unzip(installer, installPath);
  } finally {
    await deleteFile(installer);
  }
```

packages/cli/src/commands/doctor/healthchecks/jdk.ts
thymikee: Specialized function makes sense. I left an additional comment on try/finally, maybe it wasn't so clear :D

packages/cli/src/tools/downloadAndUnzip.ts
molant: Yes, is just that my brain is extra slow these last few days 😓

molant: Hopefully this is the final one 🤞

packages/cli-types/src/node-stream-zip.d.ts
```diff
@@ -0,0 +1,66 @@
+// https://github.com/antelle/node-stream-zip/issues/36#issuecomment-596249657
```
Esemesek, line 1: Maybe it would be worth to contribute those typings to `@DefinitelyTyped`? :P

molant: I had to do some modifications from the code in that comment and the discussion in that thread made me believe they were going to be added eventually in the package.

PR #1072, opened 2020-03-25
chore(breaking): use Node 10 as compile target

wtrocki:
Node 8 is out of the support. 10 and 12 LTS are the latest 2 LTS versions.
Creating PR just to verify if tests will run on Node 10 LTS. I understand that there might be reasons why Node 8 is still used.

List of the supported circle ci images:
https://circleci.com/docs/2.0/circleci-images/#nodejs

## Verification

1. Build passes


thymikee:
`react-native` module is still on Node 8, and the closest release, 0.62, is not dropping it yet: https://github.com/facebook/react-native/blob/0.62-stable/package.json#L11-L13.

I'd say we drop it when `react-native` does.

thymikee (approved):
Thanks, we're finally ready to merge :)

Issue #1054, opened 2020-03-16
Autolinking fails if android/app folder or Xcode project are missing


tido64:
## Environment

```
System:
    OS: macOS Mojave 10.14.6
    CPU: (8) x64 Intel(R) Core(TM) i7-6700HQ CPU @ 2.60GHz
    Memory: 938.13 MB / 16.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 12.16.1 - /usr/local/opt/node@12/bin/node
    Yarn: 1.22.4 - /usr/local/bin/yarn
    npm: 6.13.4 - /usr/local/opt/node@12/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  SDKs:
    iOS SDK:
      Platforms: iOS 13.2, DriverKit 19.0, macOS 10.15, tvOS 13.2, watchOS 6.1
    Android SDK:
      API Levels: 26, 27, 28, 29
      Build Tools: 27.0.3, 28.0.3, 29.0.2, 29.0.3
      System Images: android-28 | Google APIs Intel x86 Atom
      Android NDK: 20.0.5594570
  IDEs:
    Android Studio: 3.5 AI-191.8026.42.35.5900203
    Xcode: 11.3.1/11C504 - /usr/bin/xcodebuild
  npmPackages:
    react: 16.9.0 => 16.9.0
    react-native: 0.61.5 => 0.61.5
  npmGlobalPackages:
    react-native-cli: 2.0.1
```

## Description

In monorepos where packages are split by platform, `react-native config` is not able to populate `projects` because it makes assumptions on the folder structure of the package. For example, given a project like below:

```
my-component
└── packages
    ├── my-component-core
    │   ├── package.json
    │   └── src
    ├── my-component-android
    │   ├── build.gradle
    │   ├── package.json
    │   └── src
    └── my-component-ios
        ├── TestApp.xcworkspace
        ├── package.json
        └── src
```

The output of `react-native config` in `my-component-android`/`-ios` is currently:

```json
{
  "..."
  "project": {
    "ios": null,
    "android": null
  }
}
```

This had no consequences up until 3.0, when autolinking started using it in #768. Autolinking is currently broken for repos with a different folder structure.

It is possible to work around this issue by adding a `react-native.config.js` with the following content:

```js
module.exports = {
  project: {
    ios: {
      project: "ReactTestApp-Dummy.xcodeproj"
    }
  }
};
```

## Reproducible Demo

[...]

PR #1073, opened 2020-03-26
fix(android): Looking for `react-native/cli` sometimes returns `undefined`

tido64:
Summary:
---------

`require('react-native/cli').bin` returns `undefined` on certain setups:

```
% node -e "console.log(require('react-native/cli').bin);"
undefined

% node -e "console.log(require.resolve('react-native/cli'));"
/~/node_modules/react-native/cli.js
```

Test Plan:
----------

All current tests should still pass.

### DIFF
--- a/packages/platform-android/native_modules.gradle
+++ b/packages/platform-android/native_modules.gradle
@@ -217,7 +217,7 @@ class ReactNativeModules {
      *
      * @todo: `fastlane` has been reported to not work too.
      */
-    def cliResolveScript = "console.log(require('react-native/cli').bin);"
+    def cliResolveScript = "console.log(require.resolve('react-native/cli'));"
     String[] nodeCommand = ["node", "-e", cliResolveScript]
     def cliPath = this.getCommandOutput(nodeCommand, this.root)
 


thymikee:
Looking for `@react-native-community/cli` on iOS was a workaround for tests to pass. Is searching for `react-native/cli` somehow breaks your workflow?

tido64:
> Looking for `@react-native-community/cli` on iOS was a workaround for tests to pass. Is searching for `react-native/cli` somehow break your workflow?

Ah, interesting. I didn't know that we were meant to use `react-native/cli`. I think we still need to use `require.resolve()` though as `.bin` returns nothing here:

```
% node -e "console.log(require('react-native/cli').bin);"
undefined

% node -e "console.log(require.resolve('react-native/cli'));"
/~/node_modules/react-native/cli.js
```

Would you be opposed to this change?

thymikee:
Is there a chance you're running CLI that is lower than 3.x? Or maybe have duplicate CLIs `react-native/cli` resolves to wrong version? 
We've added this piece, so `bin` should always be defined:
https://github.com/react-native-community/cli/blob/74b43f63b8f35841d2c37d909d12c6c3efdb6e1d/packages/cli/src/index.ts#L243-L245

tido64:
TIL; we're on 0.60 still, so I guess it pulls in cli 2.6. We'll probably bump to react-native soon (or can just depend on cli 3.x directly).

Closing the issue. Thanks a lot!

thymikee:
Thanks for collaboration. We've made quite a bunch of changes in this area recently, but we feel it's in a pretty good shape now, supporting any project structure. Some changes may be available in 4.x version, which is currently still compatible with RN 0.61 and it likely stays this way

tido64:
> We've made quite a bunch of changes in this area recently, but we feel it's in a pretty good shape now, supporting any project structure.

Speaking of which, it would be really nice if we could support the project structure described in #1054. It would really benefit the [ReactTestApp project](https://github.com/microsoft/react-native-test-app) that we're currently working on. \*wink wink* 😉

thymikee:
It's on a roadmap :)

PR #1075, opened 2020-03-27
feat: add CocoaPods to `info` command

lucasbento:
Summary:
---------

We are relying quite a lot on CocoaPods so it makes sense to have it when running `react-native info`, will be useful for bug reports.

Test Plan:
----------

1. Run `react-native info`;
1. Be happy 😄


alloy:
:ceiling-cat: What does the output of this look like?

lucasbento:
image

@alloy: looks like this 🙂

alloy:
Nice, thanks 👌

Issue #1063, opened 2020-03-19
`npx react-native init AwesomeProject` not working


peterp:
```bash
➜ npx react-native init AwesomeProject2

               ######                ######
             ###     ####        ####     ###
            ##          ###    ###          ##
            ##             ####             ##
            ##             ####             ##
            ##           ##    ##           ##
            ##         ###      ###         ##
             ##  ########################  ##
          ######    ###            ###    ######
      ###     ##    ##              ##    ##     ###
   ###         ## ###      ####      ### ##         ###
  ##           ####      ########      ####           ##
 ##             ###     ##########     ###             ##
  ##           ####      ########      ####           ##
   ###         ## ###      ####      ### ##         ###
      ###     ##    ##              ##    ##     ###
          ######    ###            ###    ######
             ##  ########################  ##
            ##         ###      ###         ##
            ##           ##    ##           ##
            ##             ####             ##
            ##             ####             ##
            ##          ###    ###          ##
             ###     ####        ####     ###
               ######                ######


                  Welcome to React Native!
                 Learn once, write anywhere

✔ Downloading template
✖ Copying template
error Error: Couldn't find the "/var/folders/w5/bsw5vt490q78cyx2_h42ldtw0000gn/T/rncli-init-template-nzsCCY/node_modules/react-native/template.config.js file inside "react-native" template. Please make sure the template is valid. Read more: https://github.com/react-native-community/cli/blob/master/docs/init.md#creating-custom-template
```

I'm on Mac OS and don't have the `react-native` cli installed globally.

thymikee:
Maybe it's something with permissions? We've seen a few users having similar issues and reasons were varying. We can't really reproduce it ourselves. If you could debug this (clone this repo and follow CONTRIBUTING.md instructions), we'd be thankful for your help

peterp:
Running `sudo npx react-native init AwesomeProject2` made it work. The download too way longer than before, so I'm guessing that it could not stream the download into the destination correctly.

thymikee:
Awesome, that's something! I suspect we swallow an important error, which should surface earlier. Also, ideas why you would need sudo to make it work? Do you have Node.js installed globally (with sudo) and not as a user?

peterp:
I can't really think of a reason why this would fail, most `npx`/ `yarn create` scripts don't.

```bash
~/Personal/cli (master)
➜ which node
/usr/local/bin/node

~/Personal/cli (master)
➜ ll /usr/local/bin/node
[...]

PR #1077, opened 2020-03-27
fix: yarn/npm init before install template (#1603)

victorwpbastos:
Summary:
---------
When creating a new project with `npx react-native init MyApp`, the template installation step fails because there is no `package.json` inside the temp folder. This PR fixes that by running `yarn/npm init -y` before the template installation.

Fixes #1075 

Test Plan:
----------
CI passes


victorwpbastos:
Fixed #1063 too.

PR #1086, opened 2020-03-31
Fix: allow Hermes Debugger to connect from Flipper

mweststrate:
Summary:
---------

This change reflects D20526486 by @rickhanlonii  at Facebook: when Running Hermes Debugger from inside Flipper, the origin is localhost, but not the metro port. This fixes https://github.com/facebook/flipper/issues/660#issuecomment-606239984

Test Plan:
----------

* Init fresh project with React Native
* Enable Hermes
* Open Flipper, start the Hermes Debugger. This will generate the error in the metro logs: "Error: Unauthorized request from http://localhost:3000. This may happen because of a conflicting browser extension. Please try to disable it and try again."
* After this change (tested by directly editing in node_modules) and restart metro, the debugger does connect:

Screen Shot 2020-03-31 at 11 16 51

### DIFF
--- a/packages/cli/src/commands/server/middleware/getSecurityHeadersMiddleware.ts
+++ b/packages/cli/src/commands/server/middleware/getSecurityHeadersMiddleware.ts
@@ -13,7 +13,7 @@ export default function getSecurityHeadersMiddleware(
 ) {
   // Block any cross origin request.
   if (
-    typeof req.headers.origin === "string" &&
+    typeof req.headers.origin === 'string' &&
     !req.headers.origin.match(/^https?:\/\/localhost:/)
   ) {
     next(
--- a/packages/cli/src/commands/server/middleware/getSecurityHeadersMiddleware.ts
+++ b/packages/cli/src/commands/server/middleware/getSecurityHeadersMiddleware.ts
@@ -11,12 +11,9 @@ export default function getSecurityHeadersMiddleware(
   res: http.ServerResponse,
   next: (err?: any) => void,
 ) {
-  // @ts-ignore Property 'client' does not exist on type 'IncomingMessage', verify
-  const address = req.client.server.address();
-
   // Block any cross origin request.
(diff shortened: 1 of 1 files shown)


packages/cli/src/commands/server/middleware/getSecurityHeadersMiddleware.ts
```diff
@@ -17,7 +17,7 @@ export default function getSecurityHeadersMiddleware(
   // Block any cross origin request.
   if (
     req.headers.origin &&
-    req.headers.origin !== `http://localhost:${address.port}`
```
thymikee: We should be able to remove the address variable then?

mweststrate: good catch! Not sure why I still try to fix things without checking out repo's😅

thymikee:
A round of `yarn lint --fix` should fix linter :)

mweststrate:
My github edit + circle ci workflow is the coolest and least efficient one possible :P. (and I actually have local checkout, so not sure why I am doing this, my one-last-error strategy is clearly not paying off😬)

rickhanlonii:
@mweststrate does this mean that the Hermes debugger is available in OSS now?

@thymikee is it possible to get this into the 0.62.1 patch Friday?

mweststrate:
@rickhanlonii yes :)

thymikee:
@rickhanlonii almost, I'll make a release in a moment. There's no need to patch React Native, since this will be in semver range

mweststrate:
I assume since it is in the semver range, people with existing projects
won't get the update then, due to lock files, affecting only new projects
as a result?

thymikee:
@mweststrate you're right. Invalidating CLI entries in lock file is necessary.

thymikee:
Released in v4.5.1

Issue #1084, opened 2020-03-30
Attempt to launch Android emulator on Windows failed with improperly trimmed `emulatorName`


qiushihe:
## Environment

```
System:
    OS: Windows 10 10.0.18363
    CPU: (8) x64 Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz
    Memory: 19.06 GB / 31.88 GB
  Binaries:
    Node: 12.16.1 - C:\Program Files\nodejs\node.EXE
    Yarn: Not Found
    npm: 6.13.4 - C:\Program Files\nodejs\npm.CMD
    Watchman: Not Found
  SDKs:
    Android SDK: Not Found
  IDEs:
    Android Studio: Not Found
  Languages:
    Python: Not Found
  npmPackages:
    @react-native-community/cli: Not Found
    react: 16.11.0 => 16.11.0
    react-native: 0.62.0 => 0.62.0
  npmGlobalPackages:
    *react-native*: Not Found
```

## Description

On Windows, the `emulatorName` used by [tryLaunchEmulator.ts](https://github.com/react-native-community/cli/blob/master/packages/platform-android/src/commands/runAndroid/tryLaunchEmulator.ts) is not properly trimmed off its trailing `\r`, thus causing the attempt to launch Android emulator to fail.

The `getEmulators` function here: https://github.com/react-native-community/cli/blob/master/packages/platform-android/src/commands/runAndroid/tryLaunchEmulator.ts#L11 splits the output of `emulator -list-avds` by the newline character only which means, when on Windows the results from the split have a trailing `\r` character. This in turn cause the `execa` call in `launchEmulator` to fail.

## Reproducible Demo

To reproduce, you'll need to be on Windows.

For my example, the name of the AVD is `Android_API_28_-_Pixel_3_XL` and I can see it is the first result in the output of `emulator -list-avds`:

```
> C:\Users\blah\blah\blah\Android\Sdk\emulator\emulator.exe -list-avds
Android_API_28_-_Pixel_3_XL
Fossil_Gen_5_API_26
```

And to confirm my suspicion of the issue, I manually edited the `tryLaunchEmulator.js` (located at `my-project/node_modules/@react-native-community/cli-platform-android/build/commands/runAndroid/tryLaunchEmulator.js`) and inserted some logging above the `execa` call:

```
const launchEmulator = async (emulatorName: string, adbPath: string) => {
  return new Promise((resolve, reject) => {
[...]

PR #1098, opened 2020-04-05
fix: split output by EOL

Esemesek:
Summary:
---------

Windows uses different `EOL` when `UNIX` systems. We can fix this by using node API: https://nodejs.org/api/os.html#os_os_eol.

Fixes: https://github.com/react-native-community/cli/issues/1084


qiushihe:
Thank you @Esemesek !

PR #1099, opened 2020-04-05
fix: Use correct terminal on linux systems

Esemesek:
Summary:
---------

User correct `env` variable for opening new terminal on `linux`.

EDIT: This assumes that `linux` users have their `TERM` env variable exported correctly (I think this is the only reliable way to tell what terminals are used on linux systems). On my ArchLinux it was exported with value `xterm-256color` (it should be just `xterm`).

Test Plan:
----------

Tested on ArchLinux


thymikee:
Would be good to document this behavior somewhere in the docs of `start` command.

PR #1110, opened 2020-04-12
feat: Add new --dev-version option to Init

empyrical:
Summary:
---------

This pull request adds a new `--dev-version` option to the `init` command, which allows users to init a React Native project based off of a development version of React Native, or a fork.

To use it with a development version of React Native, you would be able to do this:

```sh
$ npx react-native init MyProject --verbose --dev-version https://github.com/facebook/react-native#4118d79
```

To use it with a forked version of React Native, e.g. using Microsoft's fork of React Native, you can do this:

```sh
$ npx react-native init MyProject --verbose --dev-version https://github.com/microsoft/react-native#f9df0ae
```

This functionality was added to a new command line flag, rather than overloading `--version`, because the desire is to keep the `--version` flag focused on valid RN versions on NPM. (see: #951)

Test Plan:
----------

Two additional tests were added to `__e2e__/init.test.ts`:

 * `init` with `--dev-version` set to a GitHub URL + a `#` commit
 * `init` with `--dev-version` set to a GitHub URL + a `#` commit, and a `--template` set to a local folder

### DIFF
--- a/__e2e__/init.test.ts
+++ b/__e2e__/init.test.ts
@@ -26,6 +26,28 @@ const customTemplateCopiedFiles = [
   'yarn.lock',
 ];
 
+// Files expected in the --dev-version tests
+const devVersionFiles = [
+  '.buckconfig',
+  '.eslintrc.js',
+  '.flowconfig',
+  '.gitattributes',
+  '.gitignore',
+  '.prettierrc.js',
+  '.watchmanconfig',
+  'App.js',
+  '__tests__',
+  'android',
+  'app.json',
+  'babel.config.js',
+  'index.js',
+  'ios',
+  'metro.config.js',
+  'node_modules',
+  'package.json',
+  'yarn.lock',
+];
+
 beforeEach(() => {
   cleanup(DIR);
   writeFiles(DIR, {});
@@ -122,4 +144,64 @@ test('init skips installation of dependencies with --skip-install', () => {
       file => !['node_modules', 'yarn.lock'].includes(file),
     ),
   );
+});
+
+test('init --dev-version with Git commit', () => {
+  const rnVersion = 'https://github.com/facebook/react-native#4118d79';
+  const {stdout} = runCLI(DIR, [
+    'init',
+    '--dev-version',
+    rnVersion,
+    'TestInit',
+  ]);
+
+  expect(stdout).toContain('Run instructions');
+
+  const packageJson = JSON.parse(
+    fs.readFileSync(path.join(DIR, 'TestInit', 'package.json'), 'utf8'),
+  );
+
+  // make sure that the version of `react-native` was properly updated
+  expect(packageJson.dependencies['react-native']).toEqual(rnVersion);
+
+  // make sure we don't leave garbage
+  expect(fs.readdirSync(DIR)).toEqual(['TestInit']);
+
+  let dirFiles = fs.readdirSync(path.join(DIR, 'TestInit'));
+
+  expect(dirFiles).toEqual(devVersionFiles);
+});
+
+test('init --dev-version with Git commit and --template filepath', () => {
+  createCustomTemplateFiles();
+  let templatePath = path.resolve(DIR, 'custom', 'template');
+  if (process.platform === 'win32') {
+    templatePath = templatePath.split('\\').join('/');
+  }
+
+  const rnVersion = 'https://github.com/facebook/react-native#4118d79';
+  const {stdout} = runCLI(DIR, [
+    'init',
+    '--dev-version',
+    rnVersion,
+    '--template',
+    `file://${templatePath}`,
+    'TestInit',
+  ]);
+
+  expect(stdout).toContain('Run instructions');
+
+  const packageJson = JSON.parse(
+    fs.readFileSync(path.join(DIR, 'TestInit', 'package.json'), 'utf8'),
+  );
+
+  // make sure that the version of `react-native` was properly updated
+  expect(packageJson.dependencies['react-native']).toEqual(rnVersion);
+
+  // make sure we don't leave garbage
+  expect(fs.readdirSync(DIR)).toContain('custom');
+
+  let dirFiles = fs.readdirSync(path.join(DIR, 'TestInit'));
+
+  expect(dirFiles).toEqual(customTemplateCopiedFiles);
 });
--- a/__e2e__/init.test.ts
+++ b/__e2e__/init.test.ts
@@ -26,28 +26,6 @@ const customTemplateCopiedFiles = [
   'yarn.lock',
 ];
 
-// Files expected in the --dev-version tests
-const devVersionFiles = [
-  '.buckconfig',
-  '.eslintrc.js',
-  '.flowconfig',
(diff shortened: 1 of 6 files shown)


Esemesek:
Don't know if this is the best of ideas to add a new flag for this type of functionality. I would rather have it handled as a part of `--template` flag.

empyrical:
If you try to init a project with a development version of React Native with `--template` like this:

```sh
$ npx react-native init MyProject --template https://github.com/facebook/react-native#4118d79
```

You run into this error during initialization:

```
✔ Downloading template
✔ Copying template
✔ Processing template
✖ Installing dependencies
error Error: Command failed: yarn install
error Couldn't find any versions for "react-native" that matches "1000.0.0"
```

This is because React Native uses `1000.0.0` as a placeholder version in its repo. To address this issue, this pull request adds functionality that goes in and changes the version of `react-native` in the template's `package.json` to the value of `--dev-version`

This requires at least one more command line flag to be used, because just using the value of `--template` as the "version" of `react-native` wouldn't make sense, especially if you're just trying to use something along the lines of `react-native-template-typescript` as your template. So this uses the value of `--dev-version` as the version to set in the template. This approach also allows you to both use a development version of React Native, and test out a template with it too.

packages/cli/src/commands/init/editTemplate.ts
```diff
@@ -106,3 +106,26 @@ export function changePlaceholderInTemplate({
       processDotfiles(filePath);
     });
 }
+
+export function changeReactNativeVersionInTemplate(newVersion: string) {
+  const packageJsonPath = path.join(process.cwd(), './package.json');
```
thymikee, line 111: We have access to `directory` from the callsite of this function. Let's use that instead of `process.cwd` which we try to avoid

packages/cli/src/commands/init/init.ts
```diff
@@ -170,14 +181,18 @@ async function createProject(
   version: string,
   options: Options,
 ) {
-  const templateName = options.template || `react-native@${version}`;
+  const templateName =
+    options.template ||
+    (options.devVersion && `react-native@${options.devVersion}`) ||
+    `react-native@${version}`;
```
thymikee, line 187: How about:
```js
const templateName = options.template || `react-native@${options.devVersion || version}`;
```

packages/cli/src/commands/init/index.ts
```diff
@@ -28,5 +28,10 @@ export default {
       name: '--title [string]',
       description: 'Uses a custom app title name for application',
     },
+    {
+      name: '--dev-version [string]',
+      description:
+        'Use a development version or fork of React Native. If no --template is provided, this will be used at the template as well.',
```
thymikee, line 34:
```suggestion
        'Replace the version of "react-native" dependency of the template with specified string. Useful for testing react-native or forks from the source.',
```

__e2e__/init.test.ts
```diff
@@ -94,3 +116,47 @@ test('init --template file with custom directory', () => {
...
+  ]);
+
+  expect(stdout).toContain('Run instructions');
+
+  // make sure we don't leave garbage
+  expect(fs.readdirSync(DIR)).toEqual(['TestInit']);
+
+  let dirFiles = fs.readdirSync(path.join(DIR, 'TestInit'));
+
+  expect(dirFiles).toEqual(devVersionFiles);
```
thymikee, line 135: Let's also verify that the version in package.json is set correctly

__e2e__/init.test.ts
```diff
@@ -94,3 +116,47 @@ test('init --template file with custom directory', () => {
...
+  ]);
+
+  expect(stdout).toContain('Run instructions');
+
+  // make sure we don't leave garbage
+  expect(fs.readdirSync(DIR)).toContain('custom');
+
+  let dirFiles = fs.readdirSync(path.join(DIR, 'TestInit'));
+
+  expect(dirFiles).toEqual(customTemplateCopiedFiles);
```
thymikee, line 161: Let's also verify that the version in package.json is set correctly

empyrical:
Changes made :+1:

grabbou:
Thanks for the PR! Looks solid!

So, this will init with the template (default or the one with `--template` flag) and override the RN version with the `--dev-version`?

Is there any reason why we can't just use `--version` for it? It sounds like it would be easier to extend its functionality a bit more.

I think I am a bit lost in this `init` experience as templates are bundled with RN versions too. Can you help me out a bit here?

thymikee:
Yea we could avoid adding extra flag. There would be no observable difference at the moment. However, the `--version` would normally install `react-native@version` and wouldn't affect what's inside the `./template/package.json` of that repo. With `--dev-version` we make it explicit. I'm fine with both solutions, if we documented that `--version` paired with `--template` acts this way. The complexity is going to be similar

grabbou:
Yeah, my preference is probably coming from `npm install` which accepts multiple different arguments.

Just repeating the other question to make sure it doesn't get lost:

> So, this will init with the template (default or the one with --template flag) and override the RN version with the --dev-version?

thymikee:
Yes

empyrical:
Yep, exactly how it handles the templates 👍 

If this adds too much complexity to the CLI I can find another way to accomplish this, but I thought I would see how this method would go before trying something else.

packages/cli/src/commands/init/init.ts
```diff
@@ -177,7 +188,8 @@ async function createProject(
   version: string,
   options: Options,
 ) {
-  const templateName = options.template || `react-native@${version}`;
+  const templateName =
+    options.template || `react-native@${options.devVersion || version}`;
```
grabbou, line 192: Insn't this a limitation? if we force `react-native@{options.devVersion}`, this will not work when pointed at git repository? (useful for unpublished forks and PRs)

grabbou: We don't seem to be doing such prepending in `changeReactNativeVersionInTemplate` tho.

grabbou:
I would recommend we also add an example with `desc` (https://github.com/react-native-community/cli/blob/master/docs/plugins.md#examples) and few cases, e.g. `--template` and `--dev-version` set to demonstrate "what is going to happen"

empyrical:
Yep, I can make those changes! :+1:

grabbou:
Sounds good! Sorry @empyrical that the review is taking so long! I understand that this may be a bit annoying, but I am really excited to bring this change in!

empyrical:
With React Native moving to a monorepo structure, this PR won't end up being useful for installing dev versions when that's done as Yarn and NPM don't allow you to pick a subfolder inside a git repo to install from.

Not sure if this will be useful for anything else - I can finish this pull request, but also would be OK with closing it (a much different solution will be needed for installing dev RN versions). Thoughts?

empyrical:
I don't think this will be useful anymore, so am going to close it for now

thymikee:
Thanks for all the work in this @empyrical!

empyrical:
No problem! I might revisit this when I figure out a good way to deal with the monorepo structure 👍

PR #1115, opened 2020-04-15
Allow out of tree platforms to work without custom metro configs

acoates-ms:
Summary:
---------

Currently it is hard to write out of tree platforms that do the right thing, and have a single metro config.  Basically as fallout from the removal of haste, now out of tree platforms contain a full implementation of react-native.  In order to build an out tree platform you have to redirect any `react-native` import into an import into to a different package.

`react-native-macos` and `react-native-windows` do this today by specifying custom extraNodeModules settings on the metro resolver to redirect `react-native` to their respective packages.  But that requires a different metro config to build macos vs building windows.

We have come up with a new solution that uses a custom **resolveRequest** function to rewrite the resolutions at bundle time, at which point we know which platform is being built, and can redirect to different packages for different platforms.

This change takes that code and generalizes it into the CLI.  So now any out of tree platform can just specify the npm package that metro should use as a redirect for `react-native` for that platform. -- And otherwise the default metro config should work.

One note, in addition to changing the **resolveRequest** property, I've changed the **assetRegistryPath** proeprty.  Previously it would be a full path, which would mean that it would not be correctly redirected by the new **reactNativePlatformResolver**.  This path is added as a require statement in the transformer, so it should be fine for most uses as `react-native/Libraries/Image/AssetRegistry`

Test Plan:
----------

Posting this early for feedback.  We recently applied this change to `react-native-windows` and are still in the process of validating that locally.  Then we'll verify that pushing this to the CLI will make everything cleaner.  -- So this has not been validated yet.

### DIFF
--- a/packages/cli-types/src/index.ts
+++ b/packages/cli-types/src/index.ts
@@ -63,6 +63,7 @@ interface PlatformConfig<
   DependencyConfig,
   DependencyParams
 > {
+  npmPackageName?: string;
   projectConfig: (
     projectRoot: string,
     projectParams: ProjectParams | void,
--- a/packages/cli/src/tools/loadMetroConfig.ts
+++ b/packages/cli/src/tools/loadMetroConfig.ts
@@ -55,28 +55,39 @@ export interface MetroConfig {
  * Default configuration
  */
 export const getDefaultConfig = (ctx: Config): MetroConfig => {
+  const outOfTreePlatforms = Object.keys(ctx.platforms).filter(
+    platform => ctx.platforms[platform].npmPackageName,
+  );
+
   return {
     resolver: {
       resolveRequest:
-        Object.keys(ctx.platforms).filter(
-          platform => ctx.platforms[platform].npmPackageName,
-        ).length === 0
+        outOfTreePlatforms.length === 0
           ? undefined
           : reactNativePlatformResolver(
-              Object.keys(ctx.platforms)
-                .filter(platform => ctx.platforms[platform].npmPackageName)
-                .reduce<{[platform: string]: string}>((result, platform) => {
+              outOfTreePlatforms.reduce<{[platform: string]: string}>(
+                (result, platform) => {
                   result[platform] = ctx.platforms[platform].npmPackageName!;
                   return result;
-                }, {}),
+                },
+                {},
+              ),
             ),
       resolverMainFields: ['react-native', 'browser', 'main'],
       platforms: [...Object.keys(ctx.platforms), 'native'],
     },
     serializer: {
+      // We can include multiple copies of InitializeCore here because metro will
+      // only add ones that are already part of the bundle
       getModulesRunBeforeMainModule: () => [
         require.resolve(
           path.join(ctx.reactNativePath, 'Libraries/Core/InitializeCore'),
+        ),
+        ...outOfTreePlatforms.map(platform =>
+          require.resolve(
+            `${ctx.platforms[platform]
+              .npmPackageName!}/Libraries/Core/InitializeCore`,
+          ),
         ),
       ],
       getPolyfills: () =>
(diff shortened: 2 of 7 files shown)


packages/cli-types/src/index.ts
```diff
@@ -63,6 +63,7 @@ interface PlatformConfig<
   DependencyConfig,
   DependencyParams
 > {
+  npmPackageName?: string;
```
thymikee, line 66: We need to document this field

packages/cli/src/tools/loadMetroConfig.ts
```diff
@@ -50,6 +57,19 @@ export interface MetroConfig {
...
   return {
     resolver: {
+      resolveRequest:
+        Object.keys(ctx.platforms).filter(
+          platform => ctx.platforms[platform].npmPackageName,
+        ).length === 0
+          ? undefined
+          : reactNativePlatformResolver(
+              Object.keys(ctx.platforms)
+                .filter(platform => ctx.platforms[platform].npmPackageName)
```
thymikee, line 67: This filtering is happening twice. Can we extract the result and reuse?

packages/cli/src/tools/metroPlatformResolver.ts
```diff
@@ -0,0 +1,49 @@
...
+ * A map of platform to npm package that implements that platform
+ *
+ * Ex:
+ * {
+ *    windows: 'react-native-windows'
+ *    macos: 'react-native-macos'
+ * }
+ */
+// @ts-ignore - no typed definition for the package
+import {resolve} from 'metro-resolver';
```
thymikee, line 16: we need to add `metro-resolver` to this package dependencies now

packages/cli/src/tools/metroPlatformResolver.ts
```diff
@@ -0,0 +1,49 @@
...
+          modifiedModuleName = `${
+            platformImplementations[platform]
+          }/${modifiedModuleName.slice('react-native/'.length)}`;
+        }
+      }
+      let result = resolve(context, modifiedModuleName, platform);
+      context.resolveRequest = backupResolveRequest;
+      return result;
+    } catch (e) {
+      context.resolveRequest = backupResolveRequest;
```
thymikee, line 45: We could use `finally` to bring back original resolveRequest once. Btw, is this really the official way to alias modules?

cpojer:
This is pretty smart, I like it. Nice work @acoates-ms!

packages/cli/src/tools/metroPlatformResolver.ts
acoates-ms: Normally we alias modules using extraNodeModules.  We have to use the resolveRequest option here, because we want to alias the module differently based on which platform is being bundled, so you can run a single metro config and serve up any of the platforms.

acoates-ms:
@thymikee - Getting close to this being ready I think, if you'd like to take another look.
We found an issue related to InitializeCore not getting properly run before the main module.  -- But that's fixed in the latest iteration.

thymikee:
FYI: https://github.com/react-native-community/cli/pull/1140

grabbou:
Thanks for the PR. Looks interesting! I like it
 
> Currently it is hard to write out of tree platforms that do the right thing, and have a single metro config. Basically, as fallout from the removal of haste, now out of tree platforms contain a full implementation of react-native. In order to build an out tree platform you have to redirect any react-native import into an import into to a different package.

Does this mean this PR is likely to be reverted when `out-of-tree` platforms no longer contain a full implementation of React Native? I would guess this is our long-term goal.

acoates-ms:
@grabbou - The long term goal would be that core is truly platform agnostic, and even android/iOS are in separate packages.  But there are no near term plans to get there. -- and its not clear what the metro config would end up looking like in that case. One of the nice things about this change is that the platform has to opt in within its react-native config.  So even if things change it should be easy to migrate platforms to opt back out of this behavior.

PR #1118, opened 2020-04-16
feat: add a programmatic dev server API

fson:
Summary:
---------

I work on Expo CLI, which uses React Native CLI to start the development server. Previously, we've done this by starting RN CLI as a subprocess. This has had some challenges, including slower boot-up time because we need to start another Node process, possibility to end up with a killed or zombie processes if there's an unhandled error, and having to parse all logs from JSON in stdout.

This PR refactors the `react-native start` command and introduces a new package `@react-native-community/dev-server-api`, which includes a programmatic interface for starting the dev server middleware and WebSocket servers used by React Native. This allows tools like Expo CLI to interface with Metro directly and then attach the additional features provided by React Native CLI to the Metro HTTP server. (https://github.com/expo/expo-cli/pull/1845 will make use of this package).

Test Plan:
----------

I have followed the [Testing your changes](https://github.com/react-native-community/cli/blob/66a0616415014c0e8f3da55c0cefffa882718cad/CONTRIBUTING.md#testing-your-changes) instructions in `CONTRIBUTING.md` and started the dev server with `npx react-native start --watchFolders /Users/ville/Projects/react-native-cli/ --resetCache`. After starting the server, I've verified the following features work:
- Running the app on iOS simulator
- Editing a file in the app, the app refreshed automatically
- Refreshing the app by pressing `r`
- Starting the debugger in Chrome and setting a breakpoint
- Looking at logs in Flipper

Screen Shot 2020-04-16 at 17 51 09


packages/tools/src/index.ts
```diff
@@ -3,5 +3,8 @@ export {default as groupFilesByType} from './groupFilesByType';
 export {default as isPackagerRunning} from './isPackagerRunning';
 export {default as getDefaultUserTerminal} from './getDefaultUserTerminal';
 export {fetch, fetchToTemp} from './fetch';
+export {default as launchDefaultBrowser} from './launchDefaultBrowser';
+export {default as launchDebugger} from './launchDebugger';
+export {default as launchEditor} from './launchEditor';
```
thymikee, line 8: Were these moved here to be reused by Expo?

packages/dev-server-api/package.json
```diff
@@ -0,0 +1,30 @@
...
+  },
+  "dependencies": {
+    "@react-native-community/cli-debugger-ui": "^4.2.1",
+    "@react-native-community/cli-tools": "^4.7.0",
+    "compression": "^1.7.1",
+    "connect": "^3.6.5",
+    "errorhandler": "^1.5.0",
+    "pretty-format": "^25.1.0",
+    "serve-static": "^1.13.1",
+    "ws": "^1.1.0"
```
thymikee, line 17: Would be awesome to migrate to the newest version as a followup

packages/tools/package.json
```diff
@@ -10,7 +10,9 @@
     "chalk": "^3.0.0",
     "lodash": "^4.17.15",
     "mime": "^2.4.1",
-    "node-fetch": "^2.6.0"
+    "node-fetch": "^2.6.0",
+    "open": "^6.2.0",
+    "shell-quote": "1.6.1"
```
thymikee, line 15: Can you remove these deps from `cli` as well?

packages/dev-server-api/src/index.ts
```diff
@@ -0,0 +1,92 @@
...
+    .use(systraceProfileMiddleware)
+    .use('/reload', (_req: http.IncomingMessage, res: http.ServerResponse) => {
+      broadcast('reload');
+      res.end('OK');
+    })
+    .use(errorhandler());
+
+  options.watchFolders.forEach(folder => {
+    // @ts-ignore mismatch between express and connect middleware types
+    middleware.use(serveStatic(folder));
```
thymikee, line 65: Do we really need to serve these folders?

packages/tools/src/index.ts
fson: These were moved from `cli` to avoid circular dependency because `dev-server-api` uses them.

packages/cli/src/commands/server/runServer.ts
```diff
@@ -9,13 +9,13 @@
 import Metro from 'metro';
 // @ts-ignore untyped metro
 import {Terminal} from 'metro-core';
-import http from 'http';
 import path from 'path';
+import {
+  createDevServerMiddleware,
+  indexPageMiddleware,
+} from '@react-native-community/dev-server-api';
```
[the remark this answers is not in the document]
fson: Nice catch! Added.

packages/dev-server-api/src/index.ts
fson: That's a good question. I don't know why they're being served, but I tried to keep this PR strictly as refactoring and avoid making changes to the functionality. Should I perhaps do a follow up PR to stop serving them, so then it's easy to revert in case this breaks something?

thymikee: Yep, let's do this in a followup

fson: Did this in another branch, I'll rebase it and open the followup PR once this has landed. 👍

packages/dev-server-api/package.json
fson: Upgraded to the latest version of `ws` in another branch 👍

fson:
At @thymikee's suggestion, I've split this PR up so it only includes the new `dev-server-api` package and changes to `cli-tools`. The CLI will be changed to use `dev-server-api` in a follow up PR.

packages/dev-server-api/package.json
```diff
@@ -0,0 +1,30 @@
+{
+  "name": "@react-native-community/dev-server-api",
```
thymikee, line 2: can we rename this to `@react-native-community/cli-server-api`? It's important to have a `@react-native-community/cli-` prefix, so we don't pollute the organization

thymikee: Also the directory change to `cli-server-api` would be cool

fson: Sounds good to me!

thymikee:
Published as a part of v4.8.0

PR #1120, opened 2020-04-18
[backport from 0.62] fix: pick correct TARGET_BUILD_DIR for run-ios (#1032)

hudon:
Summary:
---------

Cherry-pick from https://github.com/react-native-community/cli/pull/1032

A fix to `react-native run-ios` was merged into 0.62 but 0.61's `run-ios` command still has the bug, so this PR just backports the fix.

Test Plan:
----------

Ran `react-native run-ios` in our codebase using 0.61 with this patch, and it now works. It was previously broken because `run-ios` was confused by the `TARGET_BUILD_DIR` present in the `React` pod target's build settings (as opposed to our app's build settings)

### DIFF
--- a/packages/platform-ios/src/commands/runIOS/index.ts
+++ b/packages/platform-ios/src/commands/runIOS/index.ts
@@ -367,10 +367,17 @@ function bootSimulator(selectedSimulator: Device) {
 }
 
 function getTargetBuildDir(buildSettings: string) {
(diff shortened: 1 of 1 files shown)


grabbou:
Is the `e2e-tests` failure expected? @thymikee

thymikee:
Yea, this is 3.x branch wich has flaky e2e tests.

PR #1126, opened 2020-04-22
feat(cli-platform-ios): add macOS auto-linking support

alloy:
Summary:
---------

These changes make it so `react-native-macos` apps can use the `native_modules.rb` script as-is, but only activate pods in those targets that target the platform the pods support.

For instance, consider these npm packages that come with native modules:

* `ios-dep`
* `macos-dep`
* `ios-and-macos-dep`

…and given a `Podfile` like the following:

```ruby
require_relative '../node_modules/@react-native-community/cli-platform-ios/native_modules'

abstract_target 'Shared' do
  pod 'React', :path => "../node_modules/react-native-macos/"
  # ...

  target 'A iOS Target' do
    platform :ios, '9'
    use_native_modules!
  end

  target 'A macOS Target' do
    platform :macos, '10.14'
    use_native_modules!
  end
end
```

…the result would be that:

* `A iOS Target` has the following dependencies:
  - `ios-dep`
  - `ios-and-macos-dep`
* `A macOS Target` has the following dependencies:
  - `macos-dep`
  - `ios-and-macos-dep`

NOTE: In a future major version, the `native_modules` support will likely move to a `cli-platform-apple` package, but for now this was the least invasive change.

Test Plan:
----------

Changes have full test coverage, which have been moved from Ruby to JS, and I’ve performed test runs with new applications.


packages/cli-types/src/index.ts
```diff
@@ -127,7 +128,7 @@ export type ProjectConfig = {
  * @property platforms - Map of available platforms (build-ins and dynamically loaded)
  * @property commands - An array of commands that are present in 3rd party packages
  */
-export type Config = {
+export interface Config extends IOSNativeModulesConfig {
```
alloy, line 131: This is done purely to ensure the `Config` doesn’t deviate in a way that the otherwise uncheckable Ruby source.

packages/cli-types/src/ios.ts
```diff
@@ -6,6 +6,13 @@
  */
 export interface IOSProjectParams {
   project?: string;
+  /**
+   * @deprecated A podspec should always be at the root of a package and
+   *             have the name of the package. This property will be
+   *             removed in a future major version.
+   *
+   * @todo Log a warning when this is used.
```
alloy, line 14: Have to follow-up on this in a future iteration.

packages/cli-types/src/ios.ts
```diff
@@ -31,3 +38,55 @@ export interface IOSProjectConfig {
...
+  execution_position?: 'before_compile' | 'after_compile' | 'any';
+};
+
+/**
+ * This describes the data that is expected by `native_modules.rb`. It is only
+ * meant to ensure the `Config` interface follows exactly what is needed, so
+ * only make changes to this interface (or `IOSScriptPhase`) if the data
+ * requirements of `native_modules.rb` change.
+ */
+export interface IOSNativeModulesConfig {
```
alloy, line 67: ❗ So this should only change if the Ruby source changes and should not be refactored/shared with other typings, even if they are similar.

packages/platform-ios/native_modules.rb
```diff
@@ -54,6 +60,16 @@ def use_native_modules!(config = nil)
 
     spec = Pod::Specification.from_file(podspec_path)
 
+    # Skip pods that do not support the platform of the current target.
+    if platform = current_target_definition.platform
+      next unless spec.supported_on_platform?(platform.name)
```
alloy, line 65: This is the actual semantic change of the PR, so tiny 😂

packages/platform-ios/native_modules.rb
```diff
@@ -108,162 +124,32 @@ def use_native_modules!(config = nil)
...
-      @podfile = podfile = Object.new
-      @spec = spec = Object.new
-
-      spec.singleton_class.send(:define_method, :name) { "ios-dep" }
-
-      podfile.singleton_class.send(:define_method, :use_native_modules) do |config|
-        use_native_modules!(config)
-      end
+  require "json"
+  runInput = JSON.parse(ARGF.read)
```
alloy, line 130: The Ruby script receives a JSON payload via `stdin`…

packages/platform-ios/native_modules.rb
```diff
@@ -108,162 +124,32 @@ def use_native_modules!(config = nil)
...
-          :name => "My Name",
-          :execution_position => :before_compile,
-          :input => "string"
-        }]
-        file_read_mock.verify
-      end
-    end
+  unless runInput["capture_stdout"]
+    puts podfile.to_hash.to_json
   end
```
alloy, line 154: …and spits out a serialised version of the `Podfile` to `stdout`.

.circleci/config.yml
```diff
@@ -80,14 +75,10 @@ jobs:
     executor: node8
     steps:
       - run-lint
-  cocoa-pods:
+  unit-tests:
     executor: noderuby
```
thymikee, line 79: Do we know which version of Node is run on this image? Would be cool if it was the latest or so

packages/cli-types/src/ios.ts
thymikee: We'll need to identify libraries that follow a pattern of adding a Podspec into `ios/` dir and send PRs.

.circleci/config.yml
alloy: I couldn’t spot an explicit listing of the version in the CI output, but it’s using `circleci/ruby:2.4-node` and in [this listing](https://hub.docker.com/layers/circleci/ruby/2.4-node/images/sha256-fb869ba73c3c712621031d17a28ebb6f38b998ec22efdeec354b870b50c1bd58?context=explore) I see `ENV NODE_VERSION=12.16.2`. So probably that?

alloy:
Seeing as it’s failing on the `lts-tests`, and also not to make things too hard on collaborators, I’ll add a check and disable the tests in case the env doesn’t support ruby/cocoapods.

packages/platform-ios/native_modules.rb
grabbou: Thanks for that comment! And also, I am impressed by the CocoaPods API :D

packages/cli-types/src/ios.ts
grabbou: Why is that? Wouldn't it be better to just reference https://github.com/react-native-community/cli/blob/f570baf34072d3256064d92162333bd173c96c78/packages/cli-types/src/index.ts#L131-L153 directly? This is the exact type of the JSON output from the command.

Could you help me understand where and how it is used? Maybe it's not the right place for it then (in shared package), but next to a utility?

alloy: Because TS cannot type-check the Ruby code, so people making changes to this `Config` interface might accidentally break some input data assumption made by it. Put differently, you should see this interface as if it were the function signature declaration of the Ruby code, if it were written in TS.

alloy:
Ace, thanks for the review! I feel confident about the need for the typing, but if needed we can amend the wording in the comment to clarify in a follow-up PR, so I’ll go ahead and merge for now 👍

PR #1130, opened 2020-04-24
RegExp: more prettier

zusinShinpei:
Summary:
---------

Test Plan:
----------


thymikee:
Thanks, but we don't accept contributions to `global-cli` package for now. It's considered frozen and we discourage from using it. Use `npx react-native` instead

zusinShinpei:
@thymikee npx react-native not using this cli?

thymikee:
`npx react-native` uses this CLI https://github.com/react-native-community/cli/tree/master/packages/cli which is the official one.

PR #1131, opened 2020-04-24
fix: Config command additional output

TMaszko:
Summary:
---------

Logger should be disabled at the top of the `setupAndRun` function when config command has been used because we want only plain JSON config at the output.

Test Plan:
----------
Added test for config command


__e2e__/config.test.ts
```diff
@@ -0,0 +1,45 @@
...
+  // Register all packages to be linked
+  for (const pkg of ['platform-ios', 'platform-android']) {
+    spawnScript('yarn', ['link'], {
+      cwd: path.join(__dirname, `../packages/${pkg}`),
+    });
+  }
+
+  // Clean up folder and re-create a new project
+  cleanup(cwd);
+  writeFiles(cwd, {});
```
thymikee, line 22: cleanup should be sufficient, can we remove writeFiles?

__e2e__/config.test.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  // Initialise React Native project
+  runCLI(cwd, ['init', 'TestProject']);
```
thymikee, line 25: We avoid running init with default template in e2e tests, because it's flaky. Can you init from a local template like in `__e2e__/init.test.ts`?

__e2e__/__snapshots__/config.test.ts.snap
```diff
@@ -0,0 +1,236 @@
...
+    }
+  ],
+  \\"assets\\": [],
+  \\"platforms\\": {
+    \\"ios\\": {},
+    \\"android\\": {}
+  },
+  \\"project\\": {
+    \\"ios\\": {
+      \\"sourceDir\\": \\"/private/var/folders/zt/917v0jxx6lg3p_zfh9s_02bm0000gn/T/test_root/TestProject/ios\\",
```
thymikee, line 208: these paths are going to be different based on the CI they run in (Circle vs Azure), you'll need to replace them with something like `<<REPLACED_ROOT>>` before taking a snapshot

__e2e__/__snapshots__/config.test.ts.snap
```diff
@@ -0,0 +1,236 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`shows up current config without additional output 1`] = `
+"ctx:  {
```
thymikee, line 4: Where this came from? Looks like a broken snapshot that shouldn't end up here, e.g. when using console.log to debug test

__e2e__/config.test.ts
```diff
@@ -0,0 +1,45 @@
...
+  });
+});
+
+afterAll(() => {
+  cleanup(cwd);
+});
+
+test('shows up current config without unnecessary output', () => {
+  const {stdout} = runCLI(path.join(cwd, 'TestProject'), ['config']);
+  expect(stdout).toMatchSnapshot();
```
thymikee, line 44: Can we corrupt the `setup_env.sh` (if it's there, if not then we're fine) before running the `config` command? This way we can actually check if this change works

__e2e__/__snapshots__/config.test.ts.snap
TMaszko: Forgot console.log yikes !

__e2e__/config.test.ts
TMaszko: We can't because writeFiles function creates folder at cwd path :)

__e2e__/config.test.ts
thymikee: On the second thought, let's give it a chance, maybe it won't be flaky this time.

__e2e__/config.test.ts
TMaszko: Yes, this is exactly what we want

__e2e__/config.test.ts
```diff
@@ -0,0 +1,89 @@
...
+  const restoreOriginalSetupEnvScript = createCorruptedSetupEnvScript();
+  const {stdout, stderr} = runCLI(path.join(DIR, 'TestProject'), ['config']);
+  const configWithReplacedProjectRoots = replaceProjectRootInOutput(
+    stdout,
+    'test_root',
+  );
+
+  restoreOriginalSetupEnvScript();
+  expect(isValidJSON(configWithReplacedProjectRoots)).toBe(true);
+  expect(configWithReplacedProjectRoots).toMatchSnapshot();
```
thymikee, line 87: the check for valid JSON is enough, please remove this snapshot (it's captured in previous test anyway)

__e2e__/replaceProjectRootInOutput.test.ts
```diff
@@ -0,0 +1,13 @@
+import {replaceProjectRootInOutput} from '../jest/helpers';
```
thymikee, line 1: let's move this test to `jest/` directory, shall we?

jest/helpers.ts
```diff
@@ -196,3 +196,10 @@ ${chalk.bold('stdout:')}  ${result.stdout}
 ${chalk.bold('code:')}    ${result.code}`);
   }
 }
+
+export function replaceProjectRootInOutput(output: string, testFolder: string) {
+  return output.replace(
+    new RegExp(`(:\\s")(.*)(\\/${testFolder}.*)`, 'g'),
```
thymikee, line 202: I think it would be more reliable to replace the `process.cwd()`? Also less regexy

thymikee: I mean, such helper may be helpful (e.g. when testing tmp directories) but in this case I think replacing process.cwd is all we need.

TMaszko: Maybe the best way it's to connect both solutions and just pass `process.cwd()` as an second argument and then we would have `output.replace(testFolder, '<<REPLACED_ROOT>>')` because when we put just process.cwd it would be hard to test it ofc we can mock process.cwd but I think it's better WDYT?

__e2e__/config.test.ts
```diff
@@ -0,0 +1,93 @@
...
+  }));
+
+  const configWithReplacedProjectRoots = replaceProjectRootInOutput(
+    JSON.stringify(parsedStdout, null, 2).replace(/\\\\/g, '\\'),
+    DIR,
+  );
+  expect(wrap(configWithReplacedProjectRoots)).toMatchSnapshot();
+});
+
+test('should log only valid JSON config if setuping env throws an error', () => {
```
jamesgeorge007, line 86:
```suggestion
test('should log only valid JSON config if setting up env throws an error', () => {
```

__e2e__/config.test.ts
```diff
@@ -0,0 +1,93 @@
...
+function isValidJSON(text: string) {
+  try {
+    JSON.parse(text);
+    return true;
+  } catch {
+    return false;
+  }
+}
+
+//We have to check whether setup_env script fails if it does then we shouldn't log any info to the console
```
jamesgeorge007, line 24:
```suggestion
// We have to check whether setup_env script fails, if it does then we shouldn't log any info to the console
```

TMaszko:
Thanks for your comments I appreciate that 👍

PR #1132, opened 2020-04-25
refactor: make watchFolders read-only everywhere

fson:
Summary:
---------

This PR changes the TS type of `watchFolders` and `projectRoots` arrays to read-only everywhere. These arrays come from the Metro configuration and they've already been treated as immutable and not modified anywhere. The read-only array type signals this: the function doesn't modify the array.

The rationale for this type change is that it helps avoid type errors when `watchFolders` is passed from Metro config to `createDevServerMiddleware`, for example:
```ts
const { middleware, attachToServer } = createDevServerMiddleware({
  port: metroConfig.server.port,
  watchFolders: metroConfig.watchFolders,
});
```
When `watchFolders` was `Array<string>` , this code resulted in a type error, because `watchFolders` is `ReadonlyArray<string>` in Metro config. (A work-around was to defensively clone `watchFolders` before passing it in.)

Test Plan:
----------

Only verified that TypeScript build succeeded, as this PR doesn't change any functionality, only types.

### DIFF
--- a/packages/cli-server-api/src/devToolsMiddleware.ts
+++ b/packages/cli-server-api/src/devToolsMiddleware.ts
@@ -27,7 +27,7 @@ function escapePath(pathname: string) {
 type LaunchDevToolsOptions = {
   host?: string;
   port: number;
-  watchFolders: Array<string>;
+  watchFolders: ReadonlyArray<string>;
 };
 
 function launchDevTools(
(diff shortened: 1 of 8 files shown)


thymikee:
Mind resolving the conflict? :)

fson:
Rebased and resolved! 😄

PR #1145, opened 2020-04-29
Upgrade some dependencies versions in global-cli and cli-server-api p…

basemanabulsi:
Summary:
---------

We faces some security issues from this module which has some old versions of dependencies which lead us to upgrade some outdated dependencies in package.json files in two packages global-cli and cli-server-api.

Test Plan:
----------
I did ```yarn test``` to be sure that nothing broke from this upgrade

### DIFF
--- a/packages/cli-server-api/package.json
+++ b/packages/cli-server-api/package.json
@@ -14,7 +14,7 @@
     "errorhandler": "^1.5.0",
     "pretty-format": "^25.1.0",
     "serve-static": "^1.13.1",
-    "ws": "^1.1.0"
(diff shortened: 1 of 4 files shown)


thymikee:
I'm sorry, but we cannot do that for 2 reasons:

1. `global-cli` is frozen for now, we don't accept any updates here
2. we can't upgrade `ws` to v7 because it's incompatible with Metro, which also uses v1, which again, you'd need to fix there, which is even more complex to do.

Thanks for the efforts though, looking forward to future contributions 😊

PR #1152, opened 2020-05-06
Add Windows and mac os instructions

### LINKED ISSUE #1153: Add instructions for Windows and mac OS to react-native init
## Describe the Feature
Add instructions for Windows and mac OS to react-native init

## Possible Implementations
It's just adding a string :) see PR #1152 

## Related Issues
See [RNW issue](https://github.com/microsoft/react-native-windows/issues/4802)

asklar:
Summary:
---------
Add instructions for React Native for Windows and mac OS.
Fixes #1153 

Test Plan:
----------

react-native init test123

image


thymikee:
@Esemesek @alloy @grabbou what do you think? Is this the right time to add it? The link is to the outside resource, so at least it will be up to date.

I was also thinking whether we should output this when the template is specifically `react-native-windows` or `react-native-macos` or something, but that won't support custom templates easily.

alloy:
> @Esemesek @alloy @grabbou what do you think? Is this the right time to add it? The link is to the outside resource, so at least it will be up to date.
> 
> I was also thinking whether we should output this when the template is specifically `react-native-windows` or `react-native-macos` or something, but that won't support custom templates easily.

In neither of these cases is the CLI template one of those, i.e. they both require you to use the normal upstream template and then `react-native-(windows|macos)-init` applies _on top of that_ its own templates.

So yeah, I think it makes sense to provide a link to the latest details 👍

packages/cli/src/commands/init/printRunInstructions.ts
```diff
@@ -34,6 +34,13 @@ function printRunInstructions(projectDir: string, projectName: string) {
   ${chalk.green(`Run instructions for ${chalk.bold('Android')}`)}:
     • Have an Android emulator running (quickest way to get started), or a device connected.
     • cd "${projectDir}" && npx react-native run-android
+
+  ${chalk.magenta(
+    `Run instructions for ${chalk.bold('Windows')} and ${chalk.bold('mac OS')}`,
```
alloy, line 39:
```suggestion
    `Run instructions for ${chalk.bold('Windows')} and ${chalk.bold('macOS')}`,
```

thymikee:
We could define a new field for template config that would provide custom run instructions for a platform: https://github.com/react-native-community/cli/blob/master/docs/init.md#creating-custom-template

alloy:
@thymikee That wouldn’t help, as the user does not init from a custom CLI template; which is what I was trying to convey earlier.

thymikee:
Yup, but in not so distant future I'd like RN for Windows and macOS be compatible with CLI anyway, so we can start thinking about this problem :)

alloy:
Ah gotcha 👍 Yeah we could, but I’m not sure that displaying run instructions will be the biggest hurdle for us to overcome 😄

packages/cli/src/commands/init/printRunInstructions.ts
```diff
@@ -34,6 +34,13 @@ function printRunInstructions(projectDir: string, projectName: string) {
   ${chalk.green(`Run instructions for ${chalk.bold('Android')}`)}:
     • Have an Android emulator running (quickest way to get started), or a device connected.
     • cd "${projectDir}" && npx react-native run-android
+
+  ${chalk.magenta(
+    `Run instructions for ${chalk.bold('Windows')} and ${chalk.bold('macOS')}`,
+  )}:
+    • See ${chalk.underline(
+      'http://aka.ms/ReactNative',
```
thymikee, line 42: I prefer full URL instead of using a shortener, WDYT?
```suggestion
      'https://microsoft.github.io/react-native-windows/',
```
Or at least use `https://` :) I'm fine with both

asklar: I think we would want to be able to update the target URL or title of the project without requiring updating the CLI

thymikee: As long as it's controlled from someone from Microsoft, then I'm good. Just https then please :D

asklar: Yep I'm on the RN team at Microsoft :) and yes can do https

alloy: 👍

alloy:
Ace :shipit:

PR #1109, opened 2020-04-10
feat: install android dependencies automatically

molant:
Summary:
---------

This PR installs Android Studio, Android SDKs and the right hypervisor on Windows machines.

There are a few things still missing that I'd like to do:

* [x] **Need someone with an AMD machine** to check what's the output of  `emulator-check accel` 
* [x] Create an AVD. I'm planning on using the same configuration used by the Xamarin team but if someone has another opinion please let me know
* [x] Add more tests around Hypervisor installation to verify it gets installed correctly

There's a lot going on so if you could start reviewing this it would be super helpful.

Test Plan:
----------

Added a few tests for Android Studio and Android SDK.

Now that there is `installComponent` for Android maybe the `NDK` portion could be automated as well?

You can see it in action in here:

[RN CLI installing Android Studio video](https://youtu.be/khRal1bElIw)

PR #1151, opened 2020-05-05
chore: upgrade Jest to 26

thymikee:
Summary:
---------

New release is up https://jestjs.io/blog/2020/05/05/jest-26. 
With this upgrade I've changed the runner to `jest-circus`, which will be the default in next Jest major, and hopefully fixed flaky config tests (started to by consistent on my machine). Apparently `rimraf` doesn't remove all directories in the tree at times, which resulted in leaks from previous test cases.

Test Plan:
----------

Green CI

PR #1156, opened 2020-05-07
chore: make `cleanup` async

molant:
Summary:
---------

I'm trying to improve the reliability of Windows CI so I can get #1109 green.
I've found out that `cleanup()` uses `rimraf.sync` underneath. Windows FS is not very... reliable, and sometimes you can end up with `ENOTEMPTY` errors like [this one](https://dev.azure.com/react-native-community/cli/_build/results?buildId=1248&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9&t=5caf77c8-9b10-50ef-b5c7-ca89c63e1c86&l=417). In fact, [`rimraf`'s documentation](https://www.npmjs.com/package/rimraf#api) calls that out:

> Windows: `EBUSY` and `ENOTEMPTY` - rimraf will back off a maximum of `opts.maxBusyTries` times before giving up, adding 100ms of wait between each attempt. The default `maxBusyTries` is 3.

The problem is the above does not apply when using the sync method:

> If an `EBUSY`, `ENOTEMPTY`, or `EPERM` error code is encountered on Windows systems, then rimraf will retry with a linear backoff wait of 100ms longer on each try. The default maxBusyTries is 3.
> **Only relevant for async usage.**

I've changed it to use the async version and updated the tests. Hopefully that should make Windows more reliable.

Test Plan:
----------

CI

### DIFF
--- a/__e2e__/default.test.ts
+++ b/__e2e__/default.test.ts
@@ -7,7 +7,7 @@ import {
 
 const DIR = getTempDirectory('test_default_behavior');
 
-beforeEach(async () => {
+beforeEach(() => {
   cleanupSync(DIR);
   writeFiles(DIR, {});
 });
--- a/__e2e__/default.test.ts
+++ b/__e2e__/default.test.ts
@@ -3,11 +3,11 @@ import {runCLI, getTempDirectory, cleanup, writeFiles} from '../jest/helpers';
 const DIR = getTempDirectory('test_default_behavior');
 
 beforeEach(async () => {
-  await cleanup(DIR);
+  cleanup(DIR, false);
   writeFiles(DIR, {});
 });
 afterEach(async () => {
-  await cleanup(DIR);
+  cleanup(DIR, false);
 });
 
 test('shows up help information without passing in any args', () => {
--- a/__e2e__/default.test.ts
+++ b/__e2e__/default.test.ts
@@ -11,7 +11,7 @@ beforeEach(() => {
   cleanupSync(DIR);
   writeFiles(DIR, {});
 });
-afterEach(async () => {
+afterEach(() => {
   cleanupSync(DIR);
 });
 
--- a/__e2e__/default.test.ts
+++ b/__e2e__/default.test.ts
@@ -1,13 +1,18 @@
-import {runCLI, getTempDirectory, cleanup, writeFiles} from '../jest/helpers';
+import {
+  runCLI,
+  getTempDirectory,
+  cleanupSync,
+  writeFiles,
+} from '../jest/helpers';
 
 const DIR = getTempDirectory('test_default_behavior');
 
 beforeEach(async () => {
-  cleanup(DIR, false);
+  cleanupSync(DIR);
   writeFiles(DIR, {});
 });
 afterEach(async () => {
-  cleanup(DIR, false);
+  cleanupSync(DIR);
 });
 
 test('shows up help information without passing in any args', () => {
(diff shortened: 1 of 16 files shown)


thymikee:
Weirdly Windows e2e tests started to time out with 5s now. Even though we set the timeout separately for these tests. Looks like misconfiguration from our side, or a Jest bug.

molant:
@thymikee locally I'm having different issues with e2e tests. I've opened #1157

molant:
@thymikee this is green now. I had to make the cleanup synchronous in the e2e tests to not timeout 🤯
It might be that the before/after hooks are not finishing correctly in there when running asynchronously. I tried to debug a bit but you mentioned that some tests are going away and some others seemed to execute non windows paths so it was a bit too confusing.

Hopefully this is good enough, can get merged, and Windows is more reliable from now on 😊

jest/helpers.ts
```diff
@@ -77,7 +80,13 @@ export const makeTemplate = (
     return values[number - 1];
   });
 
-export const cleanup = (directory: string) => rimraf.sync(directory);
+export const cleanup = (directory: string, async = true) => {
```
Esemesek, line 83: Can we have 2 functions named: `cleanupSync` and `cleanup`. Passing a boolean as a function parameter is a code-smell.

packages/cli/src/tools/config/__tests__/index-test.ts
```diff
@@ -51,13 +51,13 @@ const removeString = (config, str) =>
...
 
-beforeEach(() => {
-  cleanup(DIR);
+beforeEach(async () => {
+  await cleanup(DIR);
   jest.resetModules();
   jest.clearAllMocks();
 });
 
-afterEach(() => cleanup(DIR));
```
thymikee, line 60: can you rebase to latest master? we've recently changed these tests, because I started to have consistent fails here, resulting in files not being properly cleaned on time. as a mitigation, I changed the DIR value for every test case

jest/setupE2eTests.js
```diff
@@ -1 +1 @@
-jest.setTimeout(60000);
+jest.setTimeout(120000);
```
molant, line 1: I'll revert this

packages/cli/src/tools/config/__tests__/index-test.ts
molant: I've rebased with master but don't see any changes in here 🤔

thymikee: Oh my, it's not on master yet 😅 https://github.com/react-native-community/cli/pull/1151
No issue then :)

__e2e__/default.test.ts
```diff
@@ -1,13 +1,18 @@
...
+  getTempDirectory,
+  cleanupSync,
+  writeFiles,
+} from '../jest/helpers';
 
 const DIR = getTempDirectory('test_default_behavior');
 
-beforeEach(() => {
-  cleanup(DIR);
+beforeEach(async () => {
```
thymikee, line 10:
```suggestion
beforeEach(() => {
```

__e2e__/default.test.ts
```diff
@@ -1,13 +1,18 @@
...
+  cleanupSync(DIR);
   writeFiles(DIR, {});
 });
-afterEach(() => {
-  cleanup(DIR);
+afterEach(async () => {
```
thymikee, line 14:
```suggestion
afterEach(() => {
```

PR #1143, opened 2020-04-27
feat: added --skip-install option to init command

FedericoMorrone:
Summary:
---------

Added skip install options mentioned in #1142, this option adds the ability to simply skip the installation of dependencies after the template has been set up

Test Plan:
----------

This option doesn't fundamentally change  any mechanics, I ran every test mentioned in [CONTRIBUTING.md](CONTRIBUTING.md), also tested by creating a project and manually installing the dependencies and checking that everything was in place and worked correctly

PR #1161, opened 2020-05-08
tests: e2e test case for init with --skip-install flag

jamesgeorge007:
Summary:
---------
Follow up of #1143 

Test Plan:
----------
This PR adds a test case for the recently landed feature (CLI flag).

### DIFF
--- a/__e2e__/init.test.ts
+++ b/__e2e__/init.test.ts
@@ -94,3 +94,32 @@ test('init --template file with custom directory', () => {
   let dirFiles = fs.readdirSync(path.join(DIR, customPath));
   expect(dirFiles).toEqual(customTemplateCopiedFiles);
 });
+
+test('init skips installation of dependencies with --skip-install', () => {
+  createCustomTemplateFiles();
+  let templatePath = path.resolve(DIR, 'custom', 'template');
+  if (process.platform === 'win32') {
+    templatePath = templatePath.split('\\').join('/');
+  }
+
+  const {stdout} = runCLI(DIR, [
+    'init',
+    '--template',
+    `file://${templatePath}`,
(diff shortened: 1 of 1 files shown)


__e2e__/init.test.ts
```diff
@@ -94,3 +94,32 @@ test('init --template file with custom directory', () => {
...
+  expect(stdout).toContain('Run instructions');
+
+  // make sure we don't leave garbage
+  expect(fs.readdirSync(DIR)).toContain('custom');
+
+  let dirFiles = fs.readdirSync(path.join(DIR, 'TestInit'));
+
+  expect(dirFiles).toEqual(
+    customTemplateCopiedFiles.filter(
+      file => !['node_modules', 'yarn.lock'].includes(file),
```
thymikee, line 122: can we also check if those files exist in previous test "init --template file with custom directory"

jamesgeorge007: We already have it 
https://github.com/react-native-community/cli/blob/2c08a8254cacae74ae2cbdd4e25d8506d4983d15/__e2e__/init.test.ts#L95

jamesgeorge007: I assume you're looking forward to have another test case with the custom directory and `--skip-install` flag.

thymikee: Nah, I don't care too much about it

jamesgeorge007: I'm sorry, what do you expect more from my side in the context of this PR?

thymikee: You're right, now I noticed `customTemplateCopiedFiles` includes these :)

PR #1164, opened 2020-05-11
feat: template option in the format of username/repo

### LINKED ISSUE #1163: Template option in format username/repo
## Describe the Feature
Init command doesn't accept github repo as  `--template` argument, nnly npm packages and local folders. Would be great to pass a repo as argument:

```sh
npx react-native init AwesomeProject --template username/repo
```

## Possible Implementations
??

## Related Issues
??

jamesgeorge007:
Summary:
---------
closes #1163 

Test Plan:
----------
Added a unit test case.

### DIFF
--- a/packages/cli/src/commands/init/templateName.ts
+++ b/packages/cli/src/commands/init/templateName.ts
@@ -7,7 +7,6 @@ const FILE_PROTOCOL = /file:/;
 const TARBALL = /\.tgz$/;
 const VERSION_POSTFIX = /(.*)(-\d+\.\d+\.\d+)/;
 const VERSIONED_PACKAGE = /(@?.+)(@)(.+)/;
-const GIT_REPO = /\//;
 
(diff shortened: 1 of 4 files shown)


packages/cli/src/commands/init/templateName.ts
```diff
@@ -77,6 +78,19 @@ function handleVersionedPackage(versionedPackage: string) {
   };
 }
 
+function handleGitHubRepo(templateName: string) {
+  if (['github', '@'].some(str => templateName.includes(str))) {
```
grabbou, line 82: I don't think our unit test handles this `if` statement, right? Can we have a case for it too?

grabbou: I think we should remove this and also improve the regex to not trigger on URLs, otherwise, this will break and affect those who init with a URL or git.

(author probably tested with Git URL only hence `github` and `@` here)

PR #1166, opened 2020-05-12
feat: assets' filename can contain non-latin letter in android

### LINKED ISSUE #1165: [Android] assets' filename can contain non-latin letter when build bundle.
## Describe the Feature
If the assets' filename in project is not alphabetic character or numeric character, it will load unsuccessfully.

For example:
There are two pictures: `src/images/ユーザー.png` and `src/images/ファイル.png`. Their filenames are in Japanese.
When build bundle in android platform, their dest paths are same, it is `drawable-mdpi/src_images_.png`.
One file has been overridden by another file. So that it will load unsuccessfully.

But that's OK in ios platform.

## Possible Implementations
Maybe it could change logic `getAndroidResourceIdentifier` function in file [`packages/cli/src/commands/bundle/assetPathUtils.ts`](https://github.com/react-native-community/cli/blob/master/packages/cli/src/commands/bundle/assetPathUtils.ts#L67).

```typescript
function getAndroidResourceIdentifier(asset: PackagerAsset): string {
  const folderPath = getBasePath(asset);
  return `${folderPath}/${asset.name}`
    .toLowerCase()
    .replace(/\//g, '_') // Encode folder structure in file name
    .replace(/([^a-z0-9_])/g, '') // Remove illegal chars
    .replace(/^assets_/, ''); // Remove "assets_" prefix
}
```
it should replace `asset.name` with `asset.hash` here.
In the case, the two assets in above examle:
```
{
  __packager_asset: true,
  fileSystemLocation: '/Project/src/images',
  httpServerLocation: '/assets/src/images',
  width: 40,
  height: 40,
  scales: [ 1 ],
  files: [
    '/Project/src/images/ユーザー.png'
  ],
  hash: '565eb945bdbabf6715f01965374ceea3',
  name: 'ユーザー',
  type: 'png'
}

{
  __packager_asset: true,
  fileSystemLocation: '/Project/src/images',
  httpServerLocation: '/assets/src/images',
  width: 36,
  height: 36,
  scales: [ 1 ],
  files: [
    '/Project/src/images/ファイル.png'
  ],
  hash: '9be5f04e7ca5623920cf5cc98b6d3add',
  name: 'ファイル',
  type: 'png'
}
```

It will output two files `drawable-mdpi/src_images_565eb945bdbabf6715f01965374ceea3.png` and `drawable-mdpi/src_images_9be5f04e7ca5623920cf5cc98b6d3add.png`.
So it will not override again.

## Related Issues

wusuopu:
Summary:
---------
Closes #1165 

Test Plan:
----------
Added a unit test case.

### DIFF
--- a/packages/cli/src/commands/bundle/__tests__/getAssetDestPathAndroid-test.ts
+++ b/packages/cli/src/commands/bundle/__tests__/getAssetDestPathAndroid-test.ts
@@ -17,6 +17,7 @@ const path = require('path');
 describe('getAssetDestPathAndroid', () => {
   it('should use the right destination folder', () => {
     const asset = {
+      hash: '565eb945bdbabf6715f01965374ceea3',
       name: 'icon',
       type: 'png',
       httpServerLocation: '/assets/test',
@@ -39,6 +40,7 @@ describe('getAssetDestPathAndroid', () => {
 
   it('should lowercase path', () => {
     const asset = {
+      hash: '565eb945bdbabf6715f01965374ceea3',
       name: 'Icon',
       type: 'png',
       httpServerLocation: '/assets/App/Test',
@@ -51,6 +53,7 @@ describe('getAssetDestPathAndroid', () => {
 
   it('should remove `assets/` prefix', () => {
     const asset = {
+      hash: '565eb945bdbabf6715f01965374ceea3',
       name: 'icon',
       type: 'png',
       httpServerLocation: '/assets/RKJSModules/Apps/AndroidSample/Assets',
@@ -61,6 +64,7 @@ describe('getAssetDestPathAndroid', () => {
 
   it('should put non-drawable resources to `raw/`', () => {
     const asset = {
+      hash: '565eb945bdbabf6715f01965374ceea3',
       name: 'video',
       type: 'mp4',
       httpServerLocation: '/assets/app/test',
@@ -73,6 +77,7 @@ describe('getAssetDestPathAndroid', () => {
 
   it('should handle assets with a relative path outside of root', () => {
     const asset = {
+      hash: '565eb945bdbabf6715f01965374ceea3',
       name: 'icon',
       type: 'png',
       httpServerLocation: '/assets/../../test',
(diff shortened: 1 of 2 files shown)


packages/cli/src/commands/bundle/__tests__/getAssetDestPathAndroid-test.ts
```diff
@@ -61,25 +66,44 @@ describe('getAssetDestPathAndroid', () => {
...
     const asset = {
+      hash: '565eb945bdbabf6715f01965374ceea3',
       name: 'video',
       type: 'mp4',
       httpServerLocation: '/assets/app/test',
     };
 
     expect(getAssetDestPathAndroid(asset, 1)).toBe(
-      path.normalize('raw/app_test_video.mp4'),
+      path.normalize('raw/app_test_565eb945bdbabf6715f01965374ceea3.mp4'),
```
thymikee, line 76: The change of behavior for regular files is not what we expect here I think. Can we fall back to using a hash only if `name` contains illegal characters?

wusuopu: @thymikee this is a good idea.

wusuopu:
Relation Pull Request: https://github.com/facebook/react-native/pull/28877
it also needs to change react-native.

packages/cli/src/commands/bundle/__tests__/getAssetDestPathAndroid-test.ts
```diff
@@ -39,18 +40,22 @@ describe('getAssetDestPathAndroid', () => {
...
+      hash: '565eb945bdbabf6715f01965374ceea3',
       name: 'Icon',
       type: 'png',
       httpServerLocation: '/assets/App/Test',
     };
 
     expect(getAssetDestPathAndroid(asset, 1)).toBe(
-      path.normalize('drawable-mdpi/app_test_icon.png'),
+      path.normalize(
+        'drawable-mdpi/app_test_565eb945bdbabf6715f01965374ceea3.png',
```
thymikee, line 51: The `name` here seems perfectly valid, so this hash is unexpected I'd say

packages/cli/src/commands/bundle/assetPathUtils.ts
```diff
@@ -66,7 +67,12 @@ function getAndroidResourceFolderName(
 
 function getAndroidResourceIdentifier(asset: PackagerAsset): string {
   const folderPath = getBasePath(asset);
-  return `${folderPath}/${asset.name}`
+  let name = asset.name;
+  if (!name.match(/^[a-z0-9_]+$/)) {
```
thymikee, line 71: we need to lower case the name first, or adjust the regex

thymikee (approved):

thymikee:
Do we need to coordinate releasing it with facebook/react-native#28877? There's a small edge case where this would break: someone has a single image with unsupported characters and cli would resolve to a hash, but image asset resolver would assign fallback `_` name.

wusuopu:
@thymikee

wusuopu:
@thymikee 
Due to both react-native-cli and react-native have changed behaviour that resolve asset.
So it should coordinate releasing it with https://github.com/facebook/react-native/pull/28877.

grabbou:
Would be good to ping somebody on the RN side to go ahead and test that PR. I think we're blocked here until that happens.

andriytsaryov (approved):

grabbou:
Unfortunately, I am closing this one since original PR got closed for inactivity. Happy to reopen, but let's focus on merging RN changes first. 

https://github.com/facebook/react-native/pull/28877

PR #1167, opened 2020-05-12
feat: allow uppercase in android package names

willcaul:
Summary:
---------

I'm proposing to allow uppercase letters in android package names. Oracle states that lowercase letters are used by convention here: https://docs.oracle.com/javase/tutorial/java/package/namingpkgs.html.  But the spec does not seem to prohibit the use of uppercase letters. My company doesn't follow the convention and uses an uppercase letter like `com.company.myPackage`. If possible I'd like to update the validation so that my company won't see the warning about an invalid package name on every build.

### DIFF
--- a/packages/platform-android/src/commands/runAndroid/index.ts
+++ b/packages/platform-android/src/commands/runAndroid/index.ts
@@ -25,7 +25,7 @@ import warnAboutManuallyLinkedLibs from '../../link/warnAboutManuallyLinkedLibs'
 
 // Validates that the package name is correct
 function validatePackageName(packageName: string) {
-  return /^[a-z][a-z0-9_]*(\.[a-z][a-z0-9_]*)+$/.test(packageName);
+  return /^[a-zA-Z][a-zA-Z0-9_]*(\.[a-zA-Z][a-zA-Z0-9_]*)+$/.test(packageName);
 }
 
 function displayWarnings(config: Config, args: Flags) {


thymikee:
cc @Krizzu @dulmandakh @troZee

troZee:
>  But the spec does not seem to prohibit the use of uppercase letters. 
I agree, but package names are also used in conjunction with filesystem directories, which often have restrictions on character case themselves. 

From my POV, it is a convention and you can stick with it or not. It is your decision. In CLI, we would like to kindly inform user about incorrect package name ([here you can find an example](https://gist.github.com/rishabhmhjn/8663966) ) It is just only warning, hence you can ignore it.

thymikee:
I agree with @troZee. It's a warning, so you can freely ignore it. What we could improve is the wording to be more of a recommendation, so that power-users can be sure they can ignore it without issues.

willcaul:
Ok makes sense. Thanks guys!

grabbou:
Question is whether we consider use of uppercase letters a warning? I guess the purpose of this validation was to filtrate invalid names that would make Android project not build (e.g. space).

Also, what is the state of the PR now? Working on a new implementation? Closing?

willcaul:
based on the feedback from @troZee and @thymikee, I thought we'd just close this, but I'll defer to you

thymikee:
> What we could improve is the wording to be more of a recommendation

This still holds true, but we can tackle this in another PR (maybe you'd like to send it? 🙂) Thanks for the contribution anyway!

PR #1177, opened 2020-05-25
Add an ability to add custom scope methods and configurations in android dependencies autolinking

michalchudziak:
Summary:
---------

Currently there's no way to use any scope method for autolinked dependencies except for hardcoded `implementation`. This PR adds an ability to choose custom scope method and configuration for `ReactNativeModules.addReactNativeModuleDependencies` gradle method.

Test Plan:
----------
Following calls should work. (for corresponding project configuration).

**REQUIRED**
`app/build.gradle`
```groovy
apply from: file("../../../../node_modules/@react-native-community/cli-platform-android/native_modules.gradle");
applyNativeModulesAppBuildGradle(project, "../../..")
```
Expected: Not to throw any errors.

**OPTIONAL:**
```groovy
apply from: file("../../../../node_modules/@react-native-community/cli-platform-android/native_modules.gradle");
applyNativeModulesAppBuildGradle(project, "../../..", "api")
```

Expected: Not to throw any errors if `implementation` method is replaceable with `api` for given project. 

```groovy
apply from: file("../../../../node_modules/@react-native-community/cli-platform-android/native_modules.gradle");
applyNativeModulesAppBuildGradle(project, "../../..", "embed", "default")
```

Expected: Not to throw any errors using custom scope method from https://github.com/kezong/fat-aar-android.

### DIFF
--- a/packages/platform-android/native_modules.gradle
+++ b/packages/platform-android/native_modules.gradle
@@ -100,12 +100,15 @@ class ReactNativeModules {
   /**
    * Adds the react native modules as dependencies to the users `app` project
    */
-  void addReactNativeModuleDependencies(Project appProject) {
+  void addReactNativeModuleDependencies(Project appProject, String scopeMethod, String configuration) {
     reactNativeModules.forEach { reactNativeModule ->
       def nameCleansed = reactNativeModule["nameCleansed"]
       appProject.dependencies {
-        // TODO(salakar): are other dependency scope methods such as `api` required?
-        implementation project(path: ":${nameCleansed}")
+        if (configuration != null) {
+          "${scopeMethod}" project(path: ":${nameCleansed}", configuration: "${configuration}")
+        } else {
+          "${scopeMethod}" project(path: ":${nameCleansed}")
+        }
       }
     }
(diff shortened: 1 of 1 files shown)


thymikee:
Hey @michalchudziak, thanks for the contribution. Looks like this is all-or-nothing affecting all autolinked deps. I think we could go with it for now, but this seems like an opportunity to put it into the dependency config, so we could do something like:
```js
module.exports = {
  dependencies: {
    "native-dep": {
      platforms: { 
        android: { scopeMethod: "api", scopeMethodConfiguration: "default" } 
      }
    }
  }
}
```

grabbou:
Even better, why don't we make this `implementation project(path: ":${nameCleansed}")` part of the configuration?

You can set it on per project basis.

The default value is `implementation project(path: ":${nameCleansed}")` and can be overriden by a project.

thymikee:
Converted this to a draft, because it's still very much work in progress. Hope you don't mind

grabbou:
Closing due to lack of activity here. Hope we can revisit once it becomes a bigger blocker

PR #1183, opened 2020-05-28
feat: detach doctor command

thymikee:
Summary:
---------

Detaching the `doctor` command so it's possible to run without being in react-native project like this:

```sh
npx @react-native-community/cli doctor
npx react-native doctor
```

cc @brentvatne since you brought that up recently

Test Plan:
----------

Tests updated

### DIFF
--- a/packages/cli/src/commands/index.ts
+++ b/packages/cli/src/commands/index.ts
@@ -23,6 +23,7 @@ export const projectCommands = [
   upgrade,
   info,
   config,
+  doctor,
 ] as Command[];
 
 export const detachedCommands = [init, doctor] as DetachedCommand[];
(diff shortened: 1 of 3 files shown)


packages/cli/src/commands/index.ts
```diff
@@ -26,7 +23,6 @@ export const projectCommands = [
   upgrade,
   info,
   config,
-  doctor,
```
grabbou, line 29: This will make `doctor` not visible when run inside `react-native` project. Is this expected? I think it's a handy feature to be able to troubleshoot some issues and report the environment.

thymikee: Good point, will bring it back

PR #1197, opened 2020-06-11
Allow drive letter on windows launchEditor calls

acoates-ms:
Summary:
---------

When running on windows, the launchEditor functionality doesn't work with `react-native-windows` since all the paths being used are absolute paths, which include a drive letter.  The drive letter doesn't match the WINDOWS_FILE_NAME_WHITELIST regex.

This slightly modifies the WINDOWS_FILE_NAME_WHITELIST to allow drive letter prefixes.


grabbou:
Thanks!

PR #1203, opened 2020-06-17
fix: Podfile.lock paths are no longer absolute

grabbou:
Summary:
---------

Small regression introduced by a previous change to the file. `reactNativePath` is absolute (as returned by the CLI configuration) in order to be used in various places.

The path used by `Podfile` has to be relative to its location, hence the change. I am leaving the object as a return value to avoid changing RN files and make it easy to customize in the future.

Context: The output of running `use_native_modules` is used here: https://github.com/facebook/react-native/tree/master/template/ios/Podfile#L9

### DIFF
--- a/packages/platform-ios/src/config/__tests__/native_modules.test.ts
+++ b/packages/platform-ios/src/config/__tests__/native_modules.test.ts
@@ -86,7 +86,7 @@ function describeIfSupportedEnv() {
   }
   console.warn(
     '[!] The `native_modules.rb` tests are disabled – ensure you have `ruby` ' +
-      'and the `cocoapods` gem installed in order to run them.',
+    'and the `cocoapods` gem installed in order to run them.',
   );
   return describe.skip;
 }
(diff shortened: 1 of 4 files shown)


grabbou:
Once running `pod install`, you will notice relative paths are again used in `Podfile.lock`.

grabbou:
Thanks, @alloy, I've added a simple test suite. Let me know if that works.

Unfortunately, it turned out to be rather complex change and I had to perform some changes to the logic of the testing script itself. Not really proud of them, but I don't think there was a better way.

Let me know if you have any other suggestions or this is good to go.

packages/platform-ios/src/config/__tests__/native_modules.test.ts
```diff
@@ -97,6 +114,7 @@ describeIfSupportedEnv()('native_modules.rb', () => {
   beforeEach(() => {
     runConfig = {
       dependencyConfig: {
+        reactNativePath: '/root/app/node_modules/react_native',
```
thymikee, line 117: you can reuse REACT_NATIVE_ROOT and rename `react_native` to `react-native` there

grabbou: Oh, that was my plan! IDK why I didn't replace it here

PR #1216, opened 2020-07-08
Add windows logic to guess the users preferred editor for launchEditor

acoates-ms:
Summary:
---------

There is logic for guessing an appropriate text editor on macOS.  This adds similar logic for windows.  Without this, developers have to manually set the `REACT_EDITOR` environment variable in order to navigate to code from redbox/logbox.

This change roughly aligns with the logic on macOS.  If `REACT_EDITOR` is set we default to using that.  But if the developer is running in order VSCode, SublimeText, Visual Studio, or lastly notepad, then we assume that we should use that editor.

Also fixed an issue where if the linenumber wasn't specified, the workspace would not be passed to vscode, which could cause the file to open in the incorrect vscode window.

Test Plan:
----------

Tested locally on windows with these apps running  / not running or `REACT_EDITOR` set, and clicking on a stack from within a logbox.

### DIFF
--- a/packages/tools/src/launchEditor.ts
+++ b/packages/tools/src/launchEditor.ts
@@ -42,6 +42,15 @@ const COMMON_EDITORS: Record<string, string> = {
   '/Applications/WebStorm.app/Contents/MacOS/webstorm': 'webstorm',
 };
 
+// Map of process image name used to identify already running instances of the editor
+// And an editor id which is used to determine which arguments to add to the commandline
+const COMMON_WINDOWS_EDITORS: Record<string, string> = {
+  'Code.exe': 'code',
+  'sublime_text.exe': 'subl',
+  'devenv.exe': 'devenv',
+  'notepad.exe': 'notepad',
+};
+
 // Transpiled version of: /^([a-zA-Z]+:)?[\p{L}0-9/.\-_\\]+$/u
 // Non-transpiled version requires support for Unicode property regex. Allows
 // alphanumeric characters, periods, dashes, slashes, and underscores, with an optional drive prefix
@@ -88,6 +97,8 @@ function getArgumentsForLineNumber(
         ['-g', `${fileName}:${lineNumber}`],
         workspace,
       );
+    case 'devenv':
+      return ['/EDIT', fileName];
     // For all others, drop the lineNumber until we have
     // a mapping above, since providing the lineNumber incorrectly
     // can result in errors or confusing behavior.
@@ -102,10 +113,10 @@ function guessEditor() {
     return shellQuote.parse(process.env.REACT_EDITOR);
   }
 
-  // Using `ps x` on OSX we can find out which editor is currently running.
-  // Potentially we could use similar technique for Windows and Linux
-  if (process.platform === 'darwin') {
-    try {
+  try {
+    // Using `ps x` on OSX we can find out which editor is currently running.
+    // Potentially we could use similar technique for Linux
+    if (process.platform === 'darwin') {
       const output = execSync('ps x').toString();
       const processNames = Object.keys(COMMON_EDITORS);
       for (let i = 0; i < processNames.length; i++) {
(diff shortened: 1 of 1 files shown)


packages/tools/src/launchEditor.ts
```diff
@@ -164,6 +200,58 @@ function findRootForFile(
...
+        )
+          .toString()
+          .trim();
+      }
+    }
+  } catch (error) {
+    // ignore
+  }
+
+  // Just use what the user specified.. it'll probably fail, but we'll show some help text when we fail to launch it
```
vmoroz, line 251: nit: please check the punctuation. Also, the end of the comment here.

packages/tools/src/launchEditor.ts
```diff
@@ -164,6 +200,58 @@ function findRootForFile(
...
+  } catch (error) {
+    // ignore
+  }
+
+  try {
+    const editorNames = Object.values(COMMON_WINDOWS_EDITORS);
+    const editorImageNames = Object.keys(COMMON_WINDOWS_EDITORS);
+    for (let i = 0; i < editorNames.length; i++) {
+      const editorName = editorNames[i];
+      if (editor === editorName) {
```
vmoroz, line 226: I wonder if the case-sensitive comparison is good here. I never can predict what casing is used in Windows. It often seems to be tool-specific.

packages/tools/src/launchEditor.ts
```diff
@@ -114,9 +141,21 @@ function guessEditor() {
...
       }
-    } catch (error) {
-      // Ignore...
+    } else if (process.platform === 'win32') {
+      const output = execSync(
+        'tasklist /NH /FO CSV /FI "SESSIONNAME ne Services"',
+      ).toString();
+
+      const processNames = Object.keys(COMMON_WINDOWS_EDITORS);
+      for (let i = 0; i < processNames.length; i++) {
```
NickGerleman, line 150: Nit: for of 

Same below

packages/tools/src/launchEditor.ts
```diff
@@ -114,9 +141,21 @@ function guessEditor() {
...
+        const processName = processNames[i];
+        if (output.indexOf(processName) !== -1) {
```
NickGerleman, line 152: Nit: "output.includes" instead of indexof

Though this would also seemingly catch substring. Can we split the task list and look for an exact match?

packages/tools/src/launchEditor.ts
```diff
@@ -164,6 +216,56 @@ function findRootForFile(
...
+  } catch (error) {
+    // ignore
+  }
+
+  try {
+    const editorNames = Object.values(COMMON_WINDOWS_EDITORS);
+    const editorImageNames = Object.keys(COMMON_WINDOWS_EDITORS);
+    for (let i = 0; i < editorNames.length; i++) {
+      const editorName = editorNames[i];
+      if (editor.toLowerCase() === editorName.toLowerCase()) {
```
NickGerleman, line 239: Do we also need case insensitivity when guessing the editor?

acoates-ms: I dont think so.  The case there is based on the actual case of the executable.  So unless someone went and renamed Code.exe to code.exe, (why would someone do that?) then it should match.  This one it makes sense to leave as case insensitive since someone could have set REACT_EDITOR=Code or REACT_EDITOR=code and this would go look up the correct executable path for it if it isn't already in the users path.

packages/tools/src/launchEditor.ts
acoates-ms: I was using the same logic as the macOS version.  But yes, I can go ahead and make the windows logic more correct.

packages/tools/src/launchEditor.ts
```diff
@@ -96,27 +107,56 @@ function getArgumentsForLineNumber(
...
+
+      const runningProcesses = output
+        .split('\n')
+        .map(line => line.replace(/^"|".*\r$/gm, ''));
+      const processNames = Object.keys(COMMON_WINDOWS_EDITORS);
+      for (const processName of processNames) {
+        if (runningProcesses.includes(processName)) {
+          return [COMMON_WINDOWS_EDITORS[processName]];
+        }
+      }
```
thymikee, line 156: makes sense to reuse this logic from https://github.com/react-native-community/cli/pull/1222 maybe?

alloy:
Going ahead and merge this 👍

PR #1230, opened 2020-07-29
Make run-android work properly on Linux

vijfhoek:
Summary:
---------

Some terminals did not start when `react-native run-android` was called, because they expect the arguments to not be quoted. This is defined in e.g. xterm, but xterm does not care. Some other terminals however, like alacritty and rxvt-unicode, do care, and will crash with an error.

execa was called synchronously, which didn't allow the command to continue building and running the Android app. This was already fixed by the Windows command by removing .sync, so I did the same for Linux. macOS does not need this, because `open` quits immediately after opening the terminal.

Since `.sync` is not used on the execa call, the error also does not get caught in the try-catch. This was fixed by using a Promise-style `.catch` on the returned value.

Test Plan:
----------

- Build react-native-cli: `yarn build`
- Link into a react-native project: `yarn link "@react-native-community/cli-platform-ios" "@react-native-community/cli-platform-android" "@react-native-community/cli" "@react-native-community/cli-server-api" "@react-native-community/cli-types" "@react-native-community/cli-tools" "@react-native-community/cli-debugger-ui"`
- Run `yarn android --terminal foo` to test without terminal
- Run `yarn android --terminal xterm` for a terminal that does support -e with a string
- Run `yarn android --terminal alacritty` for one that does not.

### DIFF
--- a/packages/platform-android/src/commands/runAndroid/index.ts
+++ b/packages/platform-android/src/commands/runAndroid/index.ts
@@ -325,15 +325,16 @@ function startServerInNewWindow(
   }
   if (process.platform === 'linux') {
     // Awaiting this causes the CLI to hang indefinitely, so this must execute without .sync.
-    const process = execa(terminal, ['-e', 'sh', launchPackagerScript], {
+    return execa(terminal, ['-e', 'sh', launchPackagerScript], {
       ...procConfig,
       detached: true,
-    });
-
-    return process.catch(_ => {
+    }).catch(_ => {
       // If the terminal could not be opened, fall back to running it inside the process.
       const processSh = execa('sh', [launchPackagerScript], procConfig);
+
+      // Unreference the process to avoid quitting when run-android is done.
       processSh.unref();
+
       return processSh;
     });
   }
--- a/packages/platform-android/src/commands/runAndroid/index.ts
+++ b/packages/platform-android/src/commands/runAndroid/index.ts
@@ -324,15 +324,18 @@ function startServerInNewWindow(
     }
   }
   if (process.platform === 'linux') {
-    try {
-      return execa.sync(terminal, ['-e', `sh ${launchPackagerScript}`], {
-        ...procConfig,
-        detached: true,
-      });
-    } catch (error) {
-      // By default, the child shell process will be attached to the parent
-      return execa.sync('sh', [launchPackagerScript], procConfig);
-    }
+    // Awaiting this causes the CLI to hang indefinitely, so this must execute without .sync.
+    const process = execa(terminal, ['-e', 'sh', launchPackagerScript], {
+      ...procConfig,
+      detached: true,
+    });
+
+    return process.catch(_ => {
+      // If the terminal could not be opened, fall back to running it inside the process.
+      const processSh = execa('sh', [launchPackagerScript], procConfig);
+      processSh.unref();
+      return processSh;
+    });
   }
   if (/^win/.test(process.platform)) {
     // Awaiting this causes the CLI to hang indefinitely, so this must execute without await.
--- a/packages/platform-android/src/commands/runAndroid/index.ts
+++ b/packages/platform-android/src/commands/runAndroid/index.ts
@@ -102,19 +102,18 @@ async function runAndroid(_argv: Array<string>, config: Config, args: Flags) {
     } else {
       // result == 'not_running'
       logger.info('Starting JS server...');
-      try {
-        startServerInNewWindow(
-          args.port,
-          args.terminal,
-          config.reactNativePath,
-        );
-      } catch (error) {
+
+      // Using a Promises-style catch here so the process runs in the background - awaiting it will cause it to hang.
+      startServerInNewWindow(
+        args.port,
+        args.terminal,
+        config.reactNativePath,
+      ).catch(error => {
         logger.warn(
-          `Failed to automatically start the packager server. Please run "react-native start" manually. Error details: ${
-            error.message
-          }`,
+          'Failed to automatically start the packager server. Please run "react-native start" manually. ' +
+            `Error details: ${error.message}`,
         );
-      }
+      });
     }
     return buildAndRun(args, androidProject);
   });
@@ -271,7 +270,7 @@ function installAndLaunchOnDevice(
   tryLaunchAppOnDevice(selectedDevice, packageName, adbPath, args);
 }
 
-function startServerInNewWindow(
+async function startServerInNewWindow(
   port: number,
   terminal: string,
   reactNativePath: string,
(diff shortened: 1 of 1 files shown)


packages/platform-android/src/commands/runAndroid/index.ts
```diff
@@ -324,15 +324,19 @@ function startServerInNewWindow(
...
-      });
-    } catch (error) {
-      // By default, the child shell process will be attached to the parent
-      return execa.sync('sh', [launchPackagerScript], procConfig);
-    }
+    // Awaiting this causes the CLI to hang indefinitely, so this must execute without .sync.
+    return execa(terminal, ['-e', 'sh', launchPackagerScript], {
+      ...procConfig,
+      detached: true,
+    }).catch(_ => {
```
thymikee, line 331: can we convert this to async try/catch?

packages/platform-android/src/commands/runAndroid/index.ts
```diff
@@ -324,15 +324,19 @@ function startServerInNewWindow(
...
+      // If the terminal could not be opened, fall back to running it inside the process.
+      const processSh = execa('sh', [launchPackagerScript], procConfig);
```
thymikee, line 333: we should guard this with try/catch as well

packages/platform-android/src/commands/runAndroid/index.ts
```diff
@@ -324,15 +324,19 @@ function startServerInNewWindow(
...
-      return execa.sync(terminal, ['-e', `sh ${launchPackagerScript}`], {
-        ...procConfig,
-        detached: true,
-      });
-    } catch (error) {
-      // By default, the child shell process will be attached to the parent
-      return execa.sync('sh', [launchPackagerScript], procConfig);
-    }
+    // Awaiting this causes the CLI to hang indefinitely, so this must execute without .sync.
+    return execa(terminal, ['-e', 'sh', launchPackagerScript], {
```
thymikee, line 328: The CLI hangs indefinitely because launching packager is a persistent process, right? Maybe we could run the script in the void, unref it and then return?

sijmn: The problem of unref'ing the script is that it'll keep running once run-android quits, which will start a ghost packager in the background, not attached to any shell. Don't think that's what we want either

packages/platform-android/src/commands/runAndroid/index.ts
sijmn: Actually no - the function above catches the error already, which will show a nice error for it. It didn't fully work, but since I remade this function to be async/await, I changed it to be a promise-style catch (with a comment to explain while). Push incoming

packages/platform-android/src/commands/runAndroid/index.ts
```diff
@@ -314,24 +313,34 @@ function startServerInNewWindow(
...
     try {
-      return execa.sync(terminal, ['-e', `sh ${launchPackagerScript}`], {
+      return await execa(terminal, ['-e', 'sh', launchPackagerScript], {
         ...procConfig,
         detached: true,
       });
-    } catch (error) {
-      // By default, the child shell process will be attached to the parent
-      return execa.sync('sh', [launchPackagerScript], procConfig);
+    } catch (_) {
```
thymikee, line 332: IIRC we can use optional catch binding:
```suggestion
    } catch {
```

sijmn: Oh cool, didn't know that one

packages/platform-android/src/commands/runAndroid/index.ts
```diff
@@ -314,24 +313,34 @@ function startServerInNewWindow(
...
     } catch (error) {
-      return execa.sync('open', [launchPackagerScript], procConfig);
+      return await execa('open', [launchPackagerScript], procConfig);
     }
   }
   if (process.platform === 'linux') {
+    // Awaiting this causes the CLI to hang indefinitely, so this must execute without .sync.
     try {
-      return execa.sync(terminal, ['-e', `sh ${launchPackagerScript}`], {
+      return await execa(terminal, ['-e', 'sh', launchPackagerScript], {
```
thymikee, line 328: `return await` is quite redundant, since we're returning a promise anyway. Can we simplify it to just `return`?

sijmn: Don't think so, since we want the try/catch to catch its error. Not sure if that happens if you just `return`

thymikee: Hm, ok.

sijmn:
I just found an edge-case if the fall-back happens (e.g. specifying a non-existent terminal) and the rest of the command fails and quits early, which causes the packager process to keep running after run-android quits.

thymikee:
Need to catch the fallback as well I guess. No rush, feel free to figure out the solution you're happy with :)

sijmn:
I've been thinking this over, and it might be a good solution to just bail out when it can't find the terminal, and ask the user to either specify a terminal with `--terminal`, or run `react-native start` themselves.

Other than that, we could also just try a bunch of terminals - perhaps we could ~steal~ borrow the list from [i3-sensible-terminal](https://github.com/i3/i3/blob/next/i3-sensible-terminal) - and then bail out if none can be found.

thymikee:
I'm fine with bailing. Would like to avoid adding a dependency which is not maintained for 2 years now. Maybe we can vendor this list and use similarly to the list of editors? I'm not a big fan of this idea though, because it's extra maintenance.

thymikee:
@sijmn ping :)

packages/platform-android/src/commands/runAndroid/index.ts
grabbou, line 328: Why `return await` instead of juts `return`?

packages/platform-android/src/commands/runAndroid/index.ts
```diff
@@ -102,19 +102,18 @@ async function runAndroid(_argv: Array<string>, config: Config, args: Flags) {
...
-      try {
-        startServerInNewWindow(
-          args.port,
-          args.terminal,
-          config.reactNativePath,
-        );
-      } catch (error) {
+
+      // Using a Promises-style catch here so the process runs in the background - awaiting it will cause it to hang.
+      startServerInNewWindow(
```
grabbou, line 107: So the idea is to run this function and not wait till it finishes, but return with `buildAndRun` on line 118?

packages/platform-android/src/commands/runAndroid/index.ts
thymikee: https://github.com/react-native-community/cli/pull/1230#discussion_r462405982 

Interesting behavior, isn't it?

without await:
image

with await:
image

grabbou:
Going to close this one for now due to inactivity. Let's reopen/revisit when there's more activity here.

PR #1231, opened 2020-07-30
Allow platforms and community packages to provide additional health checks

acoates-ms:
Summary:
---------

Out of tree platforms, and specific community modules may have their own set of health checks that could help developers diagnose issues and be more productive.  In particular I am looking at how to have react-native-windows provide checks to doctor.  This change allows any out of tree platform to be able to provide their own checks.

Some talking points:
- I added the extension point so that any package could add doctor health checks rather than just for platforms.  I could see that other packages might also want to add healthChecks for any specific native dependencies, or other checks.  
- This does increase the public API for the CLI.  So, there should be a discussion of the API I'm exposing.
- I slightly modified the API for RunAutomaticFix to provide the logManualInstallation method as an arg.  In addition to the Ora instance, this seems to be one of the primary APIs that RunAutomaticFixes use to interact with the CLI.
- I did a lame copy/paste of the typings for Ora in the cli-types project.  I could also just make ora a pacakge dependency of cli-types, but that seems to go against the spirit of that pacakge.

Test Plan:
----------

I'm posting this early because I expect lots of feedback.  I've run this locally with changes to react-native-windows to run through the end to end of adding health checks in react-native-windows and them showing up in doctor.


thymikee:
Nice! Kindly asking @lucasbento @molant to review :)

docs/healthChecks.md
```diff
@@ -0,0 +1,147 @@
...
+            needsToBeFixed: !isBarInstalled()
+          }),
+          runAutomaticFix: async ({loader}) => {
+            await installBar();
+            loader.succeed();
+          },
+        }
+      ],
+};
+```
```
lucasbento, line 29: This is so awesome.

packages/cli/src/commands/doctor/healthchecks/index.ts
```diff
@@ -22,33 +26,48 @@ type Options = {
...
-          label: 'iOS',
-          healthchecks: [xcode, cocoaPods, iosDeploy],
-        },
-      }
-    : {}),
-});
+export const getHealthchecks = ({contributor}: Options): Healthchecks => {
+  let additionalChecks: HealthCheckCategory[] = [];
+
+  // Todo: maybe detached commands should take the config as an arg, thats null if running in detacted.
```
lucasbento, line 32: Can you clarify what detached means in this context?

lucasbento (commented):
Sorry for taking so long to review this, I checked this PR multiple times and I was sure that I had already submitted a review.

I don't really have much to comment here besides that this is an awesome initiative.

The thing that we would need to work on the most is the documentation, I think the API can get pretty confusing to understand without some visuals (screenshots of the list, the loading part, how errors are printed out, etc).

I would love to have this merged soon as it's a major contribution and will increase the value of `react-native doctor` greatly.

packages/cli/src/commands/doctor/healthchecks/index.ts
acoates-ms: Detached is basically when you run doctor without a local RN package.  You can run `npx react-native doctor` before you init a project.

acoates-ms: Basically the config is project specific.  And when running in detached mode there is no app package, and so there is no config.  The CLI already runs loadConfig to try to get the local config and pass it to the commands.  But in the case of doctor, since it can run without an app package, it doesn't have the config passed to it currently.  The comment is saying that that means we end up loading the config twice in the non-detached case.  Once by the main part of the CLI.  Then again within doctor, since the original wasn't passed though. -- I'll open an issue on this and remove the todo.

acoates-ms:
I added a little more documentation, but I think its probably something we can add to later.  Hopefully once this merges+publishes I can add links to samples etc to at least react-native-windows where we will have a bunch of additional health checks.

packages/cli-types/src/index.ts
```diff
@@ -120,6 +122,100 @@ export type ProjectConfig = {
...
+    Memory: string;
+    Shell: AvailableInformation;
+  };
+  Binaries: {
+    Node: AvailableInformation;
+    Yarn: AvailableInformation;
+    npm: AvailableInformation;
+    Watchman: AvailableInformation;
+  };
+  SDKs: {
```
asklar, line 146: is the idea that this is what will get asked of envinfo? if so, please make sure that info continues to report the windows properties we care about: Windows SDKs, VS installs, dev mode reg keys.

packages/cli-types/src/ora.ts
```diff
@@ -0,0 +1,154 @@
+/**
+ * The types exported from here are all from the Ora package.
+ * Alternatively we could make this pacakge depend on the Ora
+ * pacakge just to export the actual types.
```
grabbou, line 4: Wouldn't re-exporting from a real package be easier from the maintenance perspective? I feel like we depend on Ora anyway somewhere, so no added overhead here.

grabbou (approved):
Love this feautre. LGTM. Just a quick question on the type defs itself (left somewhere in the review) and this one is good to go from my side!

grabbou:
PS. Looks like a little rebase to resolve conflicts might be needed

PR #1232, opened 2020-07-31
Choose which device to run the app on

lucasbento:
Summary:
---------

This PR adds a feature to output a list of available devices when building so that the user can choose which device to run the app on.

### iOS
Kapture 2020-07-31 at 10 19 21

### Android

Kapture 2020-07-31 at 10 22 57

Discussion points:
---------

1. Is this valuable to the CLI?
> The reason I'm submitting this is because I have wanted this for a while and I need to open the IDEs (Xcode & Android Studio) *every single time* I need to choose a different device to run the app on.
1. Is the `--device` flag the best way to make this feature available?
> I'm not sure and I'm definitely open for opinions, I went through the path that doesn't get in the way of the users but can still be valuable, perhaps we could do this under `--choose-device`?

Test Plan:
----------

1. In the CLI folder, run: `yarn link-packages`;
1. In your app project, run: `yarn link "@react-native-community/cli-platform-ios" "@react-native-community/cli-platform-android" "@react-native-community/cli" "@react-native-community/cli-server-api" "@react-native-community/cli-types" "@react-native-community/cli-tools" "@react-native-community/cli-debugger-ui"`;
1. Still on the app project, run `npx react-native start --watchFolders=/path/to/cli/`;
1. And then: `npx react-native run-ios --device` or `npx react-native run-android --device`.

#### Test cases

- Run on an emulator/simulator that has not been started;
- Run on an emulator/simulator that has already been started;
- [Android] Run on an emulator that has not been started and close it as soon as it opens (should show an error);
- Run on a physical device (not properly tested on Android).

### DIFF
--- a/packages/platform-android/src/commands/runAndroid/adb.ts
+++ b/packages/platform-android/src/commands/runAndroid/adb.ts
@@ -41,6 +41,55 @@ function getDevices(adbPath: string): Array<string> {
   }
 }
 
+function getAllAvailableDevices(adbPath: string) {
+  const emulators = execFileSync(
+    `${process.env.ANDROID_HOME}/emulator/emulator`,
+    ['-list-avds'],
+    {encoding: 'utf8'},
+  )
+    .replace(/\n$/, '')
+    .split('\n')
+    .filter(Boolean)
+    .map(emulator => ({
+      name: emulator,
+      type: 'emulator',
+    }));
+
+  const physicalDevicesList = execFileSync(adbPath, ['devices', '-l'], {
+    encoding: 'utf-8',
+  })
+    .replace(/\n$/, '')
+    .split('\n');
+
+  const modelKey = 'model:';
+  const physicalDevices = physicalDevicesList
+    .slice(1, physicalDevicesList.length)
+    .map((device: string) => {
+      if (device.includes('emulator')) {
+        return;
+      }
+
+      const [deviceId, ...deviceInfo] = device.split(' ');
+
+      if (!deviceId) {
+        return;
+      }
+
+      const name = deviceInfo
+        .find(info => info.includes(modelKey))!
+        .replace(modelKey, '');
+
+      return {
+        deviceId,
+        name,
+        type: 'device',
+      };
+    })
+    .filter(Boolean);
+
+  return [...physicalDevices, ...emulators];
+}
+
 /**
  * Gets available CPUs of devices from ADB
  */
@@ -69,5 +118,6 @@ function getAvailableCPUs(adbPath: string, device: string): Array<string> {
 
 export default {
   getDevices,
+  getAllAvailableDevices,
   getAvailableCPUs,
 };
--- a/packages/platform-android/src/commands/runAndroid/adb.ts
+++ b/packages/platform-android/src/commands/runAndroid/adb.ts
@@ -51,6 +51,7 @@ function getAllAvailableDevices(adbPath: string) {
     .split('\n')
     .filter(Boolean)
     .map(emulator => ({
+      deviceId: undefined,
       name: emulator,
       type: 'emulator',
     }));
@@ -58,6 +59,7 @@ function getAllAvailableDevices(adbPath: string) {
   const physicalDevicesList = execFileSync(adbPath, ['devices', '-l'], {
     encoding: 'utf-8',
   })
+    .toString()
     .replace(/\n$/, '')
     .split('\n');
 
--- a/packages/platform-android/src/commands/runAndroid/emulator.ts
+++ b/packages/platform-android/src/commands/runAndroid/emulator.ts
@@ -15,8 +15,10 @@ const getEmulators = () => {
   }
 };
 
-const launchEmulator = async (emulatorName: string, adbPath: string) => {
+export const launchEmulator = async (emulatorName: string, adbPath: string) => {
   return new Promise((resolve, reject) => {
+    const devicesList = Adb.getDevices(adbPath);
+
     const cp = execa(emulatorCommand, [`@${emulatorName}`], {
       detached: true,
       stdio: 'ignore',
@@ -31,7 +33,11 @@ const launchEmulator = async (emulatorName: string, adbPath: string) => {
     }, timeout * 1000);
 
     const bootCheckInterval = setInterval(() => {
-      if (Adb.getDevices(adbPath).length > 0) {
+      const latestDevicesList = Adb.getDevices(adbPath);
+      const hasDevicesListChanged =
+        latestDevicesList.filter(d => !devicesList.includes(d)).length > 0;
+
+      if (hasDevicesListChanged) {
         cleanup();
         resolve();
       }
@@ -54,9 +60,9 @@ const launchEmulator = async (emulatorName: string, adbPath: string) => {
   });
 };
 
-export default async function tryLaunchEmulator(
+export const tryLaunchEmulator = async (
   adbPath: string,
-): Promise<{success: boolean; error?: string}> {
+): Promise<{success: boolean; error?: string}> => {
   const emulators = getEmulators();
   if (emulators.length > 0) {
     try {
@@ -70,4 +76,4 @@ export default async function tryLaunchEmulator(
     success: false,
     error: 'No emulators found as an output of `emulator -list-avds`',
   };
-}
+};
--- a/packages/platform-android/src/commands/runAndroid/emulator.ts
+++ b/packages/platform-android/src/commands/runAndroid/emulator.ts
@@ -1,6 +1,7 @@
 import os from 'os';
 import execa from 'execa';
 import Adb from './adb';
+import {CLIError} from '@react-native-community/cli-tools';
 
 const emulatorCommand = process.env.ANDROID_HOME
   ? `${process.env.ANDROID_HOME}/emulator/emulator`
@@ -15,7 +16,10 @@ const getEmulators = () => {
   }
 };
 
-export const launchEmulator = async (emulatorName: string, adbPath: string) => {
+export const launchEmulator = async (
+  emulatorName: string,
+  adbPath: string,
+): Promise<string[]> => {
   return new Promise((resolve, reject) => {
     const devicesList = Adb.getDevices(adbPath);
 
@@ -29,7 +33,7 @@ export const launchEmulator = async (emulatorName: string, adbPath: string) => {
     // Reject command after timeout
     const rejectTimeout = setTimeout(() => {
       cleanup();
-      reject(`Could not start emulator within ${timeout} seconds.`);
+      reject({message: `Could not start emulator within ${timeout} seconds.`});
     }, timeout * 1000);
 
     const bootCheckInterval = setInterval(() => {
@@ -39,7 +43,7 @@ export const launchEmulator = async (emulatorName: string, adbPath: string) => {
 
       if (hasDevicesListChanged) {
         cleanup();
-        resolve();
+        resolve(latestDevicesList);
       }
     }, 1000);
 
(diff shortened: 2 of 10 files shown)


lucasbento:
I have fixed the issue with the Android build and everything seems to be working fine now, I did a little bit of refactoring and moving stuff around to fit the new feature.

I did quite some heavy testing on iOS and Android with the exception of starting and running on an Android physical device as I no longer have one, I would appreciate some help on testing that.

packages/platform-ios/src/commands/runIOS/index.ts
```diff
@@ -98,20 +115,48 @@ function runIOS(_: Array<string>, ctx: Config, args: FlagsT) {
...
+  }
+
+  const physicalDevices = devices.filter(d => d.type !== 'simulator');
+  const device = matchingDevice(physicalDevices, args.device);
+  if (device) {
+    return runOnDevice(device, scheme, xcodeProject, args);
   }
 }
 
+async function promptWhichDeviceToRun(devices: Device[]): Promise<Device> {
```
grabbou, line 133: How about moving this to `utils` so that both iOS/Android (and possibly other platforms in the future) can take advantage of that?

packages/platform-android/src/commands/runAndroid/index.ts
```diff
@@ -152,14 +155,79 @@ function buildAndRun(args: Flags, androidProject: AndroidProject) {
...
+    return;
+  }
+
+  const devicesList = devices.map(device => {
+    if (!device) {
+      return;
+    }
+
+    const {type, name} = device;
+
```
grabbou, line 209: Feels like there's definitely too many line breaks in this block 🤔

packages/platform-android/src/commands/runAndroid/index.ts
```diff
@@ -152,14 +155,79 @@ function buildAndRun(args: Flags, androidProject: AndroidProject) {
...
+      name: 'chosenDevice',
+      message: 'Which device to run the application?',
+      filter: (device: string) =>
+        device.substring(0, device.lastIndexOf('-')).trim(),
+      choices: devicesList,
+      loop: false,
+    },
+  ]);
+
+  return devices.find(device => device && device.name === chosenDevice)!;
```
grabbou, line 227: Will there ever be a case that we enumerate over `device` that is undefined?

packages/platform-android/src/commands/runAndroid/index.ts
```diff
@@ -152,14 +155,79 @@ function buildAndRun(args: Flags, androidProject: AndroidProject) {
...
-  } else {
-    return runOnAllDevices(args, cmd, packageName, adbPath, androidProject);
   }
+
+  return runOnEmulator(args, cmd, packageName, adbPath, androidProject);
 }
 
-function runOnSpecificDevice(
+async function promptWhichDeviceToRun(adbPath: string) {
+  const devices = adb.getAllAvailableDevices(adbPath);
```
grabbou, line 197: The iOS function is pure - doesn't ask for devices, receives an array instead. Would it make sense to move this outside of the function - might be useful to refactor the code in a follow up PR.

It looks like we're calling `adb.getAllAvailableDevices(adbPath);` and `adb.getDevices(adbPath)` definitely too many times (as opposed to just one on the iOS side of things).

packages/platform-android/src/commands/runAndroid/adb.ts
```diff
@@ -41,6 +41,57 @@ function getDevices(adbPath: string): Array<string> {
   }
 }
 
+function getAllAvailableDevices(adbPath: string) {
```
grabbou, line 44: Three questions:
1/ Is this more `getDevicesAndEmulators`, in other words, an array of devices & emulators that are not booted?
2/ Why not extend `getDevices` already? Seems a bit odd to introduce another one. If you want to have a separate one, why not to call `adb.getDevices()` on line 59?
3/ Why not use `parseDevicesResult` here, but instead custom logic? Is there something we can do to tweak that function to be useful here as well?

packages/platform-android/src/commands/runAndroid/adb.ts
```diff
@@ -41,6 +41,57 @@ function getDevices(adbPath: string): Array<string> {
...
+    .replace(/\n$/, '')
+    .split('\n')
+    .filter(Boolean)
+    .map(emulator => ({
+      deviceId: undefined,
+      name: emulator,
+      type: 'emulator',
+    }));
+
+  const physicalDevicesList = execFileSync(adbPath, ['devices', '-l'], {
```
grabbou, line 59: In other places, we use `execSync('${adbPath} devices')` and so on. No preference here, but would be good to stay consistent.

packages/platform-android/src/commands/runAndroid/adb.ts
```diff
@@ -41,6 +41,57 @@ function getDevices(adbPath: string): Array<string> {
...
+      if (device.includes('emulator')) {
+        return;
+      }
+
+      const [deviceId, ...deviceInfo] = device.split(' ');
+
+      if (!deviceId) {
+        return;
+      }
+
```
grabbou, line 79: nit: It seems to me there's a bit too much of line breaks here, following your other code in other places

packages/platform-android/src/commands/runAndroid/emulator.ts
```diff
@@ -42,32 +52,31 @@ const launchEmulator = async (emulatorName: string, adbPath: string) => {
       clearInterval(bootCheckInterval);
     };
 
-    cp.on('exit', () => {
+    cp.on('exit', (isEmulatorLaunched: 0 | 1) => {
```
grabbou, line 55: Can there be a better type than `0 | 1`? (tip: boolean?)

packages/platform-android/src/commands/runAndroid/runOnEmulator.ts
```diff
@@ -26,29 +26,62 @@ function toPascalCase(value: string) {
...
+
+  logger.info('Launching emulator...');
+
+  try {
+    const deviceListAfterAnyEmulatorLaunched = await launchAnyEmulator(adbPath);
+
+    return deviceListAfterAnyEmulatorLaunched;
+  } catch (error) {
+    console.log({error});
+    return logStartEmulatorError(error);
```
grabbou, line 72: Do we need a separate `logStartEmulatorError` function here? How about we just handle it inline?

packages/platform-android/src/commands/runAndroid/runOnEmulator.ts
```diff
@@ -26,29 +26,62 @@ function toPascalCase(value: string) {
...
+  }
+
+  logger.info('Launching emulator...');
+
+  try {
+    const deviceListAfterAnyEmulatorLaunched = await launchAnyEmulator(adbPath);
+
+    return deviceListAfterAnyEmulatorLaunched;
+  } catch (error) {
+    console.log({error});
```
grabbou, line 71: This `console.log` is extraneous here. There's already logging inside `logStartEmulatorError`.

packages/platform-android/src/commands/runAndroid/runOnEmulator.ts
```diff
@@ -26,29 +26,62 @@ function toPascalCase(value: string) {
 
 type AndroidProject = NonNullable<Config['project']['android']>;
 
-async function runOnAllDevices(
+function logStartEmulatorError(error: Error) {
```
grabbou, line 29: Idea: How about we actually throw `CLIError` in all cases here, instead of custom logging / error handling logic?

packages/platform-android/src/commands/runAndroid/runOnEmulator.ts
grabbou: I see, the reason it was done was due to line 55, where it is being referenced as well.

I would wrap `launchAnyEmulator` and `launchEmulator` in the same `try/catch` block to address this.

packages/platform-android/src/commands/runAndroid/runOnEmulator.ts
```diff
@@ -26,29 +26,62 @@ function toPascalCase(value: string) {
...
-    } else {
-      logger.error(
-        `Failed to launch emulator. Reason: ${chalk.dim(result.error || '')}.`,
-      );
-      logger.warn(
-        'Please launch an emulator manually or connect a device. Otherwise app may fail to launch.',
-      );
-    }
-  }
+  const devices = await startEmulator(adbPath, deviceName);
```
grabbou, line 84: I would prefer to stick to previous:
```js
await startEmulator(adbPath, deviceName);
```
instead of passing around a list of devices, which makes the code a bit harder to follow.

Alternatively, I would also accept:
```js
const launchedDevice = await startEmulator(adbPath, deviceName);
```
but I don't think it's useful, as we have to run `adb.getDevices()` anyway.

grabbou: When we address that, the code will become much easier in other places, due to lack of `return`

packages/platform-android/src/commands/runAndroid/emulator.ts
```diff
@@ -27,13 +33,17 @@ const launchEmulator = async (emulatorName: string, adbPath: string) => {
     // Reject command after timeout
     const rejectTimeout = setTimeout(() => {
       cleanup();
-      reject(`Could not start emulator within ${timeout} seconds.`);
+      reject({message: `Could not start emulator within ${timeout} seconds.`});
```
grabbou, line 36: How about rejecting here with an actual Error instance?

packages/platform-android/src/commands/runAndroid/emulator.ts
```diff
@@ -42,32 +52,31 @@ const launchEmulator = async (emulatorName: string, adbPath: string) => {
...
-    cp.on('exit', () => {
+    cp.on('exit', (isEmulatorLaunched: 0 | 1) => {
       cleanup();
-      reject('Emulator exited before boot.');
+
+      if (!isEmulatorLaunched) {
+        return reject({message: 'Emulator exited before boot.'});
+      }
+
+      return resolve(Adb.getDevices(adbPath));
```
grabbou, line 62: Again, no need to resolve here with `devices`. You could technically resolve with a `device` object of what you have just launched, or its UDID.

packages/platform-android/src/commands/runAndroid/runOnEmulator.ts
grabbou: To be honest, it could've been simplified to:
```js
async function startEmulator(adbPath: string, deviceName?: string) {
    try {
        if (deviceName) {
            logger.info('Launching ${deviceName} emulator...');
            return launchEmulator(
                deviceName,
                adbPath,
            );
        } else {
            logger.info('Launching first available emulator...');
            return launchAnyEmulator(adbPath);
        }
    } catch (error) {
        if (error.message) {
            logger.error(
                `Failed to launch emulator. Reason: ${chalk.dim(error.message)}`,
            );
        } else {
            logger.error('Failed to launch emulator.');
        }

        logger.warn(
            'Please launch an emulator manually or connect a device. Otherwise the app may fail to launch.',
        );
    }
}
```
and in such case, I am not even sure if we need this function - maybe we can place it directly inside `runAndroid` instead.

packages/platform-android/src/commands/runAndroid/runOnEmulator.ts
```diff
@@ -109,4 +140,4 @@ function createInstallError(error: Error & {stderr: string}) {
   return new CLIError(`Failed to install the app. ${message}.`, error);
 }
 
-export default runOnAllDevices;
+export default runOnEmulator;
```
grabbou, line 143: I think we still `runOnAllDevices`, not just on a single "emulator" - all in all you have `forEach` inside on line 110.

grabbou (changes requested):
Thank you for this PR! I love this feature and would really like to get it out in the next release. I did leave some comments tho - mainly to Android part.

grabbou:
Closing due to inactivity. Happy to revisit later, since the features in this PR were quite handy!

PR #1234, opened 2020-08-01
fix: proper range on RN peer dep

thymikee:
Summary:
---------

v4 supports both 0.62 and 0.63.

Test Plan:
----------

No peer dep warning

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -64,7 +64,7 @@
     "wcwidth": "^1.0.1"
   },
   "peerDependencies": {
-    "react-native": "^0.62.0-rc.0"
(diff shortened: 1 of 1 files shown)


SimenB:
Will e.g. `metro-react-native-babel-transformer` be bumped to match RN 0.63 as well?

thymikee:
Not for now. Major bumping Metro deps is possibly a breaking change for CLI users and just in case we tend to do it separately. Metro could do a similar change to this one for 0.58 versions, but I assume it's not a priority, since this is a warning and things work as expected.

SimenB:
Sound good. Any plans to merge this PR? Peer warnings are annoying 😅

thymikee:
Will merge and publish hopefully today :)

PR #1244, opened 2020-08-10
chore: bump joi to 17

thymikee:
Summary:
---------

Joi moved out of Hapi org.

Test Plan:
----------

Config validation works

image


packages/cli/src/tools/config/errors.ts
```diff
@@ -1,36 +1,28 @@
...
-                Unknown option ${name} with value "${value}" was found.
-                This is either a typing error or a user mistake. Fixing it will remove this message.
-              `;
-            }
-            case 'object.base':
-            case 'string.base': {
-              const expectedType = error.type.replace('.base', '');
-              const actualType = typeof (error.context && error.context.value);
-              return `
-                Option ${name} must be a ${expectedType}, instead got ${actualType}
```
thymikee, line 25: Removed this, because the only difference was "instead got ..." which is not super helpful anyway.

grabbou (approved):

PR #1246, opened 2020-08-13
feat: add a command to download the Hermes Sampling Profiler to a local machine

jessieAnhNguyen:
## Summary:
---------

This PR implements a new command to download the Hermes Sampling Profiler to a local machine.

This command is built as part of the goal to visualize the performance of JavaScript running on Hermes in a React Native app.     
@saphal1998 and I worked on this under @axemclion and @jevakallio guidance.

### 1. Purpose: 
- This new command helps users convert a Hermes tracing profile to Chrome tracing profile and pull it to the local machine
- Instructions on how to use Hermes: [Using Hermes](https://reactnative.dev/docs/hermes)
- The hermes-profile-transformer tool that converts Hermes tracing profile to Chrome tracing profile [Hermes-profile-transformer package](https://www.npmjs.com/package/hermes-profile-transformer)

### 2. Usage:

```sh
react-native profile-hermes [destinationDir] <flag>
```
- `destinationDir` is optional; if provided, pull the file to that directory
  > default: pull to the current React Native app root directory

**Options**

- `--filename [string]`

File name of the profile to be downloaded, ex. sampling-profiler-trace8593107139682635366.cpuprofile

> default: pull the latest file

- `--verbose`

Lists commands and steps that are run internally to pull the file from Android device

- `--raw`

Pulls the original Hermes tracing profile without any transformation

- `--sourcemap-path [string]`

The local path to your source map file if you generated it manually, ex. `/tmp/sourcemap.json`

- `--generate-sourcemap`

Generate the JS bundle and source map in `os.tmpdir()`

#### Notes on source map
- Source map is used to enhance the profile and associate trace events with the application code
- This step is recommended in order for the source map to be generated:

If you are planning on building a debug APK, that will run without the packager, by invoking `./gradlew assembleDebug` you can simply set `bundleInDebug: true` in your app/build.gradle file, inside the `project.ext.react` map.

## Test Plan:
----------

- ```yarn```
- ```yarn lint```
- Test on a React Native app: follow the instructions on the CLI's [Testing Plan](https://github.com/react-native-community/cli/blob/master/CONTRIBUTING.md#testing-your-changes)
- Command ```react-native profile-hermes``` as per instruction above

### Example output of running the command
image

image


grabbou:
Hey,

Thanks for submitting the PR. Let me review it shortly. 

Quick question: have you considered creating a separate package under `packages` that could group Hermes-related commands? (see what we have done for the IOS/Android). 

We're working towards making the CLI itself a really small and thin layer, whereas iOS/Android/Windows related commands and utilities are grouped separately and downloaded to the project depending on its configuration and platforms used.

In this case, we could have a "@react-native-community/cli-hermes" or related. It could be released independently of the CLI, added dynamically to a project (as its dependency) and CLI would automatically pick it up and make the command available.

jevakallio:
> Quick question: have you considered creating a separate package under packages that could group Hermes-related commands? (see what we have done for the IOS/Android).

Hey @grabbou! We just had a discussion about this with @axemclion, and we don't have any issues with moving this command into its own package.

However, for user adoption we think this should be generally available and easy to adopt for users. As far as I can see, there currently don't seem to be any additional CLI extension modules that are left for the user to install? All the `cli-*` packages are [listed in the main cli package dependencies](https://github.com/react-native-community/cli/blob/master/packages/cli/package.json#L32-L35) and the [platform packages are listed in react-native core package](https://github.com/facebook/react-native/blob/master/package.json#L89-L90).

This means that the Hermes CLI would be the first user-installed module, and as there is no precedent to this, it would likely hinder the adoptability of this feature.

To this end, I think we should either:
a) Include the new command in the core as we have already done and refactor it out later, or
b) Migrate it to its own package `cli-hermes`, and add it as a dependency to either `cli` or `react-native` to keep it at par with all the other module extensions.

We'd prefer option B, and add it as a dependency to `cli`. Any thoughts on this?

Whichever approach we take, could you please review this code for implementation now rather than after the commit that moves this (if we end up making that decision). The reason being that @jessieAnhNguyen and @saphal1998 are rolling off the MLH program, and while they can continue to work on this afterwards, it would be good to get the feedback and green light for mergeability now.

grabbou:
Yes, of course. I will do that tomorrow!

packages/cli-hermes/package.json
```diff
@@ -0,0 +1,24 @@
...
+  "version": "4.11.0",
+  "license": "MIT",
+  "main": "build/index.js",
+  "publishConfig": {
+    "access": "public"
+  },
+  "types": "build/index.d.ts",
+  "dependencies": {
+    "@react-native-community/cli-tools": "^4.11.0",
+    "hermes-profile-transformer": "^0.0.6",
```
thymikee, line 12: missing `chalk` in deps

packages/cli-hermes/src/profileHermes/index.ts
```diff
@@ -0,0 +1,79 @@
...
+    await downloadProfile(
+      ctx,
+      dstPath,
+      options.fileName,
+      options.sourcemapPath,
+      options.raw,
+      options.generateSourcemap,
+    );
+  } catch (err) {
+    logger.error(`Unable to download the Hermes Sampling Profile.\n${err}`);
```
thymikee, line 37: we tend to `throw CLIError(message, err)` in such cases

packages/cli-hermes/src/profileHermes/sourcemapUtils.ts
```diff
@@ -0,0 +1,120 @@
+import {Config} from '@react-native-community/cli-types';
+import {logger, CLIError} from '@react-native-community/cli-tools';
+import fs from 'fs';
+import path from 'path';
+import os from 'os';
+import {SourceMap} from 'hermes-profile-transformer';
+import ip from 'ip';
+import nodeFetch from 'node-fetch';
```
thymikee, line 8: We have a `fetch` utility in `cli-tools`, maybe we could reuse that? If not, `node-fetch` needs to be declared in package dependencies

packages/cli-hermes/src/profileHermes/sourcemapUtils.ts
```diff
@@ -0,0 +1,120 @@
...
+
+function writeJsonSync(targetPath: string, data: any) {
+  let json;
+  try {
+    json = JSON.stringify(data);
+  } catch (e) {
+    logger.error(
+      `Failed to serialize data to json before writing to ${targetPath}\n`,
+      e,
+    );
```
thymikee, line 22: I think we should throw CLIError here and in line 28. Currently an undefined `json` will be passed to `fs.writeFileSync` in case of parsing failure and that doesn't make too much sense, isn't it?

packages/cli-hermes/src/profileHermes/sourcemapUtils.ts
```diff
@@ -0,0 +1,120 @@
...
+  try {
+    fs.writeFileSync(targetPath, json, 'utf-8');
+  } catch (e) {
+    logger.error(`Failed to write json to ${targetPath}\n`, e);
+  }
+}
+
+async function getSourcemapFromServer(): Promise<SourceMap> {
+  logger.debug('Getting source maps from Metro packager server\n');
+  const DEBUG_SERVER_PORT = '8081';
```
thymikee, line 34: port is hardcoded. Worth to pass it from arguments and default to 8081

packages/cli-hermes/src/profileHermes/sourcemapUtils.ts
```diff
@@ -0,0 +1,120 @@
...
+ *
+ * @param ctx
+ */
+export async function findSourcemap(ctx: Config): Promise<string> {
+  const intermediateBuildPath = path.join(
+    ctx.root,
+    'android',
+    'app',
+    'build',
+    'intermediates', //'generated',
```
thymikee, line 81:
```suggestion
    'intermediates',
```

packages/cli-hermes/src/profileHermes/sourcemapUtils.ts
```diff
@@ -0,0 +1,120 @@
...
+      logger.debug(`Getting the source map from ${intermediateBuildPath}\n`);
+    }
+    return Promise.resolve(intermediateBuildPath);
+  } else {
+    const sourcemapResult = await getSourcemapFromServer();
+    if (logger.isVerbose()) {
+      logger.debug('Using source maps from Metro packager server\n');
+    }
+    const tempPath = getTempFilePath('index.map');
+    writeJsonSync(tempPath, sourcemapResult);
```
thymikee, line 116: I see, you want `writeJsonSync` to be skipped when failing inside. In such case we could return early from the function when the read is failing

packages/cli/src/commands/index.ts
```diff
@@ -11,6 +11,7 @@ import info from './info/info';
 import config from './config/config';
 import init from './init';
 import doctor from './doctor';
+import profile from '@react-native-community/cli-hermes';
```
thymikee, line 14: can we use named export instead of default one? It will be less breaking when we decided to introduce more Hermes-related commands

packages/cli-hermes/package.json
jessieAnhNguyen: Oh so since `chalk` is a dependency in the general repo, do we need to add it here too?

thymikee: Yes. You will be able to install this package alone, so all dependencies used need to be explicitly declared, because there may not be "general repo" or other packages containing this package.

packages/cli-hermes/src/profileHermes/downloadProfile.ts
```diff
@@ -0,0 +1,175 @@
...
+ * @param config
+ */
+function getPackageName(config: Config) {
+  const androidProject = config.project.android;
+
+  if (!androidProject) {
+    throw new CLIError(`
+  Android project not found. Are you sure this is a React Native project?
+  If your Android files are located in a non-standard location (e.g. not inside \'android\' folder), consider setting
+  \`project.android.sourceDir\` option to point to a new location.
```
grabbou, line 43: Note: You can write it like here: https://github.com/react-native-community/cli/blob/a90851a34b96ac0b42b1cc32b4a3f6039134da8c/packages/cli/src/commands/link/link.ts#L63-L66 (indentation on lines 41/42/43). We inline them anyway and in the future, will use dedent or similar when that changes.

packages/cli-hermes/src/profileHermes/downloadProfile.ts
```diff
@@ -0,0 +1,175 @@
...
+  if (!validatePackageName(packageName)) {
+    logger.warn(
+      `Invalid application's package name "${chalk.bgRed(
+        packageName,
+      )}" in 'AndroidManifest.xml'. Read guidelines for setting the package name here: ${chalk.underline.dim(
+        'https://developer.android.com/studio/build/application-id',
+      )}`,
+    );
+  }
+  return packageName;
```
grabbou, line 67: These two are very similar to https://github.com/react-native-community/cli/blob/f497f5a9f1bb0980488b989034d96b30ffd08747/packages/platform-android/src/commands/runAndroid/index.ts#L133-L139 and https://github.com/react-native-community/cli/blob/f497f5a9f1bb0980488b989034d96b30ffd08747/packages/platform-android/src/commands/runAndroid/index.ts#L144-L152. 

I wonder if it wouldn't be better from the maintenance perspective if they weren't moved to utilities? Maybe that's something `Android` package can expose and this package can depend on? 

I believe the project check at the top is also something that exists in `run-android` command right now.

packages/cli-hermes/src/profileHermes/downloadProfile.ts
```diff
@@ -0,0 +1,175 @@
...
+}
+
+/**
+ * Validates that the package name is correct
+ * @param packageName
+ */
+function validatePackageName(packageName: string) {
+  return /^[a-z][a-z0-9_]*(\.[a-z][a-z0-9_]*)+$/.test(packageName);
+}
```
grabbou, line 76: Same: https://github.com/react-native-community/cli/blob/f497f5a9f1bb0980488b989034d96b30ffd08747/packages/platform-android/src/commands/runAndroid/index.ts#L27-L29

I would create src/utils inside `platform-android` package and export them, so that we can use them here. What do you think?

packages/cli-hermes/src/profileHermes/downloadProfile.ts
```diff
@@ -0,0 +1,175 @@
...
+    if (!file) {
+      logger.error(
+        'There is no file in the cache/ directory. Did you record a profile from the developer menu?\n',
+      );
+      process.exit(1);
+    }
+
+    logger.info(`File to be pulled: ${file}\n`);
+
+    //if destination path is not specified, pull to the current directory
```
grabbou, line 109: Warning - nitpicking:

Comment on line 98 starts: `// if` (with space), while this one - does not `//if`. Some of the comments later in the document are `//If`. 

Let's stick to a single style if that's okay - I would suggest `// If` (space & uppercase) to keep it consistent with other places.

packages/cli-hermes/src/profileHermes/downloadProfile.ts
```diff
@@ -0,0 +1,175 @@
...
+    dstPath = dstPath || ctx.root;
+
+    if (logger.isVerbose()) {
+      logger.info('Internal commands run to pull the file:\n');
```
grabbou, line 113: For `verbose`, you can use `logger.debug` that checks for verbose mode inside itself already: https://github.com/react-native-community/cli/blob/f497f5a9f1bb0980488b989034d96b30ffd08747/packages/tools/src/logger.ts#L37

packages/cli-hermes/src/profileHermes/downloadProfile.ts
```diff
@@ -0,0 +1,175 @@
...
+        if (shouldGenerateSourcemap) {
+          sourcemapPath = await generateSourcemap();
+        } else {
+          sourcemapPath = await findSourcemap(ctx);
+        }
+
+        //Run without source map
+        if (!sourcemapPath) {
+          logger.warn(
+            'Cannot generate or find bundle and source map, running the transformer without source map\n',
```
grabbou, line 144: Is the `\n` at the end needed? Doesn't seem to be used in other places where we use `logger[method]`

packages/cli-hermes/src/profileHermes/downloadProfile.ts
```diff
@@ -0,0 +1,175 @@
...
+        transformedFilePath,
+        JSON.stringify(events, undefined, 4),
+        'utf-8',
+      );
+      logger.success(
+        `Successfully converted to Chrome tracing format and pulled the file to ${transformedFilePath}\n`,
+      );
+    }
+  } catch (e) {
+    throw new Error(e.message);
```
grabbou, line 173: You may consider providing more descriptive error message and passing the original error to the CLIError to allow CLI to handle it using its standard error handling mechanism.

Example:
https://github.com/react-native-community/cli/blob/79498d598e00e481e6058a2892eaa71870215922/packages/platform-android/src/commands/runAndroid/tryLaunchAppOnDevice.ts#L41

packages/cli-hermes/src/profileHermes/index.ts
```diff
@@ -0,0 +1,79 @@
...
+};
+
+async function profileHermes(
+  [dstPath]: Array<string>,
+  ctx: Config,
+  options: Options,
+) {
+  try {
+    logger.info(
+      'Downloading a Hermes Sampling Profiler from your Android device...\n',
```
grabbou, line 20: Similar to comment in other places, we don't use `\n` in `logger` anywhere else. If that was for styling purposes, I would welcome a separate PR to improve logging experience for all output 👯

packages/cli-hermes/src/profileHermes/index.ts
```diff
@@ -0,0 +1,79 @@
...
+    );
+    if (!options.fileName) {
+      logger.info('No filename is provided, pulling latest file\n');
+    }
+    if (options.verbose) {
+      logger.setVerbose(true);
```
grabbou, line 26: Done automatically for you - https://github.com/react-native-community/cli/blob/b7778db91e1033e8f5a6338f2e8805c4b6f6d967/packages/cli/src/index.ts#L190

packages/cli-hermes/src/profileHermes/index.ts
```diff
@@ -0,0 +1,79 @@
+import {logger} from '@react-native-community/cli-tools';
+import {Config} from '@react-native-community/cli-types';
+import {downloadProfile} from './downloadProfile';
+
+type Options = {
+  verbose: boolean;
```
grabbou, line 6: No need to define `verbose` in `Options` type, as we already do that in the main file:
https://github.com/react-native-community/cli/blob/b7778db91e1033e8f5a6338f2e8805c4b6f6d967/packages/cli/src/index.ts#L20

packages/cli-hermes/src/profileHermes/index.ts
grabbou: Yeah. I already left an example few comments earlier.

packages/cli-hermes/src/profileHermes/sourcemapUtils.ts
grabbou: Yup. Let's stick to the same error handling everywhere - rethrowing CLIErrors, while not perfect, provides a unified error handling experience. 

PS. It is okay to provide additional logging (like logger.info/error) when needed, for better UX.

grabbou (commented):
Thank you for the PR! Did a first pass and left some comments here and there. Overall, I like it! Just some minor styling/code formatting issues to make it more unified with the rest of the codebase. Let me know what you think.

grabbou:
Thank you @thymikee for helping out with reviewing this PR! 🎉

packages/cli-hermes/src/profileHermes/downloadProfile.ts
jessieAnhNguyen: Yeah I think this is a good idea. I'll go ahead with that 🚀

packages/cli-hermes/src/profileHermes/index.ts
jessieAnhNguyen: Yeah the reason why I put a new line after each sentence is that it loos very clustered, so I just wanted to give some space after each log 😆

packages/cli-hermes/src/profileHermes/sourcemapUtils.ts
jessieAnhNguyen: That's a good idea, I changed to use the `fetch` in `cli-tools` 🚀

packages/cli-hermes/src/profileHermes/sourcemapUtils.ts
jessieAnhNguyen: Yup I've added another flag `--port [number]` so user can enter their running server port (default is 8081)

packages/cli-hermes/src/profileHermes/index.ts
```diff
@@ -0,0 +1,77 @@
...
+        'The local path to your source map file, eg. /tmp/sourcemap.json',
+    },
+    {
+      name: '--generate-sourcemap',
+      description: 'Generates the JS bundle and source map',
+    },
+    {
+      name: '--port [number]',
+      default: process.env.RCT_METRO_PORT || 8081,
+      parse: (val: number) => String(val),
```
jessieAnhNguyen, line 67: Per @thymikee's suggestion about hardcoding the port, I added a new flag `--port [number]` for user to input their running server port (default: 8081).

packages/cli-hermes/src/profileHermes/sourcemapUtils.ts
```diff
@@ -0,0 +1,113 @@
...
+
+  if (sourceMapResult) {
+    logger.debug('Using source maps from Metro packager server');
+    writeJsonSync(sourceMapPath, sourceMapResult);
+    logger.debug(
+      `Successfully obtained the source map and stored it in ${sourceMapPath}`,
+    );
+    return sourceMapPath;
+  } else {
+    logger.error('Cannot obtain source maps from Metro packager server');
```
jessieAnhNguyen, line 67: I don't throw an error here since even when we don't obtain source maps, the tool can still be run without source maps. Thus, I just displayed a `logger.error` and continue.

packages/cli-hermes/src/profileHermes/sourcemapUtils.ts
```diff
@@ -0,0 +1,113 @@
...
+  );
+
+  if (fs.existsSync(generatedBuildPath)) {
+    logger.debug(`Getting the source map from ${generateSourcemap}`);
+    return generatedBuildPath;
+  } else if (fs.existsSync(intermediateBuildPath)) {
+    logger.debug(`Getting the source map from ${intermediateBuildPath}`);
+    return intermediateBuildPath;
+  } else {
+    return generateSourcemap(port);
```
jessieAnhNguyen, line 111: I called this function here since the previous code was very similar to what's going on with the  `generateSourcemap` function.

packages/platform-android/src/utils/getAndroidProject.ts
```diff
@@ -0,0 +1,56 @@
+import {Config, AndroidProjectConfig} from '@react-native-community/cli-types';
```
jessieAnhNguyen, line 1: I refactored these functions from `platform-android/src/commands/runAndroid/index.ts` per @grabbou's suggestion to this new file. Those functions are hence used in both the above `downloadProfile.ts` file and the `runAndroid/index.ts` file.

docs/commands.md
```diff
@@ -546,3 +547,48 @@ Upgrade your app's template files to the specified or latest npm version using [
...
+```
+
+Pull and convert a Hermes tracing profile to Chrome tracing profile, then store it in the directory <destinationDir> of the local machine.
+
+- `destinationDir` is optional, if provided, pull the file to that directory
+  > default: pull to the current React Native app root directory
+
+#### Options
+
+#### `--fileName [string]`
```
likern, line 566: `--fileName` option name is inconsistent with other options.
I think it should be `--filename` or `--file-name`

packages/cli-hermes/src/profileHermes/downloadProfile.ts
```diff
@@ -0,0 +1,128 @@
...
+  ctx: Config,
+  dstPath: string,
+  fileName?: string,
+  sourcemapPath?: string,
+  raw?: boolean,
+  shouldGenerateSourcemap?: boolean,
+  port?: string,
+) {
+  try {
+    const androidProject = getAndroidProject(ctx);
```
likern, line 51: If we can't find Android project or get package name - we can provide descriptive error message.
For example, when I run command outside project I get error "error Failed to download the Hermes Sampling Profile" which doesn't say what I did wrong. Especially that I knew I had cpuprofiles on a device.

likern (changes requested):

packages/platform-android/src/utils/getAndroidProject.ts
```diff
@@ -0,0 +1,56 @@
...
+export function getAndroidProject(config: Config) {
+  const androidProject = config.project.android;
+
+  if (!androidProject) {
+    throw new CLIError(`
+      Android project not found. Are you sure this is a React Native project?
+      If your Android files are located in a non-standard location (e.g. not inside \'android\' folder), consider setting
+      \`project.android.sourceDir\` option to point to a new location.
+    `);
+  }
```
jessieAnhNguyen, line 15: @likern I believe that this code throws an error when it can't find an Android project. Can you please check again if something else is the issue so the error is not thrown when you run it outside of the project?

packages/platform-android/src/utils/getAndroidProject.ts
```diff
@@ -0,0 +1,56 @@
...
+  const androidManifest = fs.readFileSync(manifestPath, 'utf8');
+
+  let packageNameMatchArray = androidManifest.match(/package="(.+?)"/);
+  if (!packageNameMatchArray || packageNameMatchArray.length === 0) {
+    throw new CLIError(
+      `Failed to build the app: No package name found. Found errors in ${chalk.underline.dim(
+        `${appFolder || appName}/src/main/AndroidManifest.xml`,
+      )}`,
+    );
+  }
```
jessieAnhNguyen, line 37: @likern Similar as above, I think when the command can't find the package name, it throws an error here. Let me know if I mistook it 😄

packages/platform-android/src/utils/getAndroidProject.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  if (!validatePackageName(packageName)) {
+    logger.warn(
+      `Invalid application's package name "${chalk.bgRed(
+        packageName,
+      )}" in 'AndroidManifest.xml'. Read guidelines for setting the package name here: ${chalk.underline.dim(
+        'https://developer.android.com/studio/build/application-id',
+      )}`,
+    );
+  }
```
jessieAnhNguyen, line 49: @likern Here checks when the packageName is not valid using the `validatePackageName` function and throw an error too.

docs/commands.md
jessieAnhNguyen: Thanks, I changed it accordingly!

packages/cli-hermes/src/profileHermes/downloadProfile.ts
jessieAnhNguyen: Hi @likern, thanks a lot for reviewing my PR! I looked at the code again, and think that the functions `getAndroidProject` and `getPackageName` already handled the error (specific details in my comments above). So can you check again whether any other issue happened when you run the command outside the project?

likern: @jessieAnhNguyen I'm not sure is it my fault or not.
I have build package using `yarn build` and later run cli directly through `node /home/me/react-native-cli/packages/cli/build/bin.js`.

If I run inside project I see this:
```bash
[me@localhost Feeve]$ pwd 
/home/me/Breeffy/Feeve
[me@localhost Feeve]$ node /home/me/react-native-cli/packages/cli/build/bin.js profile-hermes
info Downloading a Hermes Sampling Profiler from your Android device...
info No filename is provided, pulling latest file
error Failed to download the Hermes Sampling Profile. Run CLI with --verbose flag for more details.
```

while if I run this command outside (in `$HOME` directory) this erro is shown:
```bash
[me@localhost ~]$ pwd
/home/me
[me@localhost ~]$ node ~/react-native-cli/packages/cli/build/bin.js profile-hermes
error Unrecognized command "profile-hermes".
```

Not sure is it indended behaviour or not.

likern: If there is a specific way to build and run cli - please let me know, I will try to test again.

jessieAnhNguyen: Hey, can you try and test using the CLI's [testing plan](https://github.com/react-native-community/cli/blob/master/CONTRIBUTING.md#testing-your-changes). Please note that this suggests a way to link the packages, so that the React Native app that you're using to test on is linked to the CLI code. In the command `yarn link`, add this into the list of linked packages "@react-native-community/cli-hermes"

After that, you can simply run `react-native profile-hermes`. Please let me know if there's any issue!

saphal1998: Hey @likern, I found the bug and fixed it. Essentially we were throwing our error in the functions but printing a custom message in the final catch block. It's fixed now, thank you so much for pointing this out!

docs/commands.md
thymikee, line 566:
```suggestion
#### `--filename [string]`
```

packages/cli-hermes/src/profileHermes/sourcemapUtils.ts
```diff
@@ -0,0 +1,113 @@
...
+  const DEBUG_SERVER_PORT = port || '8081';
+  const IP_ADDRESS = ip.address();
+  const PLATFORM = 'android';
+
+  const requestURL = `http://${IP_ADDRESS}:${DEBUG_SERVER_PORT}/index.map?platform=${PLATFORM}&dev=true`;
+  try {
+    const {data} = await fetch(requestURL);
+    return data as SourceMap;
+  } catch (e) {
+    return undefined;
```
thymikee, line 45:
```suggestion
    logger.debug(`Failed to fetch source map from ${requestURL}`);
    return undefined;
```

thymikee (approved):
Last 2 changes and we can merge this. Awesome work :)

thymikee:
I would appreciate a followup PR to our unit/e2e tests to make sure we don't break stuff. @jessieAnhNguyen @saphal1998 would you be willing to do so? 🙂

BrendonSled:
is there no `appId` flag?

PR #1264, opened 2020-09-07
feat: augment `info` with out of tree platform packages info

asklar:
Summary:
---------

This PR will enable out of tree platforms to report their npm packages in `react-native info`

Test Plan:
----------

Verified that running the built JS on a react-native-windows app correctly reports the react-native-windows package too.
image

Verified that changing the mapping to not include win32 (so that platform never matches it) correctly does not report it.

### DIFF
--- a/packages/cli/src/tools/envinfo.ts
+++ b/packages/cli/src/tools/envinfo.ts
@@ -20,6 +20,7 @@ async function getEnvironmentInfo(
   let packages = ['react', 'react-native', '@react-native-community/cli'];
 
   const outOfTreePlatforms: {[key: string]: string} = {
+    darwin: 'react-native-macos',
     win32: 'react-native-windows',
   };
 
--- a/packages/cli/src/tools/envinfo.ts
+++ b/packages/cli/src/tools/envinfo.ts
@@ -24,7 +24,7 @@ async function getEnvironmentInfo(
     win32: 'react-native-windows',
   };
 
-  const outOfTreePlatformPackage: any = outOfTreePlatforms[platform()];
+  const outOfTreePlatformPackage = outOfTreePlatforms[platform()];
   if (outOfTreePlatformPackage) {
     packages.push(outOfTreePlatformPackage);
   }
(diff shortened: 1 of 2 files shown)


packages/cli/src/tools/envinfo.ts
```diff
@@ -16,6 +17,17 @@ async function getEnvironmentInfo(
 ): Promise<string | EnvironmentInfo> {
   const options = {json, showNotFound: true};
 
+  let packages = ['react', 'react-native', '@react-native-community/cli'];
+
+  const outOfTreePlatforms: {[key: string]: string} = {
```
thymikee, line 22: Can we rely on TS interference here? We should be able to avoid the "any" on line 26

packages/cli/src/tools/envinfo.ts
```diff
@@ -16,6 +17,17 @@ async function getEnvironmentInfo(
...
+    win32: 'react-native-windows',
+  };
```
thymikee, line 24: maybe worth adding macos straight away, at least to justify using `os.platform` usage. cc @alloy

alloy: Makes sense to me 👍

packages/cli/src/tools/envinfo.ts
```diff
@@ -16,6 +17,17 @@ async function getEnvironmentInfo(
 ): Promise<string | EnvironmentInfo> {
   const options = {json, showNotFound: true};
 
+  let packages = ['react', 'react-native', '@react-native-community/cli'];
+
+  const outOfTreePlatforms: {[key: string]: string} = {
+    win32: 'react-native-windows',
```
alloy, line 23:
```suggestion
    darwin: 'react-native-macos',
    win32: 'react-native-windows',
```

thymikee (approved):

asklar:
@thymikee looks like the test is flaky, but I don't have permissions to re-run it, mind doing it?

PR #1266, opened 2020-09-07
chore: remove await in getLatestFile function call from downloadProfile file.

luism3861:
Summary:
---------
this PR remove await in `getLatestFile `  call function from `packages/src/cli-hermes/profileHermes/downloadProfile.ts` it has no effect this function.

### DIFF
--- a/packages/cli-hermes/src/profileHermes/downloadProfile.ts
+++ b/packages/cli-hermes/src/profileHermes/downloadProfile.ts
@@ -52,7 +52,7 @@ export async function downloadProfile(
     const packageName = getPackageName(androidProject);
 
     // If file name is not specified, pull the latest file from device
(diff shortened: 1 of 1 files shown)


thymikee (approved):

packages/cli-hermes/src/profileHermes/downloadProfile.ts
```diff
@@ -52,7 +52,7 @@ export async function downloadProfile(
     const packageName = getPackageName(androidProject);
 
     // If file name is not specified, pull the latest file from device
-    const file = filename || (await getLatestFile(packageName));
+    const file = filename || (getLatestFile(packageName));
```
thymikee, line 55:
```suggestion
    const file = filename || getLatestFile(packageName);
```

thymikee:
Thanks!

Issue #1262, opened 2020-09-02
Wrong (or at least not descriptive) "error Android project not found" msg


scarlehoff:
## Description
A badly formatted AndroidManifest file will make `react-native run-android` fail with the following message:
```
error Android project not found. Are you sure this is a React Native project? If your Android files are located in a non-standard lo
cation (e.g. not inside 'android' folder), consider setting `project.android.sourceDir` option to point to a new location
```
It will be very useful to either point to the actual error or just say that it could not run _for some reason_ without giving a wrong reason.

## React Native version:
  Binaries:
    Node: 14.3.0 - /usr/bin/node
    Yarn: 1.22.4 - /usr/bin/yarn
    npm: 6.14.5 - /usr/bin/npm
    Watchman: 4.9.0 - /usr/bin/watchman
  SDKs:
    Android SDK:
      API Levels: 28, 29
      Build Tools: 28.0.3, 29.0.2, 29.0.3
      System Images: android-29 | Google APIs Intel x86 Atom
  npmPackages:
    react: 16.11.0 => 16.11.0 
    react-native: 0.62.2 => 0.62.2 
  npmGlobalPackages:
    react-native: 0.60.5

## Steps To Reproduce

1. Open `android/app/src/main/AndroidManifest.xml`
2. Add some random mistakes
3. Try to run `react-native run-android`

Great project and thank you very much for your hard work!

(originally put this issue in this repository: https://github.com/facebook/react-native/issues/29035 but was told to bring it here!)

thymikee:
Yep, this would definitely use an improvement!

PR #1270, opened 2020-09-17
fix: pass Android manifest parse errors through to the user

thymikee:
Summary:
---------

The try/catch block in `warnAboutManuallyLinkedLibs` functions swallowed errors when accessing `config.project` getter, which was the cause to the AndroidManifest parse errors not being shown to the user. There's no need for this error handling here, because if reading configuration fails, we want to know it immediately.

As a bonus I've restructured the errors to not hide important things like stack trace behind dimmed text:

Screen Shot 2020-09-17 at 14 55 10

Fixes https://github.com/react-native-community/cli/issues/1262.

Test Plan:
----------

None, check repro.


packages/cli/src/index.ts
```diff
@@ -30,17 +30,20 @@ const handleError = (err: Error) => {
   } else {
     // Some error messages (esp. custom ones) might have `.` at the end already.
     const message = err.message.replace(/\.$/, '');
-    logger.error(
-      `${message}. ${chalk.dim(
+    logger.error(`${message}.`);
+  }
+  if (err.stack) {
+    logger.log(err.stack);
```
grabbou, line 36: Are there cases when `err.stack` is missing? Just curious

grabbou: Oh, I see that's an old code anyway

thymikee: yea, a custom error can delete it

PR #1272, opened 2020-09-19
The keyword error gives a misleading message when the keyword is not "React"

ofirgeller:
Summary:
---------

to be a valid project name the name must not be a keyword, currently the error being thrown will claim the user tried using "React" as the project name even if they used another keyword.

Test Plan:
----------
There is already a test that check the error is created correctly, we can add a test with another keyword but I don't think that is needed.

There are two commits in this PR, the second one can be dropped if we don't consider the error class part of the public API since we only use the class in one place and in that place I added the new parameter.

### DIFF
--- a/packages/cli/src/commands/init/errors/ReservedNameError.ts
+++ b/packages/cli/src/commands/init/errors/ReservedNameError.ts
@@ -1,5 +1,5 @@
 export default class ReservedNameError extends Error {
-  constructor(name: string) {
+  constructor(name: string = 'React') {
     super(
       `Not a valid name for a project. Please do not use the reserved word "${name}".`,
     );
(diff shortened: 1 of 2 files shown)


packages/cli/src/commands/init/errors/ReservedNameError.ts
```diff
@@ -1,7 +1,7 @@
 export default class ReservedNameError extends Error {
-  constructor() {
+  constructor(name: string = 'React') {
```
grabbou, line 2: TBH I would remove this default value, as it's a bit misleading given the error is not `ReactReservedNameError`. 

While I greatly appreciate your backward-compatible approach here, I think we can safely remove that one. Typescript will warn us when we don't pass argument to a constructor.

grabbou (approved):
Good catch and a very nice, quick PR. Left a quick comment regarding the default constructor argument, rather than that - LGTM.

ofirgeller:
@grabbou You are correct about typescript warning us, but if the error is used by someone else we will break their code.  But I agree it is unlikely given the project.

grabbou:
Since this is going to land in the next CLI, let's annotate that as a breaking change - we're bumping major anyway. Don't want to pollute the codebase with such odd code.

PR #1273, opened 2020-09-21
Provide bundler host & port to the custom debugger

343max:
Summary:
---------

The `REACT_DEBUGGER` env variable allows a custom debugger to be used instead of the default Chrome debugger. So because I prefer React Native Debugger over Chrome I set `REACT_DEBUGGER` to `rndebugger-open --open --port 8081` in my shells profile.

This isn't ideal, because the port number is just a guess: e.g. expo uses a different port. It would be great if the middleware could pass this information to the debugger.

Because of this this PR introduces two new environment variables which it passes to the custom debugger to prevent them from having to guess.

- `REACT_BUNDLER_HOST` which is the the host where the bundler to debug runs (e.g. localhost)
- `REACT_BUNDLER_PORT` is the port (e.g. 8081)

This will allow the user to change their `REACT_DEBUGGER` script to pass the correct port to their debugger. Ideally the authors of the debugger could adopt over time and evaluate this variable directly.

Test Plan:
----------

I need to make sure that
- all the existing environment variable are being passed to the custom debugger like before
- my new environment variables are passed to the custom debugger as expected

I set my REACT_DEBUGGER env var to `rndebugger-test` and wrote a script rndebugger-test that dumps all env variables to a file so I can verify everything is there:
```
#!/bin/sh

env > /tmp/rndebugger-env-dump
```

Then I verified manually that
- the expected env vars are set correctly
- all the other env vars are still passed as expected

### DIFF
--- a/packages/cli-server-api/src/devToolsMiddleware.ts
+++ b/packages/cli-server-api/src/devToolsMiddleware.ts
@@ -8,13 +8,8 @@ import http from 'http';
 import {launchDebugger, logger} from '@react-native-community/cli-tools';
 import {exec} from 'child_process';
 
-function launchDefaultDebugger(
-  host: string | undefined,
-  port: number,
-  args = '',
-) {
-  const hostname = host || 'localhost';
-  const debuggerURL = `http://${hostname}:${port}/debugger-ui${args}`;
+function launchDefaultDebugger(host: string, port: number, args = '') {
+  const debuggerURL = `http://${host}:${port}/debugger-ui${args}`;
   logger.info('Launching Dev Tools...');
   launchDebugger(debuggerURL);
 }
@@ -34,13 +29,14 @@ function launchDevTools(
   {host, port, watchFolders}: LaunchDevToolsOptions,
   isDebuggerConnected: () => boolean,
 ) {
+  const hostname = host ?? 'localhost';
   // Explicit config always wins
   const customDebugger = process.env.REACT_DEBUGGER;
   if (customDebugger) {
-    startCustomDebugger({watchFolders, customDebugger, host, port});
+    startCustomDebugger({watchFolders, customDebugger, host: hostname, port});
   } else if (!isDebuggerConnected()) {
     // Debugger is not yet open; we need to open a session
-    launchDefaultDebugger(host, port);
+    launchDefaultDebugger(hostname, port);
   }
 }
 
@@ -52,10 +48,9 @@ function startCustomDebugger({
 }: {
   watchFolders: ReadonlyArray<string>;
   customDebugger: string;
-  host: string | undefined;
+  host: string;
   port: number;
 }) {
-  const hostname = host || 'localhost';
   const folders = watchFolders.map(escapePath).join(' ');
   const command = `${customDebugger} ${folders}`;
   logger.info('Starting custom debugger by executing:', command);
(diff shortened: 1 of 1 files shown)


packages/cli-server-api/src/devToolsMiddleware.ts
```diff
@@ -47,18 +47,36 @@ function launchDevTools(
 function startCustomDebugger({
   watchFolders,
   customDebugger,
+  host,
+  port,
 }: {
   watchFolders: ReadonlyArray<string>;
   customDebugger: string;
+  host: string | undefined;
```
grabbou, line 55: Wouldn't it be better to do `host = 'localhost'` right in the method declaration? (line 50, I believe)

grabbou: Also, we already have similar code here https://github.com/react-native-community/cli/blob/c23e16192f8d23915c14a11e54c3d5ffdfb243df/packages/cli-server-api/src/devToolsMiddleware.ts#L16. 

Potential code smell - we have default value in two different places - what if they get out of sync in the future?

Recommendation: set default value of `host` in this method https://github.com/react-native-community/cli/blob/c23e16192f8d23915c14a11e54c3d5ffdfb243df/packages/cli-server-api/src/devToolsMiddleware.ts#L34 before passing it to any `launchXDebugger`

grabbou:
Please run `yarn lint` locally to verify failing `lint` tests on CircleCI.

grabbou:
Thanks for the PR. Can you provide a bit more context into what your `REACT_DEBUGGER` flag currently is and what exactly are you having problems with?

As per the documentation:
> REACT_DEBUGGER="node /path/to/launchDebugger.js --port 2345 --type ReactNative
the command will be used to launch your debugger and that's where you can pass a custom `--port` to it.

Sure, having an environmental variable to automaticalally detect the host/port would be useful, but it would need to be adopted by every custom debuggero out there. Which makes we not like this solution that much. Always down to reduce the amount of proprietary code.

packages/cli-server-api/src/devToolsMiddleware.ts
343max: Yeah, makes sense. I was thinking the same but didn't feel confident enough to change the code more then necessary.

343max:
Hey!

Thanks for the tip about `yarn lint` - I made sure to run it but couldn't find any issues that where related to my changes.

Also I updated the description. Hopefully it explains now better what I'm. trying to accomplish.

packages/cli-server-api/src/devToolsMiddleware.ts
```diff
@@ -34,31 +29,49 @@ function launchDevTools(
   {host, port, watchFolders}: LaunchDevToolsOptions,
   isDebuggerConnected: () => boolean,
 ) {
+  const hostname = host ?? 'localhost';
```
andriytsaryov, line 32: Seems to be very confusing using both host and hostname all around in this function. Typically hostname is a host name and it cannot contain port number while host contains both host name and port number.

grabbou:
I am going to close this PR since there are some questions that remain unanswered for almost 2 years and there is no activity on this PR (which makes me believe there's not much interest for this particular feature). I believe some of the workflows have changed in the meantime, due to Flipper release.

Thank you for spending your time on making this contribution, I would be happy to revisit this PR in the future when there is more requests for it.

PR #1281, opened 2020-09-28
chore: bump Node types to match supported Node version

Simek:
Summary:
---------

Refs: #1072

This small PR bumps the Node types package (`@types/node`) to the `10.0.0` version to match the current [minimal supported version](https://github.com/react-native-community/cli/blob/master/packages/cli/package.json#L21) of Node required for the CLI.

Test Plan:
----------

Successful `yarn install` run!

### DIFF
--- a/packages/cli/src/commands/start/watchMode.ts
+++ b/packages/cli/src/commands/start/watchMode.ts
@@ -39,7 +39,7 @@ function enableWatchMode(messageSocket: any) {
           process.exit();
           break;
         case 'z':
-          process.emit('SIGTSTP');
+          process.emit('SIGTSTP', 'SIGTSTP');
           break;
       }
     } else if (name === 'r') {
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
     "@types/glob": "^7.1.1",
     "@types/jest": "^25.1.4",
     "@types/mkdirp": "^0.5.2",
-    "@types/node": "^8.0.0",
+    "@types/node": "^10.0.0",
     "@types/node-fetch": "^2.3.7",
     "babel-jest": "^25.2.4",
     "babel-plugin-module-resolver": "^3.2.0",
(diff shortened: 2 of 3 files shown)


Simek:
According to the CLI error - according to TS types it looks like emitting Signal to `process` [requires two parameters](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/process.d.ts#L322), where both of them are signals. Only REPL `emit` seems to [accept one argument](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/repl.d.ts#L309) (but not all Signals).

packages/cli/src/commands/start/watchMode.ts
```diff
@@ -39,7 +39,7 @@ function enableWatchMode(messageSocket: any) {
           process.exit();
           break;
         case 'z':
-          process.emit('SIGTSTP');
+          process.emit('SIGTSTP', 'SIGTSTP');
```
grabbou, line 42: Curious why these two (same) arguments are required? What they're supposed to do?

grabbou (approved):
LGTM, just a comment to better understand one thing.

packages/cli/src/commands/start/watchMode.ts
Simek: It looks like process emit is based on the Events emitter https://nodejs.org/api/events.html#events_emitter_emit_eventname_args and the second argument is passed to the user defined listener/handler, but I'm not 100% sure about this. 😉

Simek:
BTW. Refs. https://github.com/facebook/react-native-website/pull/2267

I think that minimum Node version bump to 12 should also be considered for the CLI in the near future.

packages/cli/src/commands/start/watchMode.ts
grabbou: I see. We could've used it then to pass some useful data but I don't think we have a use case for it right now.

PR #1286, opened 2020-10-06
docs: remove legacy init testing section

thymikee:
Summary:
---------

No need for verdaccio, we can run init command locally, without proxying it through `react-native` binary.

Test Plan:
----------

None


grabbou:
Finally!

PR #1287, opened 2020-10-06
chore: update logo banner character and colors

Simek:
Summary:
---------

This PR updates the charter used in the React Native banner logo from hash `#` to [full block](https://www.fileformat.info/info/unicode/char/2588/index.htm) `█` and updates the banner and title colors to match them better to actual logo color and to fix the readability issue in Windows PS.

#### Windows PS issue
Screenshot 2020-10-06 140140

Test Plan:
----------

I have tested the changes locally on macOS and Windows, running `init` command locally.

### macOS Terminal
Screenshot 2020-10-06 at 18 07 47

### Windows Console
Screenshot 2020-10-06 182041

### Windows PS
Screenshot 2020-10-06 182206

### Windows custom terminal
Screenshot 2020-10-06 181904


thymikee (approved):
Lovely!

thymikee:
I think Metro uses the same ascii art somewhere in their codebase? Would be good to update it there as well

cpojer:
I'm not a huge fan of the full block characters to be honest. Could we only use those on Windows to fix the existing issue?

Simek:
> I'm not a huge fan of the full block characters to be honest. Could we only use those on Windows to fix the existing issue?

@cpojer The character change was only for the visual purpose, changing color is enough to fix the PS issue. I will revert the character back to the hash.

thymikee (approved):

pranshuchittora:
Just curious to know, how to convert an image to `#` any online tool / script

PR #1288, opened 2020-10-06
feat: remove Python checks from doctor

thymikee:
Summary:
---------

Python is seemingly only necessary to _some_ Node modules that provide native extensions. We can either remove it or mark it as not required. This PR removes the code associated with Python checks, to simplify the codebase. 

cc @molant @acoates-ms @Simek 

Test Plan:
----------

Code is removed, doctor command works as expected


acoates-ms:
@molant added this, so he's the one we should try to find out from why he thought it was needed.  But I have no issue removing it.

molant:
We've been talking offline about this 😊
I added it because it was in the instructions but it looks like python is not used on Windows to create Android apps anymore (or we haven't found that case yet).
It should be safe to remove.

Simek:
I also was able to verify that the removal is safe. I have tested the starting guide steps on the brand new Windows VM skipping Python installation.

Refs: https://github.com/facebook/react-native-website/pull/2250

grabbou (approved):

grabbou:
Thank you @thymikee for sending this over!

PR #1294, opened 2020-10-12
support HTTP_PROXY environment

lotosbin:


grabbou:
Thank you for your first contribution. Could you please provide a bit more details, what is that you're trying to accomplish, is there an underlying issue here we can follow for more details?

lotosbin:
somewhere cannot direct access github.com.
when using a proxy, Failed to fetch diff.
the reason is this react native cli do not use system env HTTP_PROXY to create proxy.

packages/tools/src/fetch.ts
```diff
@@ -56,7 +57,11 @@ const fetch = async (
   url: string | Request,
   options?: FetchOptions,
 ): Promise<{status: number; data: any; headers: Headers}> => {
-  const result = await nodeFetch(url, options);
+  const instanceOptions = options || {};
```
grabbou, line 60: Instead of this, you can modify line `58` to use default value.

Example: 
```
options: FetchOptions = {}
```

grabbou:
Thanks.

I think that there are a few more factors beyond `HTTP_PROXY` that we should factor in and unfortunately, [`node-fetch` has no support for it out of the box](https://github.com/node-fetch/node-fetch/issues/195#issuecomment-259094022).

`request`, on the other hand, [does have built-in support for various env variables](https://github.com/request/request#controlling-proxy-behaviour-using-environment-variables). 

I think it would be quite trivial to change to `request`, given that we're just using it to fetch a JSON and do not even take advantage of any other library specific features. 

What do you think?

thymikee:
`request` is too big: https://packagephobia.com/result?p=request

grabbou:
Let's stick to custom implementation, as in the PR.

PR #1295, opened 2020-10-13
feat(ios): parse xtrace list devices

DomiR:
Summary:
---------
Running instruments is deprecated. We need to switch over to `xcrun xctrace list devices` in the future. This is a starting point. We can probably parse the `xcrun --version` and switch to this if it supported.
`instruments` is now deprecated in favor of 'xcrun xctrace' (see `man xctrace` for more information on its replacement)

Test Plan:
----------

Added jest test.

Please let me know if I should procede.


thymikee:
Thanks! Is xctrace available in Xcode 11?

DomiR:
I'm on xCode 12 already, so I can't tell. As I mentioned I would add reading the xcode version via `envinfo` and use xctrace only if version >= 12.

grabbou:
Hm, is this PR completed? I don't see the added function being actually used somewhere else in the code.

grabbou:
Ah, I see!

> This is a starting point.

Would you be open to complete this PR? Since `xctrace` is available as of Xcode 12, we will need to detect `Xcode` version before deciding on which approach to use. 

We will drop that `if` statement when we drop `Xcode 11` support.

grabbou (changes requested):
Ah, I see!

> This is a starting point.

Would you be open to complete this PR? Since `xctrace` is available as of Xcode 12, we will need to detect Xcode version before deciding on which approach to use.

In the current form, I can't really merge this PR as the added function is not used anywhere.

DomiR:
Yes, I'll try to completed this week. I started working on a custom solution, but then saw that there’s already tooling for detecting the Xcode version in the repository, so I’ll use that instead...
Am 3. Nov. 2020, 15:48 +0100 schrieb Mike Grabowski <notifications@github.com>:

packages/cli/src/commands/info/info.ts
```diff
@@ -6,8 +6,7 @@
  */
 
 // @ts-ignore untyped
```
thymikee, line 8:
```suggestion
```

packages/platform-ios/src/commands/runIOS/__tests__/parseXtraceIOSDevicesList.test.ts
```diff
@@ -0,0 +1,78 @@
...
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ */
+
+import parseXtraceIOSDevicesList from '../parseXtraceIOSDevicesList';
+
+jest.dontMock('../parseXtraceIOSDevicesList');
```
thymikee, line 11: We don't automock, so this is not needed
```suggestion
```

packages/platform-ios/src/commands/runIOS/__tests__/parseXtraceIOSDevicesList.test.ts
```diff
@@ -0,0 +1,78 @@
...
+
+describe('parseXtraceIOSDevicesList', () => {
+  it('parses typical output', async () => {
```
thymikee, line 14:
```suggestion
  it('parses typical output', () => {
```

packages/platform-ios/src/commands/runIOS/index.ts
```diff
@@ -83,11 +86,20 @@ function runIOS(_: Array<string>, ctx: Config, args: FlagsT) {
     );
   }
 
-  const devices = parseIOSDevicesList(
-    child_process.execFileSync('xcrun', ['instruments', '-s'], {
-      encoding: 'utf8',
-    }),
-  );
+  const env = await getEnvironmentInfo(true);
```
thymikee, line 89: Running full envinfo is pretty slow. Can we optimize it so that only Xcode info is searched for?

DomiR: envinfo exposes all it's helpers... we could add getXcodeVersion() to envinfo and run only that one helper and maybe even parse (split at '/' as it includes build number) it right there.

grabbou: I think a better way would be to revert all `envinfo` related changes in this PR and just do something like:
```js
    child_process.execSync('xcrun -f xctrace');
```
This will throw when `xctrace` is not available, which is fine.

Something like this:
```js
try {
    // This will throw when `xctrace` is not available. We fallback to `instruments`.
    child_process.execSync('xcrun -f xctrace', {stdio: 'ignore'});
    devices = parseXtraceIOSDevicesList(...);
  } catch (e) {
    devices = parseIOSDevicesList(...);
  }
```
should do the trick.

grabbou: ^ with that in mind, we don't need `semver` either.

packages/platform-ios/src/commands/runIOS/parseIOSDevicesList.ts
```diff
@@ -26,7 +26,7 @@ function parseIOSDevicesList(text: string): Array<Device> {
 
   text.split('\n').forEach((line) => {
     const device = line.match(
-      /(.*?) (\(([0-9\.]+)\) )?\[([0-9A-F-]+)\]( \(Simulator\))?/i,
+      /(.*?) (\(([0-9.]+)\) )?\[([0-9A-F-]+)\]( \(Simulator\))?/i,
```
grabbou, line 29: Is this change intentional?

packages/platform-ios/src/commands/runIOS/index.ts
```diff
@@ -83,11 +86,20 @@ function runIOS(_: Array<string>, ctx: Config, args: FlagsT) {
...
+  const xcodeVersion = semver.coerce(env.IDEs.Xcode.version.split('/')[0]);
+  const devices =
+    xcodeVersion != null && semver.satisfies(xcodeVersion, '>= 12.x')
+      ? parseXtraceIOSDevicesList(
+          child_process.execFileSync('xcrun', ['xctrace', 'list', 'devices'], {
+            encoding: 'utf8',
+          }),
```
grabbou, line 96: Also, while testing locally, I am getting output of this command printed in my terminal and `text` passed to `parseXtraceIOSDevicesList` is `null`.

Not sure why is that. Do we need to overwrite `stdio` here? The best scenario is when there's nothing outputted to the console. We just get the list and assign to `devices`.

We do the printing later anyways, on our own. Otherwise, we will print list of devices twice.

packages/platform-ios/src/commands/runIOS/index.ts
```diff
@@ -83,11 +86,20 @@ function runIOS(_: Array<string>, ctx: Config, args: FlagsT) {
...
-  const devices = parseIOSDevicesList(
-    child_process.execFileSync('xcrun', ['instruments', '-s'], {
-      encoding: 'utf8',
-    }),
-  );
+  const env = await getEnvironmentInfo(true);
+  const xcodeVersion = semver.coerce(env.IDEs.Xcode.version.split('/')[0]);
+  const devices =
+    xcodeVersion != null && semver.satisfies(xcodeVersion, '>= 12.x')
+      ? parseXtraceIOSDevicesList(
```
grabbou, line 93: Hint: It's `parseXctraceIOSDevicesList` 👍

packages/tools/src/envinfo.ts
```diff
@@ -35,7 +35,7 @@ async function getEnvironmentInfo(
       Binaries: ['Node', 'Yarn', 'npm', 'Watchman'],
       IDEs: ['Xcode', 'Android Studio', 'Visual Studio'],
       Managers: ['CocoaPods'],
-      Languages: ['Java', 'Python'],
+      Languages: ['Java'],
```
grabbou, line 38: Please ignore this change. Did it to make it easier to rebase.

grabbou: When you move `envinfo` where it was, just nevert this change too :)

grabbou:
Thank you for your work and apologies that it's taking some time! I am learning new things as I am going through this PR every time! 

Hopefully we can ship it soon, because it's needed for `0.64` and few other PRs! Keeping fingers crossed.

packages/platform-ios/src/commands/runIOS/parseIOSDevicesList.ts
DomiR: no, my editor probably removed that on save

packages/tools/src/envinfo.ts
thymikee: Python is no longer necessary on Windows, so no need to declare it here, I'd keep it removed :)

grabbou: Yes, but this code is not based on latest master and at the time of cutting the original branch, Python was still in the source. I removed it hoping that Github UI would let me resolve conflicts in the UI, otherwise it says that they're too complex.

Either way, let's just revert all changes done to this file in this PR and Github will do the rest.

grabbou:
I updated the PR with some tweaks, it's ready to be merged.

grabbou (approved):

thymikee:
@grabbou please fix linter issues

packages/platform-ios/src/commands/runIOS/index.ts
```diff
@@ -83,11 +85,19 @@ function runIOS(_: Array<string>, ctx: Config, args: FlagsT) {
...
 
-  const devices = parseIOSDevicesList(
-    child_process.execFileSync('xcrun', ['instruments', '-s'], {
-      encoding: 'utf8',
-    }),
-  );
+  let devices;
+  try {
+    devices = parseXctraceIOSDevicesList(
+      execa.sync('xcrun', ['xctrace', 'list', 'devices']).stderr,
```
bjackson, line 91: Why do we parse from stderr instead of stdout? I'm debugging an issue where no devices are returned, and I see that changing this line to: `execa.sync('xcrun', ['xctrace', 'list', 'devices']).stdout,` outputs the expected list of devices and simulators.

PR #1298, opened 2020-10-19
feat: allow adding modules only for the specific build type

lukmccall:
Summary:
---------

Currently, you can't use the auto-linking with modules that should be added only to a specific build type (for example, only to the debug builds). With those changes, people can easier install helper packages. I believe this is a really nice addition to the `react-native-cli`.

Why have I added two fields (under `ios` and `android`)?
I believe in that way we give users the most control over their packages. Moreover, users people can use everything that native platforms give them - build variants/flavors on Android, pod configuration on iOS.

Test Plan:
----------

- create a package with custom `react-native.config.js`
```
module.exports = {
    dependency: {
        platforms: {
            android: {
                buildTypes: ['debug']
            },
            ios: {
                buildTypes: ['debug']
            }
        }
    }
}
```
- link this package into your RN project 
- check if dependency is added only to the `debug` build - not to the release one

### DIFF
--- a/docs/dependencies.md
+++ b/docs/dependencies.md
@@ -111,7 +111,7 @@ See [`script_phase` options](https://www.rubydoc.info/gems/cocoapods-core/Pod/Po
 
 #### platforms.ios.configurations
 
-An array of build configurations which will include the dependency. If the array is empty, your dependency will be installed in all configurations. If you're working on a helper library, you should set debug as a value, to avoid shipping it as a part of the release process. For more details, see [`build configurations`](https://guides.cocoapods.org/syntax/podfile.html#pod).
+An array of build configurations which will include the dependency. If the array is empty, your dependency will be installed in all configurations. If you're working on a helper library that should only be included in development, such as a replacement for the React Native development menu, you should set this to `['debug']` to avoid shipping the library in a release build. For more details, see [`build configurations`](https://guides.cocoapods.org/syntax/podfile.html#pod).
 
 #### platforms.android.sourceDir
 
@@ -133,7 +133,7 @@ For settings applicable on other platforms, please consult their respective docu
 
 #### platforms.android.buildTypes
 
-An array of build variants or flavors which will include the dependency. If the array is empty, your dependency will be included in all build types. If you're working on a helper library, you should set debug as a value, to avoid shipping it as a part of the release process. For more details, see [`build variants`](https://developer.android.com/studio/build/build-variants#dependencies).
+An array of build variants or flavors which will include the dependency. If the array is empty, your dependency will be included in all build types. If you're working on a helper library that should only be included in development, such as a replacement for the React Native development menu, you should set this to `['debug']` to avoid shipping the library in a release build. For more details, see [`build variants`](https://developer.android.com/studio/build/build-variants#dependencies).
 
 ### assets
 
--- a/docs/dependencies.md
+++ b/docs/dependencies.md
@@ -58,6 +58,7 @@ type DependencyParamsIOST = {
   project?: string;
   podspecPath?: string;
   sharedLibraries?: string[];
+  buildTypes?: string[];
 };
 
 type DependencyParamsAndroidT = {
@@ -65,6 +66,7 @@ type DependencyParamsAndroidT = {
   manifestPath?: string;
   packageImportPath?: string;
   packageInstance?: string;
+  buildTypes?: string[];
 };
 ```
 
@@ -107,6 +109,10 @@ module.exports = {
 
 See [`script_phase` options](https://www.rubydoc.info/gems/cocoapods-core/Pod/Podfile/DSL#script_phase-instance_method) for a full list of available object keys.
 
+#### platforms.ios.buildTypes
+
+An array of build configurations which will include the dependency. If the array is empty, your dependency will be installed in all configurations. For more details, see [`build configurations`](https://guides.cocoapods.org/syntax/podfile.html#pod).
+
 #### platforms.android.sourceDir
 
 A relative path to a folder with Android project (Gradle root project), e.g. `./path/to/custom-android`. By default, CLI searches for `./android` as source dir.
@@ -124,6 +130,10 @@ Custom package import. For example: `import com.acme.AwesomePackage;`.
 Custom syntax to instantiate a package. By default, it's a `new AwesomePackage()`. It can be useful when your package requires additional arguments while initializing.
 
 For settings applicable on other platforms, please consult their respective documentation.
+
+#### platforms.android.buildTypes
+
+An array of build variants or flavors which will include the dependency. If the array is empty, your dependency will be included in all build types. For more details, see [`build variants`](https://developer.android.com/studio/build/build-variants#dependencies).
 
 ### assets
 
--- a/docs/dependencies.md
+++ b/docs/dependencies.md
@@ -58,7 +58,7 @@ type DependencyParamsIOST = {
   project?: string;
   podspecPath?: string;
   sharedLibraries?: string[];
-  buildTypes?: string[];
+  configurations?: string[];
 };
 
 type DependencyParamsAndroidT = {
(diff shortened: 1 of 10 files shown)


thymikee:
Haven't look yet, but the idea seems reasonable. @grabbou can you have a look?

packages/platform-ios/native_modules.rb
```diff
@@ -87,7 +88,11 @@ def use_native_modules!(config = nil)
 
     relative_path = podspec_dir_path.relative_path_from project_root
 
-    pod spec.name, :path => relative_path.to_path
+    if buildTypes.nil? || buildTypes.empty?
```
grabbou, line 91: Since this is Ruby, one could wonder if this can be simplified. 

(would something happen if the code was just: `pod spec.name, :path => relative_path.to_path, :configurations => buildTypes`)?

docs/dependencies.md
```diff
@@ -58,13 +58,15 @@ type DependencyParamsIOST = {
   project?: string;
   podspecPath?: string;
   sharedLibraries?: string[];
+  buildTypes?: string[];
```
grabbou, line 61: I would keep the original naming to not introduce additional layer of confusion. We already have `scriptPhases` that are CocoaPods-specific (and others, such as `sourceDir`) and will most likely continue doing so in the future. What you think?

docs/dependencies.md
```diff
@@ -107,6 +109,10 @@ module.exports = {
 
 See [`script_phase` options](https://www.rubydoc.info/gems/cocoapods-core/Pod/Podfile/DSL#script_phase-instance_method) for a full list of available object keys.
 
+#### platforms.ios.buildTypes
+
+An array of build configurations which will include the dependency. If the array is empty, your dependency will be installed in all configurations. For more details, see [`build configurations`](https://guides.cocoapods.org/syntax/podfile.html#pod).
```
grabbou, line 114: We could probably suggest here that if you're working on a helper, you should set `debug` as a value, to avoid shipping it as a part of the release process. 

I imagine some libraries may want to do it by default.

grabbou (commented):
Great work, thank you for submitting that. I do see a valid use case here.

CC: @alloy for your review on these

alloy:
Thanks for taking the time to contribute! The implementation looks good to me 👌 

Having said that, the premise for this change isn’t actually true:

> Currently, you can't use the auto-linking with modules that should be added only to a specific build type (for example, only to the debug builds).

You _should_ be able to specify that a specific pod is only meant for certain configurations (or any type of customisation) in your `Podfile` and the auto-linking code in this package should then simply [skip the pod](https://github.com/react-native-community/cli/blob/003be0c8cfa9bc361503eee1bb68a2a05a1ff49a/packages/platform-ios/native_modules.rb#L81-L84) in question.

_(If this doesn’t currently work, I consider it a bug and that should be addressed.)_

While I don’t have a issue against this specific case, from a design perspective I do have some hesitation. The auto-linking code was meant to do one thing, namely to identify and activate native dependencies that, solely by the virtue of having been added to a project’s `package.json`, should be linked into the application. Anything more than that means we start abstracting away the underlying tools that already do their job well and introduces the chance to end-up with leaky abstractions, requiring one to know both the abstractions _and_ underlying tools.

I’m sorry you were led to believe this use-case couldn’t be addressed before, which is likely due to a lack of documentation on that aspect, and appreciate you stepping up to try and address that. However, because of my above reasoning, I’m going to have to vote against adding this feature.

lukmccall:
@alloy, I understand that I can exclude a module from the auto-linking mechanism, but I don't fit in my use case. Maybe I wasn't clear enough. I've been currently working on a react-native module that should be used only with debug builds. It's a helper module for the developers. The distribution of that module is a lite bit painful because people need to add it to their projects manually. I can't use the auto-linking mechanism, because, from the package's point of view, I can't affect how the pod (or android library) is added to the project. That's why I decided to create this PR.

alloy:
I see. Just to be clear, when you say “manually” you mean something like the following?

```ruby
pod “some-dev-lib”, :path => “../node_modules/some-dev-lib”, :configurations => [“Debug”]
```

(Written on phone, so quotes might be wrong.)

I personally think that libs should not be in control about when they are loaded, rather the application should control that to stay flexible. (I have wanted to use dev helpers in release mode every now and then.) Thus this use-case doesn’t change my hesitation when it comes to the aforementioned design thoughts.

alloy:
Btw, one thing that I don’t know offhand is if the Android auto-linking support similarly allows one to specify a de manually with such configuration and skip it during auto-linking. Do you know?

grabbou:
I generally fully agree with @alloy on how modifications to linking should be done on per project basis - sticking to CocoaPods is most powerful and flexible way and always better than having custom abstractions.

> The distribution of that module is a lite bit painful because people need to add it to their projects manually. I can't use the auto-linking mechanism, because, from the package's point of view, I can't affect how the pod (or android library) is added to the project. That's why I decided to create this PR.

I can see this being useful for packages that should stay "debug" by default. With current setup, there will always be one additional step required - to go to a `Podfile` and define it explicitly.

That leads me to the next question - is there any way for a package to specify it's debug only in its own `podspec`?

lukmccall:
> That leads me to the next question - is there any way for a package to specify it's debug only in its own `podspec`?

I didn't find a way to do it in package `podspec`.

alloy:
Indeed, there is no way to specify that in a `podspec`. From my perspective this is a documentation matter, much like the `yarn add --dev Foo` lines you’ll see in many READMEs, not something we need to add any complexity for _or_ open the door to more complexity in the future.

brentvatne:
I think this is a reasonable addition and worth the small increase in API surface area in order to reduce the friction of installing development plugins; for example: if you wanted to install a library to replace the React Native development menu.

If you do end up wanting to include a development plugin in a release build, you could explicitly add it in your Podfile so it's skipped by auto-linking. This way the edge case requires an additional step rather than putting the work on the default path when installing a development plugin.

alloy:
> If you do end up wanting to include a development plugin in a release build, you could explicitly add it in your Podfile so it's skipped by auto-linking. This way the edge case requires an additional step rather than putting the work on the default path when installing a development plugin.

This is a good point 👍

docs/dependencies.md
```diff
@@ -107,6 +109,10 @@ module.exports = {
 
 See [`script_phase` options](https://www.rubydoc.info/gems/cocoapods-core/Pod/Podfile/DSL#script_phase-instance_method) for a full list of available object keys.
 
+#### platforms.ios.configurations
+
+An array of build configurations which will include the dependency. If the array is empty, your dependency will be installed in all configurations. If you're working on a helper library, you should set debug as a value, to avoid shipping it as a part of the release process. For more details, see [`build configurations`](https://guides.cocoapods.org/syntax/podfile.html#pod).
```
brentvatne, line 114:
```suggestion
An array of build configurations which will include the dependency. If the array is empty, your dependency will be installed in all configurations. If you're working on a helper library that should only be included in development, such as a replacement for the React Native development menu, you should set this to `['debug']` to avoid shipping the library in a release build. For more details, see [`build configurations`](https://guides.cocoapods.org/syntax/podfile.html#pod).
```

docs/dependencies.md
```diff
@@ -125,6 +131,10 @@ Custom syntax to instantiate a package. By default, it's a `new AwesomePackage()
 
 For settings applicable on other platforms, please consult their respective documentation.
 
+#### platforms.android.buildTypes
+
+An array of build variants or flavors which will include the dependency. If the array is empty, your dependency will be included in all build types. If you're working on a helper library, you should set debug as a value, to avoid shipping it as a part of the release process. For more details, see [`build variants`](https://developer.android.com/studio/build/build-variants#dependencies).
```
brentvatne, line 136:
```suggestion
An array of build variants or flavors which will include the dependency. If the array is empty, your dependency will be included in all build types. If you're working on a helper library that should only be included in development, such as a replacement for the React Native development menu, you should set this to `['debug']` to avoid shipping the library in a release build. For more details, see [`build variants`](https://developer.android.com/studio/build/build-variants#dependencies).
```

brentvatne (approved):

brentvatne:
:shipit: ?

packages/platform-ios/native_modules.rb
```diff
@@ -86,9 +87,7 @@ def use_native_modules!(config = nil)
     podspec_dir_path = Pathname.new(File.dirname(podspec_path))
 
     relative_path = podspec_dir_path.relative_path_from project_root
-
-    pod spec.name, :path => relative_path.to_path
-
+    pod spec.name, :path => relative_path.to_path, :configurations => configurations
```
grabbou, line 90: What would happen when the `configurations` is `[]`? (the default value, when nothing is passed)

The desired behaviour: dependency gets added to all configurations. 

Unwanted behaviour: dependency is not added to any configuration.

grabbou:
Looks good to me. Let's ship it.

Just left a quick comment to better understand the behaviour when no value is present.

packages/platform-ios/native_modules.rb
lukmccall: It will add it to all configurations ;)

Issue #1202, opened 2020-06-16
Support XDG base directory specification


distefam:
## Describe the Feature
The React Native CLI should place its configuration files according to the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html).

## Related Issues
[Expo CLI](https://github.com/expo/expo-cli/issues/2274)

codingjerk:
Closed by PR #1303

PR #1303, opened 2020-11-01
feat: use OS-specific paths instead of .react-native-cli/cache directory

codingjerk:
Summary:
---------

Makes react-native-cli use OS-specific paths to store cache instead of `.react-native-cli/cache`.

It uses consistent cache paths across all popular OS' and provides backward compatibility and fallback options via appdirsjs package.

Also, there is related issue requesting XDG Base Directory Support in Linux: #1202.

Test Plan:
----------

Run any react-native command, that requires cache while `~/.react-native-cli/cache` directory exists. It should work just like before PR.

Remove `~/.react-native-cli` and run same command again. It should create OS-specific directory (e.g. `~/.cache/react-native-cli` on Linux).


packages/cli/package.json
```diff
@@ -33,6 +33,7 @@
     "@react-native-community/cli-server-api": "^4.13.0",
     "@react-native-community/cli-tools": "^4.13.0",
     "@react-native-community/cli-types": "^4.10.1",
+    "appdirsjs": "^1.2.1",
```
thymikee, line 36: Adding a 17kB dependency just for that? How about using `os.tmpdir`?

codingjerk: @thymikee

Apposite remark, but `os.tmpdir` has different behaviour. It returns a path to a system's cache directory and not a user's one.

Also, there is some logic to use the legacy path for backward compatibility.

thymikee: Yea, I don't really mind breaking this, it's only used for reminding a user that there's a new version of RN once in a while, nothing critical. I do appreciate your care about this though! 

Wonder if we really need a user directory _now_? Maybe it we cached more things, then I'd be on board with using this module. But now I can't justify using it. Also, a lot of tools use system's cache dir and it works just fine, although they usually make the path configurable.

codingjerk: Alright, that's reasonable. I believe we do need a user's directory cause it is a standard de facto and we will need to store the cache there eventually. Yeah, YAGNI, but it's quite easy to use the user's cache directory.

So, I vote for storing cache there. If we don't want unnecessary dependencies, I can inline the code.

satya164: I think for linux it'll be good enough to use `process.env.XDG_CACHE_HOME` and fallback to `~/.cache` to use the standard cache directory.

thymikee:
Ok, I've just had a look at the source of the library at https://github.com/codingjerk/appdirsjs and I like it. What needs improvement is the package size. Seems like there's a bunch of unnecessary stuff published to npm, like spec files for example or some config. Ideally the project would use `files` allow list instead of block list in a form of `.npmignore`. Would you mind doing these changes to the lib and come back here with the updated numbers from https://packagephobia.com/result?p=appdirsjs ? :)

codingjerk:
@thymikee 

Sure. Package size matters, and I'll add `.npmignore`. Thanks for pointing it out

thymikee (approved):

PR #1306, opened 2020-11-04
Update default iOS simulator and fallback

kevinnguy:
Summary:
---------

Updating the default iOS simulator to `iPhone 12` and adding `iPhone 11` as a fallback simulator to support those not using Xcode 12. I'm tempted to remove `iPhone X` as a fallback since `iPhone 11` has been used as the default simulator.

Test Plan:
----------

yarn
yarn lint
yarn test


thymikee (approved):
👍

PR #1309, opened 2020-11-07
Support experimental transform profile for bundle command

janicduplessis:
Summary:
---------

This adds a new command line option `--unstable-transform-profile` and simply passes the value to metro.

Test Plan:
----------

Tested in an app by adding `extraPackagerArgs: ["--unstable-transform-profile", "hermes-canary"],` in the build.gradle react config.


janicduplessis:
Not sure if we want to merge this now since the feature is still experimental, but wanted to put the code out there.

thymikee (approved):
I don't mind. It's prefixed with "unstable" so things should be expected to change :)

PR #1314, opened 2020-11-16
fix getProductName regex (run-ios)

immackay:
Summary:
---------

`getProductName` used in the `run-ios` command from `cli-platform-ios` was failing to correctly parse build output and instead only returned the scheme name passed. It seems that the build output has a backslash before all equals signs, so this change reflects that.

Test Plan:
----------

Run `react-native run-ios` with a scheme name different from the CFBundleIdentifier.

### DIFF
--- a/packages/platform-ios/src/commands/runIOS/index.ts
+++ b/packages/platform-ios/src/commands/runIOS/index.ts
@@ -441,7 +441,7 @@ function getBuildPath(
 }
 
 function getProductName(buildOutput: string) {
-  const productNameMatch = /export FULL_PRODUCT_NAME="?(.+).app"?$/m.exec(
+  const productNameMatch = /export FULL_PRODUCT_NAME\\="?(.+).app"?$/m.exec(
     buildOutput,
   );
   return productNameMatch ? productNameMatch[1] : null;


packages/platform-ios/src/commands/runIOS/index.ts
```diff
@@ -441,7 +441,7 @@ function getBuildPath(
 }
 
 function getProductName(buildOutput: string) {
-  const productNameMatch = /export FULL_PRODUCT_NAME="?(.+).app"?$/m.exec(
+  const productNameMatch = /export FULL_PRODUCT_NAME\\="?(.+).app"?$/m.exec(
```
thymikee, line 444: Can you make this backslash optional? The regex worked earlier, so that's probably some recent change.

thymikee (changes requested):
Can you add a test or at least a video/screenshot of when it happens and how this change fixes it? :)

grabbou:
Reproduced, got the error with custom scheme name:

> xcodebuild: error: The workspace named "RNTestProject" does not contain a scheme named "RNTestProject". The "-list" option can be used to find the names of the schemes in the workspace.

The scheme name I set was `RNTestProjectTestSchemeName`. Looking into it.

grabbou:
I am going to close this PR and send a follow up PR that refactors this entire code. Looking for `appName` is not necessary and relying on it to detect tvOS is a bit brittle. 

Will cross-link. Thank you for raising our awareness on this.

PR #1315, opened 2020-11-17
chore(breaking): use Node 12 as compile target

Naturalclar:
Summary:
---------

Updated node version of CI to use v12, as supported node version for React Native will be v12 from react-native v0.64

Test Plan:
----------
CI to pass on Node v12


thymikee (approved):
Thanks!

PR #1322, opened 2020-11-23
tests: add test case for using --npm with init

jamesgeorge007:
Summary:
---------

Increase test coverage.

Test Plan:
----------

Supplying `--npm` prefers `npm` for initialization, added a test case that ensures this behavior.


thymikee (approved):
👍

PR #1324, opened 2020-11-23
fix: options for init command have required arguments

### LINKED ISSUE #1323: options for init command should have required arguments.
## Description

Not supplying a value for any of the options available as part of the `init` command fails.

## Reproducible Demo

`react-native-cli init app --template`

Screenshot 2020-11-23 at 2 15 07 PM

## Proposed Solution

```bash
error: option `--template <string>' argument missing
```

```diff
     {
-      name: '--version [string]',
+      name: '--version <string>',
       description: 'Shortcut for `--template react-native@version`',
     },
     {
-      name: '--template [string]',
+      name: '--template <string>',
       description:
         'Uses a custom template. Valid arguments are: npm package, absolute directory prefixed with `file://`, Git repository or a tarball',
     },
@@ -21,11 +21,11 @@ export default {
       description: 'Forces using npm for initialization',
     },
     {
-      name: '--directory [string]',
+      name: '--directory <string>',
       description: 'Uses a custom directory instead of `<projectName>`.',
     },
     {
-      name: '--title [string]',
+      name: '--title <string>',
       description: 'Uses a custom app title name for application',
      }
```

/cc @thymikee

jamesgeorge007:
Summary:
---------

closes #1323 

Test Plan:
----------

### DIFF
--- a/packages/cli/src/commands/init/init.ts
+++ b/packages/cli/src/commands/init/init.ts
@@ -175,12 +175,21 @@ async function installDependencies({
   loader.succeed();
 }
 
+function validateOption(optName: string, optValue: string) {
+  if (typeof optValue === 'boolean') {
+    throw new CLIError(`error: '${optName} <string>' argument missing`);
+  }
+}
+
 async function createProject(
   projectName: string,
   directory: string,
   version: string,
   options: Options,
 ) {
+  // Required arguments check
+  options.template && validateOption('template', options.template);
+
   const templateUri = options.template || `react-native@${version}`;
 
   return createFromTemplate({
@@ -206,6 +215,11 @@ export default (async function initialize(
    * We have to use `minimist` to take that directly from `process.argv`
    */
   const version: string = minimist(process.argv).version || DEFAULT_VERSION;
+
+  // Required arguments check
+  validateOption('version', version);
+
+  options.directory && validateOption('directory', options.directory);
 
   const directoryName = path.relative(root, options.directory || projectName);
 
--- a/packages/cli/src/tools/assertRequiredOptions.ts
+++ b/packages/cli/src/tools/assertRequiredOptions.ts
@@ -22,11 +22,12 @@ export default function assertRequiredOptions(
   passedOptions: PassedOptions,
 ) {
   options.forEach((opt) => {
-    const option = new Option(opt.name);
-
-    if (!option.required) {
+    // Early return if the option is not required
+    if (!opt.required) {
       return;
     }
+
+    const option = new Option(opt.name);
 
     const name = camelCase(option.long);
 
--- a/packages/cli/src/tools/assertRequiredOptions.ts
+++ b/packages/cli/src/tools/assertRequiredOptions.ts
@@ -22,12 +22,11 @@ export default function assertRequiredOptions(
   passedOptions: PassedOptions,
 ) {
   options.forEach((opt) => {
-    // Early return if the option is not required
-    if (!opt.required) {
+    const option = new Option(opt.name);
+
+    if (!option.required) {
       return;
     }
-
-    const option = new Option(opt.name);
 
     const name = camelCase(option.long);
 
--- a/packages/cli-types/src/index.ts
+++ b/packages/cli-types/src/index.ts
@@ -28,7 +28,6 @@ export type CommandOption<T = (ctx: Config) => OptionValue> = {
   description?: string;
   parse?: (val: string) => any;
   default?: OptionValue | T;
-  required?: OptionValue;
 };
 
 export type DetachedCommandFunction<Args = Object> = (
--- a/packages/cli-types/src/index.ts
+++ b/packages/cli-types/src/index.ts
@@ -28,6 +28,7 @@ export type CommandOption<T = (ctx: Config) => OptionValue> = {
   description?: string;
   parse?: (val: string) => any;
   default?: OptionValue | T;
+  required?: OptionValue;
 };
 
 export type DetachedCommandFunction<Args = Object> = (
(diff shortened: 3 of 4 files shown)


thymikee:
e2e tests are failing

jamesgeorge007:
> e2e tests are failing

Fixed.

packages/cli/src/tools/assertRequiredOptions.ts
```diff
@@ -1,38 +0,0 @@
...
-// for options. Until that gets merged, we run the checks manually
-// https://github.com/tj/commander.js/issues/230
-export default function assertRequiredOptions(
-  options: CommandOption<any>[],
-  passedOptions: PassedOptions,
-) {
-  options.forEach((opt) => {
-    const option = new Option(opt.name);
-
-    if (!option.required) {
```
thymikee, line 27: Removing this is likely a regression. We support adding a `required` key to the command's interface, and when it's set to "true" it shall throw. Can you make sure this is still supported? Also, the linked issue is closed with a comment that it was fixed in commanderjs v4. We use v2.

jamesgeorge007: Bumped commander.js to `v4`. It should be better now.

grabbou:
BTW, we could've just set `require: true` to make it work in the current form w/o updating `commander`. I am a bit nervous about doing that kind of an update. 

Do we have a list of all breaking changes between `2.x` and `4.x`?

grabbou:
Also, keep in mind that removing `assertRequiredOptions` is going to be a breaking change for all the users that currently rely on that functionality. 

That said, I would accept a quick fix that just adds `required: true` to necessary options and a follow-up that bumps `commander` and removes that one.

thymikee:
I'm also hesitant on merging this at the moment. @grabbou suggestion is a good way to unblock ourselves for now, and once we hit a stable release we can bump commander after testing it thoroughly

jamesgeorge007:
`assertRequiredOptions()` throws for not supplying a particular option, which doesn't satisfy the use case.

grabbou:
Ah, of course - arguments are different than options! Anyway, I got a bit confused here - please help me out by answering my comments and I hope we can get through this!

packages/cli/src/tools/assertRequiredOptions.ts
```diff
@@ -22,12 +22,13 @@ export default function assertRequiredOptions(
   passedOptions: PassedOptions,
 ) {
   options.forEach((opt) => {
-    const option = new Option(opt.name);
-
-    if (!option.required) {
+    // Early return if the option is not required
+    if (!opt.required) {
```
grabbou, line 26: So, the thing is... `const option = new Option(opt.name);` creates an `Option` object from the name, e.g. `--version <string>` or `--version [string]`.

Depending on `<>` or `[]` used, the `option.required` will be true or false. 

We shouldn't be adding `required` to `CommandOption` interface and doing an early return here - that's my bad, I got you confused in my other comment.

The way it works is that `commander` creates an `Option` object with `required` based on the name, but it doesn't do the actual validation and early returns in case the value is missing. That's what this code handles here.

In other words, we the code was fine as is. If we're handling arguments, we should be looking into a different place.

grabbou:
Technically, when we set `--template <name>`, we make it ALWAYS required. In other words, `react-native init app` will be invalid invocation.

Another technical detail, `--template` is as valid as `--template name` and that's just how Commander works. In that case, its value will be `true`, because it's treated as a boolean option.

The quickest way to do it would be to go here https://github.com/react-native-community/cli/blob/aad25b360efb5b678081fc401f6e8b5004d46249/packages/cli/src/commands/init/init.ts#L180 and change this line to something that checks if `options.template` is String. If it's string, it's at least 1 character long.

packages/cli/src/tools/assertRequiredOptions.ts
jamesgeorge007: The current implementation is misleading; having a `required` key would be better IMO. The intended way would be to use `requiredOption()` once we bump `commander.js`.

packages/cli/src/commands/init/init.ts
```diff
@@ -171,12 +171,22 @@ async function installDependencies({
   loader.succeed();
 }
 
+function validateOption(optName: string, optValue: string) {
+  if (typeof optValue === 'boolean') {
+    logger.error(`error: '${optName} <string>' argument missing`);
+    process.exit(1);
```
grabbou, line 177: I would prefer we `throw new CLIError` here to stay consitent with other parts of the codebase.

grabbou:
Looks good. I'll apply a tiny naming tweak and ship it. Thanks!

grabbou:
Unfortunately, while testing your PR, I have found out that we do support this feature out of the box without any additional work. All we have to do is to correct `[string]` with `<string>` when is required.

I have sent a follow-up PR to do it: https://github.com/react-native-community/cli/pull/1328

I am really sorry that we have found this out THAT LATE in the process. I really appreciate your time and contributions and while this PR didn't get merged, we have brainstormed some interesting ideas together and have found out a better way to address the underlying problem!

https://github.com/react-native-community/cli/pull/1328

jamesgeorge007:
> Unfortunately, while testing your PR, I have found out that we do support this feature out of the box without any additional work. All we have to do is to correct `[string]` with `<string>` when is required.
> 
> I have sent a follow-up PR to do it: #1328
> 
> I am really sorry that we have found this out THAT LATE in the process. I really appreciate your time and contributions and while this PR didn't get merged, we have brainstormed some interesting ideas together and have found out a better way to address the underlying problem!
> 
> #1328

In fact, this was the initial approach https://github.com/react-native-community/cli/pull/1324#discussion_r528610053

PR #1325, opened 2020-11-23
feat: refactor template name logic

grabbou:
Summary:
---------

After carefully analysing the logic behind `templateName`, I came to the conclusion that this piece of code is not necessary and can be removed.

We don't need to parse `--template` in order to generate `{uri, name}` object. In fact, we can just pass it through to an underlying package manager installation command (such as `yarn add` or `nom install`). As a result, we reduce the heavy parsing abstraction and automatically make CLI to support all sorts of different arguments that are valid. 

This is motivated by the recent commit to bring support for `username/repo` shorthand to install templates straight from Github. Unfortunately, that doesn't factor in branches, commits and other arguments (such as protocol) that can be passed and users are used to using them.

The original motivation for this code was to probably guess the `name` of the template, so that we can locate it inside `node_modules` and copy its content after running `yarn add [templatePackage]` (we would go to `tmp/node_modules/templatePackage/` and copy the contents). 

However, we already have a `adjustNameIfUrl` function that is able to detect the name of a template after its installation. It's great and we can use it all the time, not just in cases when we were not able to guess the name.

As a result, all sorts of different protocols and arguments are now supported, subject to using `npm` or `yarn`. We don't need to parse the `--template` argument ourselves.

We also avoid issues while copying templates, e.g. when tarball name is non-standard `abc.tgz` (doesn't contain version) and when contains a different name `tmp-package-1.0.0.tgz` (where name inside `package.json` is "my-template").

Test Plan:
----------

Run `init` with `--template` and pass everything that you want that Yarn supports.


thymikee:
Some lint errors slipped in

thymikee (approved):
Nice simplification! 👍

grabbou:
Test failures unrelated, let's wait for @Esemesek to give it a closer look as well! 👍

PR #1326, opened 2020-11-23
feat: better platform name detection

### LINKED ISSUE #1282: platform-ios run iOS getProductName broken in xCode 12
## Description
xCode 12 is escaping equal signs in the build output that is parsed here, this regex will not match the new format:

https://github.com/react-native-community/cli/blob/7238b2057b9c5004b46c01854ea05a61e4a3a3d1/packages/platform-ios/src/commands/runIOS/index.ts#L444

Example of the relevant line in xCode 12 build output:
```
export FULL_PRODUCT_NAME\=AppName.app
```
Current Regex: `/export FULL_PRODUCT_NAME="?(.+).app"?$/m`
Working Regex: `/export FULL_PRODUCT_NAME\\?="?(.+).app"?$/m`

grabbou:
Summary:
---------

Fixes #1282 

We used to parse the build output for `appName` (or used scheme as a default value) only to check if it contains `tvOS` to detect whether we should spin up a `tvOS` simulator.

There's a better way to do it - we can return `buildOutput` and just look for `PLATFORM_NAME` variable which will be of `appletvsimulator`, `iphonesimulator` or `iphoneos` depending on how we built the project before.

It is unlikely that we don't find the value, but because of types, I have included an `if` statement to throw in such case, which I consider unexpected.

I have also updated regular expression to factor in potential quotation marks and `=` escaping.

As a result, we can simplify the code. Related prior work #1314 

Test Plan:
----------

Run `react-native run-ios` on device/simulator and it should be installed.

### DIFF
--- a/packages/platform-ios/src/commands/runIOS/index.ts
+++ b/packages/platform-ios/src/commands/runIOS/index.ts
@@ -27,6 +27,7 @@ import {
 } from '@react-native-community/cli-tools';
 import {Device} from '../../types';
 import ora from 'ora';
+import { platform } from 'os';
 
 type FlagsT = {
   simulator?: string;
@@ -180,7 +181,7 @@ async function runOnSimulator(
     bootSimulator(selectedSimulator);
   }
 
-  const appName = await buildProject(
+  const buildOutput = await buildProject(
     xcodeProject,
     selectedSimulator.udid,
     scheme,
@@ -190,8 +191,7 @@ async function runOnSimulator(
   const appPath = getBuildPath(
     xcodeProject,
     args.configuration,
-    appName,
-    false,
+    buildOutput,
     scheme,
   );
 
@@ -247,7 +247,7 @@ async function runOnDevice(
     );
   }
 
-  const appName = await buildProject(
+  const buildOutput = await buildProject(
     xcodeProject,
     selectedDevice.udid,
     scheme,
@@ -256,7 +256,7 @@ async function runOnDevice(
 
   const iosDeployInstallArgs = [
     '--bundle',
-    getBuildPath(xcodeProject, args.configuration, appName, true, scheme),
+    getBuildPath(xcodeProject, args.configuration, buildOutput, scheme),
     '--id',
     selectedDevice.udid,
     '--justlaunch',
@@ -357,7 +357,7 @@ function buildProject(
         return;
       }
       logger.success('Successfully built the app');
-      resolve(getProductName(buildOutput) || scheme);
+      resolve(buildOutput);
     });
   });
 }
@@ -398,19 +398,10 @@ function getTargetPaths(buildSettings: string) {
 function getBuildPath(
   xcodeProject: ProjectInfo,
   configuration: string,
-  appName: string,
-  isDevice: boolean,
+  buildOutput: string,
   scheme: string,
 ) {
-  let device;
-
-  if (isDevice) {
-    device = 'iphoneos';
-  } else if (appName.toLowerCase().includes('tvos')) {
-    device = 'appletvsimulator';
-  } else {
-    device = 'iphonesimulator';
-  }
+  const device = getPlatformName(buildOutput);
 
   const buildSettings = child_process.execFileSync(
     'xcodebuild',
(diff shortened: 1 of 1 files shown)


packages/platform-ios/src/commands/runIOS/index.ts
```diff
@@ -398,20 +398,9 @@ function getTargetPaths(buildSettings: string) {
...
 ) {
-  let device;
-
-  if (isDevice) {
-    device = 'iphoneos';
-  } else if (appName.toLowerCase().includes('tvos')) {
-    device = 'appletvsimulator';
-  } else {
-    device = 'iphonesimulator';
-  }
```
grabbou, line 413: Side effect, we also support iPad and MacOS as well. 

CC: @alloy, maybe it's time to change `run-ios` to just `run-apple`.

grabbou: (I mean not sure if the command supports these platforms right now, but this is surely first step).

packages/platform-ios/src/commands/runIOS/index.ts
```diff
@@ -441,11 +430,17 @@ function getBuildPath(
...
-    buildOutput,
+function getPlatformName(buildOutput: string) {
+  // Xcode can sometimes escape `=` with a backslash or put the value in quotes
+  const platformNameMatch = /export PLATFORM_NAME\\?="?(\w+)"?$/m.exec(
+    buildOutput
   );
-  return productNameMatch ? productNameMatch[1] : null;
+  if (!platformNameMatch) {
+    throw new CLIError(
+      'Couldn\'t find "PLATFORM_NAME" variable in your build output. Please report this issue and run your project with Xcode instead.',
```
thymikee, line 440:
```suggestion
      'Couldn\'t find "PLATFORM_NAME" variable in xcodebuild output. Please report this issue and run your project with Xcode instead.',
```

grabbou:
Needs https://github.com/react-native-community/cli/pull/1295 to full testing.

packages/platform-ios/src/commands/runIOS/index.ts
```diff
@@ -420,7 +409,7 @@ function getBuildPath(
       '-scheme',
       scheme,
       '-sdk',
-      device,
+      getPlatformName(buildOutput),
```
grabbou, line 412: Potential values that can be returned: 
> iphoneos macosx appletvos watchos iphonesimulator appletvsimulator watchsimulator

dlowder-salesforce (approved):

thymikee:
@grabbou mind fixing CI? :)

thymikee (approved):

WilbertJanney:
How do I find out which version of the package this fix is in?

liamjones:
@WilbertJanney If you click through to the commit above your comment (495d067), below the commit message at the top, at the bottom of the blue box, you should see the tags (and thus releases) that contain this commit.

WilbertJanney:
@liamjones Cool, I never knew that. I see three tags, one in black and the other two grayed out. Does that mean it's in all three versions, but the latest is shown in black?

I tried installing 5.0.1-alpha.2 (@react-native-community/cli) as a dev dependency, and then running npx react-native run-ios, but I still have the same problem.

liamjones:
> I see three tags, one in black and the other two grayed out. Does that mean it's in all three versions, but the latest is shown in black?

I _think_ so. I've not seen it documented but that's what I surmised too. 5.0.1-alpha.1 was certainly when that commit was initial included (you can see this by comparing 5.0.1-alpha.1 to 5.0.1-alpha.0, the previous release: https://github.com/react-native-community/cli/compare/v5.0.1-alpha.0...v5.0.1-alpha.1)

> I saw the code update in my node_modules folder, but I still have the same error.

Maybe you're not using the version you think you are? Try `npx react-native --version` to see what version of the CLI tools are currently being used.

WilbertJanney:
> > I see three tags, one in black and the other two grayed out. Does that mean it's in all three versions, but the latest is shown in black?
> 
> I _think_ so. I've not seen it documented but that's what I surmised too. 5.0.1-alpha.1 was certainly when that commit was initial included (you can see this by comparing 5.0.1-alpha.1 to 5.0.1-alpha.0, the previous release: [v5.0.1-alpha.0...v5.0.1-alpha.1](https://github.com/react-native-community/cli/compare/v5.0.1-alpha.0...v5.0.1-alpha.1))
> 
> > I saw the code update in my node_modules folder, but I still have the same error.
> 
> Maybe you're not using the version you think you are? Try `npx react-native --version` to see what version of the CLI tools are currently being used.

I noticed that if I update @react-native-community/cli for some reason the cli-platform-ios version is not updated that is why I wasn't seeing the change. Once I targeted @react-native-community/cli-platform@5.0.1-alpha.2 I was able to see the update.

I opted to just use npx react-native in order to do any CLI actions for now. I opted to link the images from the Images.xcassets package in iOS to fix the images missing in IPA problem for now until the script issue is fixed.

Issue #1261, opened 2020-08-28
Cleanup - detached commands should get config passed to them


acoates-ms:
With #1231 the doctor command now uses the config when run in a non-detached manor.  But since doctor is a detached command it doesn't get the config from the CLIs main loadConfig call passed to it.  So, doctor ends up having to call loadConfig again.

This is mostly just a minor performance issue.  But we could fix it by passing the config along to detached commands, -- or null if it is running detached.  I don't think this would be a compat issue, since all the detached commands are internal to the CLI currently.

grabbou:
Sounds good. I'll tackle it while working on my `commander` update.

grabbou:
Update: Right now, doctor command added certain properties to `config`, such as `healthchecks` that can be set by user/dependency. 

The idea here would be to provide a new function on top of `loadConfig`, called `loadUserConfig` or `loadAdditionalConfig` that would allow loading a custom, command-specific, additional properties, without needing to update the main package.

If, a `doctor` command` wants to use and set an additional `health checks` value, it doesn't need to interfere with `config` (main), it can simply use that function instead.

grabbou:
#1527 as it's related.

PR #1327, opened 2020-11-25
feat: upgrade to latest commander and clean up interface

grabbou:
Upgrades Commander to the latest version and removes some of the hacks that were put up in place.

Changes:
- You can now write `--paths <list...>` and have your option automatically accept array values. No need for custom parsing and splitting by a comma
- Hacky `printHelpInformation` was removed

Breaking changes:
- It now matters how you define your options, e.g. `--name <string>` and `--name [string]` previously had no effect. In this version, `name <string>` means that `--name` must always include a value and passing `--name` will throw an error. This prevents issues when users invoke CLI with `--port` and not provide value. For `--name [string]`, `[]` indicate that the value is optional and can be either boolean `--name` or string `--name Mike`. This should prevent issues like #1324 
- Options should now be passed with a space, not with a comma, e.g. `--paths a b c` instead of `--paths a,b,c`. This only affects commands built-in to the CLI. Other community plugins have an opportunity to take advantage of the new interface, but it will not be turned on automatically, unless option declaration is updated.


packages/cli/src/commands/link/link.ts
```diff
@@ -93,14 +93,12 @@ export default {
   name: 'link [packageName]',
   options: [
     {
-      name: '--platforms [list]',
+      name: '--platforms <list...>',
       description: 'Scope linking to specified platforms',
-      parse: (val: string) => val.toLowerCase().split(','),
```
grabbou, line 98: This will be a breaking change - space separated instead of comma separated.

packages/cli/src/index.ts
```diff
@@ -193,6 +144,8 @@ async function setupAndRun() {
 
   logger.setVerbose(process.argv.includes('--verbose'));
```
thymikee, line 145: this shouldn't be necessary now?

packages/cli/src/index.ts
```diff
@@ -44,134 +81,50 @@ const handleError = (err: Error) => {
...
 
-  cmd.helpInformation = printHelpInformation.bind(
-    cmd,
-    command.examples,
-    command.pkg,
-  );
+  /**
+   * Extend built-in help information by displaying examples, if defined
+   */
+  cmd.on('--help', () => {
```
thymikee, line 101: can we set similar hook for `config` command where we disable logger? https://github.com/react-native-community/cli/pull/1327/files#diff-2fb448d11702324cc59d657d5945d59cd325540eb87b71f2e66a5bc01ada834cR141-R143

packages/cli/src/index.ts
```diff
@@ -183,8 +136,6 @@ async function run() {
 }
```
thymikee, line 136: because we're now not exiting with non-0 code on handleError, `run` function should call it in catch handler

grabbou: that's actually my mistake, better to return from there because `handleError` is called from different places too.

packages/cli/src/index.ts
grabbou: true!

packages/cli/src/index.ts
grabbou: Surprisingly, doesn't work `on('command:config')`. Need to try again.

packages/cli/src/index.ts
```diff
@@ -45,133 +82,51 @@ const handleError = (err: Error) => {
...
-        if (isDetachedCommand(command)) {
-          await command.func(argv, passedOptions);
-        } else {
-          await command.func(argv, rest[0] as Config, passedOptions);
-        }
-      } catch (error) {
-        handleError(error);
-      }
-    });
+function registerCommand(...[command, config]: [command: DetachedCommand] | [command: Command, config: Config]): commander.Command {
```
thymikee, line 92: Our ESLint parser is too low to handle this syntax
```suggestion
function registerCommand(...[command, config]: [DetachedCommand] | [Command, Config]): commander.Command {
```
Would be cool to upgrade in a followup

grabbou:
Related issue to handle: https://github.com/react-native-community/cli/issues/1261#issue-688204892

grabbou:
Closing as part of it already landed. Will submit a follow-up soon with a different take.

PR #1328, opened 2020-11-26
feat: respect required/optional option values

grabbou:
Summary:
---------

Historically, we used to have `assertRequiredOptions` function present in the codebase as a workaround for a broken Commander feature.

1/ It is now not needed to have it. While testing this PR, it turned out this function was necessary as Commander correctly throws on missing values. It was either a mistake while doing the original PR or the underlying issue got fixed.

2/ We don't need any additional logic to enforce values for options, e.g. `--path <value>`. All we have to do is to use `<>` instead of `[]`. I have corrected the declaration when necessary to avoid issues.

Originally started by @jamesgeorge007 

Test Plan:
----------

Run `react-native start --port` and see it throw an error of a missing `port` value.


packages/cli-hermes/src/profileHermes/index.ts
```diff
@@ -62,9 +62,8 @@ export default {
       description: 'Generates the JS bundle and source map',
     },
     {
-      name: '--port [number]',
-      default: process.env.RCT_METRO_PORT || 8081,
-      parse: (val: number) => String(val),
```
grabbou, line 67: This is incorrect. Values to be parsed are always string.

packages/cli/src/commands/link/link.ts
```diff
@@ -98,9 +98,8 @@ export default {
       parse: (val: string) => val.toLowerCase().split(','),
     },
     {
-      name: '--all [boolean]',
+      name: '--all',
       description: 'Link all native modules and assets',
-      parse: (val: string) => val.toLowerCase().split(','),
```
grabbou, line 103: Not sure about this one, we don't support comma separated values, nor `--all true`, so just sticking to `--all`.

jamesgeorge007 (commented):
CI is failing since you have removed `assertRequiredOptions()`; but not the reference to it.
https://github.com/react-native-community/cli/blob/ef98aeefa3a17261031a101ad1499f88cf49882e/packages/cli/src/index.ts#L12

grabbou:
Good catch!

thymikee (approved):
Ok, let's do it. I added @jamesgeorge007 to the description so he gets some attribution.

jamesgeorge007:
https://github.com/react-native-community/cli/pull/1324#issuecomment-732213768 Isn't this an issue?

grabbou:
It will not.

So here's what happened. 

I thought that `assertRequiredOptions` was required because of Commander bug and it was a workaround. Turns out, it was unnecessary because when an option value was missing (after setting `<name>` instead of `[name]`), it was Commander that was throwing a bug.

So they must have fixed it in the meantime and we never refactored this code out.

PR #1329, opened 2020-11-26
chore: add e2e test for when init fails if the directory already exists

jamesgeorge007:
Summary:
---------

The init functionality fails if the specified directory already exists; added a test case for this scenario.

Test Plan:
----------

This PR aims at updating the test suite.

### DIFF
--- a/__e2e__/init.test.ts
+++ b/__e2e__/init.test.ts
@@ -39,6 +39,15 @@ afterEach(() => {
   cleanupSync(DIR);
 });
 
+test('init fails if the directory already exists', () => {
+  fs.mkdirSync(path.join(DIR, 'TestInit'));
+
+  const {stderr} = runCLI(DIR, ['init', 'TestInit'], {expectedFailure: true});
+  expect(stderr).toBe(
+    'error Cannot initialize new project because directory "TestInit" already exists.',
+  );
+});
+
 test('init --template fails without package name', () => {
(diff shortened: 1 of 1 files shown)


__e2e__/init.test.ts
```diff
@@ -39,6 +39,15 @@ afterEach(() => {
   cleanupSync(DIR);
 });
 
+test('init fails if the directory already exists', () => {
+  fs.mkdirSync(path.join(DIR, 'TestInit'));
+
+  const {stderr} = runCLI(DIR, ['init', 'TestInit'], {expectedFailure: true});
```
thymikee, line 45: Let's use a custom local template, so we don't pay the cost of fetching all of those dependencies all over again for another test. See: https://github.com/react-native-community/cli/pull/1329/files#diff-f0cd47ce844b386b0a1e17b5f44a631eaf7bdc98c5b0e4d3df7dd674e92ed87dR67-R72

jamesgeorge007: What it does is creating an empty directory, and making sure, an attempt to create a project under the same name errors. Since this is the case of failure, I don't think supplying the `--template` flag is necessary.

thymikee: Ugh, brainfart

thymikee (approved):
Thanks!

PR #1331, opened 2020-11-27
docs: update commands argument types to match source

jamesgeorge007:
Summary:
---------

Follow up of #1328

Test Plan:
----------

N/A

### DIFF
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -33,13 +33,13 @@ Builds the JavaScript bundle for offline use.
 
 Path to the root JS file, either absolute or relative to JS root.
 
-#### `--platform [string]`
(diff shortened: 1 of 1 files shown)


thymikee (approved):

jamesgeorge007:
@thymikee any reason to keep https://github.com/react-native-community/cli/blob/master/__e2e__/legacyInit.test.ts

thymikee:
Yup. Although CLI itself is not gonna break it, because we don't touch this code. It serves as an early indicator for the community that something on the RN side broke the legacy integration. Ideally should be moved to RN core, but I don't mind having it here for now.

PR #1341, opened 2021-01-02
docs: update monorepo autolinking section with a disclaimer

BenLorantfy:
Summary:
---------

Adds a note explaining a caveat with auto-linking in a monorepo.  Ran into this problem today trying to get a monorepo working.

### DIFF
--- a/docs/autolinking.md
+++ b/docs/autolinking.md
@@ -118,7 +118,7 @@ correct location and update them accordingly:
 - path to `native_modules.gradle` in your `android/settings.gradle`
 - path to `native_modules.gradle` in your `android/app/build.gradle`
 
-Another note: Dependencies are only linked if they are listed in the package.json of the mobile workspace.  For example, with this file structure:
+Dependencies are only linked if they are listed in the package.json of the mobile workspace, where "react-native" dependency if defined. For example, with this file structure:
 ```
 /root
   /packages
--- a/docs/autolinking.md
+++ b/docs/autolinking.md
@@ -117,3 +117,18 @@ correct location and update them accordingly:
 - path to `native_modules.rb` in your `ios/Podfile`
 - path to `native_modules.gradle` in your `android/settings.gradle`
 - path to `native_modules.gradle` in your `android/app/build.gradle`
+
+Another note: Dependencies are only linked if they are listed in the package.json of the mobile workspace.  For example, with this file structure:
+```
+/root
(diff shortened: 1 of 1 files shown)


thymikee (approved):
Sounds right. @grabbou wanna add anything?

liamjones:
I believe another option if using yarn workspaces, is specifying the native dependency as 'nohoist', correct? If someone can confirm my belief, it might be worth adding that too?

thymikee:
There's no need for "nohoist" when using yarn workspaces.

docs/autolinking.md
```diff
@@ -117,3 +117,18 @@ correct location and update them accordingly:
...
+  /packages
+    /mobile
+      /ios
+      /android
+      package.json <-- Only dependencies listed here are auto-linked
+    /components
+      package.json <-- Dependencies here are ignored when auto-linking
+  package.json
+```
+In this example, if you add a package with native code as a dependecy of `components`, you need to also add it as a dependency of `mobile` for auto-linking to work.
```
jamesgeorge007, line 133:
```suggestion
In this example, if you add a package with native code as a dependency of `components`, you need to also add it as a dependency of `mobile` for auto-linking to work.
```

docs/autolinking.md
```diff
@@ -117,3 +117,17 @@ correct location and update them accordingly:
 - path to `native_modules.rb` in your `ios/Podfile`
 - path to `native_modules.gradle` in your `android/settings.gradle`
 - path to `native_modules.gradle` in your `android/app/build.gradle`
+
+Another note: Dependencies are only linked if they are listed in the package.json of the mobile workspace.  For example, with this file structure:
```
thymikee, line 121:
```suggestion
Dependencies are only linked if they are listed in the package.json of the mobile workspace, where "react-native" dependency if defined. For example, with this file structure:
```

thymikee (approved):
Sounds good, thanks for the contribution!

adrianha:
Just opened a [PR](#1647) that might help to solve the caveat defined here. Please let me know what you think @thymikee @BenLorantfy

liamjones:
@adrianha Your link is broken, looks like it was meant to point to PR #1647?

adrianha:
> @adrianha Your link is broken, looks like it was meant to point to PR #1647?

Thanks, updated

Issue #1344, opened 2021-01-10
Change setup_env.sh to setup_env.zsh?


workpebojot:
## Ask your Question
I want to change this from **sh** to **zsh**:
![sh](https://user-images.githubusercontent.com/38276345/104127239-e6136880-539b-11eb-9292-e10350aeee6c.png)

Because I keep receiving this error:
![bin](https://user-images.githubusercontent.com/38276345/104127256-0a6f4500-539c-11eb-8cb2-8c27d1a7677f.png)

My SDK is written inside default shell **(usr/bin/zsh)** so it doesn't read my sdk export from **./zshrc** file

PR #1345, opened 2021-01-10
Update getAdbPath.ts

workpebojot:
I updated getAdbPath.ts because it seems platform-tools/adb is not found because filesystem might interpret it as a directory running Windows Subsystem for Linux version 2. I updated it by adding **.exe** extension.

See: #1344

Summary:
---------

Test Plan:
----------

### DIFF
--- a/packages/platform-android/src/commands/runAndroid/getAdbPath.ts
+++ b/packages/platform-android/src/commands/runAndroid/getAdbPath.ts
@@ -8,7 +8,7 @@
 
 function getAdbPath() {
   return process.env.ANDROID_HOME
-    ? `${process.env.ANDROID_HOME}/platform-tools/adb`
+    ? `${process.env.ANDROID_HOME}/platform-tools/adb.exe`
     : 'adb';
 }
 


packages/platform-android/src/commands/runAndroid/getAdbPath.ts
```diff
@@ -8,7 +8,7 @@
 
 function getAdbPath() {
   return process.env.ANDROID_HOME
-    ? `${process.env.ANDROID_HOME}/platform-tools/adb`
+    ? `${process.env.ANDROID_HOME}/platform-tools/adb.exe`
```
thymikee, line 11: We're missing a check for `process.platform === 'win32'` here, as not everyone is on Windows :) Can you update it so this change only affects Windows users? Also cc @molant

thymikee (changes requested):

packages/platform-android/src/commands/runAndroid/getAdbPath.ts
workpebojot: I already fix mine by changing the line ending. So it finally works without .exe. And I think this is actually specific in my environment

thymikee: Let's close this then

Issue #1348, opened 2021-01-16
Set transformer.allowOptionalDependencies to true by default


brentvatne:
## Describe the Feature

https://github.com/facebook/metro/pull/511 added support for `allowOptionalDependencies` to Metro. This allows us write code like this:

```
try {
  require('does-not-exist');
} catch {
  console.log('could not load does-not-exist')
}
```

This is particularly useful for library authors, and would have been helpful during the "lean core" efforts (https://github.com/react-native-community/discussions-and-proposals/issues/120) but will continue to be useful in environments where libraries want to either: a) support multiple implementations of some primitive and automatically switch between them depending on which one the developer uses b) provide some behavior in their library that works if the developer has installed an optional peer dependency or fallback to something else and/or provide some warning otherwise.

## Possible Implementations

We would just need to add `allowOptionalDependencies: true` to the default `transformer` config in https://github.com/react-native-community/cli/blob/22a3c2558c2e03cc61f088807d2b09d1567c07ca/packages/cli/src/tools/loadMetroConfig.ts#L108-L116

This should not break any existing apps, but libraries that adopt optional imports will only work on newer versions of Metro/this CLI that use this config.

## Related Issues

- https://github.com/facebook/metro/pull/511
- https://github.com/react-native-community/discussions-and-proposals/issues/120

thymikee:
Sure, sounds reasonable and non-breaking. Happy to accept a PR :)

brentvatne:
@thymikee - will open a PR for this shortly!

one more question - do you think it would be possible to backport this and a version bump to depend on metro@0.59 ([the first release that includes this feature](https://github.com/facebook/metro/releases/tag/v0.59.0) - it shipped in March of last year) to a CLI version that is compatible with react-native@0.63?

thymikee:
Yup, we can do it!

brentvatne:
merged in #1350

PR #1350, opened 2021-01-21
feat: default to enableOptionalDependencies: true in metro config

brentvatne:
Summary:
---------

Motivation outlined in https://github.com/react-native-community/cli/issues/1348

Test Plan:
----------

I wasn't sure what to do for testing this because there isn't any precedent for testing `loadMetroConfig` or `getDefaultConfig` that I could find. I added a test for `getDefaultConfig` to ensure this option doesn't accidentally get disabled, although I understand that this is not the most useful test.

Note: the `Context` type is massive and difficult to construct, so I tried to narrow it to just what we use here by replacing `ctx: Context` with `ctx: ConfigLoadingContext` as defined here:

```ts
type ConfigLoadingContext = Pick<Config, 'root' | 'reactNativePath'> & {
  platforms: {
    [name: string]: {npmPackageName?: string};
  };
};
```

I tried to add a test for `loadMetroConfig` to ensure that if we override other `transformer` properties we still preserve the default `allowOptionalDependencies` value, but metro's config loading code via `cosmiconfig` didn't work out of the box with the the cli's existing `fs` mocking code so I shelved that idea for now.

Note:
----------

As discussed in https://github.com/react-native-community/cli/issues/1348 I'd love to see this backported to 4.x along with https://github.com/react-native-community/cli/pull/1351

### DIFF
--- a/packages/cli/src/tools/__tests__/loadMetroConfig-test.ts
+++ b/packages/cli/src/tools/__tests__/loadMetroConfig-test.ts
@@ -1,0 +1,16 @@
+import {getDefaultConfig} from '../loadMetroConfig';
+
+jest.mock('fs');
+jest.mock('path');
+
+describe('getDefaultConfig', () => {
+  it('should preserve transformer.allowOptionalDependencies=true when overriding other transformer options', async () => {
+    const config = getDefaultConfig({
+      root: '/',
+      reactNativePath: '',
+      platforms: {},
+    });
+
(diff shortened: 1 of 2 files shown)


thymikee (approved):
👍🏼

brentvatne:
azure pipeline test failure looks unrelated to this pr:

Screen Shot 2021-01-21 at 2 32 01 PM

PR #1351, opened 2021-01-21
feat: bump metro from 0.58->0.59 on 4.x

brentvatne:
Summary:
---------

Context in https://github.com/react-native-community/cli/issues/1348#issue-787562729 and https://github.com/react-native-community/cli/issues/1348#issuecomment-763234781

Note that react-native 0.63 (which uses cli@4.x) already uses the 0.59 versions of various metro packages: https://github.com/facebook/react-native/blob/27651720b40cab564a0cbd41be56a02584e0c73a/package.json#L101-L103

Test Plan:
----------

I initialized a new react-native@0.63 project, cherry-picked #1350 on to this branch to test it together with this, and followed the CONTRIBUTING instructions to run the local version of the CLI in the new project. I was able to successfully use this code in App.js:

```
try {
  require('does-not-exist');
} catch {
  console.log('failed to import');
}
```

Screen Shot 2021-01-20 at 9 19 42 PM


thymikee (approved):
In my testing this was safe to upgrade and use with 0.63, so I'm down with that, especially if Expo is also to benefit from this change

PR #1356, opened 2021-02-01
fix: rename file cause "no such file" error

wcly:
Summary:
---------

os: Mac OS

### I have found that there is an error occur in this situation：

First, I use `react-native` create a custom template name 'template'.
```bash
npx react-native init template --skip-install
```
Then I create a `template.config.js`.

In the end, my workspace like:

image

**And then, I use flowing commend to create my  new reate-native project**

```bash
npx react-native init test --skip-install --template /my/path/to/template
```

It cause an error like this:

image

This only happen when the template project name called `template`, if use  other name, it won'd happen.

Test Plan:
----------

I try the exactly the same to create project, an it run well.

### DIFF
--- a/packages/cli/src/commands/init/editTemplate.ts
+++ b/packages/cli/src/commands/init/editTemplate.ts
@@ -43,7 +43,9 @@ function renameFile(filePath: string, oldName: string, newName: string) {
 
   logger.debug(`Renaming ${filePath} -> file:${newFileName}`);
 
-  fs.renameSync(filePath, newFileName);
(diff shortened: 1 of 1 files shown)


grabbou:
Thanks for the PR! I do wonder why this only happens when the name is `template`, yet your fix is related to renaming logic. Sounds like either the problem is much wider, or the fix is applied to an incorrect code block?

grabbou:
Closing due to inactivity. Happy to revisit later.

PR #1366, opened 2021-02-10
Add missing project refs

janicduplessis:
Summary:
---------

I was having issues with the ts build after a fresh clone because it couldn't find the debugger-ui package. I noticed the cli-server-api package didn't have any project refs defined. Adding those and doing a clean build makes it work again.

Test Plan:
----------

Run clean + build


thymikee (approved):
Thanks! Not sure how we missed that, but glad it's there now :)

PR #1374, opened 2021-03-12
add editorconfig for the template

pvinis:
Summary:
---------

After adding `_editorconfig` in the template for 0.64.0, we should have the cli move it correctly.
(We need a better way for these files than manually 😬)

Test Plan:
----------

### DIFF
--- a/packages/cli/src/commands/init/editTemplate.ts
+++ b/packages/cli/src/commands/init/editTemplate.ts
@@ -71,6 +71,7 @@ const UNDERSCORED_DOTFILES = [
   'gitignore',
   'prettierrc.js',
   'watchmanconfig',
+  'editorconfig',
 ];
 
 function processDotfiles(filePath: string) {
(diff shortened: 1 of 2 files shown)


thymikee:
Hey, thanks for the PR! This only addresses legacy init (run with global `react-native init`). To make it work for the current init we'll need to edit this constant as well: https://github.com/react-native-community/cli/blob/77a06cec289f84b085dae7f27ba5b1f602898b96/packages/cli/src/commands/init/editTemplate.ts#L66-L74

We could update the logic to treat all underscored files and convert them to dotfiles, but this behavior is only really needed for `react-native/template` so that both legacy and new init can work together without breaking the legacy one.

pvinis:
Ops! Good catch. Fixing it now. Would it make sense to make an issue/PR to not have these manually listed, and instead have anything with `_*` become `.*`?

thymikee (approved):

grabbou:
> Hey, thanks for the PR! This only addresses legacy init (run with global `react-native init`). To make it work for the current init we'll need to edit this constant as well:
> 
> https://github.com/react-native-community/cli/blob/77a06cec289f84b085dae7f27ba5b1f602898b96/packages/cli/src/commands/init/editTemplate.ts#L66-L74
> 
> We could update the logic to treat all underscored files and convert them to dotfiles, but this behavior is only really needed for `react-native/template` so that both legacy and new init can work together without breaking the legacy one.

Great insights @thymikee as always! I am afraid that for the purpose of this PR and need for a timely fix, we may need to leave that for a future (or drop entirely, as it's low impact).

Anyways, something to keep in mind!

Issue #1376, opened 2021-03-14
Init with template using npx results in 404 for the template package


iepsen:
## Environment
```
System:
    OS: macOS 11.2.3
    CPU: (8) x64 Intel(R) Core(TM) i7-8569U CPU @ 2.80GHz
    Memory: 2.76 GB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 12.16.2 - ~/.nvm/versions/node/v12.16.2/bin/node
    Yarn: 1.22.5 - ~/.yarn/bin/yarn
    npm: 6.14.4 - ~/.nvm/versions/node/v12.16.2/bin/npm
    Watchman: Not Found
  Managers:
    CocoaPods: 1.10.1 - /usr/local/bin/pod
  SDKs:
    iOS SDK:
      Platforms: iOS 14.4, DriverKit 20.2, macOS 11.1, tvOS 14.3, watchOS 7.2
    Android SDK: Not Found
  IDEs:
    Android Studio: 4.1 AI-201.8743.12.41.6953283
    Xcode: 12.4/12D4e - /usr/bin/xcodebuild
  Languages:
    Java: 11.0.8 - /usr/local/opt/openjdk@11/bin/javac
  npmPackages:
    @react-native-community/cli: Not Found
    react: 17.0.1 => 17.0.1 
    react-native: 0.64.0 => 0.64.0 
    react-native-macos: Not Found
  npmGlobalPackages:
    *react-native*: Not Found
```

## Description
It fails to fetch the template when using `npx`. Templates are often explicitly saying to use the template with `react-native-template-<template>`, e.g.: https://github.com/react-native-community/react-native-template-typescript#arrow_forward-usage.

Looking at the code, it is relying on URL packages, otherwise it adds the prefix `react-native-template-` into the template:
https://github.com/react-native-community/cli/blob/5ec3fcf3a623b5a9faef050b72eab3778f241761/packages/cli/src/tools/generator/templates.ts#L56-L64

## Reproducible Demo
It is possible to reproduce it by running the following command:
```npx react-native init MyApp --template react-native-template-typescript```

iepsen:
It might not need to fix as the note here: https://github.com/react-native-community/react-native-template-typescript#note-on-the-legacy-cli, however, it is actually a breaking change. So I think if should be more explicit than just a note.

Feel free to close this issue and also the PR.

iepsen:
[...]

PR #1377, opened 2021-03-14
Allow prefixed template as option

iepsen:
Summary:
---------
Templates have been referenced prefixed with `react-native-template` on documentation, e.g.: https://github.com/react-native-community/react-native-template-typescript#arrow_forward-usage. This PR introduces a fix for this kind of usage, but also keep the existing non prefixed way.

Test Plan:
----------
It should work by running the cli providing a template with and without the `react-native-template` prefix.
1. Using prefixed template: `npx react-native init MyApp --template react-native-template-typescript`
2. Using no prefixed template: `npx react-native init MyApp --template typescript`

### DIFF
--- a/packages/cli/src/tools/generator/templates.ts
+++ b/packages/cli/src/tools/generator/templates.ts
@@ -44,6 +44,8 @@ async function createProjectFromTemplate(
 /**
  * The following formats are supported for the template:
  * - 'demo' -> Fetch the package react-native-template-demo from npm
+ * - 'react-native-template-demo' -> Fetch the package
+ * react-native-template-demo from npm
  * - git://..., http://..., file://... or any other URL supported by npm
  */
 async function createFromRemoteTemplate(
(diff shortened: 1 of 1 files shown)


grabbou:
Looks good.

Not sure if test failures are related. 

Side question - do we have a test case for this already somewhere and would it be possible to quickly add a test for that?

iepsen:
@grabbou,

Last time I saw it was failing because of `adb`, if you look at the top of the CI logs:
https://app.circleci.com/pipelines/github/react-native-community/cli/3516/workflows/4b0f4143-6e7d-4029-bba0-2918962edbd4/jobs/17528/parallel-runs/0/steps/0-110

```
/bin/sh: 1: adb: not found
 FAIL   unit  packages/cli/src/commands/doctor/healthchecks/__tests__/androidNDK.test.ts (63.67 s)
  ● androidNDK › returns a message if the Android SDK is not installed
```

Didn't see any tests for this case, but I'd reconsider this PR regarding my comments on the issue https://github.com/react-native-community/cli/issues/1376.

iepsen:
Closing this PR as the issue is with the legacy cli as it says on the documentation https://github.com/react-native-community/react-native-template-typescript#note-on-the-legacy-cli.

I think it makes sense to have this legacy note more clear and will open a PR there.

PR #1384, opened 2021-03-17
chore: update versions in readme

grabbou:
Summary:
---------

Apart from updating 5.x, I also removed `next` - it looks like it's not up to date right now and definitely not tracking the upcoming 6.x.


thymikee (approved):

thymikee:
Looks like we need to increase the timeout for environmentInfo calls in tests. Or create a mock and be done with it, as it's a reason for flakiness

PR #1386, opened 2021-03-18
Added appId and appIdSuffix options to cli-hermes

BrendonSled:
Summary:
---------

Very simple change.

Added the appId and appIdSuffix from the platform-android package. This allows anyone with an app suffix like `.debug` to profile hermes.

Test Plan:
----------

This code is copy paste from the platform-android package.


packages/cli-hermes/src/profileHermes/index.ts
```diff
@@ -1,3 +1,4 @@
+// @ts-ignore
```
thymikee, line 1: please revert this

packages/cli-hermes/src/profileHermes/sourcemapUtils.ts
```diff
@@ -1,11 +1,10 @@
 import {Config} from '@react-native-community/cli-types';
-import {logger, CLIError} from '@react-native-community/cli-tools';
+import {logger, CLIError,fetch} from '@react-native-community/cli-tools';
```
thymikee, line 2: need to run Prettier on that (`yarn lint --fix`)

packages/cli-hermes/src/profileHermes/downloadProfile.ts
```diff
@@ -46,13 +46,18 @@ export async function downloadProfile(
   raw?: boolean,
   shouldGenerateSourcemap?: boolean,
   port?: string,
+  appId?: string,
+  appIdSuffix?: string,
```
thymikee, line 50: can you update the jsdoc comment for this function to be up-to-date?

thymikee (changes requested):
Thanks! Why not :)

thymikee (approved):

grabbou:
Thanks, could you provide a use case for this addition? 

At the same time, could you provide a description to our documentation? Simply open `md` file and add these options together with the description.

AndrewMorsillo:
@grabbou An example of when this is necessary: my app uses react-native-firebase, in development mode it connects to a separate staging firebase project. To do that it needs to have a different android package name than the production app, so I use `--appIdSuffix=staging`. Without this PR I don't think I can use `profile-hermes` because it won't have the correct package name in development mode. I think this solves https://github.com/react-native-community/cli/issues/1438

BrendonSled:
@grabbou this is a simple port of the same feature from the cli.
https://github.com/react-native-community/cli/blob/master/docs/commands.md#--appid-string

packages/cli-hermes/src/profileHermes/downloadProfile.ts
```diff
@@ -69,7 +76,7 @@ export async function downloadProfile(
     // If --raw, pull the hermes profile to dstPath
     if (raw) {
       execSyncWithLog(
-        `adb shell run-as ${packageName} cat cache/${file} > ${dstPath}/${file}`,
+        `adb shell run-as ${packageNameWithSuffix} cat cache/${file} > ${dstPath}/${file}`,
```
grabbou, line 79: Do we need to run `as` with suffix? We're not doing this in other `adb` calls in this file.

grabbou (changes requested):
Generally looks good, but I do have a question (left a comment).

packages/cli-hermes/src/profileHermes/downloadProfile.ts
BrendonSled: That's a good question. I attempted to modify as little as possible. The prior command was `run-as` so I just used that

BrendonSled:
> Thanks, could you provide a use case for this addition?
> 
> At the same time, could you provide a description to our documentation? Simply open `md` file and add these options together with the description.

For this I just copied this code from the cli 
https://github.com/react-native-community/cli/blob/641b21f583c97e3d48ce87d5fe804f42db92fa5c/packages/platform-android/src/commands/runAndroid/tryLaunchAppOnDevice.ts#L20

This allows app suffix support. Example: com.test**.development**, com.test**.staging**,

BrendonSled:
Also @grabbou I would love to update the documentation for hermes. But I can't find the documentation for that specific cli.

The appidsuffix is already in the main cli commands, here:
https://github.com/react-native-community/cli/blob/master/docs/commands.md#--appidsuffix-string

Can you point me in the right direction?

henrymoulton:
@BrendonSled have just come across this issue/PR as I'm needing to do a bit of Hermes profiling in an app that has Debug/Staging/Production flavors. 

@elizabeth-dev are you actively still needing this feature?

@thymikee  / @grabbou anything needed to get this merged in?

packages/cli-hermes/src/profileHermes/downloadProfile.ts
grabbou: What I am saying is that on this line: https://github.com/react-native-community/cli/pull/1386/files#diff-13ca0a162291646faf5ecefc657d151b45ebf950563c9f32d38ec39ede1341b1R90 we are not using `packageNameWithSuffix`.

I guess as per your comment, that has to be modified.

grabbou:
The PR is ready to be merged, I have applied some tweaks. @henrymoulton would you mind testing that everything works as expected for you?

grabbou:
Meanwhile, I have decided to ship it. Since we're not going to make a release anytime soon and this PR is additive, I will continue working on the CLI and wait for reports in the meantime.

joaobertacchi:
@grabbou, @BrendonSled  , I'm trying to use this new feature without success. Am I doing something wrong?
```
$ npx react-native@nightly profile-hermes --appId com.myapp ~/tmp
error: unknown option `--appId'
```

PR #1387, opened 2021-03-24
fix: check stdout for xcrun output

dotconnor:
Summary:
---------

At some point the xcrun command for printing all available devices, `xcrun xctrace list devices` changed from outputting from `stderr` to `stdout`.

```bash
❯ xcrun --version
xcrun version 59.
❯ xcodebuild -version
Xcode 12.5
Build version 12E5244e
```

Test Plan:
----------

Run `react-native ios --device` with a newish (beta 3) version of Xcode and a physical device connected.

### DIFF
--- a/packages/platform-ios/src/commands/runIOS/index.ts
+++ b/packages/platform-ios/src/commands/runIOS/index.ts
@@ -87,9 +87,8 @@ function runIOS(_: Array<string>, ctx: Config, args: FlagsT) {
 
   let devices;
   try {
-    devices = parseXctraceIOSDevicesList(
-      execa.sync('xcrun', ['xctrace', 'list', 'devices']).stderr,
-    );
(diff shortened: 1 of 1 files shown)


bjackson:
I encountered this same issue, and the change here fixed my issue.

maxoumime:
I'm facing the same issue. It's only affecting XCode Beta 3 users but it could affect more when this version moves to stable, probably in the coming weeks.

maxoumime:
@dotconnor [This is fixed](https://github.com/react-native-community/cli/blob/master/packages/platform-ios/src/commands/runIOS/index.ts#L93)

grabbou:
Fixed via https://github.com/react-native-community/cli/commit/e5f4ef3442932992d748537eaad1de750f6a8232

PR #1388, opened 2021-03-24
feat: add option to build only active architectures on android

janicduplessis:
Summary:
---------

Building from source in debug takes a very long time because native builds need to run for all supported architectures. It is possible to check which architecture the devices for which we are about to launch the app on are and build only for those. For most cases we can reduce the number of architectures we build for to 1 instead of 4, resulting in a large speedup of the build.

This is inspired by iOS which has a "Build for active architecture only" option. Since android doesn't really support this natively we can implement it here and also in react-native by reading the build properties that we pass and alter the abi we build for.

With fabric / codegen coming up I suspect that we might want to default to building c++ soon. This should ease the transition as builds won't be orders of magnitude slower.

For now this is enabled only behind a flag, I think this is better since for the common case of not building from source this change doesn't have any effect. If RN moves to building c++ by default on android we can consider changing the default.

RN PR to add support for reactNativeDebugArchitectures: https://github.com/facebook/react-native/pull/31232

Test Plan:
----------

Tested this change + an incoming PR to the react-native repo with a combinaison of simulators and physical devices. Checked the build logs to see which architectures are being built.

Clean build for x86 simulator
`react-native run-android`
824.41s
`react-native run-android --active-arch-only`
299.77s

### DIFF
--- a/packages/platform-android/src/commands/runAndroid/adb.ts
+++ b/packages/platform-android/src/commands/runAndroid/adb.ts
@@ -67,7 +67,29 @@ function getAvailableCPUs(adbPath: string, device: string): Array<string> {
   }
 }
 
+/**
+ * Gets the CPU architecture of a device from ADB
+ */
+function getCPU(adbPath: string, device: string): string | null {
+  try {
+    const cpus = execFileSync(adbPath, [
+      '-s',
+      device,
+      'shell',
+      'getprop',
+      'ro.product.cpu.abi',
+    ])
+      .toString()
+      .trim();
+
+    return cpus.length > 0 ? cpus : null;
+  } catch (e) {
+    return null;
+  }
+}
+
 export default {
   getDevices,
   getAvailableCPUs,
+  getCPU,
 };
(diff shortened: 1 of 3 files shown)


liamjones:
> This is inspired by iOS which has a "Build for active architecture only" option

That's just raised a question for me - does `react-native run-ios` already do this?

janicduplessis:
@liamjones This is configured in xcode and I think is enabled by default. You can verify in your project settings here.

image

thymikee (approved):
LGTM. @grabbou mind having a look and releasing a minor version when merged?

grabbou:
Yup, this is the default on iOS. Thanks @janicduplessis for the PR and for the explanation - my first question was why we don't turn it by default, and I guess we can do it in a follow-up release or when it becomes more widely spread.

grabbou (approved):

janicduplessis:
Can probably turn it on by default on by default. I've been using this for a while without issues. I try to get the RN part merged so this can work, maybe target next RN release 0.66.

PR #1392, opened 2021-04-04
feat: use xcbeautify for xcodebuild output if installed

liamjones:
Summary:
---------

The run-ios command already detects and uses xcpretty (https://github.com/xcpretty/xcpretty) if it's available.

This commit adds support for another xcodebuild output formatter, xcbeautify (https://github.com/thii/xcbeautify).

The logic for what is used is as follows:
1. If xcbeautify is available use it
2. If xcbeautify isn't available but xcpretty is use xcpretty
3. If neither are available fallback to raw xcodebuild output

Test Plan:
----------

Tested manually with the following scenarios;
* only xcbeautify installed (xcbeautify used)
* only xcpretty installed (xcpretty used)
* both installed (xcbeautify used)
* neither installed (xcodebuild used raw)

### DIFF
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -443,7 +443,7 @@ Do not launch packager while building.
 
 #### `--verbose`
 
-Do not use `xcpretty` even if installed.
+Do not use `xcbeautify` or `xcpretty` even if installed.
 
 #### `--port <number>`
 
--- a/packages/platform-ios/src/commands/runIOS/index.ts
+++ b/packages/platform-ios/src/commands/runIOS/index.ts
@@ -313,17 +313,14 @@ function buildProject(
     );
     let xcodebuildOutputFormatter: ChildProcess | any;
     if (!args.verbose) {
-      xcodebuildOutputFormatter =
-        xcbeautifyAvailable() &&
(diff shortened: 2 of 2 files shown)


liamjones:
I looked at adding unit tests for the new code but my initial attempt was ending up pretty messy due to having to run the entirety of `runIOS` and there being lots of file system access that needed dealing with (mocked or otherwise). I assumed (maybe incorrectly?) that the project maintainers wouldn't want `buildProject` exported from its module purely for testing purposes.

thymikee (approved):
Mind sending a screenshot of how it looks with xcbeautify? :)

liamjones:
@thymikee here's an example of xcbeautify with a successful build:

Screenshot 2021-04-07 at 15 14 21

And xcbeautify with a failed build:

Screenshot 2021-04-07 at 15 17 38

(Note: PR https://github.com/react-native-community/cli/pull/1393 is also applied in the above screenshot so there's no raw xcodebuild output after the failure output from xcbeautify itself)

grabbou (approved):
LGTM. I have cleaned up the `if/else` logic for better readability.

Issue #1391, opened 2021-04-04
run-ios command ignores the verbose option


liamjones:
## Environment

```
info Fetching system and libraries information...
System:
    OS: macOS 10.15.7
    CPU: (8) x64 Intel(R) Core(TM) i7-7820HQ CPU @ 2.90GHz
    Memory: 1.47 GB / 16.00 GB
    Shell: 5.8 - /usr/local/bin/zsh
  Binaries:
    Node: 10.24.0 - /var/folders/h6/g6rldscd6h389g9w42_7k_jm0swv2d/T/yarn--1617537855315-0.38815081056993295/node
    Yarn: 1.22.10 - /var/folders/h6/g6rldscd6h389g9w42_7k_jm0swv2d/T/yarn--1617537855315-0.38815081056993295/yarn
    npm: 6.14.11 - ~/.nvm/versions/node/v10.24.0/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  Managers:
    CocoaPods: 1.10.0 - /Users/liam.jones/.rbenv/shims/pod
  SDKs:
    iOS SDK:
      Platforms: iOS 14.4, DriverKit 20.2, macOS 11.1, tvOS 14.3, watchOS 7.2
    Android SDK:
      API Levels: 23, 25, 26, 27, 28, 29
      Build Tools: 27.0.3, 28.0.3, 29.0.2, 29.0.3, 30.0.3, 31.0.0
      System Images: android-21 | Google APIs Intel x86 Atom, android-25 | Google APIs Intel x86 Atom, android-25 | Google Play Intel x86 Atom, android-26 | Google Play Intel x86 Atom, android-27 | Google Play Intel x86 Atom, android-29 | Google APIs Intel x86 Atom, android-29 | Google Play Intel x86 Atom, android-30 | Google Play Intel x86 Atom
      Android NDK: Not Found
  IDEs:
    Android Studio: 4.1.3 4.1.3
    Xcode: 12.4/12D4e - /usr/bin/xcodebuild
  Languages:
    Java: 1.8.0_282 - /usr/bin/javac
    Python: 3.9.2 - /usr/local/opt/python/libexec/bin/python
  npmPackages:
    @react-native-community/cli: Not Found
    react: 16.13.1 => 16.13.1
    react-native: 0.63.4 => 0.63.4
    react-native-macos: Not Found
  npmGlobalPackages:
    *react-native*: Not Found
```

## Description

[...]

PR #1393, opened 2021-04-04
feat: less noisy run-ios build failures with xcpretty

liamjones:
Summary:
---------

The run-ios command detects and uses xcpretty (https://github.com/xcpretty/xcpretty) if it's available. If it's not available it displays a loader/spinner animation until xcodebuild finishes.

If a build failure happens then the buildOutput and errorOutput are included in the raised CLIError and printed out to the terminal.

This makes sense when xcpretty isn't present as there has been no build output displayed up to now. However, if xcpretty is being used, this behaviour makes less sense:

Firstly, xcpretty has been doing coloured output, nicely highlighting warnings and errors as they happened. The raised CLIError will repeat the output but without colours which means you have to scroll back up past a potential mountain of monochrome xcodebuild output to find the xcpretty output (or attempt to pick out the error from the monochrome output).

Secondly, you will be prompted to "Run CLI with --verbose flag for more details." upon build failure anyway. If you use xcpretty then running the build again with the --verbose flag would force raw xcodebuild output (which would then including the buildOutput and errorOutput again).

Test Plan:
----------

Tested manually with and without xcpretty installed.

### DIFF
--- a/packages/platform-ios/src/commands/runIOS/index.ts
+++ b/packages/platform-ios/src/commands/runIOS/index.ts
@@ -360,7 +360,7 @@ function buildProject(
             logs further, consider building your app with Xcode.app, by opening
             ${xcodeProject.name}.
           `,
-            xcpretty ? undefined : buildOutput + '\n' + errorOutput,
+            xcodebuildOutputFormatter ? undefined : buildOutput + '\n' + errorOutput,
           ),
         );
         return;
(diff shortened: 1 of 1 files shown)


liamjones:
One extra point concerning this:

>Secondly, you will be prompted to "Run CLI with --verbose flag for more details." upon build failure anyway. If you use xcpretty then running the build again with the --verbose flag would force raw xcodebuild output (which would then including the buildOutput and errorOutput again).

The `--verbose` flag doesn't seem to function currently - see https://github.com/react-native-community/cli/issues/1391. If this PR is approved, you may want to hold off merging this until that bug is resolved.

thymikee:
Please send some screenshots/videos of the change :)

liamjones:
@thymikee here you go:

Before changes: https://youtu.be/ZhpZurWsAso

After changes: https://youtu.be/sjNf7eGY8Gs

grabbou:
Thank you for sending this over!

PR #1394, opened 2021-04-05
fix: remove unused --verbose option from start command

liamjones:
Summary:
---------

While investigating https://github.com/react-native-community/cli/issues/1391 I discovered that the start command had a `--verbose` option that didn't do anything. You _do_ get extra logging with `--verbose` but this comes from the parent react-native command, where it is already documented.

I had a look through the metro functions being called to see if `verbose` was meant to be passed to one of them and had been missed but couldn't find anywhere it would be relevant. The only one I could see that it would potentially apply to is the `loadMetroConfig` call. While `loadMetroConfig` doesn't document a verbose parameter of its own, it calls `loadConfig` in the `metro-config` package which *does* take a verbose parameter. However, all that does is kill the reporter if `verbose === false` (see https://github.com/facebook/metro/blob/master/packages/metro-config/src/loadConfig.js#L256), which would mean you get no output from the start command whatsoever. Currently, the loadMetroConfig code is reliant on the fact that verbose is `undefined` when calling `loadConfig` so its custom reporter isn't overwritten.

Test Plan:
----------

* Ran `react-native start --verbose` before making this change and observed extra logging
* Removed verbose option from the start command
* Ran `react-native start --verbose` again and still observed extra logging

### DIFF
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -499,10 +499,6 @@ Removes cached files
 
 Path to a JavaScript file that exports a log reporter as a replacement for TerminalReporter
 
-#### `--verbose`
-
-Enables logging
-
 #### `--https`
 
 Enables https connections to the server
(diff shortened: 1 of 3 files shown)


thymikee (approved):
Thanks for the cleanup!

grabbou:
Nice catch, thanks!

PR #1405, opened 2021-04-30
Fix for 'No iOS devices connected.'

thang2162:
Changed stderr to stdout to fix 'No iOS devices connected.' when attempting run-ios with '--device' on XCode 12.5.

Summary:
---------

Test Plan:
----------


thymikee:
Thanks! Would be great to support lower versions of Xcode for a while. Would you be so kind and first try with stdout and then fall back with stderr? `xcrun` returns empty string in error stream so we could do something like `stderr || stdout`

PR #1420, opened 2021-05-21
Bump to Metro 0.66

NickGerleman:
Summary:
---------

~1. Run `lerna version 6.0.0` to bump all package versions. This number is chosen instead of prerelease to avoid needing a new bump once no longer prerelease.~ (reverted)
2. Bump Metro to 0.66
3. A best effort scan for other dependencies where we should ideally align.

## Required after this PR
1. Tag this commit with v6.0.0 (do not locally have permissions)
2. lerna publish
3. Use this version in RN 0.65-stable + master

Test Plan:
----------

CLI version change should not have runtime affect. Developers in the RNW repo have been using metro 0.66 via resolution.


grabbou:
Looks good. We would normally do the version bump separately from Metro release, so that we can have all the tags created on Github project as well. Would it be possible to revert that particular commit and keep it only to Metro release changes?

grabbou (approved):

NickGerleman:
> Looks good. We would normally do the version bump separately from Metro release, so that we can have all the tags created on Github project as well. Would it be possible to revert that particular commit and keep it only to Metro release changes?

Sure, I can separate this to just the Metro dependency bump. I just noticed that there's also a branch per major version.

We should maybe reverse the order though, where we bump/branch before the metro change. The metro change is breaking, so we don't want in in 5.0.

NickGerleman:
@grabbou this PR is now isolated to the Metro bump.

I do not have RNC permissions, so I will probably need some assistance with version cutting/tagging, merging the change.

thymikee (approved):

thymikee:
Thank you Nick!

Issue #1418, opened 2021-05-20
Why doesn't this CLI let me create a new project in an existing, empty directory?


Pika-Pool:
## Ask your Question

When i run this command: `npx react-native init --directory . --skip-install awesomeproject`, I get the error `error Error occurred while trying to create project directory.` Note that the '.' directory is completely empty

I checked the code and it seems the cli only checks if the directory already exists, and if it does,  error is thrown [here](https://github.com/react-native-community/cli/blob/121916fca2d85d808862d10b75b787a82f683d62/packages/cli/src/commands/init/init.ts#L44)

But in create-react-app, you can just specify the directory, the command works, as long as package.json does not exist, even if the directory does.

Is there are particular reason react-native does this? If not, I think the CRA way is better and I would love to implement it in this CLI

Pika-Pool:
ok, I'll work on this then. It'll be great if you could clarify a few things for me

- whats the diff between `/package/cli/src/commands/init.ts` and `/package/cli/src/commands/initCompat.ts`?
- in what order are these files executed?
- and which file do i make the changes for this issue?

This is the workflow in my mind. 
- There are 3 scenarios for the command line arguments:
	 - `projectName` is given, `--directory` is not:
		- treat `projectName` as a directory path, so something like this:
		```js
		const projectDirectory = path.resolve(projectName);
		projectName = path.basename(root);
		```

	- `projectName` and `--directory` both are given:
		```js
		const projectDirectory = path.resolve(directory);
		// projectName remains same as the command line argument passed
		// projectName is also already validated against a regex in the existing code
		```
	- `projectName` is not given:
		- this already throws an error, so no changes

- Now,  to check if the `projectDirectory` is valid, CRA uses [function `isSafeToCreateProjectIn`](https://github.com/facebook/create-react-app/blob/7bdeced8495370b683c08b86bc84b66af83f2f0a/packages/create-react-app/createReactApp.js#L934), which uses a list of valid files that could be already present in the directory.
	- i couldn't find any other conflicting files for the react-native cli, please mention them if there are any

thymikee:
Create a PR please :)

Pika-Pool:
I'm stuck at writing tests for this. To validate a directory, the code checks the local file system. So should I use something like [mock-fs](https://github.com/tschaub/mock-fs) for this?

PR #1421, opened 2021-05-21
Init project in existing directory

Pika-Pool:
Summary:
---------

This PR makes changes to allow initializing a new react-native project in an already existing directory, only if there aren't any conflicting files/sub-directories

This behavior is directly influenced by create-react-app

Test Plan:
----------
tested on local machine(windows). All these commands work properly
- `npx react-native init --directory dummydir app`
    - directory: `.\dummydir`
    - projectName: `app`
- `npx-react-native init --directory dummydir\dir1\dir2 app`
    - directory: `.\dummydir\dir1\dir2`
    - projectName: `app`
- `npx react-native init dummydir\dir1\dir2`
    - directory: `.\dummydir\dir1\dir2`
    - projectName: `dir2`
- `npx react-native init app`
    - directory: `.\app`
    - projectName: `app`

Refs: #1418

### DIFF
--- a/packages/cli/src/commands/init/init.ts
+++ b/packages/cli/src/commands/init/init.ts
@@ -60,13 +60,19 @@ function validateProjectDirectory(directory: string) {
     'Thumbs.db',
   ];
 
-  const conflicts = fs.readdirSync(directory).filter((file) => {
-    return (
-      !validFiles.includes(file) &&
-      // IntelliJ IDEA creates module files before CRA is launched
-      !/\.iml$/.test(file)
-    );
-  });
+  const conflicts = fs
+    .readdirSync(directory)
+    .filter((file) => {
+      return (
+        !validFiles.includes(file) &&
+        // IntelliJ IDEA creates module files before CRA is launched
+        !/\.iml$/.test(file)
+      );
+    })
+    .map((file) => {
+      const stats = fs.lstatSync(path.join(directory, file));
+      return `${file}${stats.isDirectory() ? '/' : ''}`;
+    });
 
   if (conflicts.length > 0) {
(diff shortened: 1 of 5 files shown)


packages/cli/src/commands/init/init.ts
```diff
@@ -199,15 +193,28 @@ export default (async function initialize(
...
-  validateProjectName(projectName);
-
   /**
    * Commander is stripping `version` from options automatically.
    * We have to use `minimist` to take that directly from `process.argv`
    */
   const version: string = minimist(process.argv).version || DEFAULT_VERSION;
 
-  const directoryName = path.relative(root, options.directory || projectName);
+  // handles the case when '.' is passed as for projectDirectory
```
grabbou, line 202: AFAIK we don't need this special case.

1/ **We should keep `validateProjectName` where it was - on line 202**. All in all, the use case is always `react-native init MyProject --directory .` for those that want to initialise in the current directory. We should not allow `react-native init` to be run without a name (not sure if that's how it works in your PR) and then, overwrite the `projectNme`

2/ Do we actually need anything else? Does `directoryName` need to be `.`, not ``, hence special case treatment on line 207? I would imagine everything down the code works perfectly fine with an empty string, mind taking another pass at that?

grabbou (changes requested):
Thank you for sending this PR over. I really like the idea behind it. Beyond two comments/requests, it's looking great. Appreciate the test case for the added feature!

grabbou:
I am going to close this PR due to inactivity. Please feel free to ping me and we will re-open when this is ready to be merged.

I still think this feature is valuable, unfortunately, we, as a core team, don't have bandwidth to work on it right now.

PR #1430, opened 2021-06-02
feat: Added custom scope method for android dependencies

TMaszko:
### Summary:
This PR continues idea from this previous [PR](https://github.com/react-native-community/cli/pull/1177)
It allows user to define `customDependencyConfiguration` per project(`project.android`) and also per module itself inside `dependencies[packageName].android` key
```js
module.exports = {
  project: {
    android: {
      customDependencyConfiguration:
        'embed project(path: ":$dependencyName", configuration: "default")',
    },
  },
  dependencies: {
    'react-native-brownfield': {
      platforms: {
        android: {
          customDependencyConfiguration:
            'embed project(path: ":react-native-brownfield-bridge", configuration: "default")',
        },
      },
    },
  },
};
```
Test Plan:
Init new project and checked whether applied config is correct.

### DIFF
--- a/docs/projects.md
+++ b/docs/projects.md
@@ -41,7 +41,7 @@ type ProjectConfigT = {
       root: string,
       platforms: {
         [key: string]: PlatformSettingsT
-      };
+      },
       assets: string[],
       hooks: {
         [key: string]: string
--- a/docs/projects.md
+++ b/docs/projects.md
@@ -77,7 +77,7 @@ type AndroidProjectParams = {
   assetsPath?: string;
   buildGradlePath?: string;
   appName?: string; // A name of the app in the Android `sourceDir`, equivalent to Gradle project name. By default it's `app`.
-  dependencyConfiguration?: string;
+  customDependencyConfiguration?: string;
 };
 
 type IOSProjectParams = {
@@ -130,7 +130,7 @@ module.exports = {
     'react-native-brownfield': {
       platforms: {
         android: {
-          dependencyConfiguration:
+          customDependencyConfiguration:
             'embed project(path: ":react-native-brownfield-bridge", configuration: "default")',
         },
       },
--- a/docs/projects.md
+++ b/docs/projects.md
@@ -27,28 +27,28 @@ You can check all available options below.
 
 ```ts
 type ProjectConfigT = {
-  reactNativePath: ?string;
+  reactNativePath: ?string,
   project: {
-    android?: ProjectParamsAndroidT;
-    ios?: IOSProjectParams;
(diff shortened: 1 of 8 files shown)


packages/platform-android/native_modules.gradle
```diff
@@ -265,6 +275,15 @@ class ReactNativeModules {
         if (!androidConfig["buildTypes"].isEmpty()) {
           reactNativeModulesBuildVariants.put(nameCleansed, androidConfig["buildTypes"])
         }
+        if(androidConfig.containsKey("customDependencyConfiguration")) {
+          reactNativeModuleConfig.put("customDependencyConfiguration", androidConfig["customDependencyConfiguration"])
+        } else if (project.containsKey("customDependencyConfiguration")) {
```
michalchudziak, line 280: Shouldn't it have priority over `androidConfig`?

TMaszko: We've agreed that config per module should have higher priority :)

michalchudziak: Yeah, I thought `project` relates to the module. LGTM

michalchudziak (approved):

docs/projects.md
```diff
@@ -27,28 +27,28 @@ You can check all available options below.
...
-        [key: string]: string
-      }
-    },
-  },
-  commands: CommandT[]
+        [key: string]: string;
+      };
+    };
+  };
+  commands: CommandT[];
```
grabbou, line 51: Why did this happen? Different ESLint / TS versions? Which one should be correct?

CC: @thymikee

docs/projects.md
```diff
@@ -121,6 +122,25 @@ module.exports = {
...
 
+or you could set:
+
+```js
+module.exports = {
+  dependencies: {
+    'react-native-brownfield': {
+      platforms: {
+        android: {
+          customDependencyConfiguration:
```
grabbou, line 133: I think we can drop `custom` in here, defining it makes it custom already.

grabbou (approved):
It looks good, thanks for the PR! I would recommend dropping `custom` from the name, rather than that, we're good.

Would you be able to tackle that @TMaszko?

docs/projects.md
thymikee: We don't actively run prettier on MD files, so this happens when editor is configured to do so. This change is correct

grabbou (approved):
LGTM. We will need some basic tests to check that it works, especially on the project/dependency level, but I guess we can handle that in a follow-up PR to not keep this one open for too long.

PR #1441, opened 2021-07-04
Print unsupported keypresses directly to console.

BasitAli:
Summary:
---------

Whenever you are reviewing logs for any app, adding a few blank lines in the console is an often used practice that stopped working with react-native in the recent releases. This was requested at https://github.com/facebook/metro/issues/663 a few months back and I recently faced this as I moved to Windows from Mac and clearing the console isn't a supported action yet (https://github.com/microsoft/terminal/issues/1882).

Test Plan:
----------

- Verified that existing handled actions (r, d, ctrl+c and ctrl+d) still work.
- All other keypresses are logged as it is to the screen.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -64,7 +64,15 @@ commands:
     steps:
       - attach_workspace:
           at: ~/react-native-cli
+      - restore_cache:
+          keys:
+            - v2-gradle-dependencies-{{ checksum "package.json" }}
+            - v2-gradle-dependencies-
       - run: yarn test:ci:e2e
+      - save_cache:
+          key: v2-gradle-dependencies-{{ checksum "package.json" }}
+          paths:
+            - ~/.gradle
 
 jobs:
   setup:
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -4,17 +4,38 @@ trigger:
 pool:
   vmImage: 'windows-2019'
 
+variables:
+  YARN_CACHE_FOLDER: $(Pipeline.Workspace)/.yarn
+  GRADLE_USER_HOME: $(Pipeline.Workspace)/.gradle
+
 steps:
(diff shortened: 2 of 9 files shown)


grabbou:
This breaks the end to end tests that depend on the output. While logging blank lines seems like a good idea, it looks like we have some unwanted Gradle output out there too. Would you mind checking that out?

BasitAli:
> This breaks the end to end tests that depend on the output. While logging blank lines seems like a good idea, it looks like we have some unwanted Gradle output out there too. Would you mind checking that out?

Sure, I'm on it.

BasitAli:
Looks like circleci has been acting up as all latest commits in master have similar failures. The tests pass locally btw.

image

BasitAli:
@grabbou Fixed a unit test that was failing on Azure/Windows and added gradle caching in the circleci pipeline. But similar failures were noticed in Azure as well and this seems like a time/network related fail where if a test takes too long, it fails. We may need to add a gradle and node_modules cache for Azure as well.

grabbou:
Thanks! Appreciate the extra work on the Windows CI. Unfortunately, it's been broken ever since I remember, I guess we might need to do something about it in the nearest future!

BasitAli:
> Thanks! Appreciate the extra work on the Windows CI. Unfortunately, it's been broken ever since I remember, I guess we might need to do something about it in the nearest future!

It seems to have worked 🥳 

image

pratteekshaurya1994:
Adding empty spaces helped in looking at consoles in a better way. can we add this feature back? or is there a workaround which we can do to have this feature?

cortinico:
> Adding empty spaces helped in looking at consoles in a better way. can we add this feature back? or is there a workaround which we can do to have this feature?

You should open an issue for this against https://github.com/facebook/metro/issues

PR #1444, opened 2021-07-12
fix(platform-ios): fix `sourceDir` detection

tido64:
Summary:
---------

This is a minimal changes version of #1436. After speaking with @grabbou, I learned that most of the props in `config` are going away, so we agreed to make as little changes as possible and instead focus on `next`.

Resolves #1435
Resolves https://github.com/microsoft/react-native-test-app/issues/375

Test Plan:
----------

Curren tests should pass:
```
 PASS   unit  packages/platform-android/src/link/patches/__tests__/makeSettingsPatch.test.ts
 PASS   unit  packages/platform-ios/src/commands/runIOS/__tests__/findMatchingSimulator.test.ts
 PASS   unit  packages/platform-android/src/config/__tests__/getProjectConfig.test.ts
 PASS   unit  packages/cli/src/commands/init/__tests__/template.test.ts
 PASS   unit  packages/cli/src/tools/__tests__/packageManager-test.ts
 PASS   unit  packages/platform-ios/src/link/__tests__/getHeaderSearchPath.test.ts
 PASS   unit  packages/cli/src/commands/link/__tests__/link.test.ts
 PASS   unit  packages/platform-android/src/commands/runAndroid/__tests__/runOnAllDevices.test.ts
 PASS   unit  packages/cli/src/tools/windows/__tests__/androidWinHelpers.test.ts
 PASS   unit  packages/cli/src/commands/upgrade/__tests__/upgrade.test.ts (5.274 s)
 PASS   unit  packages/cli/src/commands/bundle/__tests__/getAssetDestPathAndroid-test.ts
 PASS   unit  packages/platform-ios/src/config/__tests__/findProject.test.ts
 PASS   unit  packages/cli/src/commands/init/__tests__/editTemplate.test.ts (6.21 s)
 PASS   unit  packages/platform-android/src/config/__tests__/findPackageClassName.test.ts
 PASS   unit  packages/platform-android/src/link/patches/__tests__/makeBuildPatch.test.ts
 PASS   unit  packages/platform-ios/src/commands/runIOS/__tests__/parseXctraceIOSDevicesList.test.ts
 PASS   unit  packages/platform-ios/src/commands/runIOS/__tests__/parseIOSDevicesList.test.ts
 PASS   unit  packages/platform-ios/src/link-pods/__tests__/removePodEntry.test.ts
 PASS   unit  packages/platform-ios/src/link-pods/__tests__/isInstalled.test.ts
[...]

PR #1445, opened 2021-07-13
feat: look for Podfile in the source directory too

Kiesco08:
Summary:
---------

I'm adding React Native to an existing native project. The project is fairly large, so it is split up into multiple subprojects and does not follow the typical React Native directory structure. Instead this is how it is setup:

```
ios/
| | | |_ Pods/
| | |__ **Podfile**
| |___ Project/Project.xcodeproj/
|____ ProjectReactNativeFramework/

... // Other subprojects
```

Currently, `cli` assumes that `Podfile` is a sibling of `Project.xcodeproj`. This change adds a feature enabling it to look for the `Podfile` from the source directory, if it is not found in the project directory.

Test Plan:
----------

- When I run `yarn react-native config`, the `Podfile` configuration value is no longer `null`.
- Running `yarn react-native run-ios` no longer throws the following error:

```
error Could not find "Podfile.lock" at null.lock. Did you run "pod install" in iOS directory?
```

### DIFF
--- a/packages/platform-ios/src/config/index.ts
+++ b/packages/platform-ios/src/config/index.ts
@@ -54,7 +54,7 @@ export function projectConfig(folder: string, userConfig: IOSProjectParams) {
     sourceDir,
     folder,
     pbxprojPath: path.join(projectPath, 'project.pbxproj'),
-    podfile: findPodfilePath(projectPath),
+    podfile: findPodfilePath(projectPath) || findPodfilePath(sourceDir),
     podspecPath:
       userConfig.podspecPath ||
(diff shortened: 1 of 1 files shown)


Kiesco08:
This test passes locally 🤔

grabbou:
Related to #1444 too. I will work on both of these PRs tomorrow and merge one of them (not sure which one tho) in order to resolve this issue.

grabbou:
Closing in favour of #1459, I would prefer to allow users to set the path manually, rather than extending the automatic detection (it's there to satisfy 99% of standard use cases)

PR #1447, opened 2021-07-15
Move all metro logic and commands into a metro package

afoxman:
Summary:
---------

I am working on a project where I need to do Metro bundling using the Metro APIs. The CLI has really useful Metro config and Metro bundling logic that I want to use, but it's tied up and not accessible.

This PR move the Metro configuration and bundling logic into separate packages, making them available for my project and to other external users.

Test Plan:
----------
Code is moving, with minimal changes, including tests. Verify that the build passes and that all tests pass.

Manually verify that bundling works on a sample app. Diffed bundles using existing release of CLI vs this PR. Bundle size and md5 hashes were identical.


grabbou:
I love this! It's something I wanted to do for a really really long time!

Can you do one more thing, rename `metro-config` to just `metro` and move `start` as well as `bundle` commands into that package? This will be a good first step to eventually provide an abstraction for other bundlers to plug into CLI instead of Metro.

CC: @tido64

packages/cli/src/tools/config/resolveNodeModuleDir.ts
```diff
@@ -1,15 +1,2 @@
...
-  root: string,
-  packageName: string,
-): string {
-  return path.dirname(
-    require.resolve(path.join(packageName, 'package.json'), {
-      paths: [root],
-    }),
-  );
-}
+import {resolveNodeModuleDir} from '@react-native-community/cli-tools';
```
afoxman, line 1: I kept this module in place to keep the mocking model intact. I initially tried removing it and reworking the mocks, but the global mock of `cli-tools` made that impossible.

afoxman:
> I love this! It's something I wanted to do for a really really long time!
> 
> Can you do one more thing, rename `metro-config` to just `metro` and move `start` as well as `bundle` commands into that package? This will be a good first step to eventually provide an abstraction for other bundlers to plug into CLI instead of Metro.
> 
> The idea is that one should be able too completely replace Metro with e.g. Webpack or other bundler by removing `@react-native-community/bundler-metro` package and using a different one instead.
> 
> CC: @tido64

Happy to help! This turned into a big PR, so you have a lot of reading to do :). The `metro` package now has all metro commands (`start`, `bundle`, and `ramBundle`) as well as the metro config/asset code.

I had to move a few "tools" from `cli` to `cli-tools`, so I could use them in `metro`. That made things a bit complicated and added to the PR.

grabbou:
No worries, that's something I've been waiting for quite some time to see, so I am more than happy to review it!

grabbou:
In the future release of React Native, I would like to discuss an opportunity of moving Metro dependency to a user project, so that they can define bundler and complementary CLI plugin for it.

afoxman:
> In the future release of React Native, I would like to discuss an opportunity of moving Metro dependency to a user project, so that they can define bundler and complementary CLI plugin for it.

That sounds good. We should sync up on what you have in mind, and how I can help. Sent you a friend request on Discord to work out the details in DMs.

afoxman:
@thymikee @grabbou Hi folks - checking in to see when you might merge this, or if you'd like any further changes?

thymikee:
I'm sure @grabbou will prioritize this the next time he works on this lib

grabbou:
Update: I am still in progress with this one, expected update tomorrow

packages/metro/src/bundle/buildBundle.ts
```diff
@@ -40,7 +40,7 @@ export interface AssetData {
   files: string[];
 }
 
-async function buildBundle(
+export async function buildBundle(
```
grabbou, line 43: Any particular reason for removing `export default` in favour of named export?

grabbou: I can see this being the case in almost every file. Is this some kind of a convention? In CLI, we tend to create a single file for a function and prefer `export default`s

packages/metro/src/config/index.ts
```diff
@@ -0,0 +1 @@
+export * from './loadMetroConfig';
```
grabbou, line 1: Do we need this file? I think we can either rename `loadMetroConfig` to just `config.js` (no need for `config` folder).

packages/metro/src/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from './bundle';
+export * from './config';
+export * from './start';
```
grabbou, line 3: I have mixed feelings about this approach, since it makes it hard to actually track down what this module exports at the end of the day.

Since we're always mindful about what kind of API surface we explore, how about we do it explicitly, like in other modules: https://github.com/react-native-community/cli/blob/7d460229962fff69841b6e743c73fd0b1d5dc529/packages/platform-ios/src/index.ts#L5-L7

packages/metro/src/start/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './start';
+export * from './runServer';
```
grabbou, line 2: Again, `index.js` file is not needed here in my humble opinion (sticking to the other plugins in the CLI).

Right now, since `runServer` and `getReporterImpl` are exported from `runServer`, they are also exported to the public (everything from `start` is exported to the public via `src/index.js`).

While these two functions are used internally in other files, **I wouldn't change any exports in this PR beyond the single one that you have requested.** 

For example, `getReporterImpl` is not needed anywhere outside of `runServer.ts` file - there's no need to even export it internally, not mentioning exporting it as a public interface.

packages/cli/package.json
```diff
@@ -29,6 +29,7 @@
     "@react-native-community/cli-server-api": "^6.0.0-rc.0",
     "@react-native-community/cli-tools": "^6.0.0-rc.0",
     "@react-native-community/cli-types": "^6.0.0-rc.0",
+    "@react-native-community/metro": "^6.0.0-rc.0",
```
grabbou, line 32: I would name the package `cli-plugin-metro`, since the name `metro` itself in that namespace is quite broad and doesn't necessarily address the package's purpose.

grabbou (changes requested):
Conceptually, everything looks fine. I have left some comments, but I believe this is going to be the final round.

TL;DR:

- Let's not export "all" (via `export * from 'foo.ts'`) and avoid creating `index.ts` in folders
- Prefer `export default` instead of single named exports per file (sticks consistent with the rest of the codebase, happy to revisit later in a separate PR)
- Export explicitly the public interface via `src/index.ts` (instead of `*`)
- Do not change current exported/private functions beyond the `loadMetroConfig.ts` that is subject to this PR (unless there's a need for an additional function - then, it's welcome)

I am happy to hop on a Discord/Hangouts call to discuss this in a bit more details if necessary. Thank you for working on that!

packages/metro/src/config/index.ts
afoxman: There's a 2nd file there: config/metroPlatformResolver.ts. It's used internally by loadMetroConfig.ts.

afoxman: And a lot of types are being exported. Going by a previous comment, I changed * to an explicit list of types/functions being exported for easier maintenance over time.

packages/metro/src/index.ts
afoxman: I made it explicit.

grabbou: Thanks. Are you working on the rest of the things I left as a PR review? Or is this the final round to review once again?

afoxman: And I limited the exports to only what I need to consume from this PR, as well as what the cli project needs to consume (the `command` objects).

packages/metro/src/start/index.ts
afoxman: Got rid of it and limited the exports to only what is needed for the cli project.

packages/cli/src/commands/index.ts
```diff
@@ -1,7 +1,9 @@
 import {Command, DetachedCommand} from '@react-native-community/cli-types';
-import start from './start/start';
-import bundle from './bundle/bundle';
-import ramBundle from './bundle/ramBundle';
+import {
+  startCommand as start,
+  bundleCommand as bundle,
+  ramBundleCommand as ramBundle,
+} from '@react-native-community/cli-plugin-metro';
```
grabbou, line 6: In the future, we can remove this when this plugin becomes direct React Native dependency. 

Here's a snippet:
https://github.com/facebook/react-native/blob/main/react-native.config.js#L16

I believe it should simply say `metro.commands`, but this is to be decided on the planning meeting.

grabbou:
I updated your PR with some further naming changes. The primary change was to group commands into `commands` folder (stays uniform with android and iOS plugins) and export them as an array. They're not supposed to be consumed as a public API, but used internally, and array is what the CLI infrastructure prefers.

Please let me know if you're happy with how it looks and whether the public API satisfies your use case.

packages/cli-plugin-metro/src/index.ts
```diff
@@ -1,13 +1,8 @@
-export {
-  bundleCommand,
-  ramBundleCommand,
-  buildBundleWithConfig,
```
afoxman: I need buildBundleWithConfig to be publicly exported.

afoxman: Added (along with CommandLineArgs type) in my latest push

afoxman:
> I updated your PR with some further naming changes. The primary change was to group commands into `commands` folder (stays uniform with android and iOS plugins) and export them as an array. They're not supposed to be consumed as a public API, but used internally, and array is what the CLI infrastructure prefers.
> 
> Please let me know if you're happy with how it looks and whether the public API satisfies your use case.

Everything looks ok to me. One function export and a type were removed during the shuffle, so I added them back in. I need to be able to call `buildBundleWithConfig(...)` and so I need it as well as its input type `CommandLineArgs`.

grabbou:
LGTM.

grabbou:
The CI failure looks unrelated. I will move forward and merge this PR.

Issue #1435, opened 2021-06-27
run-ios bug: Try/catch removed from `warnAboutManuallyLinkedLibs.ts` causing exceptions to propagate/crash


Saadnajmi:
## Environment

```
System:
  OS: macOS 11.4
  CPU: (8) x64 Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz
  Memory: 921.95 MB / 16.00 GB
  Shell: 3.2.57 - /bin/bash
Binaries:
  Node: 16.4.0 - /usr/local/bin/node
  Yarn: 1.22.10 - /usr/local/bin/yarn
  npm: 7.18.1 - /usr/local/bin/npm
  Watchman: 4.9.0 - /usr/local/bin/watchman
Managers:
  CocoaPods: 1.10.1 - /usr/local/bin/pod
SDKs:
  iOS SDK:
    Platforms: iOS 14.5, DriverKit 20.4, macOS 11.3, tvOS 14.5, watchOS 7.4
  Android SDK:
    API Levels: 28, 29, 30
    Build Tools: 28.0.3, 29.0.3, 30.0.1
    System Images: android-30 | Google APIs Intel x86 Atom
    Android NDK: Not Found
IDEs:
  Android Studio: 4.1.2 4.1.2
  Xcode: 12.5.1/12E507 - /usr/bin/xcodebuild
Languages:
  Java: 11.0.1 - /usr/bin/javac
npmPackages:
  @react-native-community/cli: Not Found
  react: 17.0.1 => 17.0.1 
  react-native: 0.64.1 => 0.64.1 
  react-native-macos: ^0.63.0 => 0.63.36 
npmGlobalPackages:
  *react-native*: Not Found
```

## Description

This refers to https://github.com/microsoft/react-native-test-app/issues/375

I was working on moving a community module to use [React Native Test App](https://github.com/microsoft/react-native-test-app/) to manage the test app of a community module (https://github.com/react-native-menu/menu), and ran into an error running `run-ios`. After investigation, it seems the issue happened upstream beginning in https://github.com/react-native-community/cli/pull/1270 .

A try-catch was removed from `packages/platform-ios/src/link/warnAboutManuallyLinkedLibs.ts`, causing the exception to bubble up and crash. This seems like an error, because an Android fix shouldn't have broken something downstream for iOS.

## Reproducible Demo

A little bit contrived since I reproed this using a downstream module, but:
```
git clone git@github.com:react-native-menu/menu.git
cd menu
[...]

PR #1459, opened 2021-08-04
feat(cli-patform-ios): allow setting a path to a `Podfile` 

grabbou:
Draft following explanation in https://github.com/react-native-community/cli/issues/1435#issuecomment-892999907

To be tested, CC @tido64 and @kelset, as I think I am not getting a successful project initialisation (`.xcworkspace` missing).

The `react-native.config.js` of my `example` has been customised as follows:

```js
ios: {
 project: "node_modules/react-native-test-app/ios/ReactTestApp.xcodeproj",
 podfile: "ios/Podfile"
},
```

### DIFF
--- a/packages/platform-ios/src/commands/runIOS/index.ts
+++ b/packages/platform-ios/src/commands/runIOS/index.ts
@@ -618,7 +618,7 @@ export default {
       description:
         'Path relative to project root where the Xcode project ' +
         '(.xcodeproj) lives.',
-      default: 'ios',
+      default: (ctx: Config) => ctx.project.ios ? ctx.project.ios.sourceDir : 'ios',
     },
     {
       name: '--device [string]',
--- a/packages/platform-ios/src/commands/runIOS/index.ts
+++ b/packages/platform-ios/src/commands/runIOS/index.ts
@@ -618,7 +618,7 @@ export default {
       description:
         'Path relative to project root where the Xcode project ' +
         '(.xcodeproj) lives.',
-      default: (ctx: Config) => ctx.project.ios ? ctx.project.ios.sourceDir : 'ios',
+      default: 'ios',
     },
     {
       name: '--device [string]',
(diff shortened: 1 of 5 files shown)


packages/platform-ios/src/commands/runIOS/index.ts
```diff
@@ -618,7 +618,7 @@ export default {
       description:
         'Path relative to project root where the Xcode project ' +
         '(.xcodeproj) lives.',
-      default: 'ios',
+      default: (ctx: Config) => ctx.project.ios ? ctx.project.ios.sourceDir : 'ios',
```
grabbou, line 621: `iOS` might be null when there's no `iOS` project but users run `react-native --help` and want to see all the commands. We could've used an empty string as well, but I guess sticking to `iOS` is also fine. 

No preference.

tido64:
Thanks, @grabbou. I applied the fixes directly to relevant files under `node_modules` and it works for me. I don't like that we have to specify where to find the `Podfile` but I guess this will be addressed in `next`?

Kiesco08:
Thanks for the fix! There is another [area](https://github.com/react-native-community/cli/blob/44f541332ee57e8a30feb3ca24479ab292149dd9/packages/platform-ios/native_modules.rb#L89) where it also assumes project root directory is the podfile directory. Would it be possible to apply a change there too?

```diff
+    project_pod_file_dir = File.dirname(config["project"]["ios"]["podfile"])
-    relative_path = podspec_dir_path.relative_path_from project_root
+    relative_path = podspec_dir_path.relative_path_from project_pod_file_dir
```

tido64:
On closer inspection, this doesn't fully work. I ran `pod install` before applying the fix so it worked fine. But if I run `pod install` _with_ the fix, I get this error:

```
[!] No podspec found for `ReactTestApp-DevSupport` in `..`
```

In addition to @Kiesco08's proposal, I also had to make `podfile` an absolute path, i.e.

```json
      podfile: path.join(__dirname, "ios", "Podfile"),
```

It's been so long since I looked at it but this is why `sourceDir` had to be fixed in #1444.

Issue #1371, opened 2021-03-06
Pass custom-placed metro.config.js through --config to run-android and run-ios commands


ElviraBurchik:
## Describe the Feature

At Shopify, we are using the React Native CLI in a monorepo setup:

```
react-native-packages/
  node_modules/
  fixture/
      node_modules/
      metro.config.js
      android/
      ...
  packages/
  ...
```

Since the CLI launches the packager taking the directory where React Native lives as the working directory, it causes the packager command to try reading the Metro configuration from the wrong directory - `react-native-packages/node_modules/react-native`. 

I suppose that it could be solved by adding an option to provide a path to the custom `metro.config.js` when running `run-ios` and `run-android` commands.

## Possible Implementations

I spent some time understanding the code base and find some possible implementations, but I struggle to find a place where Metro CLI actually gets called. 

How I understand the process so far: [runAndroid command](https://github.com/react-native-community/cli/blob/master/packages/platform-android/src/commands/runAndroid/index.ts#L248) runs `launchPackager.command` to start metro server. `launchPackager.command` calls [react-native-community/cli run](https://github.com/react-native-community/cli/blob/master/packages/global-cli/index.js#L264), and from here I can't connect the dots how we run Metro CLI and start the server. 

I'd be happy to submit a pr for this feature request if somebody could point me out the right place for introducing these changes 😄🤞

## Related Issues

https://github.com/facebook/metro/issues/588
https://github.com/react-native-community/cli/pull/1358

ElviraBurchik:
@grabbou moving [the discussion from this pr](https://github.com/react-native-community/cli/pull/1358#issuecomment-800453417) here since we are going to implement it withing this issue :)

> Right now, we obtain Metro configuration programatically here https://github.com/react-native-community/cli/blob/master/packages/cli/src/tools/loadMetroConfig.ts#L152 and we set `cwd` to `root` from Config. By design, Metro will look for a config there.
> 
> However, we also pass a second object, `options`, which - inside `start` and `bundle` commands, has historical `config` property ->
> 
> https://github.com/react-native-community/cli/blob/5ec3fcf3a623b5a9faef050b72eab3778f241761/packages/cli/src/commands/start/start.ts#L87-L89
> 
> You can use it to set a path to custom Metro configuration. The name probably needs to be updated, as historically, Metro/CLI config used to be a single file. While refactoring, we stopped relying on that property inside CLI, but kept it for Metro.
> 
> So, in general, I would say there are two ways going forward:
[...]

PR #1461, opened 2021-08-12
Add metro-config option

fortmarek:
Resolves https://github.com/react-native-community/cli/issues/1371

Summary:
---------

Hi 👋 

As explained in the attached issue this PR should resolve, when starting the metro bundler after running `run-android`, it cannot find the `metro.config.js` due to `react-native` being one level higher in the file hierarchy than expected.

Therefore, I have added `--metro-config` option that will allow users to specify the location of metro config when running `run-android`. Under the hood, I am creating a new launch packager script that will invoke `launchPackager.command` defined in `react-native` module - but, importantly - with `--metro-config xyz` if it was specified. `launchPackager.command` then runs `packager.sh` which then runs `"$NODE_BINARY" "$REACT_NATIVE_ROOT/cli.js" start "$@"`. Fortunately, it passes arguments by default, so our `--metro-config xyz` gets passed as well. cli is then run with the metro config specified and it doesn't fail to find it.

I don't like the fact that I have to create a temporary script to be actually able to pass commands to `launchPackager.command` as it's not possible to do that directly, so if you have ideas how to make this cleaner, I'd love to know!

Test Plan:
----------

I have tested that when I run `react-native run-android --metro-config path-to-metro.config.js`, it gets passed to the CLI when invoked from the launch packager scripts and is subsequently used. (when trying to load the bundle, I am getting `Error: Unable to resolve module @babel/runtime/helpers/interopRequireDefault` but that is IMHO unrelated to my changes and is rather tied to my local setup - if you know how to resolve this when contributing to this repo, do let me know, I'll continue investigating this further in the meantime)

If this PR is in the right direction, I should also probably add some unit tests.

### DIFF
--- a/packages/cli/src/commands/bundle/buildBundle.ts
+++ b/packages/cli/src/commands/bundle/buildBundle.ts
@@ -48,7 +48,7 @@ async function buildBundle(
   const config = await loadMetroConfig(ctx, {
     maxWorkers: args.maxWorkers,
     resetCache: args.resetCache,
-    config: args.config,
+    config: args.metroConfig ?? args.config,
   });
 
   if (config.resolver.platforms.indexOf(args.platform) === -1) {
--- a/packages/cli/src/commands/bundle/bundleCommandLineArgs.ts
+++ b/packages/cli/src/commands/bundle/bundleCommandLineArgs.ts
@@ -15,6 +15,7 @@ export interface CommandLineArgs {
   resetGlobalCache: boolean;
   transformer?: string;
   minify?: boolean;
+  metroConfig?: string;
   config?: string;
   platform: string;
   dev: boolean;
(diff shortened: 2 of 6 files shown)


fortmarek:
Also, `unit-tests` seem to be failing [here](https://app.circleci.com/pipelines/github/react-native-community/cli/3647/workflows/e9c3ebfe-e6b5-45fc-81d2-ad37d7f362b6/jobs/18153?invite=true#step-102-386) but that seems unrelated to my changes - any idea how can I fix it? 🤔

fortmarek:
Would you have time to look into this in the foreseeable future, @grabbou? 🙂

grabbou:
Closing as per my comment in #1371. Thank you for your work! Looking forward to the next PR!

PR #1464, opened 2021-08-24
fix: export `loadConfig` so it can be used for other purposes

tido64:
Summary:
---------

Exports `loadConfig` so it can be used by other libraries. Specifically, I'm working on adding the ability to target specific platforms in Jest. Being able to detect which platforms are available (especially out-of-tree ones) would be most useful.

Test Plan:
----------

There are no code changes to test.

### DIFF
--- a/packages/cli/src/index.ts
+++ b/packages/cli/src/index.ts
@@ -255,4 +255,4 @@ async function setupAndRun() {
 
 const bin = require.resolve('./bin');
 
-export {run, init, bin};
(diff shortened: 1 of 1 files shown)


tido64:
@thymikee: I would really appreciate it if you could have a look and let me know if this is something you'd want to do.

thymikee (approved):
This increases the API surface, but i guess it's pretty harmless

PR #1469, opened 2021-09-09
fix(debugger-ui): Add @babel/runtime dependency

asmundg:
Summary:
---------

The @babel/plugin-transform-runtime transform adds a runtme require of
@babel/runtime. See
https://babeljs.io/docs/en/babel-plugin-transform-runtime#installation.

Test Plan:
----------

No testing required, this only affects downstream consumers.

### DIFF
--- a/packages/debugger-ui/package.json
+++ b/packages/debugger-ui/package.json
@@ -18,6 +18,7 @@
     "parcel-bundler": "^1.12.4"
   },
   "dependencies": {
+    "@babel/runtime": "^7.4.4",
     "serve-static": "^1.13.1"
   },
   "homepage": "https://github.com/react-native-community/cli/tree/master/packages/debugger-ui",


asmundg:
Using babel-transform-runtime adds a require("@babel/runtime") to your output. From the link above (note the last line):

"The transform-runtime transformer plugin does three things:

    Automatically requires @babel/runtime/regenerator when you use generators/async functions (toggleable with the regenerator option).
    Can use core-js for helpers if necessary instead of assuming it will be polyfilled by the user (toggleable with the corejs option)
    Automatically removes the inline Babel helpers and uses the module @babel/runtime/helpers instead (toggleable with the helpers option).

What does this actually mean though? Basically, you can use built-ins such as Promise, Set, Symbol, etc., as well use all the Babel features that require a polyfill seamlessly, without global pollution, making it extremely suitable for libraries.

Make sure you include @babel/runtime as a dependency."

asmundg:
The usage of the plugin is here: https://github.com/react-native-community/cli/blob/e1da64317a1178c2b262d82c2f14210cdfa3ebe1/packages/debugger-ui/.babelrc#L2

PR #1473, opened 2021-09-22
Fix React Native peer dep

Titozzz:
There is an issue with peer deps when using NPM and 0.66.
Using '*' should be fine since the CLI is installed as a dependency of react-native, so as long as react-native install the correct version they should be fine with each other


thymikee (approved):
Thanks! We still specify supported versions in the README, and this will fix npm auto installing peer deps where we wouldn't like it to.

PR #1475, opened 2021-09-25
fix: Windows unknown command on build android

rmaafs:
Summary:
---------

### In windows, when I execute `npx react-native run-android`, I got this problem:
image

### It's because my Windows name profile is "Rodrigo Maafs" (with a space)
image

### The error is caused from `getDevices(adbPath)`, adbPath have 3 arguments, looks like:
image

### So, I fixed this adding `"` in the adb path:
image

Test Plan:
----------

1. Use Windows.
2. Use a "adb path" in folder with a space.
3. Check if the environment PATH, have the adb path correctly.
4. Execute the app, in my case I use `npx react-native run-android`

image

After fixed it, looks like
after

### DIFF
--- a/packages/platform-android/src/commands/runAndroid/adb.ts
+++ b/packages/platform-android/src/commands/runAndroid/adb.ts
@@ -34,6 +34,11 @@ function parseDevicesResult(result: string): Array<string> {
  */
 function getDevices(adbPath: string): Array<string> {
   try {
+    //If the path includes spaces (for example, the name of the USER_NAME) in Windows...
+    if (adbPath.includes(' ') && process.platform.startsWith('win')) {
+      adbPath = `"${adbPath}"`;
+    }
+
     const devicesResult = execSync(`${adbPath} devices`);
     return parseDevicesResult(devicesResult.toString());
   } catch (e) {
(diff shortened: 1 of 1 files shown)


packages/platform-android/src/commands/runAndroid/adb.ts
```diff
@@ -34,6 +34,11 @@ function parseDevicesResult(result: string): Array<string> {
  */
 function getDevices(adbPath: string): Array<string> {
   try {
+    //If the path includes spaces (for example, the name of the USER_NAME) in Windows...
+    if (adbPath.includes(' ') && process.platform.startsWith('win')) {
+      adbPath = `"${adbPath}"`;
```
grabbou, line 39: Does this conditional matter? Can't we always wrap it with " " for the safety? I think that would work on Mac too.

grabbou (changes requested):
Thanks for the PR! Your research is really impressive! I think we can further simplify the changes you have in your PR and definitely ship it, since it's going to affect quite a few with the spaces in the username.

packages/platform-android/src/commands/runAndroid/adb.ts
rmaafs: I don't have a Mac to test it, but I will find one.
The conditional is because I don't want break any thing.

rmaafs:
✔ Tested on Mac. Thanks @efrenperedo

asklar (approved):

grabbou:
My question is - why don't we wrap it with `""` all the time? I don't think we need a special `if` condition to do it only when the string contains spaces.

@thymikee

thymikee:
@grabbou sure we can go with it by default

thymikee (approved):

Issue #1047, opened 2020-03-11
Move Doctor to a separate plugin


grabbou:
Right now, Doctor command and its tools are populating the core of the CLI, making it larger and more complex to maintain.

With future improvements in mind and additional automatic resolves, I recommend we move it to a separate package, just like iOS/Android files.

thymikee:
Yes please

nickx720:
@thymikee Hi, I wasn't sure if this ticket was still being worked on, so I went ahead and made the changes. I am a newbie so I hope I got it right. Thanks in advance :).

PR #1480, opened 2021-09-29
feat: move doctor into `cli-doctor` package, extract config into `cli-config`, clean-up types

nickx720:
Summary:
---------
Implemented changes as per the [pr](https://github.com/react-native-community/cli/issues/1047)

1. Created a separate directory under packages. 
2. Labelled it as cli-doctor
3. Moved the contents from doctor to cli-doctor
4. Updated import paths
5. Set up package.json under cli-doctor and modelled it along other cli packages like cli-hermes
6. Ran yarn-install to ensure success local build

From Mike:

7. Refactored `Ora` updates and cleaned-up `getLoader`
8. Extracted `cli-config` into a separate package, reorganised tools inside `cli` so that it contains `cli`-only code
9. Checked all `package.json`'s for extraneous dependencies and made sure everything corresponds to the files used

Test Plan:
----------

1. Yarn install to ensure the dependencies worked on local.
2. Made all unit tests pass


packages/cli-doctor/src/healthchecks/__tests__/androidHomeEnvVariable.test.ts
```diff
@@ -1,5 +1,5 @@
 import androidHomeEnvVariables from '../androidHomeEnvVariable';
-import {NoopLoader} from '../../../../tools/loader';
+import {NoopLoader} from '@react-native-community/cli/src/tools/loader';
```
grabbou, line 2: You should not require files directly from source, but rather expose them via public API. TL;DR always require from `@react-native-community/cli`.

packages/cli-doctor/src/healthchecks/__tests__/androidSDK.test.ts
```diff
@@ -1,15 +1,15 @@
...
+import getEnvironmentInfo from '@react-native-community/cli/src/tools/envinfo';
+import * as downloadAndUnzip from '@react-native-community/cli/src/tools/downloadAndUnzip';
 import {EnvironmentInfo} from '@react-native-community/cli-types';
-import {NoopLoader} from '../../../../tools/loader';
+import {NoopLoader} from '@react-native-community/cli/src/tools/loader';
 import * as common from '../common';
-import * as androidWinHelpers from '../../../../tools/windows/androidWinHelpers';
-import * as environmentVariables from '../../../../tools/windows/environmentVariables';
+import * as androidWinHelpers from '@react-native-community/cli/src/tools/windows/androidWinHelpers';
+import * as environmentVariables from '@react-native-community/cli/src/tools/windows/environmentVariables';
```
grabbou, line 12: Are all these functions used outside of a Doctor command? If all these tools, such as `androidWinHelpers` or `downloadAndUnzip` are used in Doctor only, then you should move them to doctor `tools` directory as well (first you may need to create one).

Check the recent commit that moves Metro to a separate package, it's a good example of moving the tooling around as well.

https://github.com/react-native-community/cli/commit/0993f627869f4de0e5a8a9a950d38a9f15e18fb2

packages/cli-doctor/package.json
```diff
@@ -0,0 +1,31 @@
+{
+  "name": "@react-native-community/cli-doctor",
```
grabbou, line 2: I guess the name should be more of a `cli-plugin-doctor`, just like we did with Metro.

grabbou (changes requested):
Thanks for the PR. It's a great thing to do such a cleanup. I did left some notes (not a big deal) that need to be applied before this one is good to be merged. Looking forward to it!

packages/cli-doctor/src/healthchecks/__tests__/androidSDK.test.ts
nickx720: Sure will do thanks.

packages/cli-doctor/package.json
nickx720: Got it. Will do Thanks :)

packages/cli-doctor/src/healthchecks/__tests__/androidHomeEnvVariable.test.ts
nickx720: I see. Makes sense

grabbou:
Thanks for the PR. I will continue the work (apply the changes + few tweaks and rebase) and merge it in order to move forward with this one today.

grabbou:
This PR is still work in progress. I will return to it tomorrow. There are a few tools that need to be moved around to clean this up.

packages/cli-doctor/src/index.ts
```diff
@@ -0,0 +1,6 @@
+import doctor from './commands/doctor';
+import info from './commands/info';
+
+export {logManualInstallation, logError} from './tools/healthchecks/common';
```
grabbou, line 4: There are some dependencies between CLI core and Doctor, to be investigated.

grabbou: I have created a separate package `config` for certain things.

grabbou:
Created a new package: `cli-config` for things that are exclusive to `config` only. 
Cleaned up dependencies - there were a lot of dependencies missing in `package.json`, I made sure everything is covered.
Fixed up dependency on `Ora` - all packages are using `getLoader()` from `tools` package and should refer to the type from that package, rather than relying on Ora directly.

packages/cli-doctor/src/index.ts
```diff
@@ -0,0 +1,6 @@
+import doctor from './commands/doctor';
+import info from './commands/info';
+
+export const commands = {info, doctor};
+
+export {default as installPods} from './tools/installPods';
```
grabbou, line 6: Ideally, I would like to remove dependency on this file from the CLI package.

grabbou:
One issue with `Ora` is remaining and we can merge this. Ignore type error and this is ready for review.

packages/cli-doctor/src/commands/doctor.ts
```diff
@@ -264,3 +263,22 @@ export default (async (_, options) => {
     printFixOptions({onKeyPress});
   }
 }) as DetachedCommandFunction<FlagsT>;
+
+export default {
+  func: doctorCommand,
+  detached: true,
+  name: 'doctor',
+  description:
+    '[EXPERIMENTAL] Diagnose and fix common Node.js, iOS, Android & React Native issues.',
```
thymikee, line 272: It's not so experimental anymore, right? Let's get rid of this label

packages/cli-types/src/index.ts
```diff
@@ -246,7 +152,7 @@ export interface Config extends IOSNativeModulesConfig {
     [name: string]: PlatformConfig<any, any, any, any>;
   };
   commands: Command[];
-  healthChecks: HealthCheckCategory[];
+  healthChecks: [];
```
thymikee, line 155: let's add a TODO to adjust it in the next round of refactor

thymikee (approved):

Issue #1476, opened 2021-09-25
Feature request: support Apple TV


douglowder:
## Describe the Feature

I'm the maintainer for the Apple TV and Android TV RN repo (https://github.com/react-native-tvos/react-native-tvos).  I'd like to implement better CLI support for Apple TV with two features:

- New command `react-native run-tvos` (this can be done with the right flags to `run-ios`, but it would be convenient to have a separate command)
- Make `react-native upgrade` work with the `react-native-tvos` fork

## Possible Implementations

I plan on implementing these, but wanted to find out your preference on the approach I should take.

For `run-tvos`, should I make changes to the existing `platform-ios` package, or do it in a separate package (which means a lot of duplicate code)?

For `upgrade`, it seems to me the right way is to have `upgrade.ts` detect that the project uses `react-native-tvos` and use the `react-native-tvos` diffs (I have already created a repo for the diffs).  But if you prefer a separate command or package for this, that's ok.

thymikee:
CLI has a plugin subscription mechanism that allows npm packages to expose their own plugins. This way `react-native-tvos` can expose a `run-tvos` plugin, that uses platform-ios under the hood (any changes to make it more generic so it fits tvos needs better are welcome!). Check this doc: https://github.com/react-native-community/cli/blob/master/docs/plugins.md.

Regarding `upgrade`, it's not super reliable for the default template. If the change would be a simple switch to change the source of the diff, then I think we can put this in for now. And later figure out something similar to CLI plugins so this repo doesn't have to maintain too many templates.

douglowder:
Thanks for the tips, I'll do `run-tvos` inside the RN TV repo, and see if the change for `upgrade` can be done here in a not too intrusive way.

douglowder:
Fixed in release 6.1.0

PR #1481, opened 2021-10-03
feat: Allow the upgrade command to use TV repo

douglowder:
Summary:
---------

The upgrade command does not currently work for Apple TV and Android TV projects, because they use `react-native-tvos`, a fork of the main repo. (Issue #1476) (https://github.com/react-native-tvos/react-native-tvos/issues/224)

Solution:
- Provide a version of https://github.com/react-native-community/rn-diff-purge that has the diffs for the TV repo
- Make `upgrade.ts` generic by detecting whether the project is using a forked repo, and having an easily extendable list of forks that are known to work.  This list includes the main repo as well.

```ts
type ForkNameType = 'react-native' | 'react-native-tvos';

const forks = {
  'react-native': {
    rawDiffUrl:
      'https://raw.githubusercontent.com/react-native-community/rn-diff-purge/diffs/diffs',
    webDiffUrl: 'https://react-native-community.github.io/upgrade-helper',
    dependencyName: 'react-native',
  },
  'react-native-tvos': {
    rawDiffUrl:
      'https://raw.githubusercontent.com/react-native-tvos/rn-diff-purge-tv/diffs/diffs',
    webDiffUrl: 'https://react-native-community.github.io/upgrade-helper',
    dependencyName: 'react-native@npm:react-native-tvos',
  },
};

```

All methods in `upgrade.ts` now refer to this structure in their implementations.

Test Plan:
----------

The upgrade unit tests have been refactored so that the test implementations and common mocks are in a common module, allowing the main repo and TV repo to be tested separately with their repo-specific mocks.

### DIFF
--- a/packages/cli/src/commands/upgrade/__tests__/__snapshots__/upgrade-tv.test.ts.snap
+++ b/packages/cli/src/commands/upgrade/__tests__/__snapshots__/upgrade-tv.test.ts.snap
@@ -1,7 +1,7 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`Upgrade tests for react-native repo cleans up if patching fails, 1`] = `
-"info Fetching diff between v0.57.8 and v0.58.4...
+exports[`Upgrade tests for react-native-tvos repo cleans up if patching fails, 1`] = `
+"info Fetching diff between v0.62.2-1 and v0.64.2-4...
 [fs] write tmp-upgrade-rn.patch
 $ execa git rev-parse --show-prefix
 $ execa git apply --binary --check tmp-upgrade-rn.patch --exclude=package.json -p2 --3way --directory=
@@ -11,7 +11,7 @@ warn Excluding files that exist in the template, but not in your project:
 error Excluding files that failed to apply the diff:
   - ios/MyApp.xcodeproj/project.pbxproj
 Please make sure to check the actual changes after the upgrade command is finished.
-You can find them in our Upgrade Helper web app: https://react-native-community.github.io/upgrade-helper/?from=0.57.8&to=0.58.4
+You can find them in our Upgrade Helper web app: https://react-native-community.github.io/upgrade-helper/?from=0.62.2-1&to=0.64.2-4
 $ execa git apply tmp-upgrade-rn.patch --exclude=package.json --exclude=.flowconfig --exclude=ios/MyApp.xcodeproj/project.pbxproj -p2 --3way --directory=
 debug \\"git apply\\" failed. Error output:
 error: .flowconfig: does not exist in index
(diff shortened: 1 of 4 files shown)


grabbou (approved):
It's a very nice PR! Thank you for working on this. Having the ability to map custom forks is definitely useful. 

I think this will become less popular in the future, as I guess the preferred way is to have a custom project and integrate with code nicely. 

However, with the abstraction made within this PR, React Native Windows and other platforms may have an easy entry-point for providing their own diffs in the future too. CC: @acoates-ms @kelset how this is done atm?

grabbou (approved):

kelset:
looping in @NickGerleman who might know too

acoates-ms:
Can react-native-tvos not use the same mechanism for declaring itself as a platform as react-native-windows and react-native-macos does?  -- Which allows it to co-exist within a repo better.  So you could have an app that targets react-native and react-native-tvos without having to use a fork of react-native?

Similarly the information for rawDiffUrl and webDiffUrl should be something thats included in the react-native.config.js for each package, that way the CLI remains extendable for any platform, without us having to add platform specific code into the CLI for all the RN platforms.

douglowder:
@acoates-ms :

> Can react-native-tvos not use the same mechanism for declaring itself as a platform as react-native-windows and react-native-macos does? -- Which allows it to co-exist within a repo better. So you could have an app that targets react-native and react-native-tvos without having to use a fork of react-native?

The reason for the decision to make `react-native-tvos` a fork is discussed extensively in https://github.com/react-native-tvos/react-native-tvos/issues/11 . That being said, there have been enough changes in the last two years (such as the release plan changes announced today with RN 0.66) and the ability to declare yourself a platform, that maybe it is worth revisiting that decision.

> Similarly the information for rawDiffUrl and webDiffUrl should be something thats included in the react-native.config.js for each package, that way the CLI remains extendable for any platform, without us having to add platform specific code into the CLI for all the RN platforms.

I remember looking at the config structure and did consider adding those to it, but decided to try and restrict my changes just to the upgrade command, to lessen the possibility of regressions in the rest of the code.  However I think your proposal makes more sense in the long term, and I'll have a look at that at some point in the future.

acoates-ms:
react-native-macos is also implemented as a fork of react-native -- since it too shares a large part of its code with the iOS implementation.  But the end user still uses it as a seperate package.  We fixed the cli to allow it to redirect all react-native imports to specific packages based on the platform being built.

douglowder:
@acoates-ms that is amazing, and I will take a look to see how you did that 👍 

It would definitely be worth aligning so that all platforms do these things the same way

acoates-ms:
All you need is this line: https://github.com/microsoft/react-native-macos/blob/0fb16c6ee620e63e3cadc331a75208b1250dca7c/react-native.config.js#L67

Within react-native-macos there is a conditional there, since we allow react-native-macos to be run either as a seperate package, or as a fork of react-native.  And maybe you would want to do the same with react-native-tvos.

PR #1495, opened 2021-11-05
Better Handle EPERM on Windows

NickGerleman:
Summary:
---------

Use async `fs-extra` function for rename, which hooks into a `graceful-fs` implementation that can handle AV locked files on Windows. See https://github.com/isaacs/node-graceful-fs/blob/89dc1330dcd8fa218c5dff92a97d8792b7da6b12/polyfills.js#L96-L118 for details.

Test Plan:
----------

Existing test coverage.

### DIFF
--- a/packages/cli/src/commands/init/editTemplate.ts
+++ b/packages/cli/src/commands/init/editTemplate.ts
@@ -1,4 +1,4 @@
-import fs from 'fs';
+import fs from 'fs-extra';
 import path from 'path';
 import {logger} from '@react-native-community/cli-tools';
 import walk from '../../tools/walk';
@@ -44,7 +44,7 @@ function replaceNameInUTF8File(
   }
 }
 
-function renameFile(filePath: string, oldName: string, newName: string) {
+async function renameFile(filePath: string, oldName: string, newName: string) {
   const newFileName = path.join(
     path.dirname(filePath),
     path.basename(filePath).replace(new RegExp(oldName, 'g'), newName),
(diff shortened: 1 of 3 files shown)


thymikee (approved):

thymikee:
Hm, we already utilize graceful-fs in `gracefulifyFs.ts`, which hooks into `fs` in this codebase. Shouldn't this handle the case?

NickGerleman:
> Hm, we already utilize graceful-fs in `gracefulifyFs.ts`, which hooks into `fs` in this codebase. Shouldn't this handle the case?

The special handling of rename on Win32 seems to only be applied to the async version of "rename". In the source linked above, I couldn't find any special handling of the sync version.

packages/cli/src/commands/init/editTemplate.ts
```diff
@@ -1,4 +1,4 @@
-import fs from 'fs';
+import fs from 'fs-extra';
```
thymikee, line 1: can we switch this back to `fs` then and keep the rest of the changes?

NickGerleman:
One thing I may want to check is that graceful-fs patches the fs promises interface as well.

NickGerleman:
Hmm I'm not sure the built in promise api works, but I can just promisify one of the official APIs.

NickGerleman:
I looked at the `gracefullify` implementation, and it doesn't seem to monkey patch the promises API. The promises API itself internally doesn't call into the callback API. https://github.com/nodejs/node/blob/0a62026f32d513a8a5d9ed857481df5f5fa18e8b/lib/fs.js#L2763

So, we could go back to `fs-extra`, as exposing promises over `graceful-fs`, or we could promisify bits of it ourselves. If `fs-extra` is already a dep that one seems a bit better.

thymikee:
Ugh. Let's please leave a comment on why we use `fs-extra` in this place then. It's kinda unexpected, so it makes sense to do it for our future selves and others. We can consider gracefullifying `fs.promises` as a separate effort, if someone is able to properly test it for regressions.

NickGerleman:
Added

PR #1496, opened 2021-11-11
Support renamed PreactNativeArchitectures

janicduplessis:
Summary:
---------

The `reactNativeDebugArchitectures` property was renamed to `reactNativeArchitectures` in https://github.com/facebook/react-native/commit/43c38cdc8e31910e0109f85d258fba22ec9469a0. This passes both flags to be compatible with both versions.

Test Plan:
----------

Test the change locally in an app.


thymikee (approved):
Thanks!

PR #1504, opened 2021-11-15
bump lerna to 4

dulmandakh:
Summary:
---------

Bump Lerna to 4.x, which dropped support for Node 6 and 8, then supports only 10+.

Test Plan:
----------

CI is same as before


dulmandakh:
@thymikee this is ready to be merged

thymikee (approved):
change to devDeps is not necessary, but in our case it doesn't matter

PR #1509, opened 2021-11-15
bump eslint, babel, jetifier, micromatch, nocache

dulmandakh:
Summary:
---------

Bump eslint, babel, jetifier, micromatch, nocache packages. Remove core-js because it's not used anywhere.

Test Plan:
----------

CI e2e pass.


thymikee:
Jetifier 1.6.7 and 2.0.0 added deprecation notices. How does it look like now? I'd prefer keeping it as is and remove the dependency in another PR

PR #1510, opened 2021-11-16
CI: run e2e on ubuntu

dulmandakh:
Summary:
---------

Run e2e on Ubuntu using GH Actions, then remove Circle CI config. Also disable e2e on Windows.

Test Plan:
----------

CI should run on GH Actions, and Circle CI fail because it couldn't find config file.


thymikee (approved):
Let's create an issue for enabling tests on Windows, shall we?

PR #1515, opened 2021-11-19
fix: doctor breaks in newer versions of Windows

### LINKED ISSUE #1513: Newer versions of windows can't run `npx react-native info` because of the missing wmic
## Environment

```
npx react-native info
error Unable to print environment info.
Error: Command failed: wmic os get Caption
'wmic' is not recognized as an internal or external command,
operable program or batch file.
```

## Description

See above

## Reproducible Demo

This will repro on Windows 10 21H1 and above. See https://docs.microsoft.com/en-us/windows/win32/wmisdk/wmic

asklar:
Fixes #1513 

Summary:
---------

wmic is deprecated (removed) in newer version of Windows. Use powershell to find out the version of android studio

Test Plan:
----------

### DIFF
--- a/packages/cli/src/commands/doctor/healthchecks/androidStudio.ts
+++ b/packages/cli/src/commands/doctor/healthchecks/androidStudio.ts
@@ -26,8 +26,9 @@ export default {
         'bin',
         'studio.exe',
       ).replace(/\\/g, '\\\\');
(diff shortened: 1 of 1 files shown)


thymikee (approved):
Thank you!

PR #1517, opened 2021-12-09
Revert "fix: doctor breaks in newer versions of Windows"

NickGerleman:
Reverts react-native-community/cli#1515

@asklar is out for the year, so we will not see an immediate fix. Reverting for now.

The hardcoded Powershell path causes failures on several current versions of Windows. This ends up also breaking the PR pipeline for `react-native-windows` regardless of whether we update, since we run `react-native doctor` on newly created user apps which hit the issue.

### DIFF
--- a/packages/cli/src/commands/doctor/healthchecks/androidStudio.ts
+++ b/packages/cli/src/commands/doctor/healthchecks/androidStudio.ts
@@ -26,9 +26,8 @@ export default {
         'bin',
         'studio.exe',
       ).replace(/\\/g, '\\\\');
-      const powershell = `${process.env.SystemRoot}\\System32\\WindowsPowerShell\\v1.0\\powershell.exe`;
       const {stdout} = await executeCommand(
-        `${powershell} -Command "& { (Get-ChildItem \\"${androidStudioPath}\\").VersionInfo.FileVersionRaw.ToString() }"`,
+        `wmic datafile where name="${androidStudioPath}" get Version`,
       );
       const version = stdout.replace(/(\r\n|\n|\r)/gm, '').trim();
 


asklar:
I'm not convinced this is necessary. As far as I can tell, windows powershell is installed by default on all Windows SKUs. In github actions, it shows up in the requested path. I see this issue is being hit in RNW, which uses custom AzDO images, so maybe the custom images have had powershell removed... Are there any reports of this issue being hit outside RNW?

Here is a successful run in github actions:
https://github.com/asklar/xaml-islands/runs/4466745872?check_suite_focus=true
image

asklar (changes requested):
marking request changes until we understand whether this is an issue or not, afaict everything is working as expected elsewhere :)

asklar:
and here is the same command run in RNWs AzDO:

https://dev.azure.com/ms/react-native-windows/_build/results?buildId=249400&view=logs&j=b8795b18-038f-57d8-81d6-958c7de806af&t=5087b0d5-a736-5368-9025-21d346c130fd&l=19
image

Were there perhaps changes to the image?

NickGerleman:
There were not changes to the image. This showed when we tried to sync the change.

To repro you can try running doctor on the RNW image, if it does not also fail locally. I've needed to disable our CI doing this, since it was preventing us from merging an increasing number of PRs.

NickGerleman:
This is the consistent failure being seen: 
image


Reverting this commit would be a quick repro: https://github.com/microsoft/react-native-windows/commit/7e947034a41171619b331e665558b5a418fa4040

NickGerleman:
We have plenty of code that still assumes `powershell` is in PATH. I remember hearing about some insider Windows builds with a bug leading it to not be present, but I still am confused about what we selectively started adding changes to hardcode the path everywhere.

asklar:
the doctor part seems to work fine locally:
image


However I do see that envinfo still relies on wmic, which isn't present on more recent OSs, so that part's broken. But it wouldn't hit explain the issue you're seeing.

NickGerleman:
@asklar please see the above quick repro. I would not be surprised if there is a difference in behavior between different versions or installations of Windows.

asklar:
here's where we say [windows powershell is available:](https://docs.microsoft.com/en-us/powershell/scripting/windows-powershell/install/installing-windows-powershell?view=powershell-7.2) 
> Windows PowerShell comes installed by default in every Windows, starting with Windows 7 SP1 and Windows Server 2008 R2 SP1.

and here we document [how to start it and what the path it would live in is](https://docs.microsoft.com/en-us/powershell/scripting/windows-powershell/starting-windows-powershell?view=powershell-7.2#:~:text=In%20Windows%20Server%202012%20R2%201%20On%20the,PowerShell%20%28x86%29.%204%20Via%20command%20line%2C%20enter%3A%20%25SystemRoot%25%5CSysWOW64%5CWindowsPowerShell%5Cv1.0%5Cpowershell.exe)(this is in the context of finding the wow64 version, but still).

I'm out of office right now but if you are still around it'd be great if you could investigate why this is failing for RNW, as I cannot reproduce this locally (beyond the other envinfo issue).

NickGerleman:
Given this change is meant for a not-yet-released versions of Windows, the previous state, which is known to work correctly in our official CI (for a current version of WIndows used in other images) seems preferable.

I am not able to dedicate time to this issue, so if you are out for the rest of the year, we should revert and revisit this in January.

asklar:
That's not correct. The change has been live publicly for several months:
https://docs.microsoft.com/en-us/windows/win32/wmisdk/wmic

> The WMI command-line (WMIC) utility is deprecated as of Windows 10, version 21H1, and as of the 21H1 semi-annual channel release of Windows Server. This utility is superseded by Windows PowerShell for WMI (see Chapter 7—Working with WMI). This deprecation applies only to the WMI command-line (WMIC) utility; Windows Management Instrumentation (WMI) itself is not affected. Also see Windows 10 features we're no longer developing.

Please feel free to work with someone who isn't away to investigate why this fails. Thanks!

NickGerleman:
@asklar I would not like this regression to randomize more folks during the holidays than it already has.

For now I've turned off tests in our repo since this was blocking multiple PRs from the engineers who are not yet out. Like you confirm above, wmic is not yet removed in current versions of Windows, so there is not a time-criticality to switch away from it before the end of the year.

I know it's not ideal to revert changes, but I really think it is the right thing to do here, both for customers, and the other folks working on RNW who are not affiliated with this change.

NickGerleman:
This has also hosed our CI for 0.65-0.67, which pull in the same new CLI bits. 
image

NickGerleman:
As a datapoint, @tido64 mentioned the issue does not repro for stock configuration of GitHub Actions (which shares bases images with Azure Pipelines).

Other variables are Azure Pipelines vs GitHub Actions non-image behavior, or differences in the RNW image vs the public one.

The CI for RNW 0.65+ aligns a bit closer to MSFT internal infrastructure than what is public. So, our CI will act a little bit more like it would for internal teams, then for GitHub Actions/Azure Pipelines. It's not running anything particularly exotic but is running something different.

NickGerleman:
image

The file appears to be physically there, but doctor consistently errors out with ENOENT.

thymikee (approved):
Thanks @NickGerleman. Since the reverted change seemingly only affects newer (I guess pre-release?) versions of Windows, and not reverting it breaks everybody else, I'm gonna revert it for now. Until there's some better story around it. 

We could, for example, try/catch the `wmic` command and fallback to `powershell` or hunt why Doctor fails when it shouldn't.

Since you and @asklar are way more familiar with Windows platform than me, please discuss what would be the best long-term solution :)

PR #1521, opened 2021-12-20
add flow-bin for template

### LINKED ISSUE #1518: Project has flowconfig but no flow-bin
## Description
So first time running `react-native init name` and process went well but I kept getting errors that flow couldn't be booted, took me a quick minute to check output and it turns out flow-bin doesn't get installed. Seems like it might be a quick one to just add [here](https://github.com/react-native-community/cli/blob/master/packages/cli/src/commands/init/initCompat.ts#L76).

## Reproducible Demo
- Have `Flow Language Support` installed on vscode
- init a new project
- See flow fail to start over and over again

Brianzchen:
Summary:
---------
Closes #1518. Add flow-bin as a template dependency when `init`ing project. When copying the react-native template by default it's built in flowtype but because of missing deps it's not a seamless experience

Test Plan:
----------
Build and run init project locally, see that flow-bin is installed and `yarn flow` works out of the box

### DIFF
--- a/packages/cli/src/commands/init/initCompat.ts
+++ b/packages/cli/src/commands/init/initCompat.ts
@@ -74,6 +74,7 @@ async function generateProject(
       '@babel/runtime',
       '@react-native-community/eslint-config',
       'eslint',
+      'flow-bin',
       'jest',
       'babel-jest',
       'metro-react-native-babel-preset',


grabbou:
Thanks for the PR, but as I already said in the issue, all the required dependencies should be present in the template package.json. For instance here: https://github.com/facebook/react-native/blob/main/template/package.json

The change you submitted modifies already deprecated `initCompat`, which is a legacy implementation of `react-native init` that was widely used more than 2 years ago. As a result, we're not going to accept this change.

PR #1522, opened 2021-12-21
Bump ora to 6.x

stianjensen:
Summary:
---------
The main changes include a new minimum node version requirement of 12.2, and that
the module is pure ESM now:
https://github.com/sindresorhus/ora/releases


packages/cli/src/commands/init/init.ts
```diff
@@ -2,7 +2,7 @@ import os from 'os';
 import path from 'path';
 import fs from 'fs-extra';
 import minimist from 'minimist';
-import ora from 'ora';
+import {Ora} from 'ora';
```
grabbou, line 5: I don't think this change was required while bumping to 6.x, was it? I personally have no objections to either convention, but I was confused while reviewing the PR.

@thymikee please lmk if you're okay with that change.

stianjensen: I believe it's due to a change in https://github.com/sindresorhus/ora/releases/tag/v6.0.0 where the ora export was changed and is no longer a ts namespace. But I may be mistaken.
I'll try once more without this change and verify whether or not I got any errors.

thymikee: I'm all good, it's the API change

stianjensen:
Something seems wrong with the lint, but I can't really figure out what. Is there an issue with the `import/named` rule?

Also, should I worry about the E2E test being red? I can't really see how the test failures are connected to this PR.

stianjensen:
> Something seems wrong with the lint, but I can't really figure out what. Is there an issue with the import/named rule?

Nothing beats coming back to it with fresh eyes. The trick is type-only imports (which luckily is just supported from the current TS version - 3.8, https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#type-only-imports-and-export)

packages/cli/src/commands/init/init.ts
grabbou: Thanks for pointing this out!

PR #1523, opened 2021-12-22
Fix metro crash when Flipper connects

mweststrate:
Summary:
---------

Fixes per work around: https://github.com/facebook/flipper/issues/3189#issuecomment-997191720. The same fix has been applied to FB infra, and the next release will temporarily set an `origin` header as work around for existing RN versions. 

Fixes:
* https://github.com/facebook/flipper/issues/2870
* https://github.com/facebook/flipper/issues/3189

The essence of the fix is that we shouldn't check the `origin` header if it is not present. That might sound unsafe, but it actually is just a check to protect from malicious websites, and browsers ensure the origin header is set for cross site requests. For websocket connections created outside a browser context, the creator has full control over the origin header anyway, so checking it would not add any safety.

Test Plan:
----------

Before, metro quits as soon as Flipper 0.125 / 0.126 connects:

Screenshot 2021-12-22 at 11 40 28

After, with manually applying the patch per linked comment above, no crash:
Screenshot 2021-12-22 at 11 42 25

N.B.: code formatter will probably fail, will push another commit once CI fails

### DIFF
--- a/packages/cli-server-api/src/websocket/eventsSocketServer.ts
+++ b/packages/cli-server-api/src/websocket/eventsSocketServer.ts
@@ -125,9 +125,10 @@ function attachToServer(
     verifyClient({origin}: {origin: string}) {
       // This exposes the full JS logs and enables issuing commands like reload
       // so let's make sure only locally running stuff can connect to it
-      return (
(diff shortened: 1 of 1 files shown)


thymikee (approved):
Sure, thanks 👍🏼

PR #1525, opened 2022-01-10
Android native_modules.gradle cliResolveScript should resolve to @react-native-community/cli

aparedes:
Summary:
---------

Android native_modules.gradle does not correctly get the app configuration. The cliResolveScript does not match the iOS version, and it only resolves to react-native/cli instead of @react-native-community/cli if avaiable

Test Plan:
----------
Green CI.


grabbou (approved):

grabbou:
Thanks, that makes sense!

Issue #1492, opened 2021-11-01
Control the loader in custom template to hide it for prompt feature


JeremyDolle:
Hi all 👋 

Is there any way to access the ora loader in the post init script ?

The use case is that I have a custom template that prompt a question to the user. The problem is that the "Executing post init script" loader is above the response of the prompt. 

It could be cool that we can control the loader in custom template to hide it for example.
Or if someone have an idea how I can achieve this I will appreciate. 
I try a lot of things like moving the cursor position in the terminal but the "executing post init script" follow it.

Hope someone can help me 🙏
Thanks for all!

thymikee:
PRs welcome :)

JeremyDolle:
Thanks @thymikee, I will try to investigate and see if I found several solution. If it makes sense, I will provide a PR.

Btw, how can I test the `init` command locally. I don't find it in the `CONTRIBUTING.md` file.

Thanks 🙏

Shaddix:
I'm also very interested in this.
The use-case is exactly the same: my postInitiScript asks the user for the input

Shaddix:
Regrading PRs/implementation:
1. I assume it's not possible to control the `ora` from postInitScript, since it's impossible to pass the `ora` object across the scripts run via `exec`.
2. If 1. is correct, then the only way to achieve the goal is to actually NOT show the loader for postInitScript
3. To do that I see 2 possibilities:
    1. Never show the loader for `postInitScripts`, and use `logger.log('Executing post init script');` instead.
    2. Do not show the loader depending on an option/flag in `template.config.js` (e.g. `skipLoaderForPostInitScript` with default to `false`).


I could PR any of the 3.1 or 3.2, if I'm correct in 1. :)

For me 3.1 is preferred, but it breaks the current behavior. 3.2 leaves the current behavior as is, but adds a configuration option for the thing that (imo) doesn't worth a configuration option :)

JeremyDolle:
I arrived to the same conclusion ! For me 3.2 was my first choice because we leave the same behavior and only projects that need to control that loader add the flag.

Btw, Hope the PR will be accepted even if it's 3.1 ! 😊

thymikee:
I'm good with moving the responsibility of showing the user progress of the "post init script" to the implementor – i.e. going with the solution 3.1. I would like the CLI's documentation to mention this responsibility though, ideally with an example of using `ora` in a similar fashion that we do today.

Shaddix:
[...]

PR #1526, opened 2022-01-17
Postinit script loader

JeremyDolle:
Summary:
---------

Related to #1492.
The idea is to remove the `ora` loader from the core of the cli to give this responsibility to the implementor. 
That way, template creator can be more creative like prompting questions to user for better experience. 
Before this, it was difficult because `ora` keeps refreshing the last line of terminal and so the prompted question.

Test Plan:
----------

I don't find any resources on how to test the init command locally. I tried to install globally the forked cli but it fails. 
Btw, I just remove the `ora` loader only for the post init script and do not add code to the base code so it's really simple and seems to work without extra testing.
But, if you tell me how to do that I'll be happy to test it !

### DIFF
--- a/docs/init.md
+++ b/docs/init.md
@@ -91,4 +91,21 @@ module.exports = {
 };
 ```
 
+## Breaking change >6.3.1
+Now the responsibility of showing the user progress of the "Executing post init script" goes to the implementor. In the cli, the `ora` package is used to display progress. 
+For a simple usage in a custom template, `ora` can be used like this in a postInitScript :
+
+```javascript
+#!/usr/bin/env node
+const ora = require('ora');
+
+new Promise((resolve) => {
+  const spinner = ora('Executing post init script ').start();
+  // do something
+  resolve();
+}).then(() => {
+  spinner.succeed();
+});
+```
+
 You can find example custom template [here](https://github.com/Esemesek/react-native-new-template).
--- a/docs/init.md
+++ b/docs/init.md
@@ -99,12 +99,17 @@ For a simple usage in a custom template, `ora` can be used like this in a postIn
 #!/usr/bin/env node
 const ora = require('ora');
 
+const spinner = ora('Executing post init script ');
+
 new Promise((resolve) => {
-  const spinner = ora('Executing post init script ').start();
+  spinner.start();
   // do something
   resolve();
 }).then(() => {
   spinner.succeed();
+}).catch(() => {
+  spinner.fail();
+  throw new Error('Something went wrong during the post init script execution');
 });
 ```
 
--- a/docs/init.md
+++ b/docs/init.md
@@ -91,7 +91,7 @@ module.exports = {
 };
 ```
 
-## 6.3.1
+## Post init script loading
 The responsibility of showing the user progress of the "Executing post init script" goes to the implementor. In the cli, the `ora` package is used to display progress. 
 For a simple usage in a custom template, `ora` can be used like this in a postInitScript :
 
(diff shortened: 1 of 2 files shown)


grabbou:
Thank you for the PR. I do wonder whether this breaking change will have a positive effect on the overall UX. I am not sure how many `post-init` scripts are out there, but I would imagine this being rather hard to update them all.

At the same time, it would mean an unnecessary `ora` dependency for most packages, where all is needed is a simple "loading..." progress.

I actually think that it's a good design that CLI displays the loader while executing a script that it controls. 

I wonder whether a better solution would be to pass `loader` as an additional (and last) argument to a function and for those that need a bit of additional configuration, they could do the work on their own.

Shaddix:
@grabbou we discussed it in the linked issue (#1492), and considered it impossible to pass loader to a post-init script, since currently it's invoked via `execa(scriptPath, {stdio: 'inherit'});` (at least, impossible while keeping backward compatibility).

Currently there's a major blocker for postInit scripts that they basically can't ask for any user input (loader prevents that).
I came to the aforementioned issue while creating (actually a very basic) custom react-native template for internal company projects. Almost anything I thought of adding required some user input (I thought about e.g. autogenerating Android keystores, setting up Sentry integration and Logging integrations).

My initial proposal was to change the `loader.start('Executing post init script ');` into `loader.info('Executing post init script');`, which I guess for most of current postInit scripts would be enough (I assume that all current scripts are relatively fast). And if it's not enough, post init scripts authors could use their own loader if they wish.

JeremyDolle:
I can change the PR depending on the choice to take

Angelk90:
@grabbou : Are there any news?

grabbou:
Thanks for additional context.

I guess

> My initial proposal was to change the loader.start('Executing post init script '); into loader.info('Executing post init script');

Makes a lot of sense. Any reason why you didn't go that way in the beginning? I think this would be much better and completely non-breaking.

JeremyDolle:
I think this is also the best choice. We don't discuss about this solution on the [issue](https://github.com/react-native-community/cli/issues/1492). I will make the necessary change into the PR.

docs/init.md
```diff
@@ -91,4 +91,26 @@ module.exports = {
 };
 ```
 
+## Breaking change >6.3.1
```
vnovick, line 94: I suggest not calling it a breaking change as it won't break things, but rather documenting this in the following manner:

**>6.3.1**  the responsibility of showing the user progress of the "Executing post init script" goes to the implementor instead of being handled by cli. the `ora` package is used to display progress.

vnovick (changes requested):
Thanks for the PR. I suggest not calling this a breaking change as it won't break things, but rather documenting this in the following manner

docs/init.md
vnovick, line 94: I suggest not calling it a breaking change, but rather documenting this in the following manner:

>6.3.1 the responsibility of showing the user progress of the "Executing post init script" goes to the implementor instead of being handled by cli. the ora package is used to display progress.

I wonder if it not showing a loader is considered a breaking change.

thymikee: We don't consider it a breaking change, let's remove this reference :)

Angelk90:
@grabbou : Are there any news?

JeremyDolle:
@grabbou @thymikee @vnovick I made the necessary changes. Let me know if I have to change or improve something 😄

thymikee (approved):
Thanks!

PR #1537, opened 2022-01-27
refactor(breaking): remove deprecated `link`, `unlink` and associated code

grabbou:
Summary:
---------

This PR removes the code related to `link`, `unlink` and `rnpm` and is a first shot at cleaning up the codebase. It is not going to clean up all the places automatically, but it's a first step.

List of changes:

- Removed `link`, `unlink`, `install` and `uninstall` commands from the CLI
- Removed all entries regarding `rnpm`, together with migration guides
- Removed a lot of unnecessary and extraneous properties from `IOSProjectConfig`, `AndroidProjectConfig` that were used for `link`, and not utilised in a new autolinking mechanism.  **That means things such as `assets` or `hooks` are no longer allowed to be present on a configuration and will cause an error when CLI validates the structure of your config**.
- Removed a lot of unnecessary properties from `IOSProjectParams`, `AndroidProjectParams`, `IOSDependencyParams` and `AndroidDependencyParams` that were used to override settings for `link` that are no longer needed.
- Removed ability to set a custom `podspecPath` for an iOS dependency. This was already marked as deprecated and set to be removed
- Removed `--project-path` option from a `run-ios` CLI command. This was meant to set a source directory for iOS files (unless it equals to `iOS`). It is now using the detection from `config`, which makes sure that `run-ios` uses the same folder that autolinking and other commands (very important for debugging)
- Changed iOS source directory detection from looking for an Xcode project to looking for a Podfile. Historically, we used to look for an `xcodeproj` or `xcworkspace` because that is what `link` needed to work and manipulate. The new autolinking relies on a Podfile only. There are certain use cases where Podfile is the only file present in a project (e.g. Xcode files are generated programatically or abstracted away). We still keep an information about `xcodeProject` we detected, as it is used for commands such as `upgrade`. I believe this may change in the future, so I wouldn't recommend relying too much on it.

We will explain the differences in types/configuration later in a form of a changelog/blog post. Right now, if you want to learn more about the changes, simply check the `diff` for `cli-types` package or a documentation.

Todo in a follow-up PRs:
------
- Add tests for `getProjectConfig` on iOS. They are missing and its important to cover them
- Remove deprecated properties in `run-android` command. It is the right time to do it now
- Potential code sharing opportunities as per https://github.com/react-native-community/cli/pull/1537#pullrequestreview-871581196


grabbou:
Except for the `docs`, this PR is ready for a review/comments. I will draft a description of changes by tomorrow.

packages/platform-ios/src/config/findPodfilePath.ts
```diff
@@ -9,13 +9,43 @@
...
+    /**
+     * Podfile from `ios` folder will be picked up as a first one.
+     */
+    .sort((project) => (path.dirname(project) === IOS_BASE ? -1 : 1));
+
   if (podfiles.length > 0) {
-    return path.join(folder, podfiles[0]);
+    /**
+     * @todo
+     * Shall we print a warning when multiple `Podfile` were found?
```
tido64, line 46: I ran this locally and got:

```
> require("glob").sync('**/Podfile', { ignore: ['**/@(Pods|node_modules|Carthage)/**'] })
[ 'ios/Podfile', 'macos/Podfile' ]
```

I think printing a warning can be very helpful, but there can be false positives if you also have a macOS project. Maybe could print the list if `--verbose` is specified?

packages/platform-ios/src/config/index.ts
```diff
@@ -5,86 +5,67 @@
...
-  const podfile = findPodfilePath(folder, projectPath);
+  const sourceDir = path.dirname(podfile);
 
-  // This is a temporary fix for #1435. In certain repos, the Xcode project can
-  // be generated by a tool. The only file that we can assume to exist on disk
-  // is `Podfile`.
-  const sourceDir = podfile ? path.dirname(podfile) : path.dirname(projectPath);
+  /**
+   * @todo
+   * What if there are multiple `xcodeproj` found? Shall we read from a Podfile?
```
tido64, line 46: Would it make sense to find the `.xcworkspace` instead? There should only be one of it (generated from running `pod install`).

tido64:
image

This looks great 😍

packages/platform-ios/src/config/index.ts
grabbou: That function actually looks for `xcworkspace` in first place and returns `isWorkspace: true`. I guess in scenario when there's no `xcworkspace` but multiple projects, we can print a warning in verbose mode, similarly to `findPodfile`, asking users to look at custom config?

packages/platform-ios/src/config/findPodfilePath.ts
grabbou: I like this idea.

grabbou: We will favour `iOS` anyway since this is `iOSConfig` used for `run-ios`. But I guess we can adapt it in a later PR to work for `Mac` too?

Can you point me to a configuration/helpers that you use for `macos`?

packages/platform-ios/src/config/index.ts
tido64: Yeah, that sounds good to me.

packages/platform-ios/src/config/findPodfilePath.ts
tido64: `findXcodeProject` is here: https://github.com/microsoft/react-native-macos/blob/main/local-cli/runMacOS/findXcodeProject.js

It only gets called from `run-macos` ([code](https://github.com/microsoft/react-native-macos/blob/main/local-cli/runMacOS/runMacOS.js)), so it can assume that `pod install` has happened. You wouldn't be able to use it for `config`.

grabbou: BTW, do you want me to export you this helper so you can reduce some code overlap? Or prefer to keep it internal?

tido64: @HeyImChris, @Saadnajmi: What do you think?

Saadnajmi: What's the exact question? What should we do if the CLI finds two podfiles (One for iOS and one for macOS)? I think that's a valid case (though the macOS podfile will have both an iOS and a macOS target in it I believe).

tido64: I think @grabbou was asking if it makes sense for react-native-macos to reuse this function instead of the existing `findXcodeProject`. If so, he will make it public.

HeyImChris: What makes searching for podfiles better than searching for Xcode projects? They'll likely be paired 1:1 in most cases

grabbou: Question was whether you would like us to export `findXcodeProject` function from the `platform-ios` package, so you don't have to re-implement it yourself?

Saadnajmi: Ah. If we do that, then `react-native-macos` takes a dependency on `@react-native-community/cli`? My initial thought is we wouldn't want that... but code sharing sounds good. 
Dumb question, does `react-native` currently take any dependency on `@react-native-community/cli`?

Saadnajmi: @HeyImChris I believe we search for podfiles rather than Xcode projects to support scenarios like react-native-test-app, which generates an xcode project given a podfile. In effect, the source of truth becomes the podfile and not the xcode project. I remember when @tido64 pushed for this change in the CLI which in turn helped us use react-native-test-app for FluentUI React Native.

grabbou: @Saadnajmi there's a sub-package `@react-native-community/platform-ios`. React Native depends on it. CLI doesn't. CLI detects runtime packages with platforms (such as `platform-ios`, or `react-native-windows`) and adds commands / configurations based on installed packages.

So by depending on `platform-ios`, you wouldn't pull the entire CLI.

grabbou:
Documentation updated via https://github.com/react-native-community/cli/pull/1537/commits/00ccd36d16ead849ee109b9e31c797bb494d5cf2, there were a few mistakes in the documentation itself, which I have fixed/updated as I was going through them anyway.

packages/platform-ios/src/config/findPodfilePath.ts
HeyImChris: I think this change makes sense given that we wouldn't be taking on a bunch of new dependencies with the obvious value of reducing our diffs. @Saadnajmi  sound good to you?

Saadnajmi: As it turns out, we already have a [dependency on the CLI](https://github.com/microsoft/react-native-macos/blob/26a7c3e9d430ebb58b9c9e2aa220fe9019ae1e1f/package.json#L97) (thanks @HeyImChris for pointing it out) so yeah this would be great! Yay code sharing!

grabbou: Right -> https://github.com/microsoft/react-native-macos/blob/26a7c3e9d430ebb58b9c9e2aa220fe9019ae1e1f/package.json#L99.


I will open a follow-up issue to explore various code-sharing opportunities.

packages/cli-types/src/index.ts
```diff
@@ -73,94 +73,59 @@ interface PlatformConfig<
...
   // @todo this should be removed: https://github.com/react-native-community/cli/issues/1261
   healthChecks: [];
 }
 
-/**
- * Shares some structure with Config, except that root is calculated and can't
- * be defined
- */
-
+// @todo double check the following two types for correction
```
thymikee, line 128: is this actionable for now or later?

packages/platform-android/src/config/__tests__/getProjectConfig.test.ts
```diff
@@ -56,40 +56,25 @@ describe('android::getProjectConfig', () => {
...
-
       const config = getProjectConfig(folder, userConfig);
-      expect(config).toMatchObject({
-        sourceDir: '/multiple/android',
-        appName: '',
-        packageName: 'com.some.example',
-        manifestPath: '/multiple/android/src/main/AndroidManifest.xml',
-      });
+
+      expect(config).toMatchSnapshot();
```
thymikee, line 77: `toMatchInlineSnapshot` would make perfect sense here in terms of readability

thymikee (approved):
Looks good! The change scope is pretty significant so let's release this in an alpha and test in the wild. The biggest breaking change to the typical workflow that I see is the removal of `assets` config and `link` command that links those assets. Plus it breaks libraries like `react-native-vector-icons`

Maybe there's a way to reintroduce `assets` with autolinking on iOS and Android? I'm not that attached to this feature but it sure saved me some time. 

Anyway, this is moving us forward and unlocking further improvements, so I'll give it a go

grabbou:
> Plus it breaks libraries like react-native-vector-icons

Actually, I am not sure about that - they do have a manual instructions for linking libraries and my impression is that they don't support officially auto-linking just yet. 

To be honest, I would be happy to welcome this as a 3rd party command, but let's see what's the demand and feedback after the release.

packages/cli-types/src/index.ts
grabbou: This has been solved in the meantime.

mikehardy:
First feedback coming in from react-native-vector-icons user, and it's basically, "Okay so now what do I do?"

https://github.com/oblador/react-native-vector-icons/issues/1416#issuecomment-1146154143

Issue #1527, opened 2022-01-17
Plan for CLI in Q1 - umbrella issue


grabbou:
Hey,

Together with the CLI team, we have been discussing the next steps as far as the development. The overall conclusion was that before we start discussing about the feature, we want to focus on cleaning up the codebase from the legacy features, so that it is easier for everyone to get involved in the library development.

> If you would like to help with the following work at any stage, either as a contributor or just a reviewer/advisor, please let me know! You can do that by writing a comment in this issue or contacting me directly on RNC Discord.

With that in mind, I have prepared the following action plan, which is divided into weeks:

- [x]  1. Review and merge outstanding PRs [17.01 - 21.01] https://github.com/react-native-community/cli/pulls

Before starting a complex rewrite, it would be great to take a final look at the outstanding PRs and merge/resolve as much of them as possible to avoid unnecessary rebasing later on. It is going to be a great way to refresh the knowledge of the codebase. 

- [x] 2. Re-apply work from `next` [24.01 - 28.01]

Previously, `link` has already been removed under `next` branch and started a wider refactor of the CLI. Unfortunately, the work was 70% done and never completed. The branch is now impossible to merge due to complex conflicts. It will be easier to re-apply it manually.

The mistake was that each PR was "self-contained" and was ready to ship. Instead of making a new branch and repeating the same mistakes, I am going to send the PRs directly against our main branch.

The main PR is https://github.com/react-native-community/cli/pull/790, which removes `link` (along few other commands) as well as cleans up the configuration itself. This PR is going to clean up a larger part of the codebase.

Smaller PRs that were submitted (status to be verified, they might have been done in a different form already):
- https://github.com/react-native-community/cli/pull/811
- https://github.com/react-native-community/cli/pull/812
- https://github.com/react-native-community/cli/pull/843
- https://github.com/react-native-community/cli/pull/859

While this is not going to complete the "clean-up" entirely, it is going to set a foundation for further work.

- [x] 3. Release a beta version with `link` removed (and codebase cleaned-up) [31.01 - 04.02]

Once all PRs are done, we're going to release a beta version of the new CLI, without the aforementioned pieces. In the optimistic scenario, this is going to be a transparent change, since most of the things were already deprecated. Unfortunately, it may turn out that certain things were removed (or are going to be removed without prior deprecation warning - as long as we have an agreement in a PR to do so). In that case, we will annotate such a breaking change and bump version number accordingly.

[...]

PR #1539, opened 2022-01-28
update: bump version of Metro packages to 67

kelset:
Summary:
---------

This is a PR to simply bump the Metro dependencies to match the new version of Metro that has been released [0.67](https://github.com/facebook/metro/releases/tag/v0.67.0).

Most likely, this will mean that a new major of the package needs to be released too? I'm not sure how this bump will fit in the bigger picture work that is happening in the repo (see https://github.com/react-native-community/cli/issues/1527).

Test Plan:
----------

Locally, I did a quick `yarn test` and a few were failing - but wanted to double check against CI to verify if it's simply something going wrong locally.

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "jest-circus": "^26.6.2",
     "jest-snapshot-serializer-raw": "^1.1.0",
     "lerna": "^4.0.0",
-    "metro-memory-fs": "^0.66.1",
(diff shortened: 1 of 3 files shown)


thymikee:
@grabbou this shouldn't conflict with your work, right?

grabbou:
Except for `yarn.lock`, not really! I'll take care of conflicts later

PR #1540, opened 2022-02-01
chore: remove global-cli

grabbou:
Summary:
---------

It's deprecated and is causing us issues with tools since it resides in `packages` folder alongside other monorepo packages. If necessary, we can move it to a separate package and explicitly archive. So far, I think it can be safely removed, we have legacy branches (such as 6.x) where we can refer to it.


packages/platform-android/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@react-native-community/cli-platform-android",
-  "version": "7.0.1",
+  "version": "7.00.1",
```
thymikee, line 3: um?

thymikee (commented):
Yay! Next step: remove legacyInit implementation

packages/platform-android/package.json
grabbou: lol mistake

PR #1544, opened 2022-02-02
chore(breaking): remove `initCompat` a.k.a legacy "init"

grabbou:
Summary:
---------

This is a breaking change as it changes the public interface. However, it was meant to be used only by the legacy/global `CLI` which is already deprecated and no longer updated.


packages/platform-android/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@react-native-community/cli-platform-android",
-  "version": "7.0.1",
+  "version": "7.00.1",
```
thymikee, line 3: I guess this is not rebased to main?

thymikee (changes requested):
Let's remove the legacyInit e2e test as well :)

packages/platform-android/package.json
grabbou: Haha, again 🤦

Issue #1485, opened 2021-10-18
Support checking against CalVer for dependencies in the `doctor` command


tywayne:
## Describe the Feature
The `doctor` tool only supports version checking with SemVer. Watchman has switched from using SemVer to CalVer in their versioning, so it always shows as failing even when it is correctly installed to the latest version. 

It would be great if the `doctor` command would support checking against both SemVer and CalVer so that Watchman would report correctly.

## Possible Implementations
Update the `doesSoftwareNeedToBeFixed` function to support checking against SemVer as well as CalVer versioning systems.

thymikee:
Thanks for noting that! Makes sense to update the logic for Watchman then 👍🏼 cc @adamTrz

adamTrz:
Closing the issue since Watchman version check was resolved with #1700

PR #1550, opened 2022-02-10
fix(doctor, watchman): accept new date-based watchman versions

mikehardy:
Summary:
---------

fix(doctor, watchman): accept new date-based watchman versions

doctor was giving false-negatives with the new watchman versions

they use date-based versions now that violate semver with leading-zeros in numeric parts:
- https://github.com/npm/node-semver/issues/232
    
loose mode correctly pasrses them though, so add optional parameter to install check that
allows loose version range coercion

Should be compatible as watchman project is on record as being permanently backwards-compatible:
https://facebook.github.io/watchman/docs/compatibility.html

Test Plan:
----------

I wrote a test that covers watchman completely before modifying it, then modified it and added a chunk of test to check the modification - so, best reviewed one commit at a time

### DIFF
--- a/packages/cli-doctor/src/tools/checkInstallation.ts
+++ b/packages/cli-doctor/src/tools/checkInstallation.ts
@@ -19,11 +19,13 @@ const isSoftwareNotInstalled = async (command: string): Promise<boolean> => {
 const doesSoftwareNeedToBeFixed = ({
   version,
   versionRange,
+  looseRange = false,
 }: {
   version: string;
   versionRange: string;
+  looseRange?: boolean;
 }): boolean => {
-  const coercedVersion = semver.coerce(version);
+  const coercedVersion = semver.coerce(version, {loose: looseRange});
 
   return (
     version === 'Not Found' ||
--- a/packages/cli-doctor/src/tools/versionRanges.ts
+++ b/packages/cli-doctor/src/tools/versionRanges.ts
@@ -3,7 +3,7 @@ export default {
   NODE_JS: '>= 8.3',
   YARN: '>= 1.10.x',
   NPM: '>= 4.x',
-  WATCHMAN: '4.x',
+  WATCHMAN: '>= 4.x',
   JAVA: '1.8.x || >= 9',
   // Android
   ANDROID_SDK: '>= 26.x',
--- a/packages/cli-doctor/src/tools/healthchecks/__tests__/watchman.test.ts
+++ b/packages/cli-doctor/src/tools/healthchecks/__tests__/watchman.test.ts
@@ -1,0 +1,100 @@
+import execa from 'execa';
+import watchman from '../watchman';
+import getEnvironmentInfo from '../../envinfo';
+import {EnvironmentInfo} from '../../../types';
+import {NoopLoader} from '@react-native-community/cli-tools';
+import * as common from '../common';
+import * as brewInstall from '../../brewInstall';
+
+jest.mock('execa', () => jest.fn());
+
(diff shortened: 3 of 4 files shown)


packages/cli-doctor/src/tools/checkInstallation.ts
```diff
@@ -19,11 +19,13 @@ const isSoftwareNotInstalled = async (command: string): Promise<boolean> => {
 const doesSoftwareNeedToBeFixed = ({
   version,
   versionRange,
+  looseRange = false,
```
mikehardy, line 22: default value so no other code has to change, and follows the semver package historical precedent of all semver APIs taking loose as a final argument, with false being the default

packages/cli-doctor/src/tools/versionRanges.ts
```diff
@@ -3,7 +3,7 @@ export default {
   NODE_JS: '>= 8.3',
   YARN: '>= 1.10.x',
   NPM: '>= 4.x',
-  WATCHMAN: '4.x',
+  WATCHMAN: '>= 4.x',
```
mikehardy, line 6: 4.9.0 was released something like 2 years ago then they transitioned to date ranges that have leading series in the semver parts, which is technically a semver violation but still easily parsed. This let's the range slip up to the new versions, which do work

liamjones:
FYI this resolves #1485 - sort of... It doesn't directly implement support for [CalVer](https://calver.org/) but negates the need to do that.

lucasbento (approved):

packages/cli-doctor/src/tools/checkInstallation.ts
```diff
@@ -19,11 +19,13 @@ const isSoftwareNotInstalled = async (command: string): Promise<boolean> => {
...
 }: {
   version: string;
   versionRange: string;
+  looseRange?: boolean;
 }): boolean => {
-  const coercedVersion = semver.coerce(version);
+  const coercedVersion = semver.coerce(version, {loose: looseRange});
```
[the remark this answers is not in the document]
mikehardy: I'm not sure I fully understand the question so let me state what I think I understand, then please correct if I'm wrong - I *think* that if you compare a semver vs a "semver except it has leading zeros in some parts because it is YYYY.MM.DD.ZZ.etc" it will coerce and compare and work as is. The coercion takes the leading three parts and the loose allows leading zeros to still go through and be treated as integers during coercion

Which is...sort of some detail and not an answer. I was worried personally about the impact here which is why I went with an optional parameter you can opt-in too when you know a version series "style" (that is: it has changed to date, and it has changed in a way that it's previous semver-compatible will work with it's current date form) and thus can make more assumptions / tests about correct behavior vs hoping that the full loose backus-naur form works well with the semver BNF

Hopefully that's what you were curious about?

grabbou:
> #1485

Well, I guess it sort of fixes that issue. Alternative would be to enable looseMode for all the version checks. I think at this point, since we have support for CalVer, we are good to close that issue and just keep support in mind when other dependencies switch to it in the future.

What you think?

grabbou:
@mikehardy great work and thanks for adding tests. My question is - where I can look up details on what exactly is loose mode doing that makes this code work?

grabbou (approved):
LGTM!

mikehardy:
> @mikehardy great work and thanks for adding tests. My question is - where I can look up details on what exactly is loose mode doing that makes this code work?

It appears undocumented, I had to read the regexes in source 😂

Issue #1470, opened 2021-09-15
Init: Custom replacements in template / allow Hermes flag


lunaleaps:
## Describe the Feature
Want to be able to do something like:
```bash
npx react-native init MYPROJECT --replace HERMES_PLACEHOLDER:true
```
or if that's too involved, can we just add a hermes flag? 

The use-case for this is helping to improve the release tester script. Right now to test Hermes on the template app, it's a manual change to the Podfile / build.gradle

## Possible Implementations

* We add something to `template.config.js` to allow specification of placeholders it supports:
```js
module.exports = {
  placeholderName: 'HelloWorld',
  titlePlaceholder: 'Hello App Display Name',
  templateDir: './template',
  placeholders: [
     'HERMES_PLACEHOLDER',
  ], // or maybe an object if we need more information
};
```
Then we can validate the placeholders against placeholder values passed in via `cli` and just do string replace -- similar to how we're doing `placeholderName`, `placeholderTitle`: [code pointer](https://github.com/react-native-community/cli/blob/641b21f583c97e3d48ce87d5fe804f42db92fa5c/packages/cli/src/commands/init/init.ts#L117)

## Related Issues
*

thymikee:
I'd be supportive of this change. Looks like `placeholders` keys would need default values? We could make placeholders an object, as noted and even move `placeholderName` and `placeholderTitle` there:
```js
module.exports = {
  templateDir: './template',
  placeholders: {
    name: 'HelloWorld',
    title: 'Hello App Display Name',
    HERMES_PLACEHOLDER: true,
  }
};
```
And support current and new syntaxes for backwards compatibility. This way we could replace any placeholder trough the CLI:
```
init PROJECT --title "Hermes project" -- HERMES_PLACEHOLDER true
```

lunaleaps:
@thymikee Re: defaults, yes that makes sense! 

Oh so we'd make each placeholder a top level option flag?

thymikee:
Possibly, but I'm not sure if it's possible to make it work with commanderjs we use to parse input 😅

grabbou:
I like this idea. On the other hand, I wonder if placeholder is a good solution to conditionally change the template? It always felt to me it's a value to replace, like name, instead of something that goes into a flag/conditional.

Alternatively, something that can be done already today, would be to have a separate template with Hermes enabled? Going forward, we could expand this to another template, with CodeGen enabled, to test a few different scenarios. 

[...]

PR #1556, opened 2022-02-13
feat: add support for custom placeholder template

numandev1:
Summary:
---------
#1470 
we need to add the custom placeholder flags while init. so i added this custom placeholder support in cli

Test Plan:
----------
you can download this template and test by this command
https://github.com/nomi9995/react-native-template-placeholder
```
react-native init test --template file:///Users/apple/clitest/react-native-template-placeholder 
```

### DIFF
--- a/docs/init.md
+++ b/docs/init.md
@@ -76,19 +76,25 @@ Every custom template needs to have configuration file called `template.config.j
 
 ```js
 module.exports = {
-  // Placeholder name that will be replaced in package.json, index.json, android/, ios/ for a project name.
-  placeholderName: 'ProjectName',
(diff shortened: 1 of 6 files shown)


nomi9995:
@thymikee sorry, i forgot to add template link for test https://github.com/nomi9995/react-native-template-placeholder

gabrieldonadel:
@nomi9995 seems that tests are failing, do you mind taking a look?

nomi9995:
@gabrieldonadel this is an architectural change so we have to update test cases if it is functionality-wise okay then i will also write test cases.

PR #1560, opened 2022-02-18
Fix Web Debugging and Other Websocket Endpoints with Metro 0.67

NickGerleman:
Summary:
---------

Metro 0.67 included https://github.com/facebook/metro/commit/38a200e748c2e22eb34ce628d51bbdf563dbe777 which changed the way external applications should interface with it to add new WebSocket endpoints. This materializes as failures when web debugging RN 0.68 apps.

This change applies the recommended fix, of passing a list of non-server endpoints to Metro when starting the server. `ws` is updated as part of the change, along with removing type-checker suppressions, of which one was resulting in a runtime error.

Test Plan:
----------

Tested web debugging after replacing the relevant CLI packages in `node_modules`, with the edited ones.

### DIFF
--- a/packages/cli-plugin-metro/src/commands/start/runServer.ts
+++ b/packages/cli-plugin-metro/src/commands/start/runServer.ts
@@ -70,7 +70,12 @@ async function runServer(_argv: Array<string>, ctx: Config, args: Args) {
     );
   }
 
-  const {middleware, attachToServer} = createDevServerMiddleware({
+  const {
+    middleware,
+    websocketEndpoints,
+    messageSocketEndpoint,
+    eventsSocketEndpoint,
+  } = createDevServerMiddleware({
     host: args.host,
     port: metroConfig.server.port,
     watchFolders: metroConfig.watchFolders,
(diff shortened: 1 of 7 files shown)


NickGerleman:
This fixes web debugging, but the PR in the current state has added constant warnings logged to console, that I want to take a look at before this is merged. 
image

NickGerleman:
The next time I launched it, I didn't see the warnings, but the debugger count in the UI was way off. 
image

NickGerleman:
That number seems to not actually be controlled by us, looking at `packages\debugger-ui\src\ui\index.js`. Going to just remove it from the UI, if it isn't always stable.

NickGerleman:
> This fixes web debugging, but the PR in the current state has added constant warnings logged to console, that I want to take a look at before this is merged.

I tried running Metro, then web debugging, a few times, even after resetting the cache, and haven't seen the issue. Since. Not sure if I had some stale sate, or if there's a real issue here...

thymikee (approved):
Thanks, this is great <3! I've moved the creation of the sockets and debugger proxy up and removed the stubs for `isDebuggerConnected` and `broadcast`, since now they're eagerly and always (I think?) created. 

Tested various cases, disconnecting the debugger and app and server many times, and it seem to work as expected every time. Going to merge this in and release asap

grabbou:
Sorry for being late here, but why https://github.com/react-native-community/cli/commit/e4b7805a92176c5c355bdb09c568b9b7b0e30acb#diff-419c2d522450d9ec89f017caa5152add5b0fa81d97b476fe461404aafad5e1dfL52 was removed? Any particular reason?

NickGerleman:
> The next time I launched it, I didn't see the warnings, but the debugger count in the UI was way off. image

> That number seems to not actually be controlled by us, looking at `packages\debugger-ui\src\ui\index.js`. Going to just remove it from the UI, if it isn't always stable, and everything else seems to work.

The number seemed to correspond to an opaque ID given to the debugger UI by Metro. After the Metro update and networking changes, the number would fluctuate.

PR #1582, opened 2022-04-01
feat: add "clean" command

tido64:
Summary:
---------

This adds a `clean` command to the CLI for common cleaning operations.

Originally contributed by @dennismunene: https://github.com/microsoft/rnx-kit/pull/993

Test Plan:
----------

```
yarn link @react-native-community/cli @react-native-community/cli-clean
```

Test `--help`:

```
% npx react-native clean --help
react-native clean

Cleans your project by removing React Native related caches and modules.

Options:
  --include <string>       Comma-separated flag of caches to clear e.g. `npm,yarn`. If omitted, an interactive prompt will appear.
  --project-root <string>  Root path to your React Native project. When not specified, defaults to current working directory. (default: "/~/test-app")
  --verify-cache           Whether to verify the cache. Currently only applies to npm cache.
  -h, --help               output usage information
```

By default, prompts the user for caches to remove:

```
% yarn react-native clean
? Select all caches to clean ›
Instructions:
    ↑/↓: Highlight option
    ←/→/[space]: Toggle selection
    a: Toggle all
    enter/return: Complete answer
◯   android (Android build caches, e.g. Gradle)
◯   cocoapods (CocoaPods cache)
◉   metro (Metro, haste-map caches)
◯   npm (`node_modules` folder in the current package, and optionally verify npm cache)
◉   watchman (Stop Watchman and delete its cache)
◯   yarn (Yarn cache)
```

Remove `node_modules`:

```
% npx react-native clean --include npm
✔ Remove node_modules
```

Clean Gradle cache and `node_modules`:

```
% npx react-native clean --include android,npm
✔ Clean Gradle cache
✔ Remove node_modules
```

### DIFF
--- a/packages/cli-clean/src/clean.ts
+++ b/packages/cli-clean/src/clean.ts
@@ -4,10 +4,11 @@ import {spawn} from 'child_process';
 import {existsSync as fileExists, rmdir} from 'fs';
 import os from 'os';
 import path from 'path';
+import prompts from 'prompts';
 import {promisify} from 'util';
 
 type Args = {
-  include: string;
+  include?: string;
   projectRoot: string;
   verifyCache?: boolean;
 };
@@ -69,6 +70,24 @@ function findPath(startPath: string, files: string[]): string | undefined {
   }
 
   return undefined;
+}
+
+async function promptForCaches(
+  commands: CLICommand,
+): Promise<string[] | undefined> {
+  const names: string[] = Object.keys(commands);
+  const {caches} = await prompts({
+    type: 'multiselect',
+    name: 'caches',
+    message: 'Select all caches to clean',
+    choices: names.map((cmd) => ({
+      title: cmd,
+      value: cmd,
+      selected: DEFAULT_CATEGORIES.includes(cmd),
+    })),
+    min: 1,
+  });
+  return caches;
 }
 
 export async function clean(
@@ -164,8 +183,15 @@ export async function clean(
     ],
   };
 
+  const categories = include
+    ? include.split(',')
+    : await promptForCaches(COMMANDS);
+  if (!categories || categories.length === 0) {
+    return;
+  }
+
   const spinner = getLoader();
-  for (const category of include.split(',')) {
+  for (const category of categories) {
     const commands = COMMANDS[category];
     if (!commands) {
       spinner.warn(`Unknown category: ${category}`);
@@ -194,8 +220,7 @@ export default {
     {
       name: '--include <string>',
       description:
-        'Comma-separated flag of caches to clear e.g. `npm,yarn`. When not specified , only non-platform specific caches are cleared. Valid values are android, cocoapods, npm, metro, watchman, yarn.',
-      default: DEFAULT_CATEGORIES.join(','),
+        'Comma-separated flag of caches to clear e.g. `npm,yarn`. If omitted, an interactive prompt will appear.',
     },
     {
       name: '--project-root <string>',
--- a/packages/cli-clean/src/clean.ts
+++ b/packages/cli-clean/src/clean.ts
@@ -1,0 +1,213 @@
+import {getLoader} from '@react-native-community/cli-tools';
+import type {Config as CLIConfig} from '@react-native-community/cli-types';
+import {spawn} from 'child_process';
+import {existsSync as fileExists, rmdir} from 'fs';
+import os from 'os';
+import path from 'path';
+import {promisify} from 'util';
+
+type Args = {
+  include: string;
+  projectRoot: string;
+  verifyCache?: boolean;
+};
+
+type Task = {
+  label: string;
+  action: () => Promise<void>;
+};
+
+type CLICommand = {
+  [key: string]: Task[];
+};
+
+const DEFAULT_CATEGORIES = ['metro', 'npm', 'watchman', 'yarn'];
+
+const rmdirAsync = promisify(rmdir);
+
+function cleanDir(directory: string): Promise<void> {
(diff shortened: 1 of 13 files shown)


thymikee:
I'm looking at the `include` flag and wonder – can we make it interactive, with checkboxes to tick, and provide --no-interactive mode so it works the same as now?

tido64:
> I'm looking at the `include` flag and wonder – can we make it interactive, with checkboxes to tick, and provide --no-interactive mode so it works the same as now?

I made it prompt if you omit `--include`. Let me know what you think.

thymikee (approved):
This works really well! The only thing I'm missing is tests. Would you mind providing some?

packages/cli-clean/src/clean.ts
```diff
@@ -0,0 +1,238 @@
...
+  _config: CLIConfig,
+  cleanOptions: Args,
+): Promise<void> {
+  const {include, projectRoot, verifyCache} = cleanOptions;
+  if (!fileExists(projectRoot)) {
+    throw new Error(`Invalid path provided! ${projectRoot}`);
+  }
+
+  const npm = os.platform() === 'win32' ? 'npm.cmd' : 'npm';
+  const yarn = os.platform() === 'win32' ? 'yarn.cmd' : 'yarn';
```
thymikee, line 104: In other packages, e.g. `cli-doctor` we use the `execa` package to handle this for us. Mind verifying it's ok and remove the platform checks? Hope this makes it unnecessary for the `execute` helper to exist

packages/cli-clean/src/clean.ts
```diff
@@ -0,0 +1,216 @@
...
+            os.platform() === 'win32' ? 'tskill' : 'killall',
+            ['watchman'],
+            {cwd: projectRoot},
+          );
+        },
+      },
+      {
+        label: 'Delete Watchman cache',
+        action: async () => {
+          await execa('watchman', ['watch-del-all'], {cwd: projectRoot});
```
thymikee, line 150: I wonder, shouldn't we call `watchman watch-del ${cwd}`? watch-del-all will clear all the caches for unrelated projects (e.g. Jest uses it), and usually it's about this specific cwd.

tido64: Another issue is that Watchman will freeze and use 100% CPU when you're switching between projects. I don't know what the cause is, but killing watchman and resetting the cache is the only thing that works.

packages/cli-clean/src/clean.ts
tido64: Testing this on Windows now. Will reply once complete.

tido64: Works on Windows. I also hid the CocoaPods option when on Windows and added hints:
```
% yarn react-native clean
? Select all caches to clean ›
Instructions:
    ↑/↓: Highlight option
    ←/→/[space]: Toggle selection
    a: Toggle all
    enter/return: Complete answer
◯   android (Android build caches, e.g. Gradle)
◯   cocoapods (CocoaPods cache)
◉   metro (Metro, haste-map caches)
◯   npm (`node_modules` folder in the current package, and optionally verify npm cache)
◉   watchman (Stop Watchman and delete its cache)
◯   yarn (Yarn cache)
```

packages/cli-clean/src/clean.ts
thymikee: oh, I didn't take monorepos into account. Let's keep it as is then

packages/cli-clean/src/clean.ts
```diff
@@ -0,0 +1,241 @@
...
+
+async function promptForCaches(
+  groups: CleanGroups,
+): Promise<string[] | undefined> {
+  const {caches} = await prompts({
+    type: 'multiselect',
+    name: 'caches',
+    message: 'Select all caches to clean',
+    choices: Object.entries(groups).map(([cmd, group]) => ({
+      title: `${cmd} (${group.description})`,
```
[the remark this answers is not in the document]
tido64: image

Yeah, that looks better.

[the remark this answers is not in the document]
tido64: And on Windows:
image

thymikee (approved):
Beauty!

packages/cli-clean/src/clean.ts
```diff
@@ -1,5 +1,6 @@
 import {getLoader} from '@react-native-community/cli-tools';
 import type {Config as CLIConfig} from '@react-native-community/cli-types';
+import chalk from 'chalk';
```
thymikee, line 3: we'll need to add `chalk` to package.json

tido64: Whoops 😄

packages/cli-clean/src/clean.ts
```diff
@@ -0,0 +1,242 @@
...
+            const candidates =
+              os.platform() === 'win32'
+                ? ['android/gradlew.bat', 'gradlew.bat']
+                : ['android/gradlew', 'gradlew'];
+            const gradlew = findPath(projectRoot, candidates);
+            if (gradlew) {
+              const script = path.basename(gradlew);
+              await execa(
+                os.platform() === 'win32' ? script : `./${script}`,
+                ['clean'],
```
cortinico, line 96: Small FYI: clean won't work for users on New Architecture due to a bug on AGP that is causing `clean` to depend on `preBuild`. I agree that `./gradlew clean` would be the correct approach here, but maybe we should not advertise it yet? Or have a `isNewArchitectureEnabled` flag inside the CLI to check this and run a `rm`?

PR #1585, opened 2022-04-06
[Feat] Add support for .xcode.env file in init and doctor commands

cipolleschi:
Summary:
---------
This PR adds the support for `.xcode.env` file in the `init` and `doctor` commands.
This PR depends on this other one: https://github.com/facebook/react-native/pull/33546

Test Plan:
----------
This PR is quite similar to 1530. I would rely on the same tests.
If anyone has suggestions about how to test this, I will be more than happy to implement them.

### DIFF
--- a/packages/cli-doctor/package.json
+++ b/packages/cli-doctor/package.json
@@ -10,6 +10,7 @@
   "dependencies": {
     "@react-native-community/cli-config": "^8.0.0-alpha.0",
     "@react-native-community/cli-tools": "^8.0.0-alpha.0",
+    "@react-native-community/cli-platform-ios": "^8.0.0-alpha.0",
     "chalk": "^4.1.2",
     "command-exists": "^1.2.8",
     "envinfo": "^7.7.2",
--- a/packages/cli-doctor/src/tools/healthchecks/xcodeEnv.ts
+++ b/packages/cli-doctor/src/tools/healthchecks/xcodeEnv.ts
@@ -1,0 +1,61 @@
+import {HealthCheckInterface} from '../../types';
+import fs from 'fs';
+import path from 'path';
+import {promisify} from 'util';
+import {findProjectRoot} from '@react-native-community/cli-tools';
+import {findPodfilePaths} from '@react-native-community/cli-platform-ios';
+
+const xcodeEnvFile = '.xcode.env';
+const pathSeparator = '/';
+
+function removeLastPathComponent(pathString: string): string {
+  const components = pathString.split(pathSeparator);
+  components.splice(components.length - 1, 1);
+  return components.join(pathSeparator);
+}
+
+function pathHasXcodeEnvFile(pathString: string): boolean {
+  const xcodeEnvPath = pathString + pathSeparator + xcodeEnvFile;
+  return fs.existsSync(xcodeEnvPath);
+}
+
+function pathDoesNotHaveXcodeEnvFile(pathString: string): boolean {
+  return !pathHasXcodeEnvFile(pathString);
+}
+
+export default {
+  label: '.xcode.env',
+  description: 'File to customize Xcode environment',
+  getDiagnostics: async () => {
+    const projectRoot = findProjectRoot();
+    const allPathsHasXcodeEnvFile = findPodfilePaths(projectRoot)
+      .map((pathString) => {
+        const basePath = removeLastPathComponent(pathString);
+        return pathHasXcodeEnvFile(basePath);
+      })
+      .reduce((previousValue, currentValue) => previousValue && currentValue);
+    return {
+      needsToBeFixed: !allPathsHasXcodeEnvFile,
+    };
+  },
+  runAutomaticFix: async () => {
+    const templateXcodeEnv = '_xcode.env';
+    const projectRoot = findProjectRoot();
+
+    const templateIosPath = path.dirname(
+      require.resolve('react-native/template/ios'),
+    );
+
+    const src = templateIosPath + templateXcodeEnv;
+    const copyFileAsync = promisify(fs.copyFile);
+
+    findPodfilePaths(projectRoot)
+      .map(removeLastPathComponent)
+      // avoid overriding existing .xcode.env
+      .filter(pathDoesNotHaveXcodeEnvFile)
+      .forEach(async (pathString) => {
+        const destFilePath = pathString + pathSeparator + xcodeEnvFile;
+        await copyFileAsync(src, destFilePath);
+      });
+  },
+} as HealthCheckInterface;
--- a/packages/cli-doctor/tsconfig.json
+++ b/packages/cli-doctor/tsconfig.json
@@ -7,6 +7,7 @@
   "references": [
     {"path": "../tools"},
     {"path": "../cli-types"},
-    {"path": "../cli-config"}
+    {"path": "../cli-config"},
+    {"path": "../platform-ios"}
   ]
 }
--- a/packages/cli-doctor/src/tools/healthchecks/index.ts
+++ b/packages/cli-doctor/src/tools/healthchecks/index.ts
@@ -11,6 +11,7 @@ import cocoaPods from './cocoaPods';
 import iosDeploy from './iosDeploy';
 import {Healthchecks, HealthCheckCategory} from '../../types';
 import loadConfig from '@react-native-community/cli-config';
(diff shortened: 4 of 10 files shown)


cipolleschi:
The CI is failing on two tasks related to Android (they are related to `gradlew`). I don't think that my PR touched those files, though... and I can see that the master branch is failing as well.

packages/cli-doctor/src/tools/healthchecks/xcodeEnv.ts
```diff
@@ -0,0 +1,21 @@
+import {HealthCheckInterface} from '../../types';
+import fs from 'fs';
+import {promisify} from 'util';
+
+export default {
+  label: '.xcode.env',
+  description: 'File to customize Xcode environment',
+  getDiagnostics: async () => {
+    return {
+      needsToBeFixed: fs.existsSync('./ios/.xcode.env'),
```
tido64, line 10: This is making assumptions on where the iOS project lives. Should we use the config to figure out where the closes `Podfile` is? And what about macOS?

cipolleschi: I updated it using `findProjectRoot`. I'm not sure that `findPodfilePath` is the right approach, because for macOS project we have an `ios` folder and a `macos` folder, both of which have their own `Podfile`.

```sh
$ ls -all ios
total 48
drwxr-xr-x  10 cipolleschi  staff    320  6 Apr 17:00 .
drwxr-xr-x  24 cipolleschi  staff    768  6 Apr 17:03 ..
-rw-r--r--   1 cipolleschi  staff   1106  6 Apr 16:59 Podfile
// ... other files here ...

$ ls -all macos
total 56
drwxr-xr-x  10 cipolleschi  staff    320  6 Apr 17:06 .
drwxr-xr-x  24 cipolleschi  staff    768  6 Apr 17:03 ..
-rw-r--r--   1 cipolleschi  staff   1086  6 Apr 17:03 Podfile
```

packages/cli-doctor/src/tools/healthchecks/xcodeEnv.ts
```diff
@@ -0,0 +1,63 @@
...
+    return fs.existsSync(platformPath + xcodeEnvFile);
+  }
+  return null;
+}
+
+export default {
+  label: '.xcode.env',
+  description: 'File to customize Xcode environment',
+  getDiagnostics: async () => {
+    const projectRoot = findProjectRoot();
```
tido64, line 39: Somewhat unrelated (and definitely not to be fixed in this PR), I looked briefly at the options that are passed to all CLI commands and it looks like they include this information. However, it doesn't look like they are properly forwarded to the individual fixers. It's a bit silly that we have to duplicate the effort to find this info again. Do we have an issue tracking this? cc @mikehardy @thymikee

packages/cli-doctor/src/tools/healthchecks/xcodeEnv.ts
```diff
@@ -0,0 +1,63 @@
...
+    };
+  },
+  runAutomaticFix: async () => {
+    const templateFilePath = '_xcode.env';
+    const projectRoot = findProjectRoot();
+
+    const src =
+      projectRoot +
+      '/node_modules/react-native/template/ios/' +
+      templateFilePath;
```
tido64, line 52: This is assuming the installation path of `react-native`. We should probably use `require.resolve` instead?

packages/cli-doctor/src/tools/healthchecks/xcodeEnv.ts
```diff
@@ -0,0 +1,63 @@
...
+    const copyFileAsync = promisify(fs.copyFile);
+
+    applePlatformFolders.forEach(async (folder) => {
+      const destFolder = projectRoot + folder;
+      const destFilePath = destFolder + '/' + xcodeEnvFile;
+      if (fs.existsSync(destFolder) && !fs.existsSync(destFilePath)) {
+        await copyFileAsync(src, destFilePath);
+      }
+    });
```
tido64, line 61: This is assuming the location of the `.xcworkspace`. I've seen many differing setups including:

- `<project-root>/ios/MyApp.xcworkspace`
- `<project-root>/MyApp.xcworkspace`
- `<project-root>/example/ios/MyApp.xcworkspace`
- `<project-root>/apps/ios/src/MyApp.xcworkspace`

Can we refactor `findPodfilePath` to return all found `Podfile`s and add a `.xcode.env` next to them? I think that will be more accurate.

packages/cli-doctor/src/tools/healthchecks/xcodeEnv.ts
thymikee: I don't think we track this. Would make sense

packages/cli-doctor/src/tools/healthchecks/xcodeEnv.ts
thymikee: Or introduce a `findAllPodfilePaths` helper

packages/cli-doctor/src/tools/healthchecks/xcodeEnv.ts
tido64: Opened #1588

packages/cli-doctor/src/tools/healthchecks/xcodeEnv.ts
cipolleschi: Hi! Thank you all for the suggestions and @tido64 for sharing its experience.
I updated the PR with the `findAllPodfilePaths` function, as suggested by @thymikee and adding the file in all the returned paths.

packages/cli-doctor/src/tools/healthchecks/xcodeEnv.ts
cipolleschi: I hope I used it properly. Let me know if I have to update the call.

packages/cli-doctor/src/tools/healthchecks/xcodeEnv.ts
```diff
@@ -0,0 +1,61 @@
+import {HealthCheckInterface} from '../../types';
+import fs from 'fs';
+import path from 'path';
+import {promisify} from 'util';
+import {findProjectRoot} from '@react-native-community/cli-tools';
+import findAllPodfilePaths from '@react-native-community/cli-platform-ios/src/config/findAllPodfilePaths';
```
thymikee, line 6: We can't import from `src` since we transform the code before publishing and put it into `build` directory; `src` is not there in published files. We either need to export `findAllPodfilePaths` from `@react-native-community/cli-platform-ios` – and add a dependency on it for `cli-doctor` and update `"references"` in its `tsconfig.json` – or move this helper to the `cli-tools` package (and add a dependency on `glob`). I'd go with the former for now.

thymikee (commented):
Almost there! One thing to address and we're good :)

packages/cli-doctor/src/tools/healthchecks/xcodeEnv.ts
thymikee: FYI we have a `yarn watch` script that builds files on the go

cipolleschi: Hi @thymikee, thanks for the review.
I updated the PR with your suggestion. Let me know if it's ok!

thymikee (approved):
LGTM 👍🏼

packages/cli-doctor/tsconfig.json
```diff
@@ -7,6 +7,7 @@
   "references": [
     {"path": "../tools"},
     {"path": "../cli-types"},
-    {"path": "../cli-config"}
+    {"path": "../cli-config"},
+    {"path": "../platform-ios"}
```
thymikee, line 11: let's update package.json as well :D

packages/cli-doctor/package.json
```diff
@@ -10,6 +10,7 @@
   "dependencies": {
     "@react-native-community/cli-config": "^8.0.0-alpha.0",
     "@react-native-community/cli-tools": "^8.0.0-alpha.0",
+    "@react-native-community/platform-ios": "^8.0.0-alpha.0",
```
thymikee, line 13:
```suggestion
    "@react-native-community/cli-platform-ios": "^8.0.0-alpha.0",
```

PR #1446, opened 2021-07-14
Init for current/existing folders

andriytsaryov:
Summary:
---------

This PR brings the ability to create new apps in existing and/or current folder.

Issue: https://github.com/react-native-community/cli/issues/1418.

TBD:
- [x] Linter
- [x] Fix existing tests
- [ ] Cover new cases with tests

Test Plan:
----------

Current empty folder:

1. Create an empty folder `mkdir TestApp1`.
2. Go to a newly created folder `cd TestApp1`.
3. Init a new app `react-native init TestApp1 --directory .`
4. **As a result a new app gets created in `TestApp1` folder.**

Non-current empty existing folder:

1. Create an empty folder `mkdir TestApp2`.
3. Init a new app `react-native init TestApp2 --directory TestApp2`
4. **As a result a new app gets created in `TestApp2` folder.**

Current folder with conflicting files:

1. Create an empty folder `mkdir TestApp3`.
2. Go to a newly created folder `cd TestApp3`.
3. Create some conflicting files `mkdir android && touch android/gradlew && touch App.js && touch package.json`.
4. Init a new app `react-native init TestApp3 --directory .`
5. **As a result get a crash with corresponding error saying that some files are conflicting.**

Screenshot 2021-07-15 at 00 31 28

PR #1591, opened 2022-04-12
WIP: feat: Init for currently exising folders

kuzkokov:
Summary:
---------

Hey, guys. So, this is a third attempt at implementing project init in the existing folder.
Basically, I took #1446 and implement fixes from @grabbou code-review.  

Then I realized that we should take underscored file names from templates into account and implemented it too.
And **then** I realized that we also change the template placeholder on the project name, including the renaming of folders and files, so we probably should take this into account as well.  

But just now I came across [this reply](https://github.com/react-native-community/cli/issues/1418#issuecomment-844924007), and now I'm confused — do we even have to worry about any conflicts? If so, why?

So now it's more of an issue for discussion rather than PR to review because it most certainly will have some finalization or complete rewriting.

Test Plan:
----------
> Current empty folder:
> 
> 1. Create an empty folder `mkdir TestApp1`.
> 2. Go to a newly created folder `cd TestApp1`.
> 3. Init a new app `react-native init TestApp1 --directory .`
> 4. **As a result a new app gets created in `TestApp1` folder.**
> 
> Non-current empty existing folder:
> 
> 1. Create an empty folder `mkdir TestApp2`.
> 2. Init a new app `react-native init TestApp2 --directory TestApp2`
> 3. **As a result a new app gets created in `TestApp2` folder.**
> 
> Current folder with conflicting files:
> 
> 1. Create an empty folder `mkdir TestApp3`.
> 2. Go to a newly created folder `cd TestApp3`.
> 3. Create some conflicting files `mkdir android && touch android/gradlew && touch App.js && touch package.json`.
> 4. Init a new app `react-native init TestApp3 --directory .`
> 5. **As a result get a crash with corresponding error saying that some files are conflicting.**
> 
> Screenshot 2021-07-15 at 00 31 28


thymikee:
@adamTrz mind having a look here? :)

adamTrz:
Hey @kuzkokov 
Thanks for the PR :) Would you mind rebasing and resolving conflicts? 🙏 

> But just now I came across https://github.com/react-native-community/cli/issues/1418#issuecomment-844924007, and now I'm confused — do we even have to worry about any conflicts? If so, why?
>
> So now it's more of an issue for discussion rather than PR to review because it most certainly will have some finalization or complete rewriting.

I would say that maybe instead of throwing an error we could prompt the user with some kind of a warning that those files exists and ask them for confirmation? 
Sth along those lines? 
"The directory ${directoryName} contains files that could conflict, please confirm that you want to continue..." 
  WDYT?

kuzkokov:
hey, @adamTrz
sorry for the long reply, just got my hands on it.
it sounds good, but then we can go further and remove all of the file-checking logic completely and just always ask for confirmation for an existing folder.
rephrasing your own words: "The directory ${directoryName} **might contain** files that could conflict, please confirm that you want to continue..."
  
the only thing here is that we will need to add an additional flag like `--no-prompt` with a default value, or maybe just a specific one for the case like `--allow-existing-folder`.

kuzkokov:
or, even a much simpler solution. we can just add a `--allow-existing-folder` flag, and:
- if the flag is set, we don't care if the folder exists or not;
- if it's not, then we throw the error as usual, with some info about the flag in the error

thymikee:
As I understand the use case here: we're in some folder, run `init MyApp` and it turns out `MyApp` exists, right? Then I'd go with a prompt you mention instead of bailing and continue as usual. No need to overcomplicate it with flags that no one will use. Unless I'm missing something, then please let me know

kuzkokov:
Well, actually, my use case seems to be exactly something that shouldn't normally happen.
I work at an outsourcing company, and we create new projects **a lot**, so we are trying to automate as much project-creating logic as possible.
And our current flow is our system inits a repository with some basic files like readme, etc., and then it runs the project initialization through our CI system. So what I personally need is to run `npx react-native init MyApp --directory ./` inside of an already existing folder.

kuzkokov:
well, it isn't stale actually, it's still might be merged

PR #1595, opened 2022-04-21
Add generate-static-view-configs command line argument for bundle command

dmitryrykun:
Summary:
---------

`@react-native/babel-plugin-codegen` is a Babel plugin that generates static view configs for Fabric components during bundling stage. Right now we have to add this plugin manually if we want to use Fabric: https://reactnative.dev/docs/next/new-architecture-app-renderer-ios#3-add-babel-plugins
In order to get rid of this manual step, we would like to ship this plugin as a part of Metro.
`--generate-static-view-configs` argument allows to turn it on or off. The idea is to have ON for Fabric projects and OFF for non-Fabric. Fabric users can also opt out, in case they don't need this feature for some reason.

Test Plan:
----------

Tested it locally by passing this argument to cli in _react-native/scripts/react-native-xcode.sh_

### DIFF
--- a/packages/cli-plugin-metro/src/commands/bundle/buildBundle.ts
+++ b/packages/cli-plugin-metro/src/commands/bundle/buildBundle.ts
@@ -28,6 +28,7 @@ interface RequestOptions {
   minify: boolean;
   platform: string | undefined;
   unstable_transformProfile: string | undefined;
+  generateStaticViewConfigs: boolean;
 }
 
 export interface AssetData {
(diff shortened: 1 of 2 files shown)


thymikee:
Thanks @dmitryrykun! Any reason this can't be a part of user's `metro.config.js`?

thymikee:
I'm going to merge this in, but still looking forward to your reply and a followup PR hopefully :)

dmitryrykun:
@thymikee I'm sorry it took so long to reply, I was distracted by another urgent task. Thank you for the idea of using metro.config.js, I'll definitely try it, it can potentially be much cleaner solution. And thank you for merging the PR, it will help me to test how my approach works overall.

PR #1601, opened 2022-05-03
update: bump version of Metro packages to 0.70.1

fortmarek:
Summary:
---------

As raised [here](https://github.com/reactwg/react-native-releases/discussions/21#discussioncomment-2673975), the metro version is out-of-sync with `react-native`. In this PR, I am bumping that to `^0.70.1`.

I have noticed that the resolved version is in the end `0.70.2` - `react-native` has versions as strict, should we have that, too? If so, I can update the PR. Otherwise, we should be good as it is.

Test Plan:
----------

CI passing should be enough of a validation here.

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "jest-circus": "^26.6.2",
     "jest-snapshot-serializer-raw": "^1.1.0",
     "lerna": "^4.0.0",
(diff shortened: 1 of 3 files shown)


thymikee (approved):
Thank you! We should be fine with `^0.70.1`

PR #1605, opened 2022-05-09
Update reactNativePlatformResolver for new metro

acoates-ms:
Summary:
---------

Metro changed how to hook into the resolver with this commit:
https://github.com/facebook/metro/commit/d81d8877c05637eac2b4ea946a9fa1e8ae869b06

The reactNativePlatformResolver is still using the old format, which causes bundle commands to fail if the project has any out of tree platforms (Ex: react-native-windows-macOS)

Test Plan:
----------

The following bundle command fails currently.  But works after making this change locally.
npx --yes react-native@0.0.0-20220415-2010-3c1a81446 init testcli --template react-native@0.0.0-20220415-2010-3c1a81446
cd testcli
npx --yes react-native-windows-init@latest --verbose --version 0.0.0-canary.495 --overwrite --language cs --projectType app --namespace MyCompany.MyApplication.MyComponent
npx react-native bundle --entry-file index.js --platform windows --bundle-output test.bundle

### DIFF
--- a/packages/cli-plugin-metro/src/tools/metroPlatformResolver.ts
+++ b/packages/cli-plugin-metro/src/tools/metroPlatformResolver.ts
@@ -12,38 +12,21 @@
  *    macos: 'react-native-macos'
  * }
  */
-// @ts-ignore - no typed definition for the package
-import {resolve} from 'metro-resolver';
 
 export function reactNativePlatformResolver(platformImplementations: {
   [platform: string]: string;
 }) {
-  return (
-    context: any,
-    _realModuleName: string,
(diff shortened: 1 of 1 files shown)


thymikee (approved):
Thanks for catching that!

thymikee:
Looks like it doesn't cause regressions with ios and android platforms on 0.69, going to release it soon

thymikee:
Released in:
 - @react-native-community/cli-plugin-metro@8.0.0-alpha.6
 - @react-native-community/cli@8.0.0-alpha.6

PR #1607, opened 2022-05-11
(docs) mention the process of uninstalling a package

### LINKED ISSUE #1606: Mention package uninstallation in the autolinking doc
In the 1st section of [autolinking.md](https://github.com/react-native-community/cli/blob/master/docs/autolinking.md), it shows this text:
```
React Native libraries often come with platform-specific (native) code. Autolinking is a mechanism that allows your project to discover and use this code.

Add a library using your favorite package manager and run the build:

# install
yarn add react-native-webview
cd ios && pod install && cd .. # CocoaPods on iOS needs this extra step
# run
yarn react-native run-ios
yarn react-native run-android

That's it. No more editing build config files to use native code.

```

It clearly shows the process for installing a package that supports autolinking.

But it doesn't show the process for uninstalling, and in my opinion it's not a good idea to assume it's obvious and that any reader can figure it out themselves. The [popularity of this question](https://stackoverflow.com/q/38613790/13255028), and especially the popularity of [this answer](https://stackoverflow.com/a/42381023/13255028), shows that many people struggle with understanding package uninstallation. And the autolinking docs in this repo don't mention uninstallation.

---

So I suggest we add this text (to be placed below the section I showed above) to [autolinking.md](https://github.com/react-native-community/cli/blob/master/docs/autolinking.md) =>

To uninstall, you follow a similar process as installing:
```
yarn remove react-native-webview
# (iOS only) update your pods
cd ios && pod install && cd ..
```
When uninstalling a package, the package is automatically unlinked.

cloudscann-dev:
Summary:
---------

Show readers how to uninstall a package that supports autolinking since that's something they should know.

Closes #1606 

Test Plan:
----------

N/A

### DIFF
--- a/docs/autolinking.md
+++ b/docs/autolinking.md
@@ -15,6 +15,13 @@ yarn react-native run-android
 
 That's it. No more editing build config files to use native code.
 
+Also, removing a library is similar to adding a library:
+```sh
+# uninstall
+yarn remove react-native-webview
+cd ios && pod install && cd .. # CocoaPods on iOS needs this extra step
+```
+
 ## How does it work
 
 Each platform defines its own [`platforms`](./platforms.md) configuration. It instructs the CLI on how to find information about native dependencies. This information is exposed through the [`config`](./commands.md#config) command in a JSON format. It's then used by the scripts run by the platform's build tools. Each script applies the logic to link native dependencies specific to its platform.


thymikee (approved):
Thanks!

PR #1611, opened 2022-05-18
fix(ios): allow detailed scriptPhases object in config

mikehardy:
align the joi schema for scriptPhases with the actually allowed values

Summary:
---------

react-native-firebase (and react-native-google-mobile-ads) use script phrases that are more than just strings, because we need to depend on the Info.plist file being generated from compilation to run the script

This worked in react-native <= 0.68 but is failing in 0.69.0-rc series

It appears that the validation has changed in the area from t.object to t.string, when it is t.object - our modules react-native-config.js files successfully validate and the modules are auto-linked, when it is t.string they are ignored.

However, if I alter it to t.object in schema.ts, then strings fail, so that's not right.

I attempted to do a joi alternatives with a full definition but I simply could **not** get the tests to work

So I need help to complete this, please, anyone :pray: as I'm out of time to work on it, but I'm certain I'm in the right area here and I *think* I've got the start of a real / fully specified solution

Test Plan:
----------

`yarn watch` and `yarn test --watch` was how I was developing

Here is an example of a module that used to work, and should work (and does work with t.object()):

https://github.com/invertase/react-native-firebase/blob/e861bea96b5b54cf56044cb390db539d07ad8f30/packages/app/react-native.config.js#L8-L14

### DIFF
--- a/packages/cli-config/src/__tests__/__snapshots__/index-test.ts.snap
+++ b/packages/cli-config/src/__tests__/__snapshots__/index-test.ts.snap
@@ -51,7 +51,10 @@ Object {
       "configurations": Array [],
       "podspecPath": "<<REPLACED>>/node_modules/react-native-test/ReactNativeTest.podspec",
       "scriptPhases": Array [
-        "./abc",
+        Object {
+          "name": "abc",
+          "path": "./phase.sh",
+        },
       ],
     },
   },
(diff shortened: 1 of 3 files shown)


mikehardy:
@grabbou this is in the area of your refactor, and was mentioned on "road to 0.69" discussion, as I'm in release testers and this was the blocker for me to qualify 0.69 on the Invertase modules, I'm really sorry I couldn't just drop a perfect PR on you, bums me out that I couldn't get this to "just work", it seemed so easy ?

mikehardy:
to be perfectly clear: I won't be upset at anything that happens to this code, by anyone if it can be made to work. Take the commits, take just the code, submit something different yourself, force-push here, no offense taken at all! :-)

Salakar:
There are a few others in the wild using this also, ones that I know of are: https://github.com/Instabug/Instabug-React-Native/blob/master/react-native.config.js#L5-L10, https://github.com/JaneaSystems/nodejs-mobile-react-native/blob/unstable/react-native.config.js#L9-L26

mikehardy:
Of note! All using the object, not the string style, that's 4 data points (react-native-firebase, react-native-google-mobile-ads, instabug, janea systems) all using object and not string - I think "string style never worked" is actually a theory now not a hypothesis ;-), with the only counter data point in the tests / test fixtures here post-refactor :thinking: - though string certainly was documented, I just don't think it ever validated through joi before.

For these 4 consumers at least, we definitely need object style working again :pray: - if I can help more in any way besides just banging my head against the wall of code ;-) just let me know

thymikee (approved):
Thanks! This is a regression from refactoring linking configuration. Brought back `scriptPhases` to be an `object` for schema validation. No need to make it super explicit imho

thymikee:
Regarding `react-native-firebase`, it uses removed `podspecPath` config for iOS, which is discovered automatically. That's why it won't show in `react-native config` output. You should be good removing it from your custom config.

mikehardy:
Thanks a bunch @thymikee I really appreciate the help

PR #1614, opened 2022-05-20
fix(platform-android): `findComponentDescriptors` picks up directories

tido64:
Summary:
---------

`findComponentDescriptors` may pick up directories, e.g. [decimal.js](https://www.npmjs.com/package/decimal.js),  and throw when trying to read it.

Test Plan:
----------

Add `decimal.js` as a dependency and run `react-native config`


thymikee (approved):
Sounds reasonable, thanks!

Issue #1622, opened 2022-06-11
`upgrade` does not properly exclude files from patching when RN is within a subdirectory


bradjones1:
## Environment

```
System:
    OS: Linux 5.17 Debian GNU/Linux 10 (buster) 10 (buster)
    CPU: (12) x64 Intel(R) Core(TM) i7-10750H CPU @ 2.60GHz
    Memory: 435.51 MB / 15.33 GB
    Shell: 5.0.3 - /bin/bash
  Binaries:
    Node: 16.13.1 - /tmp/yarn--1654921762795-0.3391503797865518/node
    Yarn: 1.22.15 - /tmp/yarn--1654921762795-0.3391503797865518/yarn
    npm: 8.1.2 - /usr/local/bin/npm
    Watchman: Not Found
  SDKs:
    Android SDK: Not Found
  IDEs:
    Android Studio: Not Found
  Languages:
    Java: Not Found
  npmPackages:
    @react-native-community/cli: Not Found
    react: 17.0.2 => 17.0.2 
    react-native: 0.68.1 => 0.68.1 
  npmGlobalPackages:
    *react-native*: Not Found
```

## Description

When upgrading using `react-native upgrade` for a project which is in a subdirectory below the repository root, files which cannot be patched (e.g., because they have been edited from the prior version's defaults) are not properly excluded.

I believe this is a regression from https://github.com/react-native-community/cli/pull/272.

## Reproducible Demo

Upgrade a project within a subdirectory of the project repo, with files which have changed from the prior version's defaults. Files which should be excluded will fail `does not exist in index`.

TMisiukiewicz:
not stale - PR open #1623

szymonrybczak:
Hey, since the `upgrade` command was changed to just link to upgrade helper in https://github.com/react-native-community/cli/pull/1879 - I'm closing this issue because it's outdated.

PR #1623, opened 2022-06-11
Better handle subdirectoried project on `upgrade`

bradjones1:
Summary:
---------

Fixes https://github.com/react-native-community/cli/issues/1622

Test Plan:
----------

Upgrade a RN project which is in a subdirectory of a repository, and has non-patchable files (e.g., they changed from the prior version's distributed copy.) Upgrade will fail to patch any files. This also adds an environment variable to further exclude files that might have been changed and will be in a dirty state (e.g., after `expo upgrade`.)

### DIFF
--- a/packages/cli/src/commands/upgrade/upgrade.ts
+++ b/packages/cli/src/commands/upgrade/upgrade.ts
@@ -248,7 +248,8 @@ const applyPatch = async (
   tmpPatchFile: string,
   repoName: RepoNameType,
 ) => {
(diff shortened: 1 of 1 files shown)


bradjones1:
I apologize I didn't touch the test coverage here... this came about as part of an upgrade on a project I'm cranking on and my Jest is even worse than my TypeScript - which is to say, it's non-existent. Hopefully this is clear enough, though.

PR #1640, opened 2022-07-07
fix: schema validation errors disguised as warnings

thymikee:
Summary:
---------

When Joi validates the schema, it will abort early and thus prevent us from constructing the rest of the configuration, which might be correct. In such case the CLI would emit a warning that configuration for a certain package was invalid. However what this warning wouldn't tell is that the library failed to autolink completely. 

This PR changes the Joi validation to not abort early and make warnings what they really are, without preventing the schema object to be created. 

This caused issues like this one: https://github.com/reactwg/react-native-releases/discussions/23 

Test Plan:
----------

Adjusted the wording of the warning:

image

### DIFF
--- a/jest/helpers.ts
+++ b/jest/helpers.ts
@@ -53,7 +53,7 @@ export async function runUntil(
     ...options,
   });
 
-  spawnPromise.stderr.pipe(
+  spawnPromise.stderr?.pipe(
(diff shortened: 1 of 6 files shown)


thymikee:
Will work on unit tests tomorrow

PR #1647, opened 2022-07-12
feat: Resolve workspaces native dependencies

adrianha:
Summary:
---------

I would like to have autolinking respect native dependencies that defined in each workspace packages, not only on the package where we have `react-native` dep. I think it will be easier to manage monorepo dependencies since each package will define it's own dependencies.
Based on https://github.com/react-native-community/cli/blob/main/docs/autolinking.md#how-can-i-use-autolinking-in-a-monorepo, If we had library with native code as a dependency in `components`, we should also define it in `mobile` to make autolinking works.

On this PR, I modified `findDependencies` method to look up all dependencies of the workspaces so it can be included in autolinking process.

Test Plan:
----------

I've made a playground [here](https://github.com/adrianha/react-native-cli-monorepo-example) to demonstrate the flow. `@react-native-community/datetimepicker` will be included in the autolinking process even though it's defined in `util` package.


packages/cli-config/src/findDependencies.ts
```diff
@@ -1,5 +1,57 @@
...
+    return [];
+  }
+
+  const rootPjson: {
+    workspaces: Array<string> | {packages: Array<string>};
+  } = require(path.join(root, 'package.json'));
+  let dependencies: Array<string> = [];
+  const workspaces = Array.isArray(rootPjson.workspaces)
+    ? rootPjson.workspaces
+    : rootPjson.workspaces.packages;
```
thymikee, line 35: I'm not a fan of relying on the `workspaces`. There are monorepos that don't use this kind of resolution strategy. Wonder what are @grabbou thoughts on that

packages/cli-config/package.json
```diff
@@ -11,6 +11,7 @@
     "@react-native-community/cli-tools": "^9.0.0-alpha.10",
     "cosmiconfig": "^5.1.0",
     "deepmerge": "^3.2.0",
+    "find-root": "^1.1.0",
```
thymikee, line 14: we use `find-up` dep in other places, maybe it's possible to use reuse it here?

satya164:
If I understand this change correctly, this is what it's implying:

- `app` depends on `foo`
- `foo` depends on `bar` which includes native code
- `bar` should be autolinked to `app`

Currently, this behavior isn't supported by the CLI in regular projects either - I think the packages being in the monorepo shouldn't work differently. Autolinking in monorepo working differently also means people who are working on libraries that are published to the registry won't work when installed in a project that's not in a monorepo.

If CLI supports such linking in a monorepo, it should also support this in regular projects where it links dependencies of dependencies. However it's problematic as I understand.

- Since the native library is a dependency of a dependency, it's possible that there can be 2 different versions of the package installed, while this is fine with JS code, 2 versions of the same native library can't be linked separately, autolinking can't work reliably in these cases.
- Some package managers (such pnpm, yarn pnp?) don't allow you to import dependencies of dependencies that you haven't declared in your `package.json`, which is good because doing that means your package can break because a dep changed its dep. Autolinking deps of a dep is a similar case.

Currently, this is the approach we follow for native dependencies: https://github.com/callstack/react-native-builder-bob#how-do-i-add-a-react-native-library-containing-native-code-as-a-dependency-in-my-library

I think monorepos should follow the same approach as well.

This means: your `util` package should include `@react-native-community/datetimepicker` as a peer dep, not regular dep, the `mobile` package should delcare `@react-native-community/datetimepicker` in its `dependencies`. I think CLI should support this workflow and find the dependencies that are declared in `dependencies` field but hoisted in a monorepo.

thymikee:
What @satya164 said. We'd like to come up with a good system for discovering native deps that's reliable and performant. This it not that easy of a task and thanks for trying to tackle it! Hopefully we can generalize this solution to support dependencies of monorepos and other dependencies.

adrianha:
hi @thymikee @satya164, thank you for the feedback.

>If I understand this change correctly, this is what it's implying:
>
>app depends on foo
>foo depends on bar which includes native code
>bar should be autolinked to app

yes, this is what I'm trying to achieve in this PR that might help to manage dependencies with native code in a monorepo project, where each package can define its dependencies and can be autolinked without defining it on root-app package.

> Currently, this behavior isn't supported by the CLI in regular projects either - I think the packages being in the monorepo shouldn't work differently. Autolinking in monorepo working differently also means people who are working on libraries that are published to the registry won't work when installed in a project that's not in a monorepo.
> 
> If CLI supports such linking in a monorepo, it should also support this in regular projects where it links dependencies of dependencies. However it's problematic as I understand.
> 
> Since the native library is a dependency of a dependency, it's possible that there can be 2 different versions of the package installed, while this is fine with JS code, 2 versions of the same native library can't be linked separately, autolinking can't work reliably in these cases.
> Some package managers (such pnpm, yarn pnp?) don't allow you to import dependencies of dependencies that you haven't declared in your package.json, which is good because doing that means your package can break because a dep changed its dep. Autolinking deps of a dep is a similar case.

agree on this point, resolving dependencies behavior should be the same project with/without monorepo. I also haven't tried other monorepo/workspace tools (nx, pnpm, etc) with this diff.

> Currently, this is the approach we follow for native dependencies: https://github.com/callstack/react-native-builder-bob#how-do-i-add-a-react-native-library-containing-native-code-as-a-dependency-in-my-library
>
>I think monorepos should follow the same approach as well.
>
> This means: your util package should include @react-native-community/datetimepicker as a peer dep, not regular dep, the mobile package should delcare @react-native-community/datetimepicker in its dependencies. I think CLI should support this workflow and find the dependencies that are declared in dependencies field but hoisted in a monorepo.

actually, I also used this flow to handle dependency in monorepo project and add some non-hoist rules to keep autolinking works without updating project path. CLI support for finding hoisted dependencies would be great though!

> What @satya164 said. We'd like to come up with a good system for discovering native deps that's reliable and performant. This it not that easy of a task and thanks for trying to tackle it! Hopefully we can generalize this solution to support dependencies of monorepos and other dependencies.

yeah, it's not an easy task indeed, there're many things that should be considered to support this workflow considering there're many monorepo/workspace tooling.

Issue #1168, opened 2020-05-13
Package subpath './package.json' is not defined by "exports"


ctavan:
## Environment

```
System:
    OS: macOS 10.15.4
    CPU: (4) x64 Intel(R) Core(TM) i7-5557U CPU @ 3.10GHz
    Memory: 94.37 MB / 16.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 14.2.0 - /usr/local/bin/node
    Yarn: 1.22.4 - /usr/local/bin/yarn
    npm: 6.14.4 - /usr/local/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  Managers:
    CocoaPods: 1.9.1 - /usr/local/bin/pod
  SDKs:
    iOS SDK:
      Platforms: iOS 13.4, DriverKit 19.0, macOS 10.15, tvOS 13.4, watchOS 6.2
    Android SDK: Not Found
  IDEs:
    Android Studio: Not Found
    Xcode: 11.4.1/11E503a - /usr/bin/xcodebuild
  Languages:
    Java: 1.8.0_121 - /usr/bin/javac
    Python: 3.7.7 - /usr/local/opt/python/libexec/bin/python
  npmPackages:
    @react-native-community/cli: Not Found
    react: 16.13.1 => 16.13.1
    react-native: 0.62.2 => 0.62.2
  npmGlobalPackages:
    *react-native*: Not Found
```

## Description

When running `react-native` under Node.js 14.x on a project that uses npm modules which specify the new `exports` field in `package.json`, `react-native` will print the following warning (in this example the `uuid` package is used):

```
warn Package uuid has been ignored because it contains invalid configuration. Reason: Package subpath './package.json' is not defined by "exports" in /PATH_TO_CURRENT_PROJECT/node_modules/uuid/package.json
```

The same issue has been reported in other places:

* `i18next-http-backend`: https://github.com/facebook/react-native/issues/28710 for (workaround: https://github.com/i18next/i18next-http-backend/commit/1d8da7909425eb33ab286cd73ee53bb48a1a3dc4#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R8)
* `i18next-locize-backend`: https://github.com/locize/i18next-locize-backend/pull/326/files
* `nanoevents`: https://github.com/ai/nanoevents/issues/44
* `uuid`: https://github.com/uuidjs/uuid/issues/444#issuecomment-627216425

[...]

PR #1655, opened 2022-07-25
Fix resolving node modules directory with custom exports

fortmarek:
Summary:
---------

Fixes https://github.com/react-native-community/cli/issues/1168

As described in the issue, it's currently not possible to use RN CLI with packages that have `exports` in their `package.json` and do not contain `package.json` itself in the `exports` object. For example, `d3` dependency [does just that](https://github.com/d3/d3/blob/main/package.json#L32-L35). While maintainers could make their package compatible with the RN CLI, such an effort would be huge and not all maintainers want [to do that](https://github.com/d3/d3/pull/3576) – understandably so, especially if their packages are not exclusively for RN.

To fix this, I've imported [this](https://github.com/microsoft/rnx-kit/blob/f37adca5161eba66fc27de25d48f72973fff9e8e/packages/tools-node/src/package.ts#L213-L234) method from the `@rnx-kit/tools-node` package that does not depend on the `require.resolve` method (which currently does not work in the case described above). Alternatively, we could just import the method itself but the package is quite small, so in this case, it's probably better in the long run to depend on it instead.

Test Plan:
----------

- Clone [this](https://github.com/fortmarek/cli-exports-repro) sample
- cd into it and run `yarn`
- Run `yarn start`
- Observe error similar to:
```
error Failed to load configuration of your project.
Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './package.json' is not defined by "exports" in /Users/marekfort/Downloads/RN0692/node_modules/d3/package.json
    at new NodeError (node:internal/errors:387:5)
    at throwExportsNotFound (node:internal/modules/esm/resolve:439:9)
    at packageExportsResolve (node:internal/modules/esm/resolve:718:3)
    at resolveExports (node:internal/modules/cjs/loader:493:36)
    at Module._findPath (node:internal/modules/cjs/loader:533:31)
    at Module._resolveFilename (node:internal/modules/cjs/loader:942:27)
    at Function.resolve (node:internal/modules/cjs/helpers:108:19)
    at resolveNodeModuleDir (/Users/marekfort/src/github.com/fortmarek/cli/packages/cli-tools/build/resolveNodeModuleDir.js:24:42)
    at /Users/marekfort/src/github.com/fortmarek/cli/packages/cli-config/build/loadConfig.js:93:76
    at Array.reduce (<anonymous>)
info Run CLI with --verbose flag for more details.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```
- Clone the branch from this PR
- Set up CLI as per [these](https://github.com/react-native-community/cli/blob/main/CONTRIBUTING.md#testing-your-changes) instructions
- Once the custom CLI is linked inside the repro sample, run `yarn start` again
- No error should occur

### DIFF
--- a/packages/cli-tools/package.json
+++ b/packages/cli-tools/package.json
@@ -7,6 +7,7 @@
     "access": "public"
   },
   "dependencies": {
+    "@rnx-kit/tools-node": "^1.3.0",
     "appdirsjs": "^1.2.4",
     "chalk": "^4.1.2",
     "lodash": "^4.17.15",
--- a/packages/cli-tools/package.json
+++ b/packages/cli-tools/package.json
@@ -7,9 +7,9 @@
     "access": "public"
   },
   "dependencies": {
-    "@rnx-kit/tools-node": "^1.3.0",
     "appdirsjs": "^1.2.4",
     "chalk": "^4.1.2",
+    "find-up": "^5.0.0",
     "lodash": "^4.17.15",
     "mime": "^2.4.1",
     "node-fetch": "^2.6.0",
@@ -22,7 +22,7 @@
     "@react-native-community/cli-types": "^9.0.0-alpha.0",
     "@types/lodash": "^4.14.149",
     "@types/mime": "^2.0.1",
-    "@types/node": "^17.0.35",
+    "@types/node": "^12.0.0",
     "@types/node-fetch": "^2.5.5"
   },
   "files": [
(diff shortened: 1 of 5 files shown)


thymikee (approved):
Thanks! I decided to vendor the implementation of `pickValues` and `findPackageDependencyDir` due to the `@rnx-kit/tools-node` package adding roughly 42kB: https://packagephobia.com/result?p=%40rnx-kit%2Ftools-node. RN CLI already downloads a lot of stuff and we need to make a better job at keeping it lean so that `npx react-native init` good enough.

luKk9:
i did but it doesnt work. omg

PR #1658, opened 2022-07-27
chore: set node to 14

thymikee:
Summary:
---------

Aligning with RN 0.70 on setting the minimum required Node version to 14. 
Additionally: 
- removed some unnecessary babel transforms from the config
- upgraded `commander` dep to the latest (at last!)

Test Plan:
----------

CI


packages/cli/package.json
```diff
@@ -34,12 +34,11 @@
     "@react-native-community/cli-tools": "^9.0.0-alpha.5",
     "@react-native-community/cli-types": "^9.0.0-alpha.0",
     "chalk": "^4.1.2",
-    "commander": "^2.19.0",
+    "commander": "^9.4.0",
     "execa": "^1.0.0",
     "find-up": "^4.1.0",
     "fs-extra": "^8.1.0",
     "graceful-fs": "^4.1.3",
-    "leven": "^3.1.0",
```
thymikee, line 42: "did you mean" is now built into commander, so there's no need for leven

PR #1660, opened 2022-07-28
Ignore case in validatePackageName

mlazari:
Summary:
---------

Currently `validatePackageName` fails for package names containing uppercase letters. A message like this is shown during pod install:
```
warn Invalid application's package name "com.lugg.ReactNativeConfig" in 'AndroidManifest.xml'. Read guidelines for setting the package name here: https://developer.android.com/studio/build/application-id
warn Invalid application's package name "com.BV.LinearGradient" in 'AndroidManifest.xml'. Read guidelines for setting the package name here: https://developer.android.com/studio/build/application-id
warn Invalid application's package name "com.lugg.ReactNativeConfig" in 'AndroidManifest.xml'. Read guidelines for setting the package name here: https://developer.android.com/studio/build/application-id
warn Invalid application's package name "com.BV.LinearGradient" in 'AndroidManifest.xml'. Read guidelines for setting the package name here: https://developer.android.com/studio/build/application-id
```
 However both the package name (https://developer.android.com/guide/topics/manifest/manifest-element.html#package) and the application id (https://developer.android.com/studio/build/application-id) are allowed to contain upper case letters.

Screenshot 2022-07-28 at 14 29 33
Screenshot 2022-07-28 at 14 29 58

This change adds the "i" flag to the regex to ignore case when validating package name.

Related issue: https://github.com/facebook/react-native/issues/34247

Test Plan:
----------

I verified in browser console that the updated regex works correctly:
Screenshot 2022-07-28 at 15 05 24

### DIFF
--- a/packages/cli-platform-android/src/config/__tests__/getAndroidProject.test.ts
+++ b/packages/cli-platform-android/src/config/__tests__/getAndroidProject.test.ts
@@ -1,0 +1,36 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ */
+
+import {validatePackageName} from '../getAndroidProject';
+
+describe('android::getAndroidProject', () => {
+  const expectedResults = {
+    'com.app': true,
+    'com.example.app': true,
+    'com.my_app': true,
+    'org.my_app3': true,
+    'com.App': true,
+    'com.Example.APP1': true,
+    'COM.EXAMPLE.APP': true,
+    '': false,
+    com: false,
+    'com.3example.app': false,
+    'com.my_app*': false,
+    'org.my-app3': false,
+    'com.App ': false,
+    'com.Example.APP#1': false,
+  };
+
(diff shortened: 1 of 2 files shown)


matinzd:
Do you think this is gonna fix `pod install` properly? Because this should not be checked when running `pod install` anyway.

mlazari:
@matinzd It will not fix pod install calling android specific configuration. I am not familiar enough with the code in this repo to know how to do that. It fixes only it considering valid package names as invalid.

matinzd (approved):
Looks good to me, but the problem I have with this is why `getAndroidProject` should be called via `pod install`?
I think this needs to be fixed in cli bin with passing params. If it's gradle running the script android should be checked and if this is being called from `native_modules` in ruby script ios should only be checked.

thymikee:
Thanks! Looks good. Can you please add a test with this case?

thymikee:
@matinzd right, we could add a `--platform` flag to the `config` command that would filter explored platforms

mlazari:
@thymikee Added a few tests for `validatePackageName`.

thymikee (approved):

PR #1662, opened 2022-07-28
breaking: remove jetifier (`--jetifier` and `--no-jetifier` flags)

### LINKED ISSUE #1409: Post warning that jetifier is deprecated and won't be called automatically in the future
## Describe the Feature

I regret not having a link to the discussion but at some point recently we decided AndroidX transition should be done and we should sunset jetifier as an automatic thing.

I've added a deprecation notice to jetifier and now warn any time a file is translated, suggesting the user post a PR to do the AndroidX transition in the repo

https://github.com/mikehardy/jetifier/releases/tag/v1.6.7

## Possible Implementations
Just a deprecation notice at first, that jetifier will no longer be automatically called in the future.
Then later, remove the jetifier functionality here.

thymikee:
Summary:
---------

Jetifier served the community super well for the time of AndroidX migration, and today it's no longer needed for the majority of apps out there. However, for those who still need it, you can run `npx jetifier` yourselves. 

This PR removes the `--jetifier` and `--no-jetifier` flags.

It shaves off 2.9MB from the bundle size, so there's no need to stall it anymore.

Closes #1409 

Test Plan:
----------

None.


mikehardy:
Fun while it lasted yeah, happy to see it ride off in the sunset though. Cheers @thymikee !

Issue #1588, opened 2022-04-07
doctor: refactor the fixers so we can forward options and avoid duplicating efforts


tido64:
Somewhat unrelated (and definitely not to be fixed in this PR), I looked briefly at the options that are passed to all CLI commands and it looks like they include this information. However, it doesn't look like they are properly forwarded to the individual fixers. It's a bit silly that we have to duplicate the effort to find this info again. Do we have an issue tracking this? cc @mikehardy @thymikee

_Originally posted by @tido64 in https://github.com/react-native-community/cli/pull/1585#discussion_r844883409_

tido64:
Not stale.

PR #1669, opened 2022-08-04
fix: pass projectRoot path to doctors fixers

adamTrz:
Summary:
---------

Pass `projectRoot` as an argument both to `doctor`'s `getDiagnostics` and `runAutomaticFix` methods. 

Fixes: #1588 

Test Plan:
----------

`doctor` should still work.


adamTrz:
Just had a thought that this PR will add additional complexity instead of simplify things in the code and is a bit artificial. Let's leave things as is, and if any doctor command will require projectRoot it will just fetch it like it was done before. 
Closing the PR. ✌️

PR #1676, opened 2022-08-12
WIP: feat: List available apple devices

esthor:
Summary:
---------

Hey, so this is a WIP for listing available apple devices in an interactive cli prompt. The intention of this is to be able to let the user select an available simulator vs. having to paste device name/udid strings in.

I wanted to put this up as a WIP first because **I don't know which path to take next**...

1. I could have this be its own helper function, that just lists out the sims (its current state) -- Do we want this as an optional helper for users?
2. I could merge it into the `run-ios` flow (initial thoughts: if user doesn't select a sim, one isn't booted or the one they supply isn't available/existing, instead of fallbacks we give this prompt and then carry-on the run-ios command)
3. Something else? Maybe a combination of 1 & 2 (it's available standalone, and also inside `run-ios`)? 🤔 

Some advice would be really helpful here as my only experience with the CLI has been my own (and teams') usage. I don't feel I have the context to make a judgement call here for a whole community. Maybe it's super obvious/simple, I just don't know 🤷 

Test Plan:
----------

WIP

### DIFF
--- a/packages/cli-platform-ios/src/commands/index.ts
+++ b/packages/cli-platform-ios/src/commands/index.ts
@@ -1,4 +1,5 @@
+import listIOS from './listIOS';
 import logIOS from './logIOS';
 import runIOS from './runIOS';
 
-export default [logIOS, runIOS];
+export default [listIOS, logIOS, runIOS];
--- a/packages/cli-platform-ios/src/commands/listIOS/index.ts
+++ b/packages/cli-platform-ios/src/commands/listIOS/index.ts
@@ -1,0 +1,52 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ */
+
+import {execFileSync} from 'child_process';
+import {logger} from '@react-native-community/cli-tools';
+import {Device} from '../../types';
+
+function findAvailableDevices(devices: {[index: string]: Array<Device>}) {
+  // TODO -- It would be cool to separate out the Apple Watch, iOS, iPad, TV, and others...
+  let availableDevices = [];
+  for (const key of Object.keys(devices)) {
+    for (const device of devices[key]) {
+      if (
+        device.availability === '(available)' ||
+        device.isAvailable === true
+      ) {
+        availableDevices.push(device.name);
+      }
+    }
+  }
+  return availableDevices;
+}
+
+async function listIOS() {
+  const rawDevices = execFileSync(
+    'xcrun',
+    ['simctl', 'list', 'devices', '--json'],
+    {encoding: 'utf8'},
+  );
+
+  const {devices} = JSON.parse(rawDevices) as {
+    devices: {[index: string]: Array<Device>};
+  };
+
+  const availableDevices = findAvailableDevices(devices);
+  if (availableDevices === null) {
+    logger.error('No available iOS devices found');
+    return;
+  }
+  logger.info(JSON.stringify(availableDevices));
+}
+
+export default {
+  name: 'list-ios',
+  description: 'lists available iOS devices',
+  func: listIOS,
+};
--- a/packages/cli-platform-ios/src/commands/listIOS/index.ts
+++ b/packages/cli-platform-ios/src/commands/listIOS/index.ts
@@ -7,6 +7,8 @@
  */
 
 import {execFileSync} from 'child_process';
+import prompts from 'prompts';
+import chalk from 'chalk';
 import {logger} from '@react-native-community/cli-tools';
 import {Device} from '../../types';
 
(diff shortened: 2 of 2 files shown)


thymikee:
How does it look like from the UI perspective? Can you share a screenshot maybe? :)

esthor:
Yeah, definitely. It looks like this right now:

Screen Recording 2022-08-22 at 12 24 28 PM

esthor:
I was thinking about it a bit more, and it should probably not be its own function, but be added to the `run-ios` command. I think in two cases:
1. By passing an extra argument/flag like `run-ios --list-simulators`
2. Replacing the `fallbackSimulators` logic with this interactive approach. (`fallbackSimulators` seems kinda jank anyway...I'd thought of having them live together, but `fallbackSimulators` seems like a pain to maintain and is a "dumb" approach compared to just asking the user what they want in a simple way vs. assuming)

What do you think?

liamjones:
This looks nice! Any chance it could list connected real devices as well as simulators too?

adamTrz:
Thanks for the contribution @esthor ! 💪 

> I wanted to put this up as a WIP first because I don't know which path to take next...
> I could have this be its own helper function, that just lists out the sims (its current state) -- Do we want this as an optional helper for users?
I could merge it into the run-ios flow (initial thoughts: if user doesn't select a sim, one isn't booted or the one they supply isn't available/existing, instead of fallbacks we give this prompt and then carry-on the run-ios command)
Something else? Maybe a combination of 1 & 2 (it's available standalone, and also inside run-ios)? 🤔

Maybe sth like what `doctor` is doing? List available devices and if user taps "exit" key we will skip selection of device and fall back to default behaviour?  WDYT? 

Also, can you add (Booted) info for your listing to show which device is booted and can be used?

esthor:
> This looks nice! Any chance it could list connected real devices as well as simulators too?

Yes! Good call. I didn't test this out, but should be doable. I'll put them at the top, and maybe flag them in some way that makes it obvious. I'm imagining that in most cases, if you have a connected physical device, your intention is probably to use that, so I'll put those first.

esthor:
> Thanks for the contribution @esthor ! 💪
> 
> > I wanted to put this up as a WIP first because I don't know which path to take next...
> > I could have this be its own helper function, that just lists out the sims (its current state) -- Do we want this as an optional helper for users?
> > I could merge it into the run-ios flow (initial thoughts: if user doesn't select a sim, one isn't booted or the one they supply isn't available/existing, instead of fallbacks we give this prompt and then carry-on the run-ios command)
> > Something else? Maybe a combination of 1 & 2 (it's available standalone, and also inside run-ios)? 🤔
> 
> Maybe sth like what `doctor` is doing? List available devices and if user taps "exit" key we will skip selection of device and fall back to default behaviour? WDYT?
> 
> Also, can you add (Booted) info for your listing to show which device is booted and can be used?

Oh, that sounds interesting. I will take a look at how the `doctor` works, but sounds very good to me.

And yeah, will add a `Booted` and a `Connected` flag or section or some way to call those out and have them show at the top of the list.

- [ ] Add 'Booted' / 'Available'
- [ ] Add 'Connected Physical Devices'

adamTrz:
Hi @esthor . Do you need any help with wrapping this PR? 🙂 
Happy to help and continue on this PR if you are swarmed with other tasks :)

esthor:
Hey! Going to look at it this week. 🙏

packages/cli-platform-ios/src/commands/runIOS/index.ts
```diff
@@ -83,8 +82,31 @@ function runIOS(_: Array<string>, ctx: Config, args: FlagsT) {
...
+    }
+    const availableDevices = await listIOSDevices();
+    const selectedDevice = await promptForDeviceSelection(availableDevices);
+    if (!selectedDevice) {
+      throw new CLIError(
+        'Failed to select device, please try to run app without "list-devices" command.',
+      );
+    }
+    if (selectedDevice.type === 'simulator') {
+      return runOnSimulator(xcodeProject, scheme, args);
```
szymonrybczak, line 101: There's no `selectedDevice` passed to the `runOnSimulator` function. 
```suggestion
      return runOnSimulator(xcodeProject, scheme, args, selectedDevice);
```

packages/cli-platform-ios/src/commands/runIOS/index.ts
```diff
@@ -746,7 +768,12 @@ export default {
       name: '--terminal <string>',
       description:
         'Launches the Metro Bundler in a new window using the specified terminal path.',
-      default: getDefaultUserTerminal,
+      default: getDefaultUserTerminal(),
+    },
+    {
+      name: '--list-devices',
```
szymonrybczak, line 774: Let's add this flag to docs 😄

adamTrz: Done! 👍

esthor:
Oh my gosh. You guys are fantastic. Sorry for delaying this so much. Turns out having a new child can throw off my scheduling. 😂 👨‍🍼 Please lmk how I can assist today

adamTrz:
> Oh my gosh. You guys are fantastic. Sorry for delaying this so much. Turns out having a new child can throw off my scheduling. 😂 👨‍🍼 Please lmk how I can assist today

No worries, there are priorities I guess 👶🏻

You can't try to run it locally and see how it goes ;)

esthor:
Nice! Ran it locally, and it's looks great for both iOS and Android! 🚀

adamTrz (approved):

PR #1687, opened 2022-08-20
chore: remove duplicates on archs array

mateosilguero:
Summary:
---------

Remove duplicated archs from archs array:

Before:
Captura desde 2022-08-19 22-46-30

After:
Captura desde 2022-08-19 22-53-13

Test Plan:
----------

Captura desde 2022-08-19 23-10-16


thymikee (approved):
Thanks!

PR #1699, opened 2022-09-08
Add more internal callsites for better stack frames

rickhanlonii:
Summary:
---------

Adds more internal callsites to clean up log box internal stack frame collapsing.

Test Plan:
----------

### Before
Notice how the code frame is not the error in the user code:
Screen Shot 2022-09-08 at 9 20 54 AM

### After
Now it is:
Screen Shot 2022-09-08 at 11 16 31 AM


thymikee:
cc @motiz88 @huntie, mind reviewing that?
@rickhanlonii can you run `yarn lint --fix`?

rickhanlonii:
@motiz88 @huntie I'm also curious - do we know why the stack frame says "<global>" and "global code"? Is there something more relevant we can display there or is it from Hermes? And should we collapse the `.bundle` frames by default (the second frame in the "after" image)?

motiz88:
I'm not sure where the string `global code` is coming from (engine-specific?) but this actually seems pretty informative: it points to the error being in the initialisation path, and the `.bundle` means this is generated code with no source mapping. We could somewhat reasonably map this particular case to somewhere in the bundle's entry point, but technically that frame does live outside of that module, so it's an ambiguous symbolication and I'd avoid it.

I would also not collapse all `.bundle` frames ( = all unsymbolicated frames) as it's a bit too indiscriminate, especially for cases where symbolication might be broken for whatever reason. But maybe matching the `global code` string (and its equivalents in other engines) alongside `.bundle` is a good heuristic. Unsymbolicated global code = most likely bundle infrastructure that can't be mapped to any file on disk.

packages/cli-plugin-metro/src/tools/loadMetroConfig.ts
```diff
@@ -14,9 +14,14 @@ const INTERNAL_CALLSITES_REGEX = new RegExp(
     '/Libraries/YellowBox/.+\\.js$',
     '/Libraries/LogBox/.+\\.js$',
     '/Libraries/Core/Timers/.+\\.js$',
+    '/Libraries/WebSocket/.+\\.js',
```
motiz88, line 17:
```suggestion
    '/Libraries/WebSocket/.+\\.js$',
```

packages/cli-plugin-metro/src/tools/loadMetroConfig.ts
```diff
@@ -14,9 +14,14 @@ const INTERNAL_CALLSITES_REGEX = new RegExp(
...
+    '/Libraries/vendor/.+\\.js',
```
motiz88, line 18:
```suggestion
    '/Libraries/vendor/.+\\.js$',
```

packages/cli-plugin-metro/src/tools/loadMetroConfig.ts
```diff
@@ -14,9 +14,14 @@ const INTERNAL_CALLSITES_REGEX = new RegExp(
...
     '/node_modules/react-devtools-core/.+\\.js$',
     '/node_modules/react-refresh/.+\\.js$',
     '/node_modules/scheduler/.+\\.js$',
+    '/node_modules/event-target-shim/.+\\.js$',
+    '/metro-runtime/.+\\.js$',
+    '\\[native code\\]',
```
motiz88, line 24:
```suggestion
    '^\\[native code\\]$',
```

I suspect this string may be engine-specific - have you tested this on both JSC and Hermes?

Also, bikeshedding: might be useful to include native code as intermediate frames, though I agree they're less useful at the top since we have no JS code frame to show there.

packages/cli-plugin-metro/src/tools/loadMetroConfig.ts
```diff
@@ -14,9 +14,14 @@ const INTERNAL_CALLSITES_REGEX = new RegExp(
     '/Libraries/YellowBox/.+\\.js$',
```
motiz88, line 14: I wonder if all of these should be prefixed with `react-native`. (Not sure)

motiz88:
I might be missing something but how does the `invariant` stack frame get skipped? Not seeing any pattern that specifically targets it.

packages/cli-plugin-metro/src/tools/loadMetroConfig.ts
rickhanlonii: Yeah probably -- I opted to minimize the diff because I wasn't sure why it was this way.

rickhanlonii:
Nice catch, I missed this copying it over.

thymikee (approved):
Thanks @rickhanlonii and @motiz88

packages/cli-plugin-metro/src/tools/loadMetroConfig.ts
thymikee: @rickhanlonii mind following up with a PR with adjusted paths?

cpojer: The reason the `react-native` prefix is missing is because forks of React Native such as [`react-native-tvos`](https://github.com/react-native-tvos/react-native-tvos) (and `react-native-windows`?) have copies of these files but are installed within `node_modules` using their forked name. If you match on `react-native`, you'll break `YellowBox` for all forks. You might want to use `node_modules\/(react-native(?:-(.*?))?)\/` here and below where you added the match on `/node_modules/react-native/index.js`.

rickhanlonii: A wild Nakazawa appears!

Christoph used Context.

It's super effective!

PR #1701, opened 2022-09-21
feat: added last booted simulator pick & iPhone 14 as fallback + doc typos

slavikdenis:
Summary:
---------

- docs: fix missing code block langs & typos
- feat: add iPhone 14 as a fallback simulator
- feat: add last booted simulator pick, if no other matches the finder

```
** last booted simulator **

Sometimes, it's better to have the last booted simulator as a fallback, instead of the first available simulator.
```

Test Plan:
----------

After a new Xcode install, when the simulators are fresh, and CLI yet doesn't have the latest iPhone listed as a fallback simulator, CLI fails to pick an appropriate simulator. If the user however run already any preferred simulator, it gets picked instead of failing on picking the appropriate simulator.

### DIFF
--- a/docs/init.md
+++ b/docs/init.md
@@ -12,7 +12,7 @@ npx react-native init ProjectName
 
 > Note: If you have both `yarn` and `npm` installed on your machine, React Native CLI will always try to use `yarn`, so even if you use `npx` utility, only `react-native` executable will be installed using `npm` and the rest of the work will be delegated to `yarn`. You can force usage of `npm` adding `--npm` flag to the command.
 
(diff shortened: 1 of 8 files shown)


docs/init.md
```diff
@@ -32,19 +32,19 @@ npx react-native@${VERSION} init ProjectName
 
 #### Initializing project with custom template
 
-It is possible to initialise a new application with a custom template with
+It is possible to initialize a new application with a custom template with
```
adamTrz, line 35: UK v US English ;)

adamTrz (approved):
LGTM! 🏅

thymikee:
That's definitely an improvement, thanks! One thing I'd add as a followup, is to select the first available simulator if none of the fallbacks exist and a user never opened a simulator—which will happen for first-time users next year when starting with fresh Xcode 15 I suppose :)

PR #1704, opened 2022-09-26
chore: update iOS simulators

luism3861:
Summary:
---------
this PR update iOS simulators according Xcode 14 release i think it's a good idea update for iPhone 14, 13, 12 simulators

Test Plan:
----------


thymikee:
This is fixed in #1701. I'll make a patch release in a moment

PR #1708, opened 2022-10-04
feat(ios): use Bundler for pods installation

hoxyq:
Summary:
---------
- Using Bundler as a default way to handle dependencies on iOS

Test Plan:
----------
- `yarn test`
- Possible error output, when other ruby version is used:
Screenshot 2022-10-04 at 11 42 28

- Successful run:
Screenshot 2022-10-04 at 12 45 10

### DIFF
--- a/packages/cli-doctor/src/tools/runBundleInstall.ts
+++ b/packages/cli-doctor/src/tools/runBundleInstall.ts
@@ -1,0 +1,23 @@
+import execa from 'execa';
+import {logger} from '@react-native-community/cli-tools';
+
+import {Loader} from '../types';
+
+async function runBundleInstall(loader: Loader) {
+  try {
+    loader.start('Installing Bundler');
+
+    await execa('bundle', ['install']);
+  } catch (error) {
+    loader.fail();
+    logger.error(error.stderr || error.stdout);
+
+    throw new Error(
+      'Looks like your iOS environment is not properly set. Please go to https://reactnative.dev/docs/next/environment-setup and follow the React Native CLI QuickStart guide for macOS and iOS.',
+    );
+  }
+
+  loader.succeed();
+}
+
+export default runBundleInstall;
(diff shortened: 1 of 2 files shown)


packages/cli-doctor/src/tools/runBundleInstall.ts
```diff
@@ -0,0 +1,21 @@
+import execa from 'execa';
+import {logger} from '@react-native-community/cli-tools';
+
+import {Loader} from '../types';
+
+async function runBundleInstall(loader: Loader) {
+  try {
+    loader.start('Installing Bundler');
```
cipolleschi, line 8: Why this line does not appear in the successful run posted in the Test Plan? 🤔

packages/cli-doctor/src/tools/runBundleInstall.ts
```diff
@@ -0,0 +1,21 @@
...
+
+    await execa('bundle', ['install']);
+  } catch (error) {
+    loader.fail();
+    logger.error(error.stderr || error.stdout);
+
+    throw new Error(
+      'Looks like your iOS environment is not properly set. Please go to https://reactnative.dev/docs/next/environment-setup and follow the React Native CLI QuickStart for macOS and iOS.',
```
cipolleschi, line 16: nit
```suggestion
      'Looks like your iOS environment is not properly set. Please go to https://reactnative.dev/docs/next/environment-setup and follow the React Native CLI QuickStart guide for macOS and iOS.',
```

cipolleschi (approved):
Hi! Thank you so much for taking the time to implement this. I left a question and a very small suggestion, but the code looks good to me.
Waiting for the CLI maintainers to chime in.

adamTrz (approved):
Great job! Thanks for contribution  @hoxyq !

packages/cli-doctor/src/tools/runBundleInstall.ts
hoxyq: > Why this line does not appear in the successful run posted in the Test Plan? 🤔

Looks like I need to make a separate `logger.info` call, will fix that

hoxyq:
Updated:
- added `logger.succeed` call to fix appearing of `Installing Bundler` message in logs
- fixed error message

PR #1711, opened 2022-10-04
Upgrade Metro dependencies to v0.73.0 in `cli-plugin-metro`

arushikesarwani94:
Summary:
---------

Updating the versions of Metro in the `cli-plugin-metro` to incorporate the new Metro Release v0.73.0. This release switches the default minifier to `terser`, raises the minimum Node version to ^14.17, and drops support for old Watchman versions. It also improves support for Apple Silicon and will enable the use of BigInt out of the box in React Native 0.71.

Per the release notes we shared (link below), I am suggesting that this is a breaking change for RN CLI.

**Metro Release Notes:** https://github.com/facebook/metro/releases/tag/v0.73.0
**Full Changelog between v0.72.3 and v0.73.0:** https://github.com/facebook/metro/compare/v0.72.3...v0.73.0

Reminder : When React Native updates the CLI to a version that depends on metro 0.73.0, it must also update metro-runtime etc to 0.73.0 in the same commit

Test Plan:
----------

- Updated dependencies with `yarn`.
- ✅ Ran `yarn test`.

### DIFF
--- a/packages/cli-plugin-metro/package.json
+++ b/packages/cli-plugin-metro/package.json
@@ -10,12 +10,12 @@
     "@react-native-community/cli-server-api": "^9.1.0",
     "@react-native-community/cli-tools": "^9.1.0",
     "chalk": "^4.1.2",
-    "metro": "0.72.3",
-    "metro-config": "0.72.3",
-    "metro-core": "0.72.3",
-    "metro-react-native-babel-transformer": "0.72.3",
-    "metro-resolver": "0.72.3",
-    "metro-runtime": "0.72.3",
+    "metro": "0.73.0",
+    "metro-config": "0.73.0",
+    "metro-core": "0.73.0",
+    "metro-react-native-babel-transformer": "0.73.0",
+    "metro-resolver": "0.73.0",
+    "metro-runtime": "0.73.0",
     "readline": "^1.3.0"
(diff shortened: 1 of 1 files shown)


arushikesarwani94:
@thymikee Can I get a review of this PR. This is the debut of the new Metro release process.

adamTrz (approved):
Lgtm!

liamjones:
https://github.com/react-native-community/cli/pull/1700 recently removed the checks for watchman versions.

I notice this PR says

>...drops support for old Watchman versions

Does that mean a check needs to be reintroduced?

robhogan:
> Does that mean a check needs to be reintroduced?

Good question. My TLDR suggestion would be "no, leave this to Metro". 

Checking Watchman versions isn't trivial, both because new version strings are semi-opaque (see commit comment https://github.com/facebook/metro/commit/d83140095c0495112634d92f0b1319c716eab9ab), and because versions don't actually tell the whole story - for a given version Watchman may be built with different [capabilities](https://facebook.github.io/watchman/docs/capabilities.html).

Metro safely falls back to non-Watchman if the installed Watchman doesn't meet requirements, but it's still pretty conservative - in practice any release after `4.9.0` (2017) will work.

PR #1716, opened 2022-10-14
feat: improve new release checker CTA and logic

kelset:
Summary:
---------

This is a small PR to address a minor inconvenience with the script that shows up when a newer version of react-native is available.

It modifies the last line to not directly mention `react-native upgrade` since that command doesn't lead the developer down a happy path.

On top of that, a small improvement to avoid the message popping up for nightlies and other non-stable releases of RN, which might happen with 1000.0.0 and 0.0.0 versions.

Test Plan:
----------

Tested locally, following the contribution guidelines:
Screenshot 2022-10-14 at 13 05 34


adamTrz (approved):
Thanks for the update! 🎉

PR #1724, opened 2022-10-24
Update ios readme to pass xcconfig from command line

ShaswatPrabhat:
Updates commands doc with the new xcconfig option


thymikee (approved):
Thank you!

PR #1729, opened 2022-10-30
Feat: customise ios build folder location with `iosBuildFolder` flag

### LINKED ISSUE #1706: Add support for customising iOS build folder for `run-ios` command
#### Add support for customising iOS build folder for `run-ios` command.  

By default, `DerivedData` folder is used for storing build artifacts, however in CI workflows it's often changed to `ios/build` (for caching purposes) but then you need to use `xcodebuild` directly instead of `yarn react-native run-ios` since currently it doesn't support customising this path which is quite inconvenient. 

## Possible Implementations

TBA

## Related Issues

None

adamTrz:
Summary:

Fixes #1706 

Test Plan:

Clone the branch and run `react-native run-ios --buildFolder "build"` command.
It should run the app as usual but put build artifacts into `ios/build` folder. 

image

### DIFF
--- a/packages/cli-platform-ios/src/commands/runIOS/index.ts
+++ b/packages/cli-platform-ios/src/commands/runIOS/index.ts
@@ -38,6 +38,7 @@ type FlagsT = {
   port: number;
   terminal: string | undefined;
   xcconfig?: string;
+  iosBuildFolder?: string;
 };
 
 function runIOS(_: Array<string>, ctx: Config, args: FlagsT) {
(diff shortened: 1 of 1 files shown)


packages/cli-platform-ios/src/commands/runIOS/index.ts
```diff
@@ -38,6 +38,7 @@ type FlagsT = {
   port: number;
   terminal: string | undefined;
   xcconfig?: string;
+  iosBuildFolder?: string;
```
thymikee, line 41: it's already a "run ios" command, `buildFolder` would be descriptive enough :)

adamTrz: Makes sense 👍

thymikee (changes requested):
Missing the docs entry and we're good!

PR #1730, opened 2022-10-31
fix: select iPhone 14 as a default simulator

mdneyazahmad:
Summary:
---------

`npx react-native run-ios` starts iPhone 13 simulator by default. This PR tries to fix this by selecting iPhone 14 as default simulator.

Test Plan:
----------

`npx react-native run-ios` should run iPhone 14 as default simulator.


thymikee (approved):
Thanks!

PR #1738, opened 2022-11-04
Remove unwanted spaces `add_subdirectory` paths in Android-rncli.cmake

tomekzaw:
Summary:
---------

This PR removes unwanted spaces in paths in `add_subdirectory` command in autogenerated `Android-rncli.cmake`:

obraz

Along with #1734, this should fix building new-arch RN apps for Android on Windows.

Test Plan:
----------

I couldn't find any snapshot tests that would contain `add_subdirectory` so not sure how to prove it works 🤷


thymikee (approved):
Thanks!

thymikee:
We miss the tests for this part unfortunately, so it's pretty manual at this point.

Issue #1741, opened 2022-11-05
`init` command failure should yield a non-zero exit status code


kraenhansen:
## Environment

```
info Fetching system and libraries information...
System:
    OS: macOS 12.6.1
    CPU: (8) arm64 Apple M1
    Memory: 92.16 MB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.18.0 - ~/.nvm/versions/node/v16.18.0/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v16.18.0/bin/yarn
    npm: 8.19.2 - ~/.nvm/versions/node/v16.18.0/bin/npm
    Watchman: Not Found
  Managers:
    CocoaPods: 1.11.3 - /Users/kraen.hansen/.rvm/gems/ruby-2.7.5/bin/pod
  SDKs:
    iOS SDK:
      Platforms: DriverKit 22.1, iOS 16.1, macOS 13.0, tvOS 16.1, watchOS 9.1
    Android SDK:
      API Levels: 28, 31, 33
      Build Tools: 30.0.3, 31.0.0, 33.0.0
      System Images: android-29 | Intel x86 Atom, android-33 | Google APIs ARM 64 v8a
      Android NDK: 21.0.6113669
  IDEs:
    Android Studio: 2021.2 AI-212.5712.43.2112.8815526
    Xcode: 14.1/14B47b - /usr/bin/xcodebuild
  Languages:
    Java: 18.0.1 - /usr/bin/javac
  npmPackages:
    @react-native-community/cli: Not Found
    react: Not Found
    react-native: Not Found
    react-native-macos: Not Found
  npmGlobalPackages:
    *react-native*: Not Found
```

## Description

Any failure from the "init" command will result in a clean (0) exit code.
This is unfortunate if calling this from a script or on CI, as it leaves no chance for the caller to detect the failure mode.

## Reproducible Demo

1. Call the CLI to initialize a new React Native app: `npx react-native init MyApp`
2. Call the CLI to initialize the same app again: `npx react-native init MyApp`.
    This will fail and print the following error:
    > error Cannot initialize new project because directory "MyApp" already exists.
3. Echo'ing the exit code reveals the failure didn't yield a non-zero exit code (as I would expect): `echo $?`

PR #1742, opened 2022-11-05
fix: avoid catching `init` errors and avoid printing their stack

kraenhansen:
Summary:
---------

I experienced the following issue: https://github.com/react-native-community/cli/issues/1741 and decided to fix it too.

I've made four changes:
- I removed the try-catch around the body of the init command, to allow errors to bubble up and be handled by the `handleError` in `src/index.ts`.
- I made all the init related errors extend `CLIError` to avoid printing the stack, since "we know exactly what went wrong" and expects the user to have the means to fix it.
- Fixed a failing e2e test (due to a trailing `\n` in the stderr)
- Added a branch to `jest/helpers.ts`'s `handleTestFailure` to assert a non-zero exit code if expectedFailure is true. This could have caught this (or a similar issue) earlier.

Test Plan:
----------

I suggest following the instructions for reproducing this on the original issue, only this time, notice how the command exits with status code 1 as expected and it still doesn't print the error stack.

### DIFF
--- a/jest/helpers.ts
+++ b/jest/helpers.ts
@@ -204,6 +204,8 @@ ${chalk.bold('args:')}    ${(args || []).join(' ')}
 ${chalk.bold('stderr:')}  ${result.stderr}
 ${chalk.bold('stdout:')}  ${result.stdout}
 ${chalk.bold('code:')}    ${result.code}`);
+  } else if (options.expectedFailure && result.code === 0) {
+    throw new Error("Expected command to fail, but it didn't");
   }
 }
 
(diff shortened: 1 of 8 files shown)


kraenhansen:
I'll go ahead and fix the e2e test.

Looking at these tests, I think adding a branch to [this code](https://github.com/react-native-community/cli/blob/main/jest/helpers.ts#L199-L207) that asserts a non-zero exit code if `expectedFailure` is true, could have caught this (or a similar issue) earlier.

jest/helpers.ts
```diff
@@ -204,6 +204,8 @@ ${chalk.bold('args:')}    ${(args || []).join(' ')}
 ${chalk.bold('stderr:')}  ${result.stderr}
 ${chalk.bold('stdout:')}  ${result.stdout}
 ${chalk.bold('code:')}    ${result.code}`);
+  } else if (options.expectedFailure && result.code === 0) {
+    throw new Error("Expected command to fail, but it didn't");
```
kraenhansen, line 208: This could have caught this (or a similar issue) earlier.

thymikee (approved):
Makes sense, thanks!

PR #1743, opened 2022-11-07
feat: redirect TS template to r-n

### LINKED ISSUE #1740: Redirect react-native-template-typescript to react-native
## Describe the Feature

From RN 0.71 RN will be the default template language, which removes the need for the `react-native-template-typescript` template. To ease the migration, as this way of starting a TS project is referenced all over the docs/tutorials/guides/etc, we could detect users using this template and treat it as `react-native`. Outputting extra warning about what's happening under the hood, so that users are aware.

Schedule for 11.0 release

adamTrz:
Summary:
---------

Fixes #1740 

Test Plan:
----------

Fork the branch and try to `init` new project with `react-native-template-typescript`. 
It should print the warning and default to normal r-n template.

### DIFF
--- a/packages/cli/src/commands/init/init.ts
+++ b/packages/cli/src/commands/init/init.ts
@@ -183,7 +183,7 @@ function createTemplateUri(options: Options, version: string): string {
     logger.warn(
       "Ignoring custom template: 'react-native-template-typescript'. Starting from React Native v0.71 TypeScript is used by default.",
     );
-    return `react-native@${version}`;
+    return 'react-native';
   }
 
   return options.template || `react-native@${version}`;
--- a/packages/cli/src/commands/init/init.ts
+++ b/packages/cli/src/commands/init/init.ts
@@ -21,7 +21,6 @@ import * as PackageManager from '../../tools/packageManager';
 import {installPods} from '@react-native-community/cli-doctor';
 import banner from './banner';
 import TemplateAndVersionError from './errors/TemplateAndVersionError';
-import semver from 'semver';
 
 const DEFAULT_VERSION = 'latest';
 
@@ -180,14 +179,11 @@ function createTemplateUri(options: Options, version: string): string {
   const isTypescriptTemplate =
     options.template === 'react-native-template-typescript';
 
-  if (isTypescriptTemplate && options.template?.includes('@')) {
-    const typescriptTemplateVersion = options.template.split('@')[1];
-    if (semver.gte(typescriptTemplateVersion, '0.71.0')) {
-      logger.warn(
-        "Ignoring custom template: 'react-native-template-typescript'. Starting from React Native v0.71 TypeScript is used by default.",
-      );
-      return `react-native@${version}`;
-    }
+  if (isTypescriptTemplate) {
+    logger.warn(
+      "Ignoring custom template: 'react-native-template-typescript'. Starting from React Native v0.71 TypeScript is used by default.",
+    );
+    return `react-native@${version}`;
   }
 
   return options.template || `react-native@${version}`;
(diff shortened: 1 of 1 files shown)


packages/cli/src/commands/init/init.ts
```diff
@@ -175,13 +176,31 @@ async function installDependencies({
   loader.succeed();
 }
 
+function checkVersionAndTemplate(options: Options, version: string): string {
+  const isTypescriptTemplate = options.template?.includes(
```
thymikee, line 180: Make it strict, not includes. We don't want to target other templates by accident

packages/cli/src/commands/init/init.ts
```diff
@@ -175,13 +176,31 @@ async function installDependencies({
...
+    'react-native-template-typescript',
+  );
+
+  if (isTypescriptTemplate && options.template?.includes('@')) {
+    const typescriptTemplateVersion = options.template.split('@')[1];
+    if (semver.gte(typescriptTemplateVersion, '0.71.0')) {
+      logger.warn(
+        "Starting from verision '0.71' typescript is used by default.\nDiregarding template and using default react-native",
```
p-syche, line 188: You're missing an "S" in "Disregarding" :) 

But honestly for me, the first sentence alone is enough.

thymikee, line 188: There's also a typo in "verision". My suggestion:
```suggestion
        "Ignoring custom template: 'react-native-template-typescript'. Starting from React Native v0.71 TypeScript is used by default.",
```

packages/cli/src/commands/init/init.ts
```diff
@@ -175,13 +176,31 @@ async function installDependencies({
...
+}
+
 async function createProject(
   projectName: string,
   directory: string,
   version: string,
   options: Options,
 ) {
-  const templateUri = options.template || `react-native@${version}`;
+  const templateUri = checkVersionAndTemplate(options, version);
```
thymikee, line 203: let's rename this fn to `createTemplateUri`

MateWW (approved):
I've tested and it works perfectly!
Thanks! 🚀🥳
Screenshot 2022-11-07 at 14 20 59

packages/cli/src/commands/init/init.ts
```diff
@@ -175,13 +176,31 @@ async function installDependencies({
...
 }
 
+function createTemplateUri(options: Options, version: string): string {
+  const isTypescriptTemplate = options.template?.includes(
+    'react-native-template-typescript',
+  );
+
+  if (isTypescriptTemplate && options.template?.includes('@')) {
+    const typescriptTemplateVersion = options.template.split('@')[1];
+    if (semver.gte(typescriptTemplateVersion, '0.71.0')) {
```
radko93, line 186: Not sure if I get this right. Is it the RN version inside the template? The TS template does not follow RN versioning. You can look up what are the versions here https://github.com/react-native-community/react-native-template-typescript#react-native--template-version

If I'm wrong just ignore this comment.

adamTrz: > Not sure if I get this right. Is it the RN version inside the template? The TS template does not follow RN versioning. You can look up what are the versions here https://github.com/react-native-community/react-native-template-typescript#react-native--template-version
> 
> If I'm wrong just ignore this comment.

No, you're 💯 correct.
Actually now I'm tending to just doing the check if there's TS template passed but without any version
If users go with `init --template react-native-template-typescript` once r-n 0.71 is released we will be using `react-native`. But if they pass version with template I would assume that they want lower react native version but with TS so we will allow that? 
wdyt?

radko93: Yeah, looks like that's the way to go!

radko93 (approved):

TMisiukiewicz:
image

tested the expected behavior, works correctly ✅

TMisiukiewicz (approved):

TMisiukiewicz (approved):

PR #1748, opened 2022-11-15
fix: move bundle install above pod version check

MateWW:
Summary:
---------

This change should fix the issue with falling `react-native init` for people with `rvm` installed.
I believe this logic was invalid since we have `Gemfile` in the template root responsible for installing cocoa pods it seems to have no sense to verify its version before.

Test Plan:
----------

- clone repo
- run `yarn build`
- `node ./packages/cli/build/bin.js init ReactNativeUnicorn`

Test scenarios:
- [x] with `rvm` installed
- [ ] without ruby version managers
- [ ] with other ruby version managers

### DIFF
--- a/packages/cli-doctor/src/tools/installPods.ts
+++ b/packages/cli-doctor/src/tools/installPods.ts
@@ -183,6 +183,10 @@ async function installPods({
       return;
     }
 
+    if (fs.existsSync('../Gemfile')) {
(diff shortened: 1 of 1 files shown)


thymikee (approved):
Yeah, makes sense to run bundler first 😅

adamTrz (approved):

Issue #1750, opened 2022-11-17
CLI fails when dev dependencies are not installed


elliotdickison:
## Environment

```
System:
    OS: macOS 12.6.1
    CPU: (8) arm64 Apple M1
    Memory: 588.30 MB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.15.0 - ~/.volta/tools/image/node/16.15.0/bin/node
    Yarn: 1.22.17 - ~/.volta/tools/image/yarn/1.22.17/bin/yarn
    npm: 8.5.5 - ~/.volta/tools/image/node/16.15.0/bin/npm
    Watchman: Not Found
  Managers:
    CocoaPods: 1.11.3 - /opt/homebrew/bin/pod
  SDKs:
    iOS SDK:
      Platforms: DriverKit 22.1, iOS 16.1, macOS 13.0, tvOS 16.1, watchOS 9.1
    Android SDK:
      API Levels: 31, 33
      Build Tools: 28.0.3, 29.0.2, 30.0.2, 30.0.3, 31.0.0
      System Images: android-31 | ARM 64 v8a, android-S | Google APIs ARM 64 v8a, android-S | Google Play ARM 64 v8a
      Android NDK: Not Found
  IDEs:
    Android Studio: 2021.1 AI-211.7628.21.2111.8193401
    Xcode: 14.1/14B47b - /usr/bin/xcodebuild
  Languages:
    Java: 11.0.11 - /usr/bin/javac
  npmPackages:
    @react-native-community/cli: Not Found
    react: 18.1.0 => 18.1.0
    react-native: 0.70.5 => 0.70.5
    react-native-macos: Not Found
  npmGlobalPackages:
    *react-native*: Not Found
```

## Description

The CLI fails with the error `Failed to load configuration of your project.` if any development dependencies listed in `package.json` are not installed (even if those dependencies are not actually referenced anywhere in the project).

Potentially related issues: #1169 #1681

## Reproducible Demo

1. `npx create-expo-app test-rn-app`
2. `cd ./test-rn-app`
5. `npm install left-pad --save-dev`
6. `npm install --production`
7. `npx react-native info` (fails)

elliotdickison:
Here's my temporary workaround (wrapping the config logic that searches through dependencies in a try/catch):

```
diff --git a/build/loadConfig.js b/build/loadConfig.js
[...]

PR #1752, opened 2022-11-18
fix: run cli when dev dependencies are not installed

TMisiukiewicz:
Summary:
---------

Relates to #1750 
When user is running `yarn install --production` command, the `devDependencies` are not installed. Because of that, they cannot be found in the `node_modules` when loading  CLI's configuration and it ends up with an error `Failed to load configuration of your project.` while running any command. This PR prevents CLI from failing when `devDependencies` are not installed.

Test Plan:
----------

1. Clone the repo and checkout this branch
2. Run `yarn install` in the root catalog
3. Create new React Native project
4. Add any dev dependency, e.g. `yarn add left-pad --dev`
5. Run `yarn install --production`
6. Run `yarn start` - at this stage the command should fail
7. Link packages using [this setup guide](https://github.com/react-native-community/cli/blob/main/CONTRIBUTING.md#setup)
8. Run `yarn start` - at this stage the command should work

Can be tested with other CLI commands as well

### DIFF
--- a/packages/cli-config/src/loadConfig.ts
+++ b/packages/cli-config/src/loadConfig.ts
@@ -98,32 +98,38 @@ function loadConfig(projectRoot: string = findProjectRoot()): Config {
     const localDependencyRoot =
       userConfig.dependencies[dependencyName] &&
       userConfig.dependencies[dependencyName].root;
-    let root =
-      localDependencyRoot || resolveNodeModuleDir(projectRoot, dependencyName);
-    let config = readDependencyConfigFromDisk(root, dependencyName);
 
-    const isPlatform = Object.keys(config.platforms).length > 0;
+    try {
+      let root =
+        localDependencyRoot ||
+        resolveNodeModuleDir(projectRoot, dependencyName);
+      let config = readDependencyConfigFromDisk(root, dependencyName);
 
-    return assign({}, acc, {
-      dependencies: assign({}, acc.dependencies, {
-        get [dependencyName](): DependencyConfig {
-          return getDependencyConfig(
-            root,
-            dependencyName,
-            finalConfig,
-            config,
-            userConfig,
-            isPlatform,
-          );
+      const isPlatform = Object.keys(config.platforms).length > 0;
+
+      return assign({}, acc, {
+        dependencies: assign({}, acc.dependencies, {
+          get [dependencyName](): DependencyConfig {
+            return getDependencyConfig(
+              root,
+              dependencyName,
+              finalConfig,
+              config,
+              userConfig,
+              isPlatform,
+            );
+          },
+        }),
+        commands: [...acc.commands, ...config.commands],
+        platforms: {
+          ...acc.platforms,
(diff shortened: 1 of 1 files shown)


adamTrz:
If I recall correctly the issue was present because this guy was throwing an error... 

https://github.com/react-native-community/cli/blob/464c07be08c53292aa2e3d139d01376c718fe211/packages/cli-tools/src/resolveNodeModuleDir.ts#L11-L17
 
If we wrap the whole block I think it will exit at the first error resulting in `finalConfig` being only partially computed. 
Could you please check if that's true? 🙏 
If so, then we would have to come up with different solution I think.

TMisiukiewicz:
> If I recall correctly the issue was present because this guy was throwing an error...
> 
> https://github.com/react-native-community/cli/blob/464c07be08c53292aa2e3d139d01376c718fe211/packages/cli-tools/src/resolveNodeModuleDir.ts#L11-L17
> 
> If we wrap the whole block I think it will exit at the first error resulting in `finalConfig` being only partially computed. Could you please check if that's true? 🙏 If so, then we would have to come up with different solution I think.

checked that and looks like it works correctly, added some more dev dependencies and `finalConfig.dependencies` for `--production=false` is:

```
{
  react: [Getter],
  'react-native': [Getter],
  '@babel/core': [Getter],
  '@babel/runtime': [Getter],
  '@react-native-community/eslint-config': [Getter],
  '@types/axios': [Getter],
  'babel-jest': [Getter],
  'cross-env': [Getter],
  eslint: [Getter],
  'http-status-codes': [Getter],
  jest: [Getter],
  'left-pad': [Getter],
  'metro-react-native-babel-preset': [Getter],
  'react-native-dotenv': [Getter],
  'react-native-svg-transformer': [Getter],
  'react-native-svg-web': [Getter],
  'react-test-renderer': [Getter]
}
```

and for `--production=true`:

```
{
  react: [Getter],
  'react-native': [Getter],
  '@babel/core': [Getter],
  '@babel/runtime': [Getter],
  'metro-react-native-babel-preset': [Getter]
}
```

IMO in this case `finalConfig` computes correctly

adamTrz (approved):

thymikee:
What's the failure of e.g. `yarn start` look like now? This function failed to present you a detailed error. If the user is instructed to install the dependencies, then I think this should be ok.

TMisiukiewicz:
@thymikee right now the failure ends up only with `Failed to load configuration of your project.` message. 
Just realised that after that change it's possible to run metro server without installing any dependencies using `react-native start` command - not sure if it should work that way, since it is still impossible to use `run-ios` and `run-android`? In this case any warning for the user about dependencies would appear for each cli command, including `start`

TMisiukiewicz:
For now we should avoid adding this warning because it would appear on each command when user has no dependencies installed/only prod dependencies installed. Should be verified in the future

thymikee (approved):
Let's give it a try 👍🏼

PR #1758, opened 2022-11-25
increase number of operations per run in stale action

TMisiukiewicz:
Summary:
---------

Noticed that stale action has a default value of operations per run set to 30. In that case, each time the action runs, it goes through the same 30 issues. Increasing the value solves this problem and should go through all the issues and PRs next time.

Test Plan:
----------

Get it merged and run a job manually or wait until cron runs it automatically, then compare the result with the [action from the past](https://github.com/react-native-community/cli/actions/runs/3537470057/jobs/5937467704#step:2:698)


thymikee:
Ok!

Issue #1762, opened 2022-12-03
npx react-native init MyApp --template on custom template causes: Error: ENOENT: no such file or directory


leinstay:
## Environment
I'm using npx, react-native not installed globally. Problem appears on Windows 10 and Debian.

## Description

Can't initialize project from my own template while ios folder is present. I receive this error in different environments and with different --template values (git, npm, local storage)

![Screenshot_1](https://user-images.githubusercontent.com/8215580/205453393-e2dac99f-0ec8-43a2-967e-119053840aeb.png)
![Screenshot_2](https://user-images.githubusercontent.com/8215580/205453394-ac651ea0-bc8a-4c3a-8dc5-1a265de67aff.png)

From what I see when I compare:
```bash
npx react-native init MyApp --template react-native-template-typescript --verbose
# and
npx react-native init MyApp --template rnbnr-template --verbose
```

In both cases script runs *replacing* 2 times for each file (which is strange), but as for *renaming* it runs 2 times on my template and 1 time on community template. So it is trying to rename file which was already renamed previously.

![Screenshot_3](https://user-images.githubusercontent.com/8215580/205453978-daf47126-a2f4-435a-a727-79f3633cd097.png)
![Screenshot_4](https://user-images.githubusercontent.com/8215580/205453979-0898d03b-ac0f-4f5a-b6cf-ba7d550a012b.png)

## Reproducible Demo

```bash
# without ios folder (error)
npx react-native init test1 --template rnbnr-template@0.0.5 --verbose
# with ios folder (ok)
npx react-native init test2 --template rnbnr-template@0.06 --verbose
```

Project files could be found here:
https://github.com/leinstay/rnbnr-template

PR #1763, opened 2022-12-04
fix: avoid renaming a file if it is already renamed

leinstay:
Fixes https://github.com/react-native-community/cli/issues/1762

Summary:
---------

When using `init` command with `--template` parameter and a custom template in which `placeholderName` as well as the attached files (such as .xcscheme) are written in lowercase, an error occurs. 

205453978-daf47126-a2f4-435a-a727-79f3633cd097

This happens because both `shouldRenameFile(filePath, placeholderName)` and `shouldRenameFile(filePath, placeholderName.toLowerCase())` return True and `renameFile()` tries to rename non-existinf file. This commit implements a workaround so that the lowercase function is only executed if the normal function returns False.

Test Plan:
----------
Given that the change is minimal, I think it is enough that I have tested it on several templates: 

yarn react-native init App --template rnbnr-template@0.0.5 --npm --skip-install --verbose
yarn react-native init App --template react-native-template-typescript --npm --skip-install --verbose

In both cases the application was initialized successfully and the files and folders inside /ios/ were renamed properly.

### DIFF
--- a/packages/cli/src/commands/init/editTemplate.ts
+++ b/packages/cli/src/commands/init/editTemplate.ts
@@ -101,8 +101,7 @@ export async function changePlaceholderInTemplate({
     }
     if (shouldRenameFile(filePath, placeholderName)) {
       await renameFile(filePath, placeholderName, projectName);
-    }
-    if (shouldRenameFile(filePath, placeholderName.toLowerCase())) {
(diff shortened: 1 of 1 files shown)


adamTrz (approved):

thymikee (approved):

thymikee:
Nice catch, thanks!

PR #1773, opened 2022-12-07
feat: read build configuration from scheme file

### LINKED ISSUE #1454: run-ios command does not look for the default build configuration when --scheme is specified
## Describe the Feature
Currently, when we use the flag --scheme to specify a scheme to be run while invoking the react-native run-ios command, the cli is not looking for the build configuration that is associated with the specified scheme and uses the 'Debug' configuration by default.

As we can specify a default build configuration for each scheme in Xcode, the cli should use that configuration as the default instead of using the 'Debug' configuration.

## Possible Implementations
Take a look at fastlane's implementation in gym action source code (file code_signing_mapping.rb, function 'detect_configuration_for_archive'). What it does is, if a configuration is not specified by the user on the command-line, then it will find the corresponding build configuration for the archive action of the specified scheme. The cli could use the build configuration used for the Run action of the specified scheme.

TMisiukiewicz:
Summary:
---------

Closes #1454 
This PR implements using build configuration value associated to scheme provided with a `--scheme` flag.

Test Plan:
----------
Create new scheme in XCode with build configuration set to Release

In the app:

- [x]  run `yarn ios` - it should use default scheme and Debug configuration
- [x] run `yarn ios --scheme YourSchemeName` - it should use newly created scheme and use Release configuration
- [x] run `yarn ios --scheme YourSchemeName --configuration Debug` - it should use newly created scheme and use Debug configuration
- [x] run `yarn ios --configuration Release` - it should use default scheme with Release configuration

When running commands, watch for `info Building...` message and check if configuration and scheme values are properly set in `xcodebuild` command

### DIFF
--- a/packages/cli-platform-ios/src/commands/runIOS/index.ts
+++ b/packages/cli-platform-ios/src/commands/runIOS/index.ts
@@ -6,7 +6,7 @@
  *
  */
 
-import child_process, {SpawnOptionsWithoutStdio} from 'child_process';
+import child_process from 'child_process';
 import path from 'path';
 import fs from 'fs';
 import {XMLParser} from 'fast-xml-parser';
@@ -538,34 +538,6 @@ function printFoundDevices(devices: Array<Device>) {
   ].join('\n');
 }
 
-function getProcessOptions({
-  packager,
-  terminal,
-  port,
-}: {
-  packager: boolean;
-  terminal: string | undefined;
-  port: number;
-}): SpawnOptionsWithoutStdio {
-  if (packager) {
-    return {
-      env: {
-        ...process.env,
-        RCT_TERMINAL: terminal,
-        RCT_METRO_PORT: port.toString(),
-      },
-    };
-  }
-
-  return {
-    env: {
-      ...process.env,
-      RCT_TERMINAL: terminal,
-      RCT_NO_LAUNCH_PACKAGER: 'true',
-    },
-  };
-}
-
 function getBuildConfigurationFromXcScheme(
   {scheme, configuration}: FlagsT,
   sourceDir: string,
@@ -594,7 +566,9 @@ function getBuildConfigurationFromXcScheme(
       return Scheme.LaunchAction['@_buildConfiguration'];
     }
   } catch {
-    logger.error(`Could not find scheme ${scheme}.`);
+    throw new CLIError(
+      `Could not find scheme ${scheme}. Please make sure the schema you want to run exists.`,
+    );
   }
 
   return configuration;
--- a/packages/cli-platform-ios/src/commands/runIOS/index.ts
+++ b/packages/cli-platform-ios/src/commands/runIOS/index.ts
@@ -9,7 +9,6 @@
 import child_process from 'child_process';
 import path from 'path';
 import fs from 'fs';
-import {XMLParser} from 'fast-xml-parser';
 import chalk from 'chalk';
 import {Config, IOSProjectInfo} from '@react-native-community/cli-types';
 import {getDestinationSimulator} from '../../tools/getDestinationSimulator';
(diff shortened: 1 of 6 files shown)


adamTrz (commented):
Did a quick check and it looks good:
```bash
❯ node ../cli/packages/cli/build/bin.js run-ios
info Found Xcode workspace "Bar.xcworkspace"
info Building (using "xcodebuild -workspace Bar.xcworkspace -configuration Debug -scheme Bar -destination id=D83F179F-6C0B-45BA-9104-45397BA3FFB9")

❯ node ../cli/packages/cli/build/bin.js run-ios --scheme Baz
info Found Xcode workspace "Bar.xcworkspace"
info Building (using "xcodebuild -workspace Bar.xcworkspace -configuration Release -scheme Baz -destination id=D83F179F-6C0B-45BA-9104-45397BA3FFB9")

❯ node ../cli/packages/cli/build/bin.js run-ios --scheme Bar --configuration Debug
info Found Xcode workspace "Bar.xcworkspace"
info Building (using "xcodebuild -workspace Bar.xcworkspace -configuration Debug -scheme Bar -destination id=D83F179F-6C0B-45BA-9104-45397BA3FFB9")

❯ node ../cli/packages/cli/build/bin.js run-ios --configuration Release           
info Found Xcode workspace "Bar.xcworkspace"
info Building (using "xcodebuild -workspace Bar.xcworkspace -configuration Release -scheme Bar -destination id=D83F179F-6C0B-45BA-9104-45397BA3FFB9")
```

packages/cli-platform-ios/src/commands/runIOS/index.ts
```diff
@@ -573,6 +579,50 @@ function getProcessOptions({
...
+          encoding: 'utf-8',
+        },
+      );
+
+      const {Scheme} = xmlParser.parse(xmlScheme);
+
+      return Scheme.LaunchAction['@_buildConfiguration'];
+    }
+  } catch {
+    logger.error(`Could not find scheme ${scheme}.`);
```
adamTrz, line 610: I think this should throw. 
When you pass dummy scheme it will log error but still proceed with the build:
```bash 
❯ node ../cli/packages/cli/build/bin.js run-ios --scheme Foo                      
error Could not find scheme Foo.
info Found Xcode workspace "Bar.xcworkspace"
info Building (using "xcodebuild -workspace Bar.xcworkspace -configuration  -scheme Foo -destination id=D83F179F-6C0B-45BA-9104-45397BA3FFB9")
```

TMisiukiewicz: > I think this should throw.
> When you pass dummy scheme it will log error but still proceed with the build:

Thanks for catching that, done!

> And there's also a case when user passes good scheme (or no scheme at all and we select default one) but not existing configuration... Do you think we can detect this as well?
> (Although this is neither scope of this PR and is also working like that with current state of CLI, so we can fix it as follow up or move responsibility to pass good configuration to the user 😉 )

If i understand correctly, the XCode allows only two configurations: Debug and Release. I think the easiest way to detect it is to just check if configuration value is one of those and throw error/run in Debug as default if it's not correct. But agree - it's not in the scope of this PR so we can fix it separately if we decide what solution fits us best

adamTrz (approved):

packages/cli-platform-ios/src/commands/runIOS/index.ts
adamTrz: > If i understand correctly, the XCode allows only two configurations: Debug and Release

Not really, you can have custom ones I think as well :)

liamjones: Yes, you can have others

TMisiukiewicz:
Added support for `build-ios` command and updated code to use `--mode` flag instead of `--configuration`

packages/cli-platform-ios/src/tools/getBuildConfigurationFromXcScheme.ts
```diff
@@ -0,0 +1,43 @@
...
+          'xcshareddata',
+          'xcschemes',
+          `${scheme}.xcscheme`,
+        ),
+        {
+          encoding: 'utf-8',
+        },
+      );
+
+      const {Scheme} = xmlParser.parse(xmlScheme);
```
thymikee, line 32: did you get a chance to asses how much extra time parsing this XML adds to a sample project?

TMisiukiewicz: good idea, thanks - looks like on my machine the first read of the schema file takes ~5.5ms and every next try is around ~2.2ms - seems quick

thymikee (approved):

PR #1774, opened 2022-12-07
chore: added release-drafter GH Action

adamTrz:
Summary:
---------

Added Release Drafter GH Action to create nicer release notes. 

Test Plan:
----------

Once merged we need to run test release and see how it goes 💪


thymikee:
Release Drafter CI failed, is this expected?

thymikee (approved):
Let's give it a try. Please merge whenever you like

adamTrz:
> Release Drafter CI failed, is this expected?

I think yes. As docs state, config file needs to be at default branch, since it is not yet there it fails. 
Only surprised a bit that it launched already :D

PR #1778, opened 2022-12-09
fix: check cocoapods version in doctor

### LINKED ISSUE #1112: Check CocoaPods version in doctor
## Describe the Feature

Currently we are only checking if CocoaPods is installed, it would be useful to check the version.

As seem in react-native-community/upgrade-support#20 there may be errors tied to a CocoaPods version which can possibly be fixed by having a minimum supported version.

## Possible Implementations

Check that CocoaPods is installed and has a version `>= 1.9.1` somewhere around [here](https://github.com/react-native-community/cli/blob/master/packages/cli/src/commands/doctor/healthchecks/cocoaPods.ts#L17), [Node.js health check](https://github.com/react-native-community/cli/blob/master/packages/cli/src/commands/doctor/healthchecks/nodeJS.ts) can be used as a reference on how to implement this.

adamTrz:
Summary:
---------

Checks for minimal version on CocoaPods in `doctor`

Fixes #1112

### DIFF
--- a/packages/cli-doctor/src/tools/versionRanges.ts
+++ b/packages/cli-doctor/src/tools/versionRanges.ts
@@ -9,4 +9,5 @@ export default {
   ANDROID_NDK: '>= 19.x',
   // iOS
   XCODE: '>= 12.x',
+  COCOAPODS: '>= 1.9.1',
 };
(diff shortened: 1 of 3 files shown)


packages/cli-doctor/src/tools/versionRanges.ts
```diff
@@ -9,4 +9,5 @@ export default {
   ANDROID_NDK: '>= 19.x',
   // iOS
   XCODE: '>= 12.x',
+  COCOAPODS: '>= 1.9.1',
```
thymikee, line 12: let's verify the latest recommended version. I recall it was 1.10.x a few versions ago

thymikee (approved):
Double check the min version and we're good. Nice we can rely more on envinfo instead of checking whether the command exists ourselves

PR #1779, opened 2022-12-15
feat: run app on iOS and Android with a hotkey

TMisiukiewicz:
Summary:
---------

This PR introduces running an app from terminal using Metro watch mode. The `run-ios` and `run-android` commands are running when `i` or `a` keys are pressed in the Terminal.

iOS:

https://user-images.githubusercontent.com/13985840/207847756-75ce1e13-00b9-425e-a015-1feee21bc4fe.mov

Android:

https://user-images.githubusercontent.com/13985840/207847801-9c13f152-bc4f-4efd-bf92-70fd840a6a5e.mov

Test Plan:
----------
2. Run `yarn start --watchFolders path/to/cloned/cli`
3. Press `a` to run the app on Android
4. Press `i` to run the app on iOS

### DIFF
--- a/packages/cli-plugin-metro/src/commands/start/watchMode.ts
+++ b/packages/cli-plugin-metro/src/commands/start/watchMode.ts
@@ -1,9 +1,16 @@
 import readline from 'readline';
+import fs from 'fs';
 import {logger, hookStdout} from '@react-native-community/cli-tools';
+import execa from 'execa';
+import chalk from 'chalk';
 
 function printWatchModeInstructions() {
   logger.log(
-    '\n\nTo reload the app press "r"\nTo open developer menu press "d"',
+    `${chalk.bold('r')} - reload the app\n${chalk.bold(
+      'd',
+    )} - open developer menu\n${chalk.bold('i')} - run on iOS\n${chalk.bold(
+      'a',
+    )} - run on Android`,
   );
 }
 
@@ -47,6 +54,14 @@ function enableWatchMode(messageSocket: any) {
     } else if (name === 'd') {
       messageSocket.broadcast('devMenu', null);
       logger.info('Opening developer menu...');
+    } else if (name === 'i' || name === 'a') {
+      const isUsingYarn = fs
+        .readdirSync(`${process.cwd()}`)
+        .includes('yarn.lock');
+      logger.info(`Opening the app on ${name === 'i' ? 'iOS' : 'Android'}...`);
+      execa(isUsingYarn ? 'yarn' : 'npm run', [
+        name === 'i' ? 'ios' : 'android',
+      ]).stdout?.pipe(process.stdout);
     } else {
       console.log(_key);
     }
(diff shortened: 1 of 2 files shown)


packages/cli-plugin-metro/src/commands/start/watchMode.ts
```diff
@@ -47,6 +54,14 @@ function enableWatchMode(messageSocket: any) {
     } else if (name === 'd') {
       messageSocket.broadcast('devMenu', null);
       logger.info('Opening developer menu...');
+    } else if (name === 'i' || name === 'a') {
+      const isUsingYarn = fs
+        .readdirSync(`${process.cwd()}`)
+        .includes('yarn.lock');
+      logger.info(`Opening the app on ${name === 'i' ? 'iOS' : 'Android'}...`);
+      execa(isUsingYarn ? 'yarn' : 'npm run', [
```
thymikee, line 62: We're using `npx` to run regardless npm/yarn

liamjones:
Oh, nice! ❤️ 

Is there potentially a way for "default command line options to use" to be exposed to these commands?

For example, in our mono-repo we have app package.jsons setup like:

```json
{
    "scripts": {
        "android": "react-native run-android --no-jetifier --appIdSuffix 'dev' --port 9081 --active-arch-only",
        "ios": "react-native run-ios --port 9081"
    }
}
```

So I'm assuming this wouldn't work for us currently - it'd be like running `react-native run-android`, etc with no parameters?

TMisiukiewicz:
> Oh, nice! ❤️
> 
> Is there potentially a way for "default command line options to use" to be exposed to these commands?
> 
> For example, in our mono-repo we have app package.jsons setup like:
> 
> ```json
> {
>     "scripts": {
>         "android": "react-native run-android --no-jetifier --appIdSuffix 'dev' --port 9081 --active-arch-only",
>         "ios": "react-native run-ios --port 9081"
>     }
> }
> ```
> 
> So I'm assuming this wouldn't work for us currently - it'd be like running `react-native run-android`, etc with no parameters?

hi @liamjones , thanks for your comment! Since the user might not have `android` and `ios` commands at all, using `npx react-native run-ios/android` under the hood seems like a better option there. Unfortunately, it also means the script is going to run with the default options. I'm going to think if there is any possibility to make it work the way you want, happy to hear any suggestions if anyone has an idea!

liamjones:
@TMisiukiewicz yeah, I appreciate most people probably(?) just use the default commands. 😄 

> I'm going to think if there is any possibility to make it work the way you want, happy to hear any suggestions if anyone has an idea!

Maybe defaults could be configured via the [`react-native.config.js`](https://github.com/react-native-community/cli/blob/main/docs/configuration.md)? I can see that requiring more discussion/being a bigger change so maybe that should be saved for another time/PR? 🤔 I don't want to hold this up and potentially generate a bunch of extra work for something that will already be useful to a lot of people in its current state.

adamTrz:
> @TMisiukiewicz yeah, I appreciate most people probably(?) just use the default commands. 😄
> 
> > I'm going to think if there is any possibility to make it work the way you want, happy to hear any suggestions if anyone has an idea!
> 
> Maybe defaults could be configured via the [`react-native.config.js`](https://github.com/react-native-community/cli/blob/main/docs/configuration.md)? I can see that requiring more discussion/being a bigger change so maybe that should be saved for another time/PR? 🤔 I don't want to hold this up and potentially generate a bunch of extra work for something that will already be useful to a lot of people in its current state.

Super agree this would be useful :)

adamTrz (approved):

Issue #1768, opened 2022-12-05
`npx react-native init` fails with an unclear error on unsupported Node versions


motiz88:
See https://github.com/facebook/react-native/issues/35561.

## Environment

Mac, using Node 12. `react-native info` is broken (see below)

## Description

When I try to initialise a React Native project on an old, unsupported version of Node, it fails with this unclear error message:

```
$ npx react-native init RN0710RC3 --version 0.71.0-rc.3


npx: installed 601 in 15.314s
Unexpected token '?'
```

**Expected behaviour:** There should be a message saying that my Node version is unsupported and which minimum version I should upgrade to.

`npx react-native info` itself fails with the same output:

```
$ npx react-native info
npx: installed 601 in 22.114s
Unexpected token '?'
```

## Reproducible Demo

1. Install an old version of Node (e.g. `nvm install 12`)
2. `npx react-native init RN0710RC3 --version 0.71.0-rc.3`

PR #1782, opened 2022-12-20
feat: add condition for min node version

szymonrybczak:
Summary:
---------

Closes https://github.com/react-native-community/cli/issues/1768

Test Plan:
----------

2. Download old version of `node` e.g. `v12.22.12`
3. Run `node packages/cli/build/bin.js start` inside repo and, you should get an error `You need at least Node 14 to run CLI.`

### DIFF
--- a/packages/cli/src/bin.ts
+++ b/packages/cli/src/bin.ts
@@ -3,8 +3,9 @@
 import './tools/gracefulifyFs';
 import semver from 'semver';
 import chalk from 'chalk';
+import {versionRanges} from '@react-native-community/cli-doctor';
 
-if (semver.satisfies(process.version, '>=14')) {
+if (semver.satisfies(process.version, versionRanges.NODE_JS)) {
   const {run} = require('./');
   run();
 } else {
--- a/packages/cli/src/bin.ts
+++ b/packages/cli/src/bin.ts
@@ -1,7 +1,15 @@
 #!/usr/bin/env node
 
 import './tools/gracefulifyFs';
+import semver from 'semver';
+import execa from 'execa';
+import chalk from 'chalk';
 
-import {run} from './';
+const nodeVersion = execa.sync('node', ['--version'], {}).stdout;
 
-run();
+if (semver.satisfies(nodeVersion, '>=14')) {
+  const {run} = require('./');
+  run();
+} else {
+  console.error(`${chalk.red('You need at least Node 14 to run CLI.')}`);
+}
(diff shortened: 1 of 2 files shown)


packages/cli/src/bin.ts
```diff
@@ -1,7 +1,15 @@
 #!/usr/bin/env node
 
 import './tools/gracefulifyFs';
+import semver from 'semver';
+import execa from 'execa';
+import chalk from 'chalk';
 
-import {run} from './';
+const nodeVersion = execa.sync('node', ['--version'], {}).stdout;
```
thymikee, line 8: you can get node's version from `process.version`

thymikee (changes requested):
I think it's weird that npx doesn't respect `engines` entry in package.json. Until it does, I'm good with this check, just let's make it less of a perf impact

packages/cli/src/bin.ts
```diff
@@ -1,7 +1,12 @@
 #!/usr/bin/env node
 
 import './tools/gracefulifyFs';
+import semver from 'semver';
+import chalk from 'chalk';
 
-import {run} from './';
-
-run();
+if (semver.satisfies(process.version, '>=14')) {
```
adamTrz, line 7: Maybe worth taking it from `versionRanges.ts` file?  🤔 
https://github.com/react-native-community/cli/blob/b06993e72dccc2658551b76a90b9f063e9cb8731/packages/cli-doctor/src/tools/versionRanges.ts#L3

thymikee: even better

packages/cli/src/bin.ts
```diff
@@ -1,7 +1,12 @@
...
+  const {run} = require('./');
+  run();
+} else {
+  console.error(`${chalk.red('You need at least Node 14 to run CLI.')}`);
```
thymikee, line 11: How about something like this:
> React Native needs Node.js >=14. You're currently on version 12.3.4. Please upgrade Node.js to a supported version and try again.
```suggestion
  console.error(`${chalk.red('React Native needs Node.js ${nodeVersionRange}. You're currently on version ${process.version}. Please upgrade Node.js to a supported version and try again.')}`);
```

packages/cli/src/bin.ts
szymonrybczak: fixed ✅

packages/cli/src/bin.ts
szymonrybczak: Looks good. Updated 😄

thymikee (approved):
Can you confirm this still works after extra imports? Ideally we'd have a test for that but It may be tricky to come up with a good one. Mocking process.version will not take module resolution into account and otherwise we'd need to run this through node@12

szymonrybczak:
Yeah, I tested that with node@12 and it still works 😄

adamTrz (approved):

PR #1786, opened 2022-12-31
Remove useless simulator check in parseXctraceIOSDevicesList

simonkoeck:
Summary:
---------
I was just sitting 4h in front of the computer debugging why my metro bundler couldn't find the device.  :D

Although the `xcrun xctrace list devices` returned me all the correct devices.
```bash
== Devices ==
Pauls & Simons Mac (*****)
iPhone von Simon (15.4.1) (****)
```

Right before that I uninstalled all my simulators. Now because of the simulator check, the function returned an empty array.

### DIFF
--- a/packages/cli-platform-ios/src/commands/runIOS/parseXctraceIOSDevicesList.ts
+++ b/packages/cli-platform-ios/src/commands/runIOS/parseXctraceIOSDevicesList.ts
@@ -26,9 +26,7 @@ import {Device} from '../../types';
 function parseIOSDevicesList(text: string): Array<Device> {
   const devices: Array<Device> = [];
   let isSimulator = false;
-  if (text.indexOf('== Simulators ==') === -1) {
-    return [];
-  }
+
   text.split('\n').forEach((line) => {
     if (line === '== Simulators ==') {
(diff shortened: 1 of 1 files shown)


adamTrz:
Thank's for the PR @simonkoeck 
Looks like super edge case having deleted all your simulators, but don't see it would brake anything 😉

simonkoeck:
Fixed :)

adamTrz (approved):

adamTrz:
One question @simonkoeck - what would happen if there're also no physical devices connected?

simonkoeck:
> One question @simonkoeck - what would happen if there're also no physical devices connected?

The function would still return the correct simulators. Only if no simulators and no devices are connected, the function returns an empty array

adamTrz:
> > One question @simonkoeck - what would happen if there're also no physical devices connected?
> 
> The function would still return the correct simulators. Only if no simulators and no devices are connected, the function returns an empty array

Yeah yeah, I know that 😉 
Sorry, not being more precise, would would happen for whole `run-ios` command? Since we don't have any devices nor simulators available it would definitely fail, but where?

PR #1793, opened 2023-01-13
refactor: add missing type definitions

TMisiukiewicz:
Summary:
---------

Added some of the missing type definitions and removed `@ts-ignore` comments where relevant. These changes do not include `metro`, as it probably requires updating types definitions in DefinitelyTyped repo first - together with @szymonrybczak we have a plan to investigate and update it as well

Test Plan:
----------
n/a


thymikee (approved):
Thanks!

PR #1806, opened 2023-01-23
fix: passing simulator via name

szymonrybczak:
Summary:
---------
Fixes: https://github.com/facebook/react-native/issues/35907

Test Plan:
----------
`yarn react-native run-ios --simulator "iPhone 14"` -> should run app on iPhone 14
`yarn react-native run-ios --device "iPhone Name"` -> should run app on iPhone Name
`yarn react-native run-ios --udid "xyz"` -> should run app on device with id xyz
`yarn react-native run-ios` -> should run on all booted devices

### DIFF
--- a/packages/cli-platform-ios/src/commands/runIOS/index.ts
+++ b/packages/cli-platform-ios/src/commands/runIOS/index.ts
@@ -236,7 +236,7 @@ async function runOnSimulator(
       },
       findMatchingSimulator(simulators, {
         ...args,
-        simulator: args.simulator ?? defaultSimulator,
+        simulator: args.simulator ?? fallbackSimulators[0],
       }),
     );
   }
--- a/packages/cli-platform-ios/src/commands/runIOS/index.ts
+++ b/packages/cli-platform-ios/src/commands/runIOS/index.ts
@@ -85,7 +85,7 @@ function runIOS(_: Array<string>, ctx: Config, args: FlagsT) {
 
   const devices = getDevices();
 
-  if (!args.device && !args.udid) {
+  if (!args.device && !args.udid && !args.simulator) {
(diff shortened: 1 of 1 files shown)


packages/cli-platform-ios/src/commands/runIOS/index.ts
```diff
@@ -209,6 +211,8 @@ async function runOnSimulator(
   } else {
     const simulators = getSimulators();
 
+    const defaultSimulator = 'iPhone 14';
```
thymikee, line 214: We can now use this variable on line 224, or calculate default simulator based on the `fallbackSimulators`, like `const defaultSimulator = fallbackSimulators[0];`

packages/cli-platform-ios/src/commands/runIOS/index.ts
```diff
@@ -230,7 +234,10 @@ async function runOnSimulator(
           findMatchingSimulator(simulators, {simulator: fallback})
         );
       },
-      findMatchingSimulator(simulators, args),
+      findMatchingSimulator(simulators, {
+        ...args,
```
thymikee, line 238: The only other key in this object argument is `uuid`, we can be explicit instead of spreading `args` in this case I guess

packages/cli-platform-ios/src/commands/runIOS/index.ts
```diff
@@ -230,7 +234,10 @@ async function runOnSimulator(
...
+        simulator: args.simulator ?? defaultSimulator,
```
thymikee, line 239: Something along the lines
```suggestion
        simulator: args.simulator ?? fallbackSimulators[0],
```

adamTrz (approved):
Lgtm! 💪

thymikee (approved):

Issue #1111, opened 2020-04-13
init with --package option


OkancanCosar:
First of all tool is great.
But react-native init  has a option for packagename.
why npx react-native init getting --package option?
Thank you :)

thymikee:
Hey! Could you explain what do you expect from `--package` option?

OkancanCosar:
for setting android app package name and ios
ex. [stackoverflow](https://stackoverflow.com/questions/40281569/use-specific-package-name-when-setting-up-a-react-native-project#answer-40282452)

victorwpbastos:
It is super necessary for Android apps.

thymikee:
I think it's only super necessary for you. There was close to zero interest in such request since a few years now. You can quickly find/replace the package name anyway.

victorwpbastos:
@thymikee Is not a simple search/replace. You need to rename folders too. Is super annoying. BTW this feature was in CLI before and was removed.

thymikee:
I'm sorry, but I have no idea why it was removed (according to SO answer linked here, it was removed in RN 0.40+ and the CLI was extracted around RN 0.59, same time I started contributing), never used that. Maybe you could dig and find the rationale behind that?

arniu:
> I think it's only super necessary for you. There was close to zero interest in such request since a few years now. You can quickly find/replace the package name anyway.

There is a special tool [react-native-rename](https://github.com/junedomingo/react-native-rename)  for doing that, which has be downloaded 199k times for now.

I think it's a common requirement.

thymikee:
@arniu thanks, didn't know about this project! I feel we should embrace it and mention somewhere in the docs, what do you think?

artyorsh:
@thymikee there is no need to mention a tool if you have it out of the box. Initializing projects with Android Studio have a step to input the package name, so I think this should be implemented in CLI, but maybe with another option (`--packageName`?). I'm free to contribute on this one if you want.

thymikee:
I still think it would be nice to give a shoutout to a tool that's helpful when one want to change it. @artyorsh Let's give it a shot then, hope it's not gonna add too much of maintenance burden.

artyorsh:
@thymikee I think it's necessary to have it since it is officially supported by both IDEs when starting a new project. Also, since neither Android Studio nor Xcode has no `package rename` option, this shouldn't be included in CLI, but it's good to know there're community tools for that :)

Will try adding this option then 👍

victorwpbastos:
[...]

PR #1808, opened 2023-01-24
feat: allow initialising project with custom package name with `init --package-name`

### LINKED ISSUE #1269: Support for bundleid/packagename when initializing a new app
## Describe the Feature
Add package name and bundle id support to init command

## Possible Implementations
Add `--package` options to `init` command which allows the user to specify bundleid/packagename

TMisiukiewicz:
Summary:
---------
Closes #1269 
Relates to #1111 

Added `--package-name` option to `init` command, which allows initialising project with custom package name for Android and iOS.

Test Plan:
----------
Use `init` command with `--package-name` option:
1. If package name is incorrect e.g. it contains any special chars except dots or have only one segment it should end up with error. Examples: `example`, `com.example@2.app`
2. Initialise project with `--package-name example.app` - project should be initialised with `com.example.app` Android package name and iOS bundleID
3. Initialise project with `--package-name com.example.app` - project should be initialised with `com.example.app` Android package name and iOS bundleID

For 2. and 3., app should be installed on the device/emulator named by project name - not package name.

### DIFF
--- a/packages/cli/src/commands/init/editTemplate.ts
+++ b/packages/cli/src/commands/init/editTemplate.ts
@@ -1,6 +1,7 @@
 import path from 'path';
-import {logger} from '@react-native-community/cli-tools';
+import {CLIError, logger} from '@react-native-community/cli-tools';
 import walk from '../../tools/walk';
+import chalk from 'chalk';
 
 // We need `graceful-fs` behavior around async file renames on Win32.
 // `gracefulify` does not support patching `fs.promises`. Use `fs-extra`, which
@@ -12,6 +13,7 @@ interface PlaceholderConfig {
   placeholderName: string;
   placeholderTitle?: string;
   projectTitle?: string;
+  packageName?: string;
 }
 
 /**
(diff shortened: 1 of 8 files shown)


thymikee:
Can we have some tests for this?

packages/cli/src/commands/init/index.ts
```diff
@@ -32,5 +32,10 @@ export default {
       name: '--skip-install',
       description: 'Skips dependencies installation step',
     },
+    {
+      name: '--package-name <string>',
+      description:
+        'Inits a project with a custom package name (Android) and bundle ID (iOS), e.g. com.example.app',
+    },
```
szymonrybczak, line 39: Let's add this to docs 😄

szymonrybczak (approved):
LGTM! 🚀

packages/cli/src/commands/init/index.ts
TMisiukiewicz: done ✅

TMisiukiewicz:
> Can we have some tests for this?

tests added ✅

adamTrz (approved):
🚀

packages/cli/src/commands/init/editTemplate.ts
```diff
@@ -83,32 +101,180 @@ async function processDotfiles(filePath: string) {
...
+
+      for (const segment of segmentsList) {
+        fs.mkdirSync(segment);
+        process.chdir(segment);
+      }
+
+      fs.rename(
+        `${filePath}/${segmentsList[segmentsList.length - 1]}`,
+        process.cwd(),
+      );
```
thymikee, line 147: Those renames can fail and `createAndroidPackagePaths` also isn't guarded by try/catch, so the user would be left blind when `fs.rename` throws. Please make sure to guard this case

packages/cli/src/commands/init/editTemplate.ts
```diff
@@ -83,32 +101,180 @@ async function processDotfiles(filePath: string) {
...
+  }
+}
+
+export async function replacePlaceholderWithPackageName({
   projectName,
   placeholderName,
-  placeholderTitle = DEFAULT_TITLE_PLACEHOLDER,
-  projectTitle = projectName,
-}: PlaceholderConfig) {
-  logger.debug(`Changing ${placeholderName} for ${projectName} in template`);
```
thymikee, line 92: why is this removed?

bryanltobing:
can we have an example for this?

I tried this 

```bash
npx react-native@latest init [APP_NAME] --package-name [PACKAGE_NAME] --title [APP_TITLE]
```

but it doesn't work

szymonrybczak:
Hey @bryanltobing, what error you get if you try this command? And also could you please provide output from `npx react-native@latest -v` just to make sure that you're working on correct version 😅

bryanltobing:
it does work now. sorry for the inconvenient

javier8340:
packages only can start by "com"? can be changed as user sets? for example if domain start with "org" or by country domain

TMisiukiewicz:
hi @javier8340, it was fixed in #1915

PR #1810, opened 2023-01-26
feat: check if configuration for iOS project exists

TMisiukiewicz:
Summary:
---------

When running `run-ios` or `build-ios` command with `--mode` option, throw error if the configuration does not exist in the project.

Test Plan:
----------

1. Use `run-ios --mode Release` - should run correctly
2. Use `run-ios --mode Test` - should throw an error
3. Use `build-ios --mode Release` - should run correctly
4. Use `build-ios --mode Test` - should throw an error

### DIFF
--- a/packages/cli-platform-ios/src/commands/runIOS/index.ts
+++ b/packages/cli-platform-ios/src/commands/runIOS/index.ts
@@ -11,19 +11,15 @@ import path from 'path';
 import fs from 'fs';
 import chalk from 'chalk';
 import {Config, IOSProjectInfo} from '@react-native-community/cli-types';
-import parseIOSDevicesList from './parseIOSDevicesList';
-import parseXctraceIOSDevicesList from './parseXctraceIOSDevicesList';
-import findMatchingSimulator from './findMatchingSimulator';
-import {
-  logger,
-  CLIError,
-  getDefaultUserTerminal,
-} from '@react-native-community/cli-tools';
-import {Device, IosProjectInfo} from '../../types';
-import ora from 'ora';
-import execa from 'execa';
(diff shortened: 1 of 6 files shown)


adamTrz (approved):
LGTM 👍

szymonrybczak (approved):
Good job! 💪

packages/cli-platform-ios/src/tools/getProjectInfo.ts
```diff
@@ -0,0 +1,12 @@
+import execa from 'execa';
+import {IosProjectInfo} from '../types';
+
+export function getProjectInfo(): IosProjectInfo {
+  process.chdir('./ios');
```
thymikee, line 5: this is assuming we know where the ios folder is, which may not always be true, e.g. in monorepos. You should be able to use `config.platform.ios.sourceDir` instead

packages/cli-platform-ios/src/commands/runIOS/index.ts
```diff
@@ -57,6 +59,8 @@ async function runIOS(_: Array<string>, ctx: Config, args: FlagsT) {
     args.mode = args.configuration;
   }
   const {xcodeProject, sourceDir} = ctx.project.ios;
+  const projectInfo = getProjectInfo();
```
thymikee, line 62: You can pass the `sourceDir` from here

thymikee (changes requested):
I'd appreciate a test for that

TMisiukiewicz:
updated `projectInfo` function and added tests

thymikee (approved):

PR #1811, opened 2023-01-26
feat: interactive mode for selecting mode and scheme for iOS

TMisiukiewicz:
Summary:
---------

Interactively select configuration and scheme for `run-ios` and `build-ios`

https://user-images.githubusercontent.com/13985840/214860549-fc965f99-66c6-4505-8c56-57dc1454b69b.mov

Test Plan:
----------

Use `run-ios --explicit` command - select desired configuration and scheme. If there is only one configuration/scheme, the prompt will be ignored. You can create additional schemes and configurations by opening the project in Xcode.


TMisiukiewicz:
one thing that comes to my mind is to integrate it in the future with device selection prompt which was merged lately. WDYT?

thymikee:
This is looking slick! Thanks for putting it behind the flag, let's experiment with it first. Regarding the flag name, I'd go with `--interactive` instead of `--explicit`, which is used in many CLI tools out there, with `--no-interactive` mode targeted for CI platforms where prompts don't make sense.

packages/cli-platform-ios/src/commands/buildIOS/index.ts
```diff
@@ -18,6 +18,7 @@ import {Device} from '../../types';
 import {BuildFlags, buildProject} from './buildProject';
 import {getDestinationSimulator} from '../../tools/getDestinationSimulator';
 import {getDevices} from '../../tools/getDevices';
+import {runExplicitMode} from '../../tools/runExplicitMode';
```
thymikee, line 21: let's rename the name to "interactive" internally as well

packages/cli-platform-ios/src/commands/buildIOS/index.ts
```diff
@@ -44,6 +45,21 @@ function buildIOS(_: Array<string>, ctx: Config, args: FlagsT) {
...
+      scheme: args.scheme,
+      mode: args.mode,
+    });
+
+    if (scheme) {
+      args.scheme = scheme;
+    }
+
+    if (mode) {
+      args.mode = mode;
```
thymikee, line 59: Let's avoid mutating args. Instead create variables and use that for the rest of the code below

packages/cli-platform-ios/src/commands/runIOS/index.ts
```diff
@@ -58,6 +59,21 @@ async function runIOS(_: Array<string>, ctx: Config, args: FlagsT) {
...
+    });
+
+    if (scheme) {
+      args.scheme = scheme;
+    }
+
+    if (mode) {
+      args.mode = mode;
+    }
+  }
```
thymikee, line 75: let's deduplicate this after build-ios is merged

packages/cli-platform-ios/src/tools/runExplicitMode.ts
```diff
@@ -0,0 +1,42 @@
...
+  promptForConfigurationSelection,
+  promptForSchemeSelection,
+} from './prompts';
+
+interface Args {
+  scheme?: string;
+  mode?: string;
+}
+
+export async function runExplicitMode(args: Args): Promise<Args> {
```
thymikee, line 14:
```suggestion
export async function runExplicitMode({scheme, mode}: Args): Promise<Args> {
```

adamTrz:
> I'd wait for this to merge after the #1744

@thymikee  - it's is already merged 😉

adamTrz:
> one thing that comes to my mind is to integrate it in the future with device selection prompt which was merged lately. WDYT?

100% agree. And renaming to `--interactive` would make even more sense there I would say.

packages/cli-platform-ios/src/tools/prompts.ts
```diff
@@ -0,0 +1,34 @@
+import prompts from 'prompts';
```
adamTrz, line 1: Not super sure about file name here...
There's also `promptForDeviceSelection` in there somewhere so we should either rename this file or move device selection prompt here. WDYT?

packages/cli-platform-ios/src/tools/runExplicitMode.ts
adamTrz, line 14: I would rename this function since it's not actually doing any `run` related stuff. It's just returning scheme and mode right?

packages/cli-platform-ios/src/tools/prompts.ts
TMisiukiewicz: good point, decided to move `promptForDeviceSelection` to this file to stay consistent

TMisiukiewicz:
@thymikee @adamTrz thanks for your responses, could you re-check the updates when you have a moment please?

adamTrz:
> @thymikee @adamTrz thanks for your responses, could you re-check the updates when you have a moment please?

Lookin' good! 

But please fix tests 🙂

TMisiukiewicz:
> 

all the tests are green when i run it locally, so i assume it will be fixed once #1804 gets merged, let's wait and see if i am correct 😄

PR #1833, opened 2023-02-16
download correct sourceMap for dev/prod profiles, add warning for Expo Dev menu profiles

### LINKED ISSUE #1831: `profile-hermes` command always downloads `dev` sourceMaps
## Description

When calling `profile-hermes`, currently it downloads the `dev` sourcemaps as a hardcoded string - if the profile has been created with a production bundle the resulting source maps are completely useless.

This is the code in question:

https://github.com/react-native-community/cli/blob/a8ff4ff0406d44ebd1cfec603affda1c7d6e1642/packages/cli-hermes/src/profileHermes/sourcemapUtils.ts#L39

I have changed this to `dev=false` in my local `node_modules` and can verify that I am getting correct line numbers now. 

Would a PR that adds this as a command line option be welcome? 

Alternatively, this could probably be auto-detected, since a lot of the `name` fields of the original profile will contain substrings like `index.bundle?platform=android&dev=false&minify=false&app=my.amazing.app&modulesOnly=false&runModule=true:107805:5`, so this could be taken as a hint on what to actually pass to hermes to get the right sourceMaps.

phryneas:
This fixes #1831

Summary:
---------

As described over in #1831, until now the `profile-hermes` command would always load the `dev` mode soureMaps, even if a production build was being profiled - this lead to wrong line numbers (at best).

This PR detects if the profile has been recorded with a `dev` or `prod` build, which platform was used and if `minify` was enabled.
Unfortunately, at this point, something further down the line seems to fail with a profile that was produced for a minifed build, so I could not really test that - that seems to be some problem with the general process and not with my changes, though.

With `minify: false`, I verified this to work correctly for both `dev=true` and `dev=false`.

If no URL information is present in the profiler, the PR just defaults to the values that were in the code before this change.

This PR also adds a warning if no URL information is present, but the profile contains calls to `EXDevMenuApp`. That is usually the case when the profiler does not record the user application, but the Expo dev menu. In that case, the user has to reload their app a few times before recording another profile - Hermes always profiles the JS bundle that was loaded last. I'm adding this warning because this is documented nowhere else, and it cost me a day until I found this out for myself.

Test Plan:
----------

Record Profiles with different combinations of "__DEV__" and "minify" selected in the settings reachable from the React Native Dev menu, call `npx react-native profile-hermes --verbose` after each recording and compare the resulting profile line numbers with the actual source code.

### DIFF
--- a/packages/cli-hermes/src/profileHermes/downloadProfile.ts
+++ b/packages/cli-hermes/src/profileHermes/downloadProfile.ts
@@ -7,6 +7,7 @@ import os from 'os';
 import transformer from 'hermes-profile-transformer';
 import {findSourcemap, generateSourcemap} from './sourcemapUtils';
 import {getAndroidProject} from '@react-native-community/cli-platform-android';
+import {getMetroBundleOptions} from './metroBundleOptions';
 /**
  * Get the last modified hermes profile
  * @param packageNameWithSuffix
@@ -44,7 +45,7 @@ export async function downloadProfile(
   sourcemapPath?: string,
   raw?: boolean,
   shouldGenerateSourcemap?: boolean,
-  port?: string,
+  port: string = '8081',
   appId?: string,
   appIdSuffix?: string,
 ) {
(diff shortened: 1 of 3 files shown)


thymikee:
@huntie would you mind taking a look at this?

phryneas:
This is still absolutely relevant, but the bot is right, it's been 3 months - could someone maybe take a look at this?

huntie (approved):
@phryneas Looks good to me ✅ Sorry about the delay on this review! 🙏🏻

PR #1842, opened 2023-02-20
feat: Bump minimum Node JS requirement to 16 for RN 0.72 in `doctor`

robhogan:
Summary:
---------

React Native is increasing the minimum Node JS requirement from 14 to 16 in https://github.com/facebook/react-native/pull/36217, https://github.com/facebook/react-native-website/pull/3580 for 0.72 onwards. This makes the corresponding change in `doctor`.

Test Plan:
----------

CI

PR #1844, opened 2023-02-20
breaking: node 16; typescript 4

thymikee:
Summary:
---------

Started off #1842 by @robhogan. Wanted to bump min Node to 16, turned out we need to upgrade TS to 4 (about time tbh). Hence I moved it to a sperate PR.

Test Plan:
----------

CI

### DIFF
--- a/packages/cli-config/package.json
+++ b/packages/cli-config/package.json
@@ -11,7 +11,7 @@
     "@react-native-community/cli-tools": "^10.1.1",
     "chalk": "^4.1.2",
     "cosmiconfig": "^5.1.0",
-    "deepmerge": "^3.2.0",
+    "deepmerge": "^4.3.0",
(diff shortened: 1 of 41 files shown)


packages/cli-config/package.json
```diff
@@ -11,7 +11,7 @@
     "@react-native-community/cli-tools": "^10.1.1",
     "chalk": "^4.1.2",
     "cosmiconfig": "^5.1.0",
-    "deepmerge": "^3.2.0",
+    "deepmerge": "^4.3.0",
```
thymikee, line 14: this turned out to be unnecessary, but checking the release notes the only breaking change is umd -> cjs, which is fine for us

adamTrz (approved):
TS v4! 🎉

PR #1869, opened 2023-03-10
feat: added --user flag to run-android command

### LINKED ISSUE #765: Handle Work Profile / Several Android Users

adamTrz:
Summary:
---------

Added option to pass `userId` to `run-android` command so the app will be run for given user profile. 
Closes #765  

run-android-userid

Test Plan:
----------

You have to have created User Profile (either by adding new user or using [Android For Work](https://www.android.com/enterprise/employees/)).
Clone the PR, link as described in Contributing guid and try to run the app either with `--interactive` mode or passing `--userId` flag.


docs/commands.md
```diff
@@ -425,6 +425,11 @@ react-native build-android --extra-params "-x lint -x test"
 
 Installs passed binary instead of building a fresh one. This command is not compatible with `--tasks`.
 
+
+#### `--deviceId` <number>
```
homersimpsons, line 429:
```suggestion
#### `--userId` <number>
```

docs/commands.md
```diff
@@ -425,6 +425,11 @@ react-native build-android --extra-params "-x lint -x test"
 
 Installs passed binary instead of building a fresh one. This command is not compatible with `--tasks`.
 
+
+#### `--userId` <number>
+
+Id of the User Profile you want to install the app on.
```
thymikee, line 431: since this feature is named "user profiles", shouldn't the flag be `--userProfile`? Id is quite ambiguous, but idk

thymikee: Or just `--user`, since this is how adb names it

packages/cli-platform-android/src/commands/runAndroid/listAndroidUsers.ts
```diff
@@ -0,0 +1,61 @@
...
+  name: string;
+};
+
+export function checkUsers(device: string, adbPath: string) {
+  try {
+    const adbArgs = ['-s', device, 'shell', 'pm', 'list', 'users'];
+
+    logger.debug(`Checking users on "${device}"...`);
+    const {stdout} = spawnSync(adbPath, adbArgs, {encoding: 'utf-8'});
+    const regex = new RegExp(/UserInfo{([0-9]*):([^:]*):[\w]*}/, 'g');
```
thymikee, line 16: is there a way to list users in a json or other structured format?

thymikee: If not, and we need to keep the regex, please include a unit test with sample output from adb

homersimpsons: For the record I did not find any way to fetch users in a structured way.

Here is the output on my environment:

```
❯ adb -s ... shell pm list users
Users:
        UserInfo{0:Homersimpsons:c13} running
        UserInfo{10:Guest:404}
```

So I guess the best RegEx could be (https://regex101.com/r/8VjP87/1):
```
/^\s*UserInfo\{(?<userId>\d+):(?<userName>.*):(?<userFlags>[0-9a-f]*)}/
```

And in fact a test should be added.

For reference here is the link where this string is produced in Android's source (It might be interesting to include it, the format is unlikely to change given the coment):
https://android.googlesource.com/platform/frameworks/base/+/refs/heads/master/core/java/android/content/pm/UserInfo.java#461

Also the `pm` command is defined here (I could not found where the `cmd` comes from):
https://android.googlesource.com/platform/frameworks/base/+/refs/heads/master/cmds/pm/pm.sh

adamTrz:
Hi @homersimpsons , sorry for delay, got swarmed in other projects. 

> Can you tell me how to test the changes easily?

1. Clone my fork (you can use GH CLI to do this as well 🙂 ) and run `yarn link-packages` command
2. In new react-native project link CLI packages as described [here](https://github.com/react-native-community/cli/blob/main/CONTRIBUTING.md#setup)
3. Run the app :)

packages/cli-platform-android/src/commands/runAndroid/listAndroidUsers.ts
adamTrz: Done :)

packages/cli-platform-android/src/commands/runAndroid/index.ts
```diff
@@ -121,6 +123,18 @@ async function buildAndRun(args: Flags, androidProject: AndroidProject) {
       );
     }
 
+    if (args.interactive) {
+      const users = checkUsers(device.deviceId as string, adbPath);
+      console.log('users', users);
```
thymikee, line 128: Let's remove excess logging

homersimpsons:
Thank you for merging, sorry I did not spend more time to try to test this.

PR #1870, opened 2023-03-10
feat: add deeplinks to our documentation

blakef:
Summary:
---------

Everything is in, so this should be good to tests against.
~~These changes will only start to work once our documentation is updated to support this feature:~~
- [X] facebook/react-native-website#3618 
- [X]  facebook/react-native-website#3619
- [X]  facebook/react-native-website#3620
- [X]  facebook/react-native-website/pull/3626

The outcome is that we now write links like this:

```typescript
import {findProjectRoot, logger, link} from '@react-native-community/cli-tools';
link.docs('environment-setup', 'ruby');
```
And get links this:

> https://reactnative.dev/docs/environment-setup#ruby?platform=ios&os=macos

For more complicated examples you'd write this:

```typescript
link.setVersion('0.69');
link.docs('environment-setup', {hash:'ruby', platform: 'ios', guide: 'native', very: 'extensible'})
```
> https://reactnative.dev/docs/0.69/environment-setup#ruby?platform=ios&os=macos&guide=native&very=extensible

Depending on how the user has run the project.  With this result if clicked on:

CleanShot 2023-03-10 at 18 49 43

Test Plan:
----------

I've run this against a locally generated project from `init`.  I'll do something more comprehensive, but wanted to get some direction feedback before.

### DIFF
--- a/packages/cli-config/src/loadConfig.ts
+++ b/packages/cli-config/src/loadConfig.ts
@@ -1,4 +1,6 @@
 import path from 'path';
+import fs from 'fs';
+import semver from 'semver';
 import {
   UserDependencyConfig,
   ProjectConfig,
@@ -65,6 +67,7 @@ function loadConfig(projectRoot: string = findProjectRoot()): Config {
         ? path.resolve(projectRoot, userConfig.reactNativePath)
         : resolveReactNativePath(projectRoot);
     },
+    reactNativeVersion: 'unknown',
     dependencies: userConfig.dependencies,
     commands: userConfig.commands,
     healthChecks: [],
@@ -88,6 +91,25 @@ function loadConfig(projectRoot: string = findProjectRoot()): Config {
       return lazyProject;
     },
   };
+
+  // Try our best to figure out what version of React Native we're running. This is
+  // currently being used to get our deeplinks working, so it's only worried with
+  // the major and minor version.
+  try {
+    const {version} = JSON.parse(
+      fs.readFileSync(
+        path.join(initialConfig.reactNativePath, 'package.json'),
+        {encoding: 'utf8'},
+      ),
+    );
+    const out = semver.parse(version);
+    if (out) {
+      // Retain only these version, since they correspond with our documentation.
+      initialConfig.reactNativeVersion = `${out.major}.${out.minor}`;
+    }
+  } catch (_) {
+    // We don't seem to be in a well formed project, give up quietly.
+  }
 
   const finalConfig = Array.from(
     new Set([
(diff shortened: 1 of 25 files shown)


packages/cli-tools/src/index.ts
```diff
@@ -11,5 +11,6 @@ export {default as hookStdout} from './hookStdout';
 export {getLoader, NoopLoader, Loader} from './loader';
 export {default as findProjectRoot} from './findProjectRoot';
 export {default as printRunDoctorTip} from './printRunDoctorTip';
+export * as link from './doclink';
```
blakef, line 14: I had to bring in the babel plugin to export like this.  Am I being silly here?

packages/cli-config/src/loadConfig.ts
```diff
@@ -89,6 +92,24 @@ function loadConfig(projectRoot: string = findProjectRoot()): Config {
...
+    );
+    const out = semver.parse(version);
+    if (out) {
+      // Retain only these version, since they correspond with our documentation.
+      initialConfig.reactNativeVersion = `${out.major}.${out.minor}`;
+    }
+  } catch (_) {
+    // We don't seem to be in a well formed project, give up.
+  }
+
```
blakef, line 112: This is a bit cheeky admittedly, I'm wondering if this is hiding somewhere (or there's a helper method that I should be using).

adamTrz (approved):
Massive work, thanks for the PR! 🎉

packages/cli-config/src/loadConfig.ts
blakef: Thanks @adamTrz  for the pointer.  I've added [another commit](ab21873c067f72904c98c0e71e5311aeb0e723dc), since `releaseChecker` doesn't quite do what I wanted here.  In that commit I've refactored that module to provide 3 things:

1. what is the current version in this project → but as a parsed semver object
2. what is the the next stable upgrade from the current version → effectively `releaseChecker` original, excluding the built in logging.
3. log the next available stable upgrade from the current version → `releaseChecker` as it was being used.

I've also cleaned up small things like consistent return types.

## TL;DR:

**I'm happy to submit [this change](ab21873c067f72904c98c0e71e5311aeb0e723dc) as another PR if the original one goes in (so I'd revert that commit on this branch).  This PR is already getting quite big.  Does that allays your original concern?**

adamTrz:
@blakef Mind resolving conflicts so we can merge? 🙏

thymikee:
This is the feature I've been waiting for years :D Thank you @blakef

PR #1877, opened 2023-03-17
feat: update cli-platform-android for find android dir on current directory

lynnkoo:
I created a project by exists android project, I think that andoid-cli can find current dirctory likes ios-cli.

### DIFF
--- a/packages/cli-platform-android/src/config/findAndroidDir.ts
+++ b/packages/cli-platform-android/src/config/findAndroidDir.ts
@@ -14,5 +14,12 @@ export default function findAndroidDir(root: string) {
     return 'android';
   }
 
+  if (
+    fs.existsSync(path.join(root, 'build.gradle'))
+    || fs.existsSync(path.join(root, 'build.gradle.kts'))
+  ) {
+    return '.';
+  }
+
   return null;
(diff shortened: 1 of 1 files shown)


packages/cli-platform-android/src/config/findAndroidDir.ts
```diff
@@ -14,5 +14,12 @@ export default function findAndroidDir(root: string) {
     return 'android';
   }
 
+  if (
+    fs.existsSync(path.join(root, 'build.gradle'))
```
cortinico, line 18: Why is this needed?

lynnkoo: When I created a android app but not used `npx react-native` script, added a `package.json` and `yarn install` only on the android root directory, then run script `npx react-native config` can't found `project.android` info.
In the iOS project, I apply the same operations,  It's can work.
image
image
The different is `sourceDir` property when get from default value.

lynnkoo: This changes is not necessary, I can create a `react-native.config.js` config fie in android root directory alos.

PR #1882, opened 2023-03-21
fix: doclinks should not select a default platform

blakef:
Previously this would default to Android, which isn't intuitive. Also fixed some links to be more specific.

Summary:
---------
## Before
CleanShot 2023-03-21 at 10 15 43
## After
CleanShot 2023-03-21 at 10 13 35

Test Plan:
----------

Built and ran the cli to `init` a new project.

### DIFF
--- a/packages/cli-tools/src/doclink.ts
+++ b/packages/cli-tools/src/doclink.ts
@@ -1,7 +1,11 @@
 import os from 'os';
 import assert from 'assert';
 
-type Platforms = 'android' | 'ios';
+type Platforms =
+  | 'android'
+  | 'ios'
+  | 'inherit' // Expect the platform to be defined before this link is called with the link.setPlatform()
+  | 'none'; // No platform specific documentation
 
 export function getOS(): string {
   // Using os.platform instead of process.platform so we can test more easily. Once jest upgrades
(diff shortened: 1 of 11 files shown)


packages/cli-tools/src/doclink.ts
```diff
@@ -62,7 +62,10 @@ function doclink(
     : `${section}/${path}`;
 
   url.searchParams.set('os', OS);
-  url.searchParams.set('platform', platform);
+
+  if (platform) {
+    url.searchParams.set('platform', platform);
+  }
```
motiz88, line 68: I don't have context on this code, but is there a way we can force relevant call sites to specify a platform where one is needed?

blakef: I'm not sure I understand the question, but giving it a go.  In some places where we take care of this.  For example when the user runs `npx react-native run-android`, we set the [platform to android](https://github.com/react-native-community/cli/blob/main/packages/cli-platform-android/src/commands/runAndroid/index.ts#L41).

Is that what you're asking about?

motiz88: What I mean is: Is it useful that `platform` is optional on line 31? Could we, for example, make it required in every call to `doclink`?

thymikee:
@blakef would you mind revisit this PR and check the suggestions?

blakef:
Thanks for the nudge @thymikee. I'll take a look at it.

blakef:
@thymikee I think we're GTG.

thymikee (approved):
🚀

PR #1890, opened 2023-03-27
fix: `log-ios` command

szymonrybczak:
Summary:
---------

Closes  #1675

This PR connects output of two commands`xcrun simctl list --json devices` and `xcrun xcdevice list`. One command outputs `state`, but doesn't return `available`. But second command `xcrun xcdevice list` outputs `available` but doesn't output `state`. So I connected outputs from both commands.

Test Plan:
----------
2. Run this command:
```
node /path/to/react-native-cli/packages/cli/build/bin.js log-ios
```

### DIFF
--- a/packages/cli-platform-ios/src/commands/logIOS/index.ts
+++ b/packages/cli-platform-ios/src/commands/logIOS/index.ts
@@ -11,12 +11,19 @@ import os from 'os';
 import path from 'path';
 import {logger} from '@react-native-community/cli-tools';
 import listIOSDevices from '../../tools/listIOSDevices';
-import {getSimulators} from '../runIOS';
+import getSimulators from '../../tools/getSimulators';
+import {Config} from '@react-native-community/cli-types';
+import prompts from 'prompts';
 
 /**
  * Starts iOS device syslog tail
  */
-async function logIOS() {
+
+type Args = {
+  interactive: boolean;
+};
+
+async function logIOS(_argv: Array<string>, _ctx: Config, args: Args) {
   // Here we're using two command because first command `xcrun simctl list --json devices` outputs `state` but doesn't return `available`. But second command `xcrun xcdevice list` outputs `available` but doesn't output `state`. So we need to connect outputs of both commands.
   const simulators = getSimulators();
   const bootedSimulators = Object.keys(simulators.devices)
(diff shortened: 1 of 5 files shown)


packages/cli-platform-ios/src/commands/logIOS/index.ts
```diff
@@ -6,44 +6,42 @@
...
 
-  const device = findAvailableDevice(devices);
-  if (device === null) {
+  if (bootedAndAvailableSimulators.length === 0) {
     logger.error('No active iOS device found');
     return;
   }
 
-  tailDeviceLogs(device.udid);
+  tailDeviceLogs(bootedAndAvailableSimulators[0].udid);
```
szymonrybczak, line 44: 💡 I think we can implement here prompt if the list of booted and available simulators is greater than 1. WDYT?

packages/cli-platform-ios/src/commands/logIOS/index.ts
```diff
@@ -6,44 +6,42 @@
...
-    for (const device of devices[key]) {
-      if (device.availability === '(available)' && device.state === 'Booted') {
-        return device;
-      }
-    }
-  }
-  return null;
-}
+import listIOSDevices from '../../tools/listIOSDevices';
+import {getSimulators} from '../runIOS';
```
thymikee, line 14: Let's extract `getSimulators` to a separate file so that commands like `run-ios` and `log-ios` don't depend on each other.

packages/cli-platform-ios/src/commands/logIOS/index.ts
thymikee: Whenever we add prompts we need to think of a way for the command to be usable on CI as well, which won't answer the prompt. I'm ok with having a prompt and fall back to `[0]` if in non-interactive environments

thymikee: Also, let's add that separately

thymikee (approved):

packages/cli-platform-ios/src/commands/logIOS/index.ts
szymonrybczak: Created `--interactive` flag to select simulators, by default it selects first one.

adamTrz:
@szymonrybczak mind resolving conflicts? 🙏

PR #1895, opened 2023-03-29
fix: replace `launchPackager.sh` with `launchPackager.command`

szymonrybczak:
Summary:
---------
This PR replaces `launchPackager.sh` with `launchPackager.command`, we though that we can use `launchPackager.sh`, but actually it doesn't work. Because in same cases when we ran `open -a Terminal.app /path/to/launchPackager.sh` it doesn't opens the Terminal.app with the script but the Code Editor with this file.

Test Plan:
----------

2. Run this command:
```
node /path/to/react-native-cli/packages/cli/build/bin.js run-android
```
Metro bundler should be opened in right place in new terminal window.

### DIFF
--- a/packages/cli-platform-android/src/commands/buildAndroid/startServerInNewWindow.ts
+++ b/packages/cli-platform-android/src/commands/buildAndroid/startServerInNewWindow.ts
@@ -17,7 +17,9 @@ export function startServerInNewWindow(
    * Set up OS-specific filenames and commands
    */
   const isWindows = /^win/.test(process.platform);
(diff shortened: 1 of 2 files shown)


tido64:
> Hey @tido64 @huntie, I'm adding missing file (which is mandatory), could you please look at this PR? 🙏

I'm out and about, but had a quick look. Not sure I understand the issue. Is this because someone set up associations to open .sh files in an editor instead of running them? Instead of introducing a third file, can we just rename the .sh file to .command?

szymonrybczak:
Yeah, you're right. Fixed ✅

thymikee (approved):
Alright

PR #1875, opened 2023-03-15
rn metro config stack 1

huntie:
- chore: Update cli-plugin-metro to use external Metro types
- [RFC] feat: Update cli-plugin-metro to use defaults in app metro.config.js from RN 0.72

Summary:
---------

Test Plan:
----------

PR #1896, opened 2023-03-31
fix: Fix merge of CLI Metro config overrides, add soft fallback to RN defaults

huntie:
### React Native Metro config → React Native repo (https://github.com/facebook/react-native/pull/36502)

Follow ups to https://github.com/react-native-community/cli/pull/1875.

## Changes

**Fixes**

- Fix a subtle bug where `loadConfig({cwd: ctx.root, ...options}, overrideConfig);` ([see original file diff](https://github.com/react-native-community/cli/pull/1875/files#diff-3eb3e1e81b924a57b22a630c633912f2f8777b5a201992591c8a3517fb63f525L128)) did not correctly merge `overrideConfig` (the extra config applied by RN CLI) over the newly migrated defaults (which were promoted into the project-level config file).
    - After changes, `overrideConfig` will now be merged after all config, _including any user-set config_. This makes sense (as before), given the CLI applies very sparse override options based on required CLI inputs.
    - **See test plan for E2E bugs fixed.**

**Upgrade improvements**

Provide a soft upgrade path for users on React Native 0.72 to migrate to the new `metro.config.js` setup in 0.73.

- Update the CLI's `loadMetroConfig` implementation to conditionally fall back to internal defaults. The intent is to remove this behaviour (and dependency in CLI) for React Native 0.73.
    - This conceptually brings the implementation back to the [first version](https://github.com/react-native-community/cli/commit/598f1a0dec7875f63e6a2a1ec677849305944357) of the original PR where we had "getDefaultConfigPre72". Instead, this now depends on the common `@react-native/metro-config` source.
- Improve warning guidance to reference template `metro.config.js` source.
    - Note: This may be later replaced with a React Native changelog deep link, once published.

Changelog: Fix bug where CLI-overridden Metro config might not have been applied

---

**📦 Please publish!**: We'd like to integrate this change into the 0.72 RC1 early next week. Apologies for the churn 🙏🏻.

---

## Test Plan

Using `rn-tester` configured with a locally `yarn link`ed copy of RN CLI.

✅ In all test cases, I added temporary debug logs for the entire final config object, which were compared and produce no diff between either config setup.

### New 0.72 config setup (project extends `@react-native/metro-config`)

**✅  App (rn-tester) runs with no CLI warnings**

image

**✅ Correctly merged `overrideConfig`**

In the new config setup, the CLI `overrideConfig` was previously clobbered by the project-level `metro.config.js` which broke new template apps. This is now fixed.

**FIXED**: `serializer.getModulesRunBeforeMainModule` now overrides [the default value in `metro-config` (metro)](https://github.com/facebook/metro/blob/1e47cb5b3cc289530fb18e402891f9d2816611dd/packages/metro-config/src/defaults/index.js#L66), cc @cipolleschi 

image

**FIXED**: `resolver.platforms` now overrides [the default value in `@react-native/metro-config`](https://github.com/facebook/react-native/blob/950472018d137f266358e439ffd17e374ec3686b/packages/metro-config/index.js#L44).

Final config in RN CLI compared to before this PR:

```diff
- platforms: [ 'android', 'ios' ],
+ platforms: [ 'ios', 'android', 'native' ],
```

(`@react-native/metro-config` uses an intentionally simpler default for this option. The former value will be read by standalone Metro CLI, and the latter is overridden by RN CLI based on `ctx.platforms`).

### Fallback 0.71 config setup (project does not extend `@react-native/metro-config`)

**✅ Warning and fallback config behaviour are present**

image

### DIFF
--- a/packages/cli-plugin-metro/src/tools/loadMetroConfig.ts
+++ b/packages/cli-plugin-metro/src/tools/loadMetroConfig.ts
@@ -1,5 +1,13 @@
+import fs from 'fs';
 import path from 'path';
-import {ConfigT, InputConfigT, loadConfig, resolveConfig} from 'metro-config';
+import {
+  ConfigT,
+  InputConfigT,
+  loadConfig,
+  mergeConfig,
+  resolveConfig,
+} from 'metro-config';
+import {getDefaultConfig} from '@react-native/metro-config';
 import {CLIError, logger} from '@react-native-community/cli-tools';
 import type {Config} from '@react-native-community/cli-types';
 import {reactNativePlatformResolver} from './metroPlatformResolver';
@@ -67,8 +75,8 @@ export interface ConfigOptionsT {
 /**
  * Load Metro config.
  *
- * Allows the CLI to override certain defaults in the base `metro.config.js`
- * based on dynamic user options in `ctx`.
+ * Allows the CLI to override select values in `metro.config.js` based on
+ * dynamic user options in `ctx`.
  */
 export default async function loadMetroConfig(
   ctx: ConfigLoadingContext,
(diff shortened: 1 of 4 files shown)


huntie:
**🚫 Planning changes**

packages/cli-plugin-metro/src/tools/loadMetroConfig.ts
```diff
@@ -89,17 +97,28 @@ export default async function loadMetroConfig(
...
+    logger.warn('Falling back to internal defaults.');
+
+    return loadConfig(
+      {cwd: ctx.root, ...options},
+      mergeConfig(getDefaultConfig(ctx.root), overrideConfig),
     );
   }
 
+  // Load config file from disk, merging dynamic CLI config before project
+  // config using `defaultConfigOverrides` arg.
```
huntie, line 122: Comment added to highlight this important `loadConfig` behaviour. Earlier versions of this PR made use of `mergeConfig`, which would be incorrect.

huntie:
Ready for review.

huntie:
✅ Ready for review

packages/cli-plugin-metro/src/tools/loadMetroConfig.ts
```diff
@@ -89,17 +97,32 @@ export default async function loadMetroConfig(
...
+    )
+  ) {
     logger.warn(
-      "From React Native 0.72, your 'metro.config.js' file should " +
-        "extend '@react-native/metro-config', however it's not present in your " +
-        "project's devDependencies. Please install '@react-native/metro-config'.",
+      'From React Native 0.72, your metro.config.js file should extend' +
+        "'@react-native/metro-config'. Please see the React Native 0.72 " +
+        'changelog, or copy the template at:\n' +
+        'https://github.com/facebook/react-native/blob/main/packages/react-native/template/metro.config.js',
```
thymikee, line 111: Thanks for adding a reference to copy the config

thymikee (approved):

packages/cli-plugin-metro/src/tools/loadMetroConfig.ts
huntie: Ideal link is once we have a URL for the changelog/upgrade helper :)

PR #1899, opened 2023-04-04
fix: Drop @react-native/metro-config dep for RN 0.72 config fallback

huntie:
### React Native Metro config → React Native repo (https://github.com/facebook/react-native/pull/36502)

This is a follow up to https://github.com/react-native-community/cli/pull/1896.

Unfortunately the most recent CLI release was auto-consumed in React Native's CI, exposing a blocker where we could not consume the (repo-)circular dependency `react-native` → `@react-native-community/cli-plugin-metro` → `@react-native/metro-config` — I 100% jumped the gun on this setup(!).

While this is now mitigated by the new CLI release process, and possible in practice, **we're opting to remove this circular setup** as it would introduce maintenance headaches while we are working across two repos.

Changelog: Internal

---

This is an architectural simplification targeting React Native `main` (0.73). However it is also worth cherry picking this commit to RN CLI `10.x` (targeting RN 0.72), so that we can once again bump all local Metro dependencies here (and replicate those bumps in both `11.x` and `10.x`). cc @robhogan @cipolleschi 

**Open as https://github.com/react-native-community/cli/pull/1901**.

---

## Changes

- Drop `@react-native/metro-config` dependency from `cli-plugin-metro`.
- Restore reproduced `@react-native/metro-config` default config within `cli-plugin-metro` (to be used in the fallback case only).
    - This is now equivalent to the [original (abandoned) version of #1875](https://github.com/react-native-community/cli/commit/598f1a0dec7875f63e6a2a1ec677849305944357).
    - Ideally, we will only keep these values in this repo for the next 1-2 versions of React Native. Once removed, we will still have the throw-descriptive-error fallback for unmigrated projects.

Note: An alternative approach for this is `peerDependencies`, but given the recent issues and how this might affect strict internal and open source CI, I'm opting for the absolutely bulletproof solution.

## Test Plan

(Repeats test plan instructions from https://github.com/react-native-community/cli/pull/1896).

| New 0.72 config setup (project extends `@react-native/metro-config`) | Fallback 0.71 config setup (project does not extend `@react-native/metro-config`) |
| --- | --- |
| image | image |

✅ Identical configs are produced
*(With only path changes from `yarn link`)*

image

### DIFF
--- a/packages/cli-plugin-metro/src/tools/getDefaultMetroConfig.ts
+++ b/packages/cli-plugin-metro/src/tools/getDefaultMetroConfig.ts
@@ -1,0 +1,87 @@
+import type {InputConfigT} from 'metro-config';
+import path from 'path';
+import type {ConfigLoadingContext} from './loadMetroConfig';
+
+/**
+ * This module reproduces defaults from the @react-native/metro-config package,
+ * used in ./loadMetroConfig.js to provide a soft upgrade process when upgrading
+ * to React Native 0.72.
+ *
+ * These values will be used when:
+ * - RN CLI 11.x or greater is present in a project (from React Native 0.72).
+ * - The project has not yet followed the upgrade instructions to update
+ *   metro.config.js to extend '@react-native/metro-config'.
+ *
+ * Until we remove this file in a future release, updates should be made both
+ * here and in '@react-native/metro-config'. (Note: Updates to these values are
+ * generally rare.)
+ *
+ * TODO(@huntie): Remove this file in a future React Native release.
+ */
+
+/**
+ * @deprecated (React Native 0.72.0) Defaults should be updated here and in
+ *   https://github.com/facebook/react-native/tree/main/package/metro-config/index.js
+ */
+const INTERNAL_CALLSITES_REGEX = new RegExp(
+  [
+    '/Libraries/Renderer/implementations/.+\\.js$',
+    '/Libraries/BatchedBridge/MessageQueue\\.js$',
+    '/Libraries/YellowBox/.+\\.js$',
+    '/Libraries/LogBox/.+\\.js$',
+    '/Libraries/Core/Timers/.+\\.js$',
+    '/Libraries/WebSocket/.+\\.js$',
+    '/Libraries/vendor/.+\\.js$',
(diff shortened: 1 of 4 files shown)


packages/cli-plugin-metro/src/tools/getDefaultMetroConfig.ts
```diff
@@ -0,0 +1,87 @@
...
+  ctx: ConfigLoadingContext,
+): InputConfigT {
+  return {
+    resolver: {
+      resolverMainFields: ['react-native', 'browser', 'main'],
+      unstable_conditionNames: ['require', 'react-native'],
+    },
+    serializer: {
+      getPolyfills: () =>
+        require(path.join(ctx.reactNativePath, 'rn-get-polyfills'))(),
```
huntie, line 62: Restored (https://github.com/react-native-community/cli/pull/1875/files#diff-3eb3e1e81b924a57b22a630c633912f2f8777b5a201992591c8a3517fb63f525L76-L77).

Deviates from `@react-native/metro-config`, which references the standalone package `@react-native/js-polyfills`. These are equivalent, but allows this repo to not reference the RN-hosted package (see removal in `yarn.lock`).

yarn.lock
```diff
@@ -2673,21 +2673,6 @@
...
-"@react-native/metro-config@^0.73.0":
-  version "0.73.0"
-  resolved "https://registry.yarnpkg.com/@react-native/metro-config/-/metro-config-0.73.0.tgz#7601d2f4ba2252a4e359f8fda21008daf5f31eb2"
-  integrity sha512-b+q8Ot6nxPE2PXEua20Q0vhXYpVtVjUaTX/hv+7mL5hLzKeIWq/bIxF04GLWT93BBjGSlL0xaZ7iGIQNN75Hog==
-  dependencies:
-    "@react-native/js-polyfills" "^0.73.0"
-    metro-config "0.76.0"
-    metro-react-native-babel-transformer "0.76.0"
-    metro-runtime "0.76.0"
-
```
huntie, line 2690: All dependencies on packages published from `facebook/react-native` are dropped.

robhogan (approved):

thymikee (approved):

PR #1904, opened 2023-04-07
fix: change permission in `launchPackager.command` when building packages

szymonrybczak:
Summary:
---------
By default the permission is `0o777`, but the `launchPackager.command` needs to have `0o755` to launch correctly.

Test Plan:
----------
Run `yarn build` and check the permission for the file `packages/cli-plugin-metro/build/launchPackager.command`. This file needs to have `-rwxr-xr-x` to launch correctly.


adamTrz (approved):
Thanks a bunch! 🏅

TMisiukiewicz (approved):

Issue #1908, opened 2023-04-13
CLI does not support subrojects


AnnatarHe:
## Environment

```bash
# xcodebuild -version
Xcode 14.3
Build version 14E222b
```

```bash
System:
    OS: macOS 13.3.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 224.83 MB / 32.00 GB
    Shell: 3.6.1 - /opt/homebrew/bin/fish
  Binaries:
    Node: 19.1.0 - ~/Library/Caches/fnm_multishells/67662_1681404637528/bin/node
    Yarn: 1.22.19 - ~/Library/Caches/fnm_multishells/67662_1681404637528/bin/yarn
    npm: 8.19.3 - ~/Library/Caches/fnm_multishells/67662_1681404637528/bin/npm
    Watchman: 2023.04.03.00 - /opt/homebrew/bin/watchman
  Managers:
    CocoaPods: 1.11.3 - /Users/USERNAME/.rbenv/shims/pod
  SDKs:
    iOS SDK:
      Platforms: DriverKit 22.4, iOS 16.4, macOS 13.3, tvOS 16.4, watchOS 9.4
    Android SDK: Not Found
  IDEs:
    Android Studio: 2022.1 AI-221.6008.13.2211.9477386
    Xcode: 14.3/14E222b - /usr/bin/xcodebuild
  Languages:
    Java: 11.0.18 - /usr/bin/javac
  npmPackages:
    @react-native-community/cli: 10.2.2
    react: 18.2.0 => 18.2.0 
    react-native: 0.71.6 => 0.71.6 
    react-native-macos: Not Found
  npmGlobalPackages:
    *react-native*: Not Found
```

## Description

got error: `error str.replace is not a function.
TypeError: str.replace is not a function`

<img width="1215" alt="image" src="https://user-images.githubusercontent.com/8704175/231839684-c0d5e2d6-6e5e-4fce-bbc7-ab4763d89128.png">

run at `ios` folder: `xcodebuild -list -json`

<img width="916" alt="image" src="https://user-images.githubusercontent.com/8704175/231840030-4d1dd2d0-f46e-4089-aec0-8d89a6e40ec1.png">

## Reproducible Demo

if a react native project has more than 1 project(1 main app + 1 widget). it will throw error like this.

my ios folder like this:

[...]

Issue #1883, opened 2023-03-21
Error: str.replace is not a function


JosueOlivero:
## Environment

System:
    OS: macOS 13.2.1
    CPU: (8) x64 Intel(R) Core(TM) i5-8259U CPU @ 2.30GHz
    Memory: 116.00 MB / 8.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 14.17.0 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 6.14.13 - /usr/local/bin/npm
    Watchman: 2023.03.13.00 - /usr/local/bin/watchman
  Managers:
    CocoaPods: 1.12.0 - /Users/jolivero/.rbenv/shims/pod
  SDKs:
    iOS SDK:
      Platforms: DriverKit 22.2, iOS 16.2, macOS 13.1, tvOS 16.1, watchOS 9.1
    Android SDK: Not Found
  IDEs:
    Android Studio: 2022.1 AI-221.6008.13.2211.9619390
    Xcode: 14.2/14C18 - /usr/bin/xcodebuild
  Languages:
    Java: 18.0.1.1 - /usr/bin/javac
  npmPackages:
    @react-native-community/cli: ^10.2.1 => 10.2.1
    react: ^18.2.0 => 18.2.0
    react-native: ^0.71.4 => 0.71.4
    react-native-macos: Not Found
  npmGlobalPackages:
    *react-native*: Not Found

## Description

Hello there,
I'm trying a new app, and I got this error.
Any idea?

<img width="1440" alt="image" src="https://user-images.githubusercontent.com/16158669/226716108-8627ac1b-18ec-4042-9661-431d3023e0d6.png">

Thanks!

berikiushi:
@JosueOlivero 

try this, it helped me - https://stackoverflow.com/a/68579858/2005679

JosueOlivero:
> @JosueOlivero
> 
> try this, it helped me - https://stackoverflow.com/a/68579858/2005679

I already have Command Line Tools selected. Just in case I re-select it again, but it didn't work.

emilstepanian:
Same issue here

pen1064:
same here, but I am using Windows
<img width="815" alt="image" src="https://user-images.githubusercontent.com/45325095/227679001-a2486c43-6630-44c1-8c84-1f308dd7094d.png">

DEVARSHI-GOHEL:
[...]

PR #1909, opened 2023-04-14
fix: str.replace not a function issue

adamTrz:
Summary:
---------

`xcodebuild` can throw but when it does inside `JSON.parse` we won't get any error in `catch` block which in turn will cause another error `Error: str.replace is not a function` in CLIError's constructor. 

Partially related to #1883 and #1908  -not a real fix for those issues but at least it will make debugging bit better :)


TMisiukiewicz (approved):

szymonrybczak (approved):
👍

Issue #1912, opened 2023-04-15
Cannot find module './xcode'


benlieb:
## Environment

Run `react-native info`  returns same error:

```
npx react-native info
node:internal/modules/cjs/loader:936
  throw err;
  ^

Error: Cannot find module './xcode'
Require stack:

```

## Description

```
pod install: 

Error: Cannot find module './xcode'
Require stack:
- /Users/user/myapp/node_modules/@react-native-community/cli-doctor/build/tools/healthchecks/index.js
```

The basic init doesn't finish installing. I've tried with a number of version of react-native, and none work. It was working last week when I installed a sample app.

```
 npx react-native@0.71.3 init beatsperminute --version 0.71.3
```

I can in fact see that `./xcode` is not in the build dir:

```
    ▾ [   ]cli-doctor/
      ▾ [   ]build/
        ▸ [   ]commands/
        ▾ [   ]tools/
          ▾ [   ]healthchecks/
              [   ]androidHomeEnvVariable.d.ts
              [   ]androidHomeEnvVariable.d.ts.map
              [   ]androidHomeEnvVariable.js
              [   ]androidHomeEnvVariable.js.map
              [   ]androidNDK.d.ts
              [   ]androidNDK.d.ts.map
              [   ]androidNDK.js
              [   ]androidNDK.js.map
              [   ]androidSDK.d.ts
              [   ]androidSDK.d.ts.map
              [   ]androidSDK.js
              [   ]androidSDK.js.map
              [   ]androidStudio.d.ts
              [   ]androidStudio.d.ts.map
              [   ]androidStudio.js
              [   ]androidStudio.js.map
              [   ]cocoaPods.d.ts
              [   ]cocoaPods.d.ts.map
              [   ]cocoaPods.js
              [   ]cocoaPods.js.map
              [   ]common.d.ts
              [   ]common.d.ts.map
              [   ]common.js
              [   ]common.js.map
              [   ]index.d.ts
              [   ]index.d.ts.map
              [   ]index.js
              [   ]index.js.map
              [   ]iosDeploy.d.ts
              [   ]iosDeploy.d.ts.map
              [   ]iosDeploy.js
              [   ]iosDeploy.js.map
              [   ]jdk.d.ts
              [   ]jdk.d.ts.map
              [   ]jdk.js
              [   ]jdk.js.map
              [   ]nodeJS.d.ts
              [   ]nodeJS.d.ts.map
              [   ]nodeJS.js
[...]

Issue #1637, opened 2022-07-06
Couldn't find template.config.js file inside "react-native" template. Please make sure the template is valid


mhmdalii:
## Ask your Question

Hi , 
i face this isusse once i get new macbook m1 and try all solution and doen't get any one to solve this , can any one help in this ?
<img width="663" alt="Screen Shot 2022-07-06 at 2 46 36 PM" src="https://user-images.githubusercontent.com/33903920/177555496-689faf07-2fc1-4665-b4c4-6ef2bd436de5.png">

ReAlign:
+1

icefee:
add --npm at the end of your command

StebanDev:
@icefee's solution did the trick for me.

mrlexz:
@icefee thanks

hclj37:
> add --npm at the end of your command

It works!

The reason for adding `--npm`:

> If you have both `yarn` and `npm` installed on your machine, React Native CLI will always try to use `yarn`, so even if you use `npx` utility, only `react-native` executable will be installed using `npm` and the rest of the work will be delegated to `yarn`. You can force usage of npm adding `--npm` flag to the command.

Reference:
https://github.com/react-native-community/cli/blob/main/docs/init.md#using-npx-utility

Murtala-1:
npx react-native init YourAwesomeProject --npm

you can also uninstall yarn globally
npm uninstall yarn

jesualdo13:
Definitely, Yarn is the problem. 

npx react-native init YourAwesomeProject --npm  

That worked for me

szymonrybczak:
Hey, the problem is with Yarn, to be exact the problem is Plug’n’Play, we don't support it. The solution for this is:
1. Downgrade `yarn` to v1,
2. Disable Plug’n’Play in your `yarn` config,
3. Pass `--npm` flag to init command.

programerTala:
please help me i have issue to create new react native applictaion
for more information please check [this link ](https://github.com/facebook/react-native/issues/37301)

adamTrz:
Closing the issue for now since we seem to have found a workaround for now (while we still work on proper solution 🙂 )

amilgasimov:
> add --npm at the end of your command

thanks

atlj:
For those who don't want to uninstall yarn globally:

If you have followed the docs here https://yarnpkg.com/getting-started/install and used corepack, just run the following command:

`corepack prepare yarn@1.22.19 --activate`

[...]

PR #1931, opened 2023-05-05
fix: improve flow in `init` command on higher Yarn versions

szymonrybczak:
Summary:
---------
Closes https://github.com/react-native-community/cli/issues/1637
Closes https://github.com/react-native-community/cli/issues/1912

This PR improves flow creating project when user has higher Yarn version.

Test Plan:
----------
1. Clone the repository and do all the required steps from the [Contributing guide](https://github.com/react-native-community/cli/blob/main/CONTRIBUTING.md).
2. Install yarn at version 2.x.x.
3. Run this command:
```other
```
It should ask to create `.yarnrc.yml` inside project or to use Yarn Classic.
4. Install yarn at version 3.x.x.
5. Run this command:
```other
```
It should detect your `.yarnrc.yml` and check what `nodeLinker` is specified. If it isn't `node-modules`, CLI should ask to change it.
6. Install yarn at version 1.x.x.
7. Run this command:
```other
```
It should create project correctly.

### DIFF
--- a/packages/cli/src/commands/init/init.ts
+++ b/packages/cli/src/commands/init/init.ts
@@ -22,6 +22,11 @@ import {installPods} from '@react-native-community/cli-doctor';
 import banner from './banner';
 import TemplateAndVersionError from './errors/TemplateAndVersionError';
 import addNodeLinker from '../../tools/addNodeLinker';
+import {getYarnVersionIfAvailable} from '../../tools/yarn';
+import semver from 'semver';
+import prompts from 'prompts';
+import findUp from 'find-up';
+import {parse, stringify} from 'yaml';
 
 const DEFAULT_VERSION = 'latest';
 
@@ -79,6 +84,32 @@ function getTemplateName(cwd: string) {
   return name;
 }
 
+async function overwriteYarnrcFile(filePath: string, fileContent: any) {
+  logger.info(
+    'Detected ".yarnrc.yml" file but it doesn\'t contain "nodeLinker: node-modules". To create React Native project you need to use "node-modules" as "nodeLinker"',
+  );
+
+  const {value} = await prompts({
+    type: 'confirm',
+    name: 'value',
+    message:
+      'Do you want to add "nodeLinker: node-modules" to "yarnrc.yml" file?',
+    initial: true,
+  });
+
+  if (value) {
+    fileContent.nodeLinker = 'node-modules';
+    fs.writeFileSync(filePath, stringify(fileContent), {
+      encoding: 'utf8',
+      flag: 'w',
+    });
+  } else {
+    throw new CLIError(
+      'In order to create React Native app you need to use "node-modules" as "nodeLinker" in ".yarnrc.yml", or use Yarn Classic.',
+    );
+  }
+}
+
 async function createFromTemplate({
   projectName,
   templateUri,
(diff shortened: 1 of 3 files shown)


packages/cli/src/commands/init/init.ts
```diff
@@ -163,6 +164,10 @@ async function installDependencies({
 }) {
   loader.start('Installing dependencies');
 
+  if (!npm) {
+    addNodeLinker(root);
+  }
+
```
szymonrybczak, line 170: 💡 Maybe we can add the `.yarnrc.yml` directly to the official template?

packages/cli/src/tools/addNodeLinker.ts
```diff
@@ -0,0 +1,17 @@
...
+import fs from 'fs';
+
+/**
+ Creates a `.yarnrc.yml` file in the root of the project to force Yarn to use the `node-modules` linker, because React Native doesn't support the Plug'n'Play node linker.
+ */
+
+const addNodeLinker = (root: string) => {
+  const yarnrcFileContent = 'nodeLinker: node-modules\n';
+
+  fs.writeFileSync(path.join(root, '.yarnrc.yml'), yarnrcFileContent, {
```
tido64, line 11: I think this will force users onto Yarn v3. I don't think we want to do that. Can we instead try to detect the version and suggest solutions?

kelset (changes requested):
blocking for merging until the yarn v3 concern is addressed

packages/cli/src/tools/addNodeLinker.ts
tido64: I did some testing locally and it looks like just adding a `.yarnrc.yml` won't trigger Yarn to use v3, which is good. But I still think we want to avoid any confusion having this extra file may introduce.

tido64: Yeah, basically:
- If Yarn Classic, all is good.
- Else check if `.yarnrc.yml` exists and whether `nodeLinker` is correctly set (and suggest/update if not)

If you're using Yarn v3, you can assume that `.yarnrc.yml` exists as you are required to specify the path to Yarn. You can find an example here: https://github.com/microsoft/react-native-test-app/blob/trunk/.yarnrc.yml

packages/cli/src/commands/init/init.ts
```diff
@@ -98,6 +130,79 @@ async function createFromTemplate({
...
+          );
+
+          if (
+            !yarnYmlContent.nodeLinker ||
+            yarnYmlContent.nodeLinker !== 'node-modules'
+          ) {
+            await overwriteYarnrcFile(yarnrcYmlPathLocation, yarnYmlContent);
+          }
+
+          addNodeLinker(templateSourceDir);
```
tido64, line 159: Won't this just overwrite the content in `overwriteYarnrcFile`?

satya164:
The ideal approach would be to set the Yarn version in the project itself than relying on global Yarn. There being fundamental differences and breaking changes between the Yarn versions, this approach allows users to work on multiple projects needing different Yarn versions. This is also recommended workflow for Yarn >= 2 - if people want to use newer Yarn, then they need to run `yarn set version berry`.

However we can do similar for Yarn 1 too, something like: `yarn set version 1`. imo it'd best to settle on the Yarn version we support and adjust the template with appropriate config file and run `yarn set version` instead of using the globally installed version.

packages/cli/src/commands/init/init.ts
szymonrybczak: Nope, because to this function I'm passing template directory, `overwriteYarnrcFile` is changing content inside `.yarnrc.yml` file which lives in project.

szymonrybczak:
I would like to renew discussion under this pull request, because it is getting a little bit stale. 

@satya164 Thanks for feedback! I think your point is very interesting, but it may be hard to implement. The current implementation is more flexible to me. I think we cannot specify Yarn version in template, and we actually don't know what version user wants to use because it depends on preferences, what Yarn version project is using etc.
Right now we detect Yarn version by doing `yarn -v`, and if the version is >= 2 we're checking what `nodeLinker` they have in `.yarnrnc.yml` and we tell them to use `nodeLinker: node-modules`.  With that we don't care what Yarn version user is using, we only care about `nodeLinker` because this is important for us. If the `nodeLiner` is not `node-modules` project cannot be initialised correctly.

satya164:
> I think your point is very interesting, but it may be hard to implement

Why is it hard to implement? It's just running one command one in the project.

> we actually don't know what version user wants to use because it depends on preferences

IMO trying to support multiple versions of Yarn means more effort maintaining it. is there a reason anyone actually would want to stay on old version of Yarn? I have only seen people stay on Yarn 1 because they thought that Yarn >= 2 doesn't work with React Native.

szymonrybczak:
After some internal discussion with @satya164, we came to conclusion that actually it would be better and easier in long term to provide default Yarn version in React Native template, but are there any blockers from providing default Yarn version >=2 (of course with `nodeLinker: node-modules`) in template? cc. @kelset @cortinico

Issue #1936, opened 2023-05-11
☂️ Improve React Native onboarding experience


arushikesarwani94:
## Description
Hi everyone!

This is an umbrella issue to collect a set of tasks aimed to improve the React Native [onboarding](https://reactnative.dev/docs/getting-started) experience. Onboarding experience as part of the DX pillar aims to improve user workflows like :

* creating or starting their first React Native app;
* trying out built-in React Native APIs for the first time;
* installing their first third-party package in a React Native project;
* collaborating on a React Native project with others for the first time;
* setting up a new machine to work on an existing React Native app

We believe that everyone would benefit by improving the core developer experience, hence this umbrella issue to help us cleanup and restructure some of our codebase.

## How to Test
It’s crucial that you test the changes you submit. The proper way of doing this is:
1. Build cli codebase using : `node ./scripts/build.js && yarn build:debugger`
2. Link the cli-doctor using : `yarn link`
3. In the example React Native app(Awesome Project), run `yarn link "@react-native-community/cli-doctor"`
4. Run the CLI Doctor using : `npx @react-native-community/cli doctor`

## The Tasks
The following is a list of tasks we think might aid the bootstrapping experience of React Native. The list is a living one: the more we improve with the situation, the more patterns we can discover. Whenever we found some pattern that we can add, we will attach this to the list.

If you want to tackle some specific task which is not listed, feel free to post a comment with a brief description of the task and I'll make sure to add it to the list of tasks so it is tracked.

Please comment below if you're willing to take the stance on any of those task and wait for a confirmation before start working on it. Please note that if you claim a task, but don't submit a PR within 2 week, we're going to free up that task for someone else.

* [ ] [feat : Custom Adb ports #1864](https://github.com/react-native-community/cli/issues/1864)

**cli-doctor :**

* [ ] [feat : detect wrong app dependencies #1848](https://github.com/react-native-community/cli/issues/1848)
* [ ] Java version check in [versionRanges](https://github.com/react-native-community/cli/blob/main/packages/cli-doctor/src/tools/versionRanges.ts) should be [11 - 18] (i.e. 19 is not supported).
* [ ] feat : doctor should detect & report if packager is running correctly :
https://github.com/react-native-community/cli/blob/main/packages/cli-tools/src/isPackagerRunning.ts

**Open ended discussions :**

* [ ] Reduce “Installing Bundler” step noise

During the “Installing Bundler” step, seemingly unnecessary warnings are printed, e.g.
[...]

PR #1940, opened 2023-05-17
chore (doctor): update Java version range to be strictly 11-18

tarunrajput:
Summary:
---------

Updated Java version check in [versionRanges](https://github.com/react-native-community/cli/blob/main/packages/cli-doctor/src/tools/versionRanges.ts) to be strictly [11 - 18] (i.e. 19 is not supported).

Part of #1936

Test Plan:
----------

Updated tests and run `yarn test`


adamTrz (approved):

tarunrajput:
@arushikesarwani94, I did try to build with jdk19 and it worked fine on both old and new arch. However, Gradle is not fully compatible with jdk20 yet because Kotlin lacks support for it.
@cortinico may have more accurate information on this

PR #1934, opened 2023-05-10
fix: add missing argument to xcodebuild shell (destination)

MichelBahl:
- destination is used for to specify the arch for iOS build
- this is needed for modules which require arch x84_64

Summary:
---------

With the destination build parameter it can be specified what destination (arch x86_64) gonna be used to build iOS

In case you are using an apple silicon Mac to build your app and you  are using Pod dependencies which were not made for apple silicon it were not possible to change the arch for these dependencies from cli.

Error:
`could not find module 'Adyen' for target 'arm64-apple-ios-simulator'; found: x86_64-apple-ios-simulator, at:`

xcode would prompt to Build for Rosetta and add en config file

buildforrosetta

And create a settings file (which has no effect for cli build):
Bildschirmfoto 2023-05-10 um 18 15 45

Bildschirmfoto 2023-05-10 um 18 46 16

Test Plan:
----------

1. Install this fork
2. Run following command `npx react-native run-ios --simulator="iPhone 14 Pro" --destination arch=x86_64`

Issue #1918, opened 2023-04-18
Cannot specify run-ios to use (Rosetta) destination


uloco:
## Environment

System:
    OS: macOS 13.3.1
    CPU: (8) arm64 Apple M1
    Memory: 198.61 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 18.14.0 - ~/Library/Caches/fnm_multishells/93529_1681811907411/bin/node
    Yarn: 1.22.19 - ~/Library/Caches/fnm_multishells/93529_1681811907411/bin/yarn
    npm: 9.6.4 - ~/Library/Caches/fnm_multishells/93529_1681811907411/bin/npm
    Watchman: Not Found
  Managers:
    CocoaPods: 1.12.0 - /opt/homebrew/bin/pod
  SDKs:
    iOS SDK:
      Platforms: DriverKit 22.4, iOS 16.4, macOS 13.3, tvOS 16.4, watchOS 9.4
    Android SDK: Not Found
  IDEs:
    Android Studio: 2022.1 AI-221.6008.13.2211.9514443
    Xcode: 14.3/14E222b - /usr/bin/xcodebuild
  Languages:
    Java: 11.0.14.1 - /usr/bin/javac
  npmPackages:
    @react-native-community/cli: Not Found
    react: 18.2.0 => 18.2.0 
    react-native: 0.71.6 => 0.71.6 
    react-native-macos: Not Found
  npmGlobalPackages:
    *react-native*: Not Found

## Description

Since Xcode 14.3 there is no longer a way to run Xcode with the rosetta flag enabled.
Also using `arch -x86_64 react-native run-ios` has no effect as it seems.
We need to run our simulators on rosetta though, since some of our dependencies are outdated and not maintained (and there is no option of swichting because hardware...)

In Xcode, there is a way around this. You can enable Rosetta in Xcode as a destination for the simulators. After enabling these in `Product > Destination > Destination Architectures > Show Rosetta Destinations` you can choose those in Xcode and the builds start working again.`

Unfortuantely you cannot specify the iOS Simulator name with (Rosetta) addition in `run-ios --simulator`. It ignores that part completely and uses the default simulator.

### React Native Version

0.71.6

### Steps to reproduce

* Enable Rosetta as destination in Xcode
* Try running your app with `npx react-native run-ios --simulator "iPhone 8 (Rosetta)`
* Build is not run on the specified destination simulator

## Reproducible Demo

[...]

PR #1947, opened 2023-05-30
fix: add missing argument to xcodebuild shell (destination) for v8.x

jtparrett:
Summary:
---------

This is a fix to allow the run-ios command to accept a destination arg so that rosetta devices can be used via the terminal build command on projects using v8 of the cli.

This is a v8 backport fix of this PR https://github.com/react-native-community/cli/pull/1934

Please also read: https://github.com/react-native-community/cli/issues/1918

### DIFF
--- a/packages/platform-ios/src/commands/runIOS/index.ts
+++ b/packages/platform-ios/src/commands/runIOS/index.ts
@@ -37,6 +37,7 @@ type FlagsT = {
   verbose: boolean;
   port: number;
   terminal: string | undefined;
+  destination?: string;
 };
 
 function runIOS(_: Array<string>, ctx: Config, args: FlagsT) {
@@ -317,7 +318,7 @@ function buildProject(
       '-scheme',
       scheme,
       '-destination',
-      `id=${udid}`,
+      `id=${udid}` + (args.destination ? ',' + args.destination : ''),
     ];
     // @todo use `getLoader` from cli-tools package
     const loader = ora();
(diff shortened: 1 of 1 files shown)


thymikee:
It's gonna be weird if we had this in v8.x and not in 9 or 10. What's blocking you from upgrading to the latest?

thymikee:
Yeah, it would be best to backport to other versions as well. I could release that next week I guess

thymikee:
And also, since it's a small change, you could probably use `patch-package` in your project to avoid waiting

jtparrett:
@thymikee Awesome. We have a patch in place. But I will create backports for v9 and v10 also soon.

Thanks for the support!

packages/platform-ios/src/commands/runIOS/index.ts
```diff
@@ -604,6 +605,10 @@ export default {
       description:
         'Explicitly set device to use by name.  The value is not required if you have a single device connected.',
     },
+    {
+      name: '--destination <string>',
+      description: 'Explicitly extend distination e.g. "arch=x86_64"',
```
JulioCVaz, line 610: **suggestion**: do you mean `destination` instead of `distination`

senkevichdv:
Hello! Any updates on this improvements? We are not able to safely update react-native version to the latest and would love to get this fix.

<details>
<summary>PS</summary>

Also I was not able to properly use `patch-package` to update it directly so maybe there is a some kind of README anywhere? Steps from the root README are not working for me

</details>

packages/platform-ios/src/commands/runIOS/index.ts
szymonrybczak: https://github.com/react-native-community/cli/pull/2025

jtparrett: Sorry, been away and missed this. Thanks for the PR 👍

szymonrybczak:
I'm closing this PR since `0.69` (the version that `8.x` corresponds to) is not longer supported, so Core Team will never release this change anyway.

PR #1946, opened 2023-05-30
fix: `--verbose` option not working for `run-ios`

szymonrybczak:
Summary:
---------
Fixes https://github.com/react-native-community/cli/issues/1391
Closes https://github.com/react-native-community/cli/issues/1391

Added `.enablePositionalOptions();` to `CommanderCommand`.

Read [here](https://github.com/tj/commander.js#parsing-configuration) more about this option.

Test Plan:
----------

1. Clone the repository and do all the required steps from the [Contributing guide](https://github.com/react-native-community/cli/blob/main/CONTRIBUTING.md)
2. Run this command:

```other
node /path/to/react-native-cli/packages/cli/build/bin.js run-ios --verbose
```
Should correctly pass the --verbose flag to the `run-ios` command.

PR #1953, opened 2023-06-05
fix: always add `--verbose` to the command

szymonrybczak:
Summary:
---------

Recently we [added](https://github.com/react-native-community/cli/pull/1946) `.enablePositionalOptions()` to the commander, and this caused one problem, because we specified `--verbose` directly to the `react-native` so we were able to do `react-native --verbose`, but if we did `react-native run-android --verbose` it didn't work - so the solution for this is to add `--verbose` to every command.

Test Plan:
----------

`npx react-native -v` <- should give version of CLI
`npx react-native --verbose` <- shouldn't work
`npx react-native run-android --verbose` <- should pass verbose to `run-android`

### DIFF
--- a/__e2e__/__snapshots__/default.test.ts.snap
+++ b/__e2e__/__snapshots__/default.test.ts.snap
@@ -5,7 +5,6 @@ exports[`shows up help information without passing in any args 1`] = `
 
 Options:
   -v                            Output the current version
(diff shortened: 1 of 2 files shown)


thymikee:
Need to update snapshots

thymikee (approved):

adamTrz (approved):

PR #1957, opened 2023-06-08
fix(ios): fix Xcode defaulting to 'Release'

tido64:
Summary:
---------

`getBuildConfigurationFromXcScheme` currently assumes where the `.xcodeproj` lives (it could be anywhere). The proper fix is to get the path from `.xcworkspace`, but that's a much more intrusive change. The least we can do for now is to fall back to `Debug`.

Test Plan:
----------

```
cd react-native-test-app
npm run set-react-version 0.72
yarn
cd example
yarn ios
```

Take note of whether `-configuration` is correctly set when `xcodebuild` is run. This is what it looks like without this fix:

```
info Building (using "xcodebuild -workspace Example.xcworkspace -configuration  -scheme Example -destination id=0")
```

### DIFF
--- a/packages/cli-platform-ios/src/tools/getConfigurationScheme.ts
+++ b/packages/cli-platform-ios/src/tools/getConfigurationScheme.ts
@@ -11,9 +11,12 @@ export function getConfigurationScheme(
 ) {
   if (scheme && mode) {
     return mode;
-  } else if (scheme) {
-    return getBuildConfigurationFromXcScheme(scheme, mode, sourceDir);
   }
 
-  return mode || 'Debug';
+  const configuration = mode || 'Debug';
+  if (scheme) {
+    return getBuildConfigurationFromXcScheme(scheme, configuration, sourceDir);
+  }
+
+  return configuration;
 }


tido64:
@adamTrz, @thymikee: Is this something we can backport to 10.x and 11.x?

cc @kelset

kelset:
yeah most likely we need to backport to both so that we can then have in RN 72 and 71.

szymonrybczak (approved):

thymikee:
Released v12.0.0-alpha.4, v11.3.2 and v10.2.4 with this fix. Btw, the `getBuildConfigurationFromXcScheme` fn accepts `sourceDir`, which is configurable from the `project.ios` config in `react-native.config.js`, not quite assumed :D

tido64:
> Released v12.0.0-alpha.4, v11.3.2 and v10.2.4 with this fix.

Thank you ❤️ 

> Btw, the `getBuildConfigurationFromXcScheme` fn accepts `sourceDir`, which is configurable from the `project.ios` config in `react-native.config.js`, not quite assumed :D

`sourceDir` is where `Podfile` and the generated `.xcworkspace` live, but `.xcodeproj` can live anywhere. Otherwise that's what I would've suggested as workaround 😄

thymikee:
Let me know if you have ideas on how to improve it (e.g. extra configuration) 👍🏼

PR #1958, opened 2023-06-08
fix(ios): fix crash when `--mode` is specified

tido64:
Summary:
---------

Assumptions about where the `.xcodeproj` lives causes `checkIfConfigurationExists` to crash because `undefined` was passed as project.

Test Plan:
----------

```
cd react-native-test-app
npm run set-react-version 0.72
yarn
cd example
yarn ios --mode 'Debug'
```

The last command should not crash.

### DIFF
--- a/packages/cli-platform-ios/src/tools/checkIfConfigurationExists.ts
+++ b/packages/cli-platform-ios/src/tools/checkIfConfigurationExists.ts
@@ -1,10 +1,15 @@
-import {CLIError} from '@react-native-community/cli-tools';
+import {CLIError, logger} from '@react-native-community/cli-tools';
 import {IosProjectInfo} from '../types';
 
 export function checkIfConfigurationExists(
   project: IosProjectInfo,
   mode: string,
 ) {
+  if (!project) {
+    logger.warn(`Unable to check whether "${mode}" exists in your project`);
+    return;
+  }
+
   if (!project.configurations.includes(mode)) {
     throw new CLIError(
       `Configuration "${mode}" does not exist in your project. Please use one of the existing configurations: ${project.configurations.join(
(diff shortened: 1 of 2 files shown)


packages/cli-platform-ios/src/tools/checkIfConfigurationExists.ts
```diff
@@ -1,10 +1,15 @@
-import {CLIError} from '@react-native-community/cli-tools';
+import {CLIError, logger} from '@react-native-community/cli-tools';
 import {IosProjectInfo} from '../types';
 
 export function checkIfConfigurationExists(
   project: IosProjectInfo,
   mode: string,
 ) {
+  if (!project) {
+    logger.warn(`Unable to check whether "${mode}" exists in your project`);
```
tido64, line 9: I don't know if this warning is actually useful. I'd prefer to not have it here, but I'll let you decide.

tido64:
> Super strange that example workspace don't have `configurations` field present when we list them... 🤔

That is because the CLI is trying to look for `.xcodeproj` where `.xcworkspace` currently lives. That isn't always the case. `Podfile` and `.xcworkspace` are the only two files you can make assumptions about. `.xcodeproj` can live anywhere. The only you can find it is by parsing the `.xcworkspace`, but it may contain several.

adamTrz:
> > Super strange that example workspace don't have `configurations` field present when we list them... 🤔
> 
> That is because the CLI is trying to look for `.xcodeproj` where `.xcworkspace` currently lives. That isn't always the case. `Podfile` and `.xcworkspace` are the only two files you can make assumptions about. `.xcodeproj` can live anywhere. The only you can find it is by parsing the `.xcworkspace`, but it may contain several.

Yeah, got it. Still, one would hope that `xcodebuild -list` would output all configurations but it does not :| 

I'll try to look at your PR ASPA, seems like `interactive` mode is broken as well since it's also calling `getProjectInfo` function...

adamTrz (approved):

PR #1971, opened 2023-06-16
chore: Remove fs-extra reference from cli-plugin-metro

huntie:
## Summary

Refactor to reference `fs` instead of `fs-extra` within `cli-plugin-metro`, removing this (implicit) npm dependency.

Changelog: None

## Test Plan

[Incoming]

### DIFF
--- a/packages/cli-plugin-metro/src/commands/bundle/assetCatalogIOS.ts
+++ b/packages/cli-plugin-metro/src/commands/bundle/assetCatalogIOS.ts
@@ -7,7 +7,7 @@
  */
 
 import path from 'path';
-import fs from 'fs-extra';
+import fs from 'fs';
 import type {AssetData} from 'metro';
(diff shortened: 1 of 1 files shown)


thymikee (approved):

packages/cli-plugin-metro/src/commands/bundle/assetCatalogIOS.ts
```diff
@@ -49,22 +49,25 @@ export function isCatalogAsset(asset: AssetData): boolean {
 }
 
 export function writeImageSet(imageSet: ImageSet): void {
-  fs.mkdirsSync(imageSet.basePath);
+  fs.mkdirSync(imageSet.basePath);
```
robhogan, line 52: @huntie This should have `{ recursive: true }` to be equivalent to `mkdirs`.

PR #1972, opened 2023-06-16
breaking: Remove fallback flow for Metro config defaults (0.73)

huntie:
# Summary

This follows up [the `@react-native/metro-config` setup added in React Native 0.72](https://github.com/facebook/react-native/pull/36502). It removes the fallback to a replicated internal copy of Metro config defaults in `cli-plugin-metro`, effectively reverting https://github.com/react-native-community/cli/pull/1896.

Rollout plan recap:

- **0.72**: Extending `@react-native/metro-config` is recommended, but falls back if users have not updated their `metro.config.js`, logging a warning.
- **0.73 (this diff**): Updating `metro.config.js` to the new format is **required** — log warning and (later on) fail.

Changelog: **[Breaking]** Remove fallback flow for Metro config defaults (React Native 0.73)

# Test Plan

Using `rn-tester` configured with a locally `yarn link`ed copy of RN CLI.

**✅ Correct config logs no warning and builds successfully**

image

**✅ Incorrect config logs updated warning and fails**

~~⬇️ FYI @cortinico @kelset, in case you have opinions for better DevX. It _is_ marginally restrictive that we require a reference to `@react-native/metro-config` in the immediate `metro.config.js` file — perhaps if we detect this more strongly we can fail rather than warn. However I can't think of a simple way to do this — and we will have had the 0.72 release cycle for people to update their projects by this point.~~ **Addressed with more visible warning**.

image


thymikee:
This ties the move of the `cli-plugin-metro` effort to the core with v0.73, right? Maybe we should start merging those changes to a parallel branch with the plugin removal?

huntie:
@thymikee It does — I was aiming to stack a sequence of simplification changes, including this PR, here in the CLI repo first. This is to slightly reduce the surface area that we are migrating into the RN repo (and from TypeScript into Flow 🙈).

A change like this also gets better visibility in this repo for now.

kelset:
added a mention here https://github.com/reactwg/react-native-releases/discussions/64 so that we/the 73 release crew won't forget that there's BC on this angle

huntie:
@kelset Sidenote: I'm looking to merge this change ahead of work to migrate and collapse `cli-plugin-metro` back into the React Native repository, with the intent being to minimise the number of files/surface area before this migration. 

---

With some over-the-weekend reflection — the flow in this PR should probably hard-error (helpful in particular to pin down this error cause in CI). ➡️ **update incoming**.

For the user, it might be better if we allowed the entire `metro.config.js` file to be optional (especially after https://github.com/facebook/metro/issues/977 ships). We can look at this post-migration, since it'll be possible to refer to `@react-native/metro-config` in the same repo.

packages/cli-plugin-metro/src/tools/loadMetroConfig.ts
```diff
@@ -97,26 +99,22 @@ export default async function loadMetroConfig(
...
+=================================================================================================
 
-    const loadedConfig = await loadConfig(
-      {cwd: ctx.root, ...options},
-      // Provide React Native defaults on top of Metro defaults
-      getDefaultMetroConfig(ctx),
-    );
+From React Native 0.73, your project's Metro config should extend '@react-native/metro-config'
+or it will fail to build. Please copy the template at:
+https://github.com/facebook/react-native/blob/main/packages/react-native/template/metro.config.js
```
motiz88, line 108: Wonder if this should be pinned (at least to a branch) so it remains accurate in the future.

motiz88 (approved):

packages/cli-plugin-metro/src/tools/loadMetroConfig.ts
huntie: Yeah fair. We don't have `0.73-stable` yet — and the counter-risk is pointing to a stale copy. Will keep an eye on it.

szymonrybczak (approved):

PR #1978, opened 2023-06-20
Update `react-native run-ios` command error output.

cpojer:
Summary:
---------

This PR cleans up the error message when running `react-native run-ios`. It is now shorter, variables are escaped, and most importantly it removes the "We" which has no meaning in the context of running a command on my computer.

Test Plan:
----------

Check the output of `react-native run-ios` when it throws an error.


thymikee (approved):
Makes sense 👍🏼

PR #1980, opened 2023-06-21
fix: init when there's a package.json in folder

thymikee:
Summary:
---------

When `init` in a folder that contains any kind of `package.json` the failed with:
```
error Failed to load configuration of your project.
```

This is due to `reactNativeVersion` introduced in https://github.com/react-native-community/cli/pull/1870 accessing the `initialConfig.reactNativePath` getter, which throws in situations where there's no `package.json` file available in the directory. Reading this in `finalConfig` is fine, because when there's no `package.json`, it skips the logic inside the `.reduce`, which accesses `initialConfig.reactNativePath` variable. 

Tbh, this is rather convoluted logic and because of that during code review we failed identifying it as a potential problem. Let's treat this as a band-aid for now and as a followup think about refactoring this piece. 

Test Plan:
----------

I haven't really tested the versioned link generation, so please make sure to verify it

### DIFF
--- a/packages/cli-config/src/loadConfig.ts
+++ b/packages/cli-config/src/loadConfig.ts
@@ -53,6 +53,25 @@ function getDependencyConfig(
   ) as DependencyConfig;
 }
 
+// Try our best to figure out what version of React Native we're running. This is
+// currently being used to get our deeplinks working, so it's only worried with
+// the major and minor version.
+function getReactNativeVersion(reactNativePath: string) {
+  try {
+    let semver = version.current(reactNativePath);
+    if (semver) {
+      // Retain only these version, since they correspond with our documentation.
+      return `${semver.major}.${semver.minor}`;
+    }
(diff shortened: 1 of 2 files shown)


TMisiukiewicz (approved):
👍

szymonrybczak (approved):

szymonrybczak:
FYI I checked if links generation works correctly. It works same as before. ✅
CleanShot 2023-06-22 at 11 05 00@2x

kelset (approved):
can confirm that the fix work, managed to test locally to generate the diff patch for upgrade-helper and by locally using this fix everything worked 👍

https://github.com/react-native-community/rn-diff-purge/commit/6ba5199fd52c9e100b23e65079e4309fb7364b5a

let's merge this one, cp into 11.x and create a new CLI release :) no rush though, atm we're aiming at doing 0.72.1 early next week

PR #1982, opened 2023-06-22
fix: resolver.resolveRequest is always undefined

ammarahm-ed:
Summary:
---------

I just upgraded to react-native 0.72.0 in our project and found out that `resolver.resolveRequest` from the project config is always getting overridden by the `getOverrideConfig` function. The PR helps fix the issue so that we are able to define our custom `resolveRequest` functions.

For example here's the `metro.config.js` file of our project where we are using `resolveRequest`. This was working in RN 0.71.11 but stopped working after upgrading to 0.72.0. 
```js
const {getDefaultConfig, mergeConfig} = require('@react-native/metro-config');
/**
 * Metro configuration for React Native
 * https://github.com/facebook/react-native
 *
 * @format
 */
// eslint-disable-next-line @typescript-eslint/no-var-requires
const path = require('path');
const config = {
  ...
};
const mergedConfig = mergeConfig(getDefaultConfig(__dirname), config);

mergedConfig.resolver = {
  ...
  resolveRequest: (context, moduleName, platform) => {
    if (moduleName ==='react') {
    console.log("LOADING REACT!");
      // Resolve react package from mobile app's node_modules folder always.
      return {
        filePath: path.resolve(path.join(__dirname, '../node_modules', "react","index.js")),
        type: 'sourceFile',
      };
    }
    return context.resolveRequest(context, moduleName, platform);
  }
};

module.exports = mergedConfig;
```

The problematic code is here:
https://github.com/react-native-community/cli/blob/a1cc7986a711e8b8ea1da07ddfec476c2c97c58f/packages/cli-plugin-metro/src/tools/loadMetroConfig.ts#L32-L36

and then the `undefined value replaces the project config value for `resolveConfig here:

https://github.com/react-native-community/cli/blob/a1cc7986a711e8b8ea1da07ddfec476c2c97c58f/packages/cli-plugin-metro/src/tools/loadMetroConfig.ts#L120-L123

Test Plan:
----------

To test the issue, create a new react-native 0.72.0 project & replace contents of `metro.config.js` with the following:

```js
const {getDefaultConfig, mergeConfig} = require('@react-native/metro-config');
/**
 * Metro configuration for React Native
 * https://github.com/facebook/react-native
 *
 * @format
 */
// eslint-disable-next-line @typescript-eslint/no-var-requires
const mergedConfig = mergeConfig(getDefaultConfig(__dirname),{});

mergedConfig.resolver = {
  resolveRequest: (context, moduleName, platform) => {

    console.log("Resolve Request Received"):

    return context.resolveRequest(context, moduleName, platform);
  }
};

module.exports = mergedConfig;
```
Without the fix `"Resolve Request Received"` never get's logged since `resolver.resolveRequest` is always undefined in the final config used by metro bundler.

### DIFF
--- a/packages/cli-plugin-metro/src/tools/loadMetroConfig.ts
+++ b/packages/cli-plugin-metro/src/tools/loadMetroConfig.ts
@@ -6,6 +6,7 @@ import {
   loadConfig,
   mergeConfig,
   resolveConfig,
+  ResolverConfigT,
 } from 'metro-config';
 import {CLIError, logger} from '@react-native-community/cli-tools';
 import type {Config} from '@react-native-community/cli-types';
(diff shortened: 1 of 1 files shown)


huntie (approved):
LGTM, thanks :)

szymonrybczak (approved):
Cool, thanks! Tests are failing, because yesterday was released `0.72` - in snapshots we still have `0.71`, so no worries :)

thymikee (approved):

PR #1996, opened 2023-07-03
npmPackageName should be resolved from project root, not from cli-plugin-metro

acoates-ms:
Summary:
---------

When resolving npmPackageName currently, we are resolving it from the cli-plugin-metro package. But cli-plugin-metro does not have a direct dependency on any of the out of tree platform packages (say react-native-windows or react-native-macos).  Instead we should be resolving npmPackageName from the project root, which will have the out of tree package as a dependency.

This generally works today because so many package managers hoist everything and so packages can resolve other packages that are used in the repo even if they are not declared as dependencies.  But some package managers are moving to be more strict about this.

Test Plan:
----------

Checklist
----------


thymikee (changes requested):
Thanks! Mind run the linter to fix formatting?

PR #1999, opened 2023-07-05
breaking: Fix ability to override commands via config

huntie:
## Summary

I noticed a non-obvious bug when attempting to use `react-native.config.js` to override `commands` (i.e. the plugin system) over the CLI's built-in defaults.

While the below printout shows the project-configured commands being successfully read after the internal commands, **these are never applied** — due to a bug in all versions of the [commander](https://www.npmjs.com/package/commander) library where `commands` are stored as an array structure where the **first registered command with a given `name` is used** (and silently failing upon re-registering a command later 😐).

image

Changelog:
[Breaking] Custom `commands` in `react-native.config.js` files in libraries or projects will now override built-in CLI commands if defined with the same command `name`

## Test Plan

- Set up more `console.log`s
- Build
- `yarn link` to local React Native project with extended config ⬇️

```js
module.exports = {
  commands: [
    ...ios.commands,
    ...android.commands,
    bundleCommand,
    ramBundleCommand,
    startCommand,
  ],
```

image

✅ Command implementation for `react-native start` can be successfully overridden

cc @tido64 for any potential downstream impact on `rnx-kit`

### DIFF
--- a/packages/cli/src/index.ts
+++ b/packages/cli/src/index.ts
@@ -82,6 +82,19 @@ function attachCommand<C extends Command<boolean>>(
   command: C,
   config: C extends DetachedCommand ? Config | undefined : Config,
 ): void {
+  // commander@9.x will internally push commands into an array structure!
+  // Commands with duplicate names (e.g. from config) must be reduced before
+  // calling this function.
+  // https://unpkg.com/browse/commander@9.4.1/lib/command.js#L1308
+  if (program.commands.find((cmd) => cmd.name() === command.name)) {
+    throw new Error(
+      'Invariant Violation: Attempted to override an already registered ' +
+        `command: '${command.name}'. This is not supported by the underlying ` +
+        'library and will cause bugs. Ensure a command with this `name` is ' +
+        'only registered once.',
+    );
+  }
+
   const cmd = program
     .command(command.name)
     .option('--verbose', 'Increase logging verbosity')
@@ -165,7 +178,14 @@ async function setupAndRun() {
 
     logger.enable();
 
+    const commands: Record<string, Command> = {};
+
+    // Reduce overridden commands before registering
     for (const command of [...projectCommands, ...config.commands]) {
+      commands[command.name] = command;
+    }
+
+    for (const command of Object.values(commands)) {
       attachCommand(command, config);
     }
   } catch (error) {


packages/cli/src/index.ts
```diff
@@ -82,6 +82,19 @@ function attachCommand<C extends Command<boolean>>(
...
+  // commander@9.x will internally push commands into an array structure!
+  // Commands with duplicate names (e.g. from config) must be reduced before
+  // calling this function.
+  // https://unpkg.com/browse/commander@9.4.1/lib/command.js#L1308
+  // @ts-ignore
+  if (program._findCommand(command.name)) {
+    throw new Error(
+      'Invariant Violation: Attempted to override an already registered ' +
+        'command. This is not supported by the underlying library and will ' +
+        'cause bugs. Ensure a command with this `name` is only registered once.',
```
tido64, line 94: Can you add the name of the command to the error message?

packages/cli/src/index.ts
```diff
@@ -82,6 +82,20 @@ function attachCommand<C extends Command<boolean>>(
   command: C,
   config: C extends DetachedCommand ? Config | undefined : Config,
 ): void {
+  // commander@9.x will internally push commands into an array structure!
+  // Commands with duplicate names (e.g. from config) must be reduced before
+  // calling this function.
+  // https://unpkg.com/browse/commander@9.4.1/lib/command.js#L1308
+  // @ts-ignore
+  if (program._findCommand(command.name)) {
```
shadowspawn, line 90: `_findCommand` is private. You might want to copy the implementation using public methods and properties rather than risk it changing.

huntie: Definitely a small evil (note: `cmd._name` and `cmd._alias` are also private) — but I judged safe enough for this use (with a large comment).

Alternative would be tracking which commands we've added inside this module / via a wrapper.

shadowspawn: `cmd.name()` and `cmd.aliases()` are public. Both routines are setter/getter. With no argument, they return the current value.

huntie: I see. Also checked we only have `name` in our input `Command` type, so this can be simplified 👍🏻

huntie:
@tido64 @thymikee Bump :)

tido64 (approved):
You have my symbolic vote. 😄

szymonrybczak (approved):
🙌

PR #2003, opened 2023-07-05
fix: Add flipper protocol support

lunaleaps:
Summary:
---------

Allow opening `flipper:` protocol urls

See https://github.com/facebook/react-native/issues/38006#issuecomment-1601963049

This is because I was trying to open the debugger via the dev menu on an iOS Hermes project. 

For a Hermes project on iOS, we open Flipper to leverage direct debugging. See code pointer: https://github.com/facebook/react-native/blob/2910f87b580a51d997458441c9f6a5684a9d3dfa/packages/react-native/React/CoreModules/RCTDevMenu.mm#L271

The issue is that CLI currently doesn't support `flipper:` protocol. This PR adds that support and renames the function to accurately reflect intention. 

Test Plan:
----------

Verified in a locally generated 0.72 React Native file by changing the cli source:
1. Update `./node_modules/@react-native-community/cli-tools/build/throwIfNonHttpProtocol.js` to accept the `flipper:` protcol
2. re-install pods under `My72Project/ios`
3. `yarn run ios` under `My72Project`
4. Open dev menu via "Shake" on iOS simulator
5. Click "open debugger"
6. Flipper app opens

Checklist
----------

- [ ] Documentation is up to date to reflect these changes. (Does it need it) 


packages/cli-tools/src/throwIfNonAllowedProtocol.ts
```diff
@@ -0,0 +1,18 @@
...
+ * Check if a url uses an allowed protocol
+ */
+
+const ALLOWED_PROTOCOLS = new Set(['http:', 'https:', 'flipper:']);
+
+export default function throwIfNonAllowedProtocol(url: string) {
+  const _url = new URL(url);
+  const urlProtocol = _url.protocol;
+
+  if (!ALLOWED_PROTOCOLS.has(urlProtocol)) {
```
tido64, line 11: Nit: You can also have an array and use `includes` here. `Set` is a bit overkill for this use case.

lunaleaps:
For context, this isn't a major blocking issue. You can still access iOS direct debugging via Chrome

thymikee (approved):

PR #2007, opened 2023-07-07
chore: Upgrading Metro dependencies to 0.77.0

arushikesarwani94:
Summary:
---------

Updating the versions of Metro in the cli-plugin-metro to incorporate the new Metro Release v0.77.0.
https://github.com/facebook/metro/releases/tag/v0.77.0

Per the release notes shared below, this is a breaking change for RN CLI including bumping minimum Node version to 18, using Terser's default for comments and removing `metro-minify-uglify` from `metro` dependencies

Metro release notes:

* **[Breaking]**: Increase minimum Node version from 16 to 18. (https://github.com/facebook/metro/commit/ebaf27ab49eb8b9394107bd7a796b24bc342c483 by @arushikesarwani94)
* **[Breaking]**: Use Terser's default for comments. (https://github.com/facebook/metro/pull/972 by @tido64)
* **[Breaking]**: Remove `metro-minify-uglify` from `metro` dependencies. (https://github.com/facebook/metro/commit/7a8fcd5ecf5ccbe760c22be7fa924f9cb7c2fe1e by @robhogan)
* **[Fix]**: Fixed a bug in the handling of path separators in `blockList` on Windows. (https://github.com/facebook/metro/commit/2eecfc767095f32bbea35a62c2dc85dcc9c4f839 by @motiz88)
* **[Fix]**: Protect against unsafe merging of `blockList` patterns with inconsistent regex flags. (https://github.com/facebook/metro/commit/c80904b64b8fa5d152da176ce668b77fa3e0ba25 by @motiz88)
* **[Fix]**: Incorrect worker resolution when multiple `metro` versions are installed. (https://github.com/facebook/metro/commit/6d46078e74ae9a43aa90bed46dbd6610e2696cd0 by @robhogan)
* **[Performance]**: Improve resolution performance for files outside the project root. (https://github.com/facebook/metro/commit/dc3cddf21803eb3addb084627e4d536318e3da0a by @robhogan)
* **[Performance]**: Improved startup speed via a new file map cache format. (https://github.com/facebook/metro/commit/166477e9154b3aec4d3081a759cc40f8e3d53c4a by @robhogan)

> NOTE: Experimental features are not covered by semver and can change at any time.

* **[Experimental]**: `metro-file-map`: Normalize root-relative paths for cross-platform cache compatibility. (https://github.com/facebook/metro/commit/d282a0818b840f1d5b0a2086051ec2c7febf1986 by @motiz88)
[...]

PR #2015, opened 2023-07-11
chore: Changing node version missed during metro dependencies upgrade

arushikesarwani94:
Summary:
---------

Updating the node version missed in https://github.com/react-native-community/cli/pull/2007/ as 18 is the Current version, will be LTS in October.

Test Plan:
----------

Tests pass

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,10 +12,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - name: Use Node.js 18
+      - name: Use Node.js 20
         uses: actions/setup-node@v3
         with:
-          node-version: 18
+          node-version: 20
           cache: yarn
       - name: Install dependencies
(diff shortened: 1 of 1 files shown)


thymikee (approved):
Thanks!

thymikee:
Install fails on node 20. I actually had the same locally and went back to 18 as I didn't have time to investigate

szymonrybczak:
I think this is problem related to `node-gyp`, as we're on old version -`node-gyp` is one of dependencies that comes with `lerna`. I bumped `leran` locally from `4.0.0` to `5.0.0` and the project is building for me correctly.

szymonrybczak:
@arushikesarwani94 mind rebasing on top of main?

arushikesarwani94:
> Hm, looks like CI is failing on creating project on node 20 on Ubuntu, interesting. 🤔

Not sure why this would just fail on Ubuntu for Node 20. 

cc: @thymikee Any idea what could be going on ?

PR #2016, opened 2023-07-12
chore: replace `glob` with `fast-glob`

szymonrybczak:
Summary:
---------

Replaced `glob` package with `fast-glob`, because:
- `fast-glob` is a lot faster than `glob`,
- `fast-glob` is few times lighter than `glob`.

TODO:
----------
- [ ] Make sure that these changes don't broke running CLI on Windows.

Test Plan:
----------

CI Green ✅

Checklist
----------

### DIFF
--- a/packages/cli-platform-android/src/config/findComponentDescriptors.ts
+++ b/packages/cli-platform-android/src/config/findComponentDescriptors.ts
@@ -1,13 +1,13 @@
 import fs from 'fs';
 import path from 'path';
-import glob from 'glob';
+import fg from 'fast-glob';
 import {extractComponentDescriptors} from './extractComponentDescriptors';
 
 export function findComponentDescriptors(packageRoot: string) {
-  const files = glob.sync('**/+(*.js|*.jsx|*.ts|*.tsx)', {
+  const files = fg.sync('**/+(*.js|*.jsx|*.ts|*.tsx)', {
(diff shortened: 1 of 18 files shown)


szymonrybczak:
I changed a lot of tests because when `fs.__setMockFilesystem({});` wasn't specifed in `beforeAll` function, but in every test suite - the tests were failing. It's very strange, but it is working correctly when mocking in `beforeAll`.

thymikee:
Do you have some measurements for speed improvement?

szymonrybczak:
> Do you have some measurements for speed improvement?

Before these changes running `getAndroidProject` took 1.462 second right now it's 0.819 🏎️

packages/cli-platform-android/src/config/findComponentDescriptors.ts
```diff
@@ -1,13 +1,13 @@
 import fs from 'fs';
 import path from 'path';
-import glob from 'glob';
+import fg from 'fast-glob';
```
thymikee, line 3: let's avoid abbreviations. go with `glob` or `fastGlob`. I'd pick the first one

thymikee:
`glob` added a few perf improvements over the last few versions. Maybe we should try upgrading it to latest first and later on migrate to fast-glob for package size optimizations

PR #2021, opened 2023-07-18
feat: add invoking CLI's scripts for launching Metro in `run-ios` command

szymonrybczak:
Summary:
---------
Recently @huntie [removed "Start packager" phase from template](https://github.com/facebook/react-native/commit/dc6845739e29ea8cf63583d530f67c2498286a3e), which means that right now when running `run-ios` command, packager won't be started. I synced implementations between platforms (`run-android` command was using CLI's script for some time). 
This change also fixes problem with starting Metro in monorepos setups (see #1799).

Test Plan:
----------
2. Run this command:

`run-ios` - Should start Metro from script located in `node_modules/.bin/launchPackager.command` ✅

TODO:
----------
- [ ] Windows support

Checklist
----------


packages/cli-platform-ios/src/commands/buildIOS/index.ts
```diff
@@ -95,6 +96,14 @@ async function buildIOS(_: Array<string>, ctx: Config, args: FlagsT) {
     } "${chalk.bold(xcodeProject.name)}"`,
   );
 
+  await runPackager(
```
adamTrz, line 99: Do we have to run packager to just do build? 🤔

packages/cli-platform-ios/src/commands/runIOS/index.ts
```diff
@@ -117,6 +118,14 @@ async function runIOS(_: Array<string>, ctx: Config, args: FlagsT) {
     } "${chalk.bold(xcodeProject.name)}"`,
   );
 
+  await runPackager(
```
tido64, line 121: We should be able to disable this with `--no-packager`.

Unrelated: I was actually hoping we could remove the packager from the `run-android` command instead :P

packages/cli-plugin-metro/src/commands/start/runPackager.ts
```diff
@@ -0,0 +1,33 @@
...
+  if (!packager) {
+    return;
+  }
+  const result = await isPackagerRunning(port);
+  if (result === 'running') {
+    logger.info('JS server already running.');
+  } else if (result === 'unrecognized') {
+    logger.warn('JS server not recognized, continuing with build...');
+  } else {
+    // result == 'not_running'
```
tido64, line 20: Should we explicitly compare against `'not_running'`? What other results are possible?

packages/cli-platform-ios/src/commands/runIOS/index.ts
szymonrybczak: > We should be able to disable this with --no-packager.

We're already able to this :) See here:
https://github.com/react-native-community/cli/blob/ba362e65e2a7feebf49a2b392f9900ca3260efdf/packages/cli-platform-ios/src/commands/runIOS/index.ts#L609-L612

And in function body I check for this:
https://github.com/react-native-community/cli/blob/b3741451e5b5c757aeaccc2fb66edaa46faa3d72/packages/cli-plugin-metro/src/commands/start/runPackager.ts#L11-L13

packages/cli-platform-ios/README.md
```diff
@@ -171,10 +171,6 @@ Explicitly set device to use by name. The value is not required if you have a si
 
 Explicitly set device to use by udid.
 
-#### `--no-packager`
```
thymikee, line 174: we removed the packager flag, but left the port which is also tied with Metro

packages/cli-plugin-metro/src/commands/start/runPackager.ts
tido64: We should also log unhandled results in case there's something that needs to be explicitly handled later.

packages/cli-platform-ios/src/commands/runIOS/index.ts
thymikee: Question is, maybe we should consider disabling running packager by default? I think this would require a poll on how people use it in the wild. I myself run the server myself in a terminal tab that suits me. But I've also seen devs relying on automatic opening of dev server and not sure if they would even realize it's missing

packages/cli-platform-ios/src/commands/buildIOS/buildProject.ts
```diff
@@ -14,12 +11,11 @@ import {
 export type BuildFlags = {
   mode: string;
   target: string;
-  packager: boolean;
   verbose: boolean;
   xcconfig?: string;
   buildFolder?: string;
   port: number;
```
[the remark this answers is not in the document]
liamjones: Doesn't port end up inside `RCT_METRO_PORT` at compile time for iOS (so may still be needed)? https://github.com/facebook/react-native/blob/080a3d300dc357fec62d767d140f964f9db66b2e/packages/react-native/React/Base/RCTDefines.h#L130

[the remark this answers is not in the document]
thymikee: It's still going to be set for `run-ios`. For `build-ios` we don't expect the Metro server to run because the JS bundle will be, well, bundled into the app executable

packages/cli-platform-ios/src/commands/runIOS/index.ts
szymonrybczak: Other options that comes to my mind is to run the packager in the same window where we start `run-ios/android` command.

packages/cli-platform-ios/src/commands/buildIOS/buildProject.ts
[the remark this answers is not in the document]
szymonrybczak: Yup, it doesn't make sense to pass these parameters to `build-ios` command as it doesn't use it.

huntie:
✅ Endorsed!

packages/cli-plugin-metro/src/commands/start/runPackager.ts
```diff
@@ -0,0 +1,30 @@
+import {isPackagerRunning, logger} from '@react-native-community/cli-tools';
+import {startServerInNewWindow} from './startServerInNewWindow';
+
+export async function runPackager(
```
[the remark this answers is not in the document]
szymonrybczak: > Can startServerInNewWindow be removed from src/index.js now that it's no longer referenced by the Android and iOS plugins?

Yes, it is still used in `cli-doctor` package, but  I think we can replace this usage with `runPackager` and use it through exposed command. 

> Can we instead expose runPackager as a command, e.g. react-native start-for-build-hook? This will enable the other plugins to invoke this handling without a direct code dependency.

Done ✅

> Can you check for any other existing imports from cli-plugin-metro? Ideally we eliminate/replace all of them 🙏🏻.

After latest commit, there are no imports from `cli-plugin-metro` 🎉 (there's only one that imports commands from package)

packages/cli-plugin-metro/src/commands/runPackager/index.ts
```diff
@@ -0,0 +1,18 @@
+import {runPackager} from './runPackager';
+
+export default {
+  name: 'run-packager-hook',
+  description: 'Starts the packager',
```
szymonrybczak, line 5: For sure this should be named in better way, do you have any ideas how we could call this?

packages/cli-doctor/src/tools/healthchecks/packager.ts
```diff
@@ -4,7 +4,8 @@ import {
 } from '@react-native-community/cli-tools';
 import {HealthCheckInterface} from '../../types';
 import {logManualInstallation} from './common';
-import {startServerInNewWindow} from '@react-native-community/cli-plugin-metro';
```
huntie, line 7: Does this enable removal of `@react-native-community/cli-plugin-metro` from `package.json` (and other packages touched)? :)

packages/cli-plugin-metro/src/commands/start/startServerInNewWindow.ts
```diff
@@ -9,10 +9,16 @@ import {
 
 export function startServerInNewWindow(
```
huntie, line 10: Let's relocate this file within `src/commands/runPackagerHook/`.

packages/cli-plugin-metro/src/commands/runPackager/index.ts
```diff
@@ -0,0 +1,18 @@
...
+  func: runPackager,
+  options: [
+    {
+      name: '--port <number>',
+      description: 'Port to run packager on.',
+    },
+    {
+      name: '--terminal <string>',
+      description:
+        'Launches packager in a new window using the specified terminal path.',
```
huntie, line 15: **Main feedback ⬇️**

Interesting, can we merge this all into the `start` command — aligning with your proposal in https://github.com/react-native-community/discussions-and-proposals/pull/613? Specifically:

- The improved error message behaviour when `start` (RN CLI `start`) is called and something is already on the configured Metro port.
- `react-native start --terminal` switches modes to integrate the `startServerInNewWindow.js` behaviour.

Then, we only expose `startCommand` and it can be used by `doctor`, `run-android`, `run-ios`.

cc @thymikee Does this seem feasible?

huntie (changes requested):

packages/cli-plugin-metro/src/commands/runPackager/index.ts
tido64: `ensure-dev-server` / `ensure-packager`?

szymonrybczak: Thanks @tido64 for proposition! We're mergining this logic to one command - `start`, so that everything will be in one place (see [here](https://github.com/react-native-community/cli/pull/2021#discussion_r1277710939)), and I will remove this command soon as it wouldn't make sense to keep it.

szymonrybczak:
👋, update from my side. I implemented whole feedback, and I added features discussed under this [RFC](https://github.com/react-native-community/discussions-and-proposals/pull/613).  I changed implementation of handling keystrokes. Before it wasn't possible to pause interactivity (to show the prompt), but right now it was possible. Without this change, after showing prompt in watch mode the user wouldn't be able to handle any keystroke.

---
Rigth now flow looks like now: 

- `run-android` starts bundler via CLI (`node cli/path/ start` command)
- `run-ios` starts bundler via CLI (`node /cli/path/ start` command)

> [!NOTE]
> When starting Android or iOS app from watchMode it runs also `run-android` and `run-ios` command so it would run again `start` command, but right now CLI is checking against port is busy or not so user won't see nothing.

Added interactive part 👀
- When starting bundler _and port is busy_ CLI will ask whether to terminate process that is running on port user requested. If user wants to kill process, CLI will kill process and starts CLI at request port.

https://github.com/react-native-community/cli/assets/63900941/529761f1-380a-4957-a3d8-68e124af2ad0

- When user don't want to kill proces, CLI will find the next free port and will ask user whether to start bundler at found port.

https://github.com/react-native-community/cli/assets/63900941/12e32911-7e45-4957-9837-aaca721469c3

Also added log to inform user about on which port Metro bundler is running:
CleanShot 2023-08-01 at 13 47 19@2x

packages/cli-platform-ios/src/commands/runIOS/index.ts
```diff
@@ -610,6 +630,21 @@ export default {
...
+    },
+    {
+      name: '--terminal <string>',
+      description:
+        'Launches the Metro Bundler in a new window using the specified terminal path.',
+      default: getDefaultUserTerminal(),
+    },
+    {
+      name: '--no-packager',
+      description: 'Do not launch packager while building',
```
TMisiukiewicz, line 646: since `--no-packager` was moved from `build` commands, it should have appropriate description (both iOS and Android)

packages/cli-plugin-metro/src/commands/start/runServer.ts
```diff
@@ -35,11 +46,70 @@ export type Args = {
...
+  const packagerStatus = await isPackagerRunning(port);
+
+  const handleSomethingRunningOnPort = async () => {
+    const {change: kill} = await askForProcessKill(port);
+    if (kill) {
+      const pid = await getProcessIdFromPort(port);
+
+      if (pid) {
+        execa.sync('kill', [pid]);
+      }
```
TMisiukiewicz, line 81: i think both `getProcessIdFromPort` and the following if statement should also cover running server on Windows

packages/cli-tools/src/getPidFromPort.ts
```diff
@@ -0,0 +1,22 @@
...
+      .split('\n')[0]
+      .trim();
+    const pid = Number(results);
+    logger.debug(`Pid: ${pid} for port: ${port}`);
+    return pid;
+  } catch (error) {
+    logger.debug(`No pid found for port: ${port}. Error: ${error}`);
+    return null;
+  }
+};
```
TMisiukiewicz, line 20: this also needs to cover behavior on Windows

packages/cli-plugin-metro/src/commands/start/runServer.ts
szymonrybczak: correct! ✅

packages/cli-tools/src/getPidFromPort.ts
szymonrybczak: correct!

packages/cli-plugin-metro/src/commands/start/index.ts
```diff
@@ -88,6 +88,10 @@ export default {
       name: '--no-interactive',
       description: 'Disables interactive mode',
     },
+    {
+      name: '--terminal',
+      description: 'Specify terminal to use for running Metro',
```
huntie, line 93: ~~Can we update this description like so,~~ (**update**: see top-level feedback, let's relocate this) and demonstrate the bail-on-port-exists behaviour in the test plan?

```suggestion
      description: 'When this option is provided, starts the dev server in a separate terminal window using the provided terminal app path. In this mode, the command will immediately bail if another process is running on the same port.',
```

packages/cli-plugin-metro/src/commands/start/watchMode.ts
```diff
@@ -37,38 +42,46 @@ function enableWatchMode(messageSocket: any, ctx: Config) {
...
+      case 'r':
+        messageSocket.broadcast('reload', null);
+        logger.info('Reloading app...');
+        break;
+      case 'd':
+        messageSocket.broadcast('devMenu', null);
+        logger.info('Opening developer menu...');
+        break;
+      case 'i':
+        logger.info('Opening the app on iOS...');
```
huntie, line 63: nit
```suggestion
        logger.info('Opening app on iOS...');
```

packages/cli-plugin-metro/src/commands/start/watchMode.ts
```diff
@@ -37,38 +42,46 @@ function enableWatchMode(messageSocket: any, ctx: Config) {
...
+      case CTRL_Z:
+        process.emit('SIGTSTP', 'SIGTSTP');
+        break;
+      case 'r':
+        messageSocket.broadcast('reload', null);
+        logger.info('Reloading app...');
+        break;
+      case 'd':
+        messageSocket.broadcast('devMenu', null);
+        logger.info('Opening developer menu...');
```
huntie, line 60: nit: ["Dev Menu"](https://reactnative.dev/docs/next/debugging#accessing-the-dev-menu)
```suggestion
        logger.info('Opening Dev Menu...');
```

packages/cli-plugin-metro/src/commands/start/watchMode.ts
```diff
@@ -37,38 +42,46 @@ function enableWatchMode(messageSocket: any, ctx: Config) {
...
+      case 'i':
+        logger.info('Opening the app on iOS...');
+        execa('npx', [
+          'react-native',
+          'run-ios',
+          ...(ctx.project.ios?.watchModeCommandParams ?? []),
+        ]).stdout?.pipe(process.stdout);
+        break;
+      case 'a':
+        logger.info('Opening the app on Android...');
```
huntie, line 71: nit
```suggestion
        logger.info('Opening app on Android...');
```

packages/cli-plugin-metro/src/commands/start/watchMode.ts
```diff
@@ -37,38 +42,46 @@ function enableWatchMode(messageSocket: any, ctx: Config) {
...
-      execa('npx', [
-        'react-native',
-        name === 'i' ? 'run-ios' : 'run-android',
-        ...(params ?? []),
-      ]).stdout?.pipe(process.stdout);
-    } else {
-      console.log(_key);
+  const onPressAsync = async (key: string) => {
+    if (key === CTRL_C) {
+      process.exit();
```
huntie, line 47: nit: This can be moved into the `switch` statement.

packages/cli-plugin-metro/src/commands/start/runServer.ts
```diff
@@ -123,6 +193,7 @@ async function runServer(_argv: Array<string>, ctx: Config, args: Args) {
   serverInstance.keepAliveTimeout = 30000;
 
   await version.logIfUpdateAvailable(ctx.root);
+  logger.info(`Started Metro Bundler at ${chalk.bold(port)} port`);
```
huntie, line 196: nit
```suggestion
  logger.info(`Started dev server on port ${chalk.bold(port)}`);
```

packages/cli-plugin-metro/src/tools/getProcessIdFromPort.ts
```diff
@@ -0,0 +1,28 @@
+import {execFileSync} from 'child_process';
+
+/**
+ *
+ * @param port The port to find the process id for
+ * @returns The process id for the port or null if not found
+ */
+
```
huntie, line 8: nit
```suggestion
 * @param port The port to find the process id for
 * @returns The process id for the port or null if not found
 */
```

packages/cli-plugin-metro/src/tools/getNextPort.ts
```diff
@@ -0,0 +1,35 @@
+/**
+ * Increases by one the port number until it finds an available port.
+ * @param port Port number to start with.
+ * @param root Root of the project.
+ */
```
huntie, line 5: nit: Move doc comment above function on line 14.

huntie (changes requested):
This is awesome work 💯

**High-level feedback**

- 1/ We can probably achieve a better organisation of responsibilities here.
  -  Maybe `--terminal` on the `start` command is the wrong place to do full-on process management (my bad). It would be really good to keep `cli-plugin-metro`'s responsibilities as clean as possible.
  - Therefore:
     - `cli-plugin-metro`: Let's reduce the `start` command to starting the server only, and erroring generically to say "Could not start dev server: The specified port <port> was taken".
     - `cli-plugin-android/ios` Let's do process detection **here**, and move the background terminal creation logic here (likely shared via `cli-tools`) also. Therefore it's these commands which will wrap `start`, which doesn't need to know where it's running.
- 2/ Being able to kill another process is probably not surface area we want to add to CLI (being mindful of when we introduce user expectations). For the `run-android`/`run-ios` commands, bailing from starting Metro when the port is taken (with an informational log message) is most likely adequate.
- 3/ I'd suggest separating the improved `KeyPressHandler` logic into a new PR — potentially this should live in `cli-tools`.

huntie:
@szymonrybczak 

> For this we also will need to move relevant scripts, as it wouldn't make sense for them to live in cli-plugin-metro and to access this from cli-tools, and I'm not sure if this would be doable if we're migrating this part of codebase to Core.

Sure, let's move as much logic as necessary into `cli-tools`, where necessary to support `run-android`/`run-ios` and as not needed by `cli-plugin-metro`.

> is `cli-tools` actually good place to keep `launchPackager.command`, `launchPackager.bat`?

I think yes for now :)

packages/cli-tools/src/launchPackager.bat
huntie, line 1: - Since these aren't _source code_ (or sort of are) shall we locate these scripts one folder above?
- `package.json` `"files"` entries need to be aligned for these (moved from `cli-plugin-metro` to `cli-tools`).

packages/cli-plugin-metro/src/commands/start/runServer.ts
```diff
@@ -40,6 +49,39 @@ export type Args = {
 };
 
 async function runServer(_argv: Array<string>, ctx: Config, args: Args) {
+  let port = args.port ?? 8081;
+  const packagerStatus = await isPackagerRunning(port);
+
+  const handleSomethingRunningOnPort = async () => {
```
huntie, line 55: nit: Simplify naming.
```suggestion
  const handlePortUnavailable = async () => {
```

huntie (approved):
Looks good to me. Might be some tidying we can do in future, but I'm keen to unblock the move of `cli-plugin-metro`. Huge thanks for this! 🙌🏻

packages/cli-tools/src/launchPackager.bat
szymonrybczak: > * Since these aren't _source code_ (or sort of are) shall we locate these scripts one folder above?

No because when script is building CLI it is looking inside `src` folder, so if these script would be in root directory of package they wouldn't be moved to `build` folder.

packages/cli-tools/src/startServerInNewWindow.ts
```diff
@@ -69,8 +73,9 @@ export function startServerInNewWindow(
       );
     }
   } catch (error) {
+    console.log('error', error);
```
huntie, line 76: Checking this is wanted ❗

packages/cli-plugin-metro/src/commands/start/watchMode.ts
```diff
@@ -3,6 +3,11 @@ import {logger, hookStdout} from '@react-native-community/cli-tools';
 import execa from 'execa';
 import chalk from 'chalk';
 import {Config} from '@react-native-community/cli-types';
+import {KeyPressHandler} from '../../tools/KeyPressHandler';
+import {addInteractionListener} from '@react-native-community/cli-tools';
```
huntie, line 7: nit: Merge with existing `'@react-native-community/cli-tools'` import.

packages/cli-tools/src/startServerInNewWindow.ts
szymonrybczak: Ah, of course not 😅

packages/cli-plugin-metro/src/commands/start/watchMode.ts
szymonrybczak: yeah, I will do it after https://github.com/react-native-community/cli/pull/2041 will be merged 👍

packages/cli-doctor/package.json
huntie, line 1: Please remove the `@react-native-community/cli-plugin-metro` dep in this package too :)

huntie (approved):

packages/cli-doctor/package.json
szymonrybczak: done ✅

packages/cli-platform-ios/src/commands/runIOS/index.ts
huntie: > I think this would require a poll on how people use it in the wild

FYI I'm up for this, and/or making the most informed decision. Originally I was with @tido64 in thinking we'd drop the behaviour on Android to align. But @szymonrybczak's changes, with an opt-out flag, look pretty useful to me.

While this item of feedback is still unresolved, I'd like to push for this PR to be merged and released anyway, since it is a direct dependency for https://github.com/facebook/react-native/pull/38944 and incoming Debugging functionality in React Native. These are all alpha releases, and we can follow up on enhancing this behaviour in a follow-up PR.|

cc @thymikee

packages/cli-platform-android/src/commands/runAndroid/index.ts
```diff
@@ -42,6 +56,50 @@ export type AndroidProject = NonNullable<Config['project']['android']>;
 async function runAndroid(_argv: Array<string>, config: Config, args: Flags) {
   link.setPlatform('android');
 
+  let {packager, port} = args;
+
+  const packagerStatus = await isPackagerRunning(port);
+
+  const handlePortUnavailable = async () => {
```
thymikee, line 63: `handlePortUnavailable` seems to be duplicated with minimal changes across 3 packages. Let's see if we can refactor this

thymikee (changes requested):
I think we're one small refactor away from merging this. Nice work @szymonrybczak and @huntie!

packages/cli-platform-android/src/commands/runAndroid/index.ts
szymonrybczak: done 👍

thymikee (approved):

packages/cli-tools/src/startServerInNewWindow.ts
```diff
@@ -1,18 +1,22 @@
...
 
-export function startServerInNewWindow(
+function startServerInNewWindow(
   port: number,
-  terminal: string,
   projectRoot: string,
   reactNativePath: string,
+  terminal?: string,
 ) {
+  if (!terminal) {
```
aajahid, line 14: This check is preventing android build on windows. On windows - so far we did not need any `terminal` value. It always used to fall back to default `cmd.exe` ( line no 106 )

In this case I think the proper fix should be fixing the `getDefaultUserTerminal` which currently doesn't have any default terminal for windows.

aajahid (commented):
Bug - #2219

packages/cli-tools/src/startServerInNewWindow.ts
aajahid: Or only apply this check on platforms other than windows.

packages/cli-platform-ios/src/commands/runIOS/index.ts
```diff
@@ -32,11 +41,43 @@ export interface FlagsT extends BuildFlags {
...
+      const result = await handlePortUnavailable(port, ctx.root, packager);
+      [port, packager] = [result.port, result.packager];
+    }
+  } else if (packagerStatus === 'unrecognized') {
+    const result = await handlePortUnavailable(port, ctx.root, packager);
+    [port, packager] = [result.port, result.packager];
+  }
+
+  if (packager) {
+    await startServerInNewWindow(
```
aajahid, line 73: await causes the CLI to hang indefinitely. Must be called without await. 
Check existing comment at line no 105

aajahid (commented):
Reason of another issue 

> This works for me. However it only opens a new window and doesn't automatically run andoid, it acts as if you've run react-native start only

#2219

packages/cli-platform-android/src/commands/runAndroid/index.ts
```diff
@@ -42,6 +54,35 @@ export type AndroidProject = NonNullable<Config['project']['android']>;
...
+      const result = await handlePortUnavailable(port, config.root, packager);
+      [port, packager] = [result.port, result.packager];
+    }
+  } else if (packagerStatus === 'unrecognized') {
+    const result = await handlePortUnavailable(port, config.root, packager);
+    [port, packager] = [result.port, result.packager];
+  }
+
+  if (packager) {
+    await startServerInNewWindow(
```
aajahid, line 78: await causes the CLI to hang indefinitely. Must be called without await.
Check existing comment at line no 105

PR #2041, opened 2023-08-01
feat: enable pausing handling keystrokes in watch mode

szymonrybczak:
Summary:
---------

In this PR I created `KeyPressHandler` that allows us to pause handling keystrokes because in nearest future we will add interactive prompts to `start` command, so they will run in watch mode. If we do not pause the handling keystrokes, the "focus" will be on the prompt. I created a function that wraps `prompts` and pauses and resumes interactivity in watchMode. 

Test Plan:
----------

2. Run this command:

```sh
node /path/to/react-native-cli/packages/cli/build/bin.js start
```
3. Press key that we're handling (`a`,` i`, `d`, `r`) and check if keystrokes are properly handled. 
 

Checklist
----------

### DIFF
--- a/packages/cli-plugin-metro/src/commands/start/watchMode.ts
+++ b/packages/cli-plugin-metro/src/commands/start/watchMode.ts
@@ -4,7 +4,6 @@ import execa from 'execa';
 import chalk from 'chalk';
 import {Config} from '@react-native-community/cli-types';
 import {KeyPressHandler} from '../../tools/KeyPressHandler';
-import {addInteractionListener} from '@react-native-community/cli-tools';
 
 const CTRL_C = '\u0003';
 const CTRL_Z = '\u0026';
@@ -42,7 +41,7 @@ function enableWatchMode(messageSocket: any, ctx: Config) {
     }
   });
 
-  const onPressAsync = async (key: string) => {
+  const onPress = (key: string) => {
     switch (key) {
       case 'r':
         messageSocket.broadcast('reload', null);
@@ -76,9 +75,8 @@ function enableWatchMode(messageSocket: any, ctx: Config) {
     }
   };
 
-  const keyPressHandler = new KeyPressHandler(onPressAsync);
-  const listener = keyPressHandler.createInteractionListener();
-  addInteractionListener(listener);
+  const keyPressHandler = new KeyPressHandler(onPress);
+  keyPressHandler.createInteractionListener();
   keyPressHandler.startInterceptingKeyStrokes();
 }
 
--- a/packages/cli-plugin-metro/src/commands/start/watchMode.ts
+++ b/packages/cli-plugin-metro/src/commands/start/watchMode.ts
@@ -3,6 +3,11 @@ import {logger, hookStdout} from '@react-native-community/cli-tools';
 import execa from 'execa';
 import chalk from 'chalk';
 import {Config} from '@react-native-community/cli-types';
+import {KeyPressHandler} from '../../tools/KeyPressHandler';
+import {addInteractionListener} from '@react-native-community/cli-tools';
+
+const CTRL_C = '\u0003';
+const CTRL_Z = '\u0026';
 
 function printWatchModeInstructions() {
   logger.log(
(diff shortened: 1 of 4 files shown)


packages/cli-plugin-metro/src/tools/KeyPressHandler.ts
```diff
@@ -0,0 +1,73 @@
+import {CLIError, logger} from '@react-native-community/cli-tools';
+
+const CTRL_C = '\u0003';
+
+/** An abstract key stroke interceptor. */
+export class KeyPressHandler {
+  private isInterceptingKeyStrokes = false;
+  private isHandlingKeyPress = false;
+
+  constructor(public onPress: (key: string) => Promise<any>) {}
```
thymikee, line 10: I think we're don't intend to return anything from this fn and if it must be either sync or async, let's state that in the type
```suggestion
  constructor(public onPress: (key: string) => Promise<void> | void) {}
```

packages/cli-plugin-metro/src/commands/start/watchMode.ts
```diff
@@ -37,38 +42,44 @@ function enableWatchMode(messageSocket: any, ctx: Config) {
...
-          ? ctx.project.ios?.watchModeCommandParams
-          : ctx.project.android?.watchModeCommandParams;
-      execa('npx', [
-        'react-native',
-        name === 'i' ? 'run-ios' : 'run-android',
-        ...(params ?? []),
-      ]).stdout?.pipe(process.stdout);
-    } else {
-      console.log(_key);
+  const onPressAsync = async (key: string) => {
```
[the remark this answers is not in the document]
szymonrybczak: we're awaiting it here: 
https://github.com/react-native-community/cli/blob/e8f7148c5d6ca71acf65b62848f05be5321f0d88/packages/cli-plugin-metro/src/tools/KeyPressHandler.ts#L40

packages/cli-plugin-metro/src/tools/KeyPressHandler.ts
```diff
@@ -0,0 +1,73 @@
...
+    // Prevent sending another event until the previous event has finished.
+    if (this.isHandlingKeyPress && key !== CTRL_C) {
+      return;
+    }
+    this.isHandlingKeyPress = true;
+    try {
+      logger.debug(`Key pressed: ${key}`);
+      await this.onPress(key);
+    } catch (error) {
+      return new CLIError('There was an error with the key press handler.');
```
thymikee, line 42: we should pass the `error` to CLIError to get user some context. This is execa that can fail here right? Let's make sure the output is readable

packages/cli-plugin-metro/src/commands/start/watchMode.ts
```diff
@@ -37,38 +42,44 @@ function enableWatchMode(messageSocket: any, ctx: Config) {
...
+        break;
+      case CTRL_C:
+        process.exit();
     }
-  });
+  };
+
+  const keyPressHandler = new KeyPressHandler(onPressAsync);
+  const listener = keyPressHandler.createInteractionListener();
+  addInteractionListener(listener);
```
thymikee, line 81: would it be a bad idea to move this call to `createInteractionListener()`?

packages/cli-plugin-metro/src/commands/start/watchMode.ts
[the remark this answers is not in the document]
thymikee: yup, however there's no async code inside this fn so there's no point in doing that now. if the handler body becomes async, we could extend the listener type then

packages/cli-plugin-metro/src/tools/KeyPressHandler.ts
```diff
@@ -0,0 +1,73 @@
...
+
+  private handleKeypress = async (key: string) => {
+    // Prevent sending another event until the previous event has finished.
+    if (this.isHandlingKeyPress && key !== CTRL_C) {
+      return;
+    }
+    this.isHandlingKeyPress = true;
+    try {
+      logger.debug(`Key pressed: ${key}`);
+      await this.onPress(key);
```
thymikee, line 40: when this gets sync (which it is right now in fact), `this.isHandlingKeyPress` gets unnecessary

thymikee (changes requested):
small nits and this is good to go

packages/cli-plugin-metro/src/commands/start/watchMode.ts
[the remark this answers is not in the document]
szymonrybczak: ah, right

thymikee (approved):

PR #2047, opened 2023-08-07
chore: Upgrade Metro to 0.78.0

motiz88:
Upgrades Metro to 0.78.0, removes the unreferenced dependencies on `metro-react-native-babel-transformer` and `metro-runtime`, and demotes `metro-resolver` to a `devDependency` (as it is used only for types).

Metro 0.78.0 contains the following breaking changes that we **do not believe are breaking** for CLI users:

* **[Breaking]** Update default server port to 8081. (https://github.com/facebook/metro/commit/891fe9674cb6bc63aea4cea365f78fa2584ab138 by @huntie)
* **[Breaking]** Remove `metro-react-native-babel-transformer` and `metro-react-native-babel-preset`, which will be published as `@react-native/metro-babel-transformer` and `@react-native/babel-preset` going forward. (https://github.com/facebook/metro/commit/799aa0a546e39546e2d0463f3fb1a963dc102bdf by @dmytrorykun)
* **[Breaking]** Remove `runInspectorProxy` option from server. (https://github.com/facebook/metro/commit/2b0b15d733c47768304569c800b3aceca3ef5b1f by @huntie)
* **[Breaking]**: `DependencyGraph.resolveDependency` now takes a dependency object instead of a string. (https://github.com/facebook/metro/commit/fbdd295c5820152487f3331852ce91824808ce81 by @motiz88)
* **[Breaking]**: Stop passing `inlineRequires` and `nonInlinedRequires` to Babel transformers. (https://github.com/facebook/metro/commit/e339c8fae4c2e925aedbd3755e18a4a780bd0fb7 by @motiz88)
  * Note that this merely removes a confusing, dead API from Metro. It does *not* functionally change how [inline requires](https://reactnative.dev/docs/ram-bundles-inline-requires#inline-requires) work in React Native.

See https://github.com/facebook/metro/releases/tag/v0.78.0 for the full changelog.

As a reminder, when updating the CLI version in React Native, `metro-runtime` and other Metro dependencies declared directly by React Native must be updated and pinned to the exact version used in the CLI (e.g. here `"0.78.0"`).

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,7 +51,7 @@
     "jest-snapshot-serializer-raw": "^1.1.0",
     "lerna": "^4.0.0",
     "lint-staged": "^13.1.0",
-    "metro-memory-fs": "0.77.0",
+    "metro-memory-fs": "0.78.0",
     "micromatch": "^4.0.4",
     "rimraf": "^3.0.2",
     "slash": "^3.0.0",
--- a/packages/cli-plugin-metro/package.json
+++ b/packages/cli-plugin-metro/package.json
@@ -11,16 +11,14 @@
     "@react-native-community/cli-tools": "12.0.0-alpha.7",
     "chalk": "^4.1.2",
     "execa": "^5.0.0",
-    "metro": "0.77.0",
-    "metro-config": "0.77.0",
-    "metro-core": "0.77.0",
-    "metro-react-native-babel-transformer": "0.77.0",
-    "metro-resolver": "0.77.0",
-    "metro-runtime": "0.77.0",
+    "metro": "0.78.0",
+    "metro-config": "0.78.0",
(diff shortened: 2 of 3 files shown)


szymonrybczak:
Since the default port is `8081`, do we want to delete adding default port on CLI side?

huntie:
@szymonrybczak Yes, we can :). I haven't tackled this personally yet as there is the `.packager.env`/`RCT_METRO_PORT` setup which people may rely on (note: this may be slightly simplified since https://github.com/facebook/react-native/pull/38242). But we can certainly remove any internally set defaults from CLI 👍🏻. Non-urgent.

szymonrybczak (approved):

PR #2052, opened 2023-08-08
fix: Fix watchMode process logs not showing color

SubsequentlySneeds:
Summary:
---------

When selecting "run on Android" from the Metro menu, the first info label is bolded and cyan, but the second one prints without color or bolding:

Code_ojM1HKCg96
(VSCode terminal)

This is because the child process spawns without color by default, even if the parent has color enabled. This causes chalk to print without color escape sequences to the child stdout, which is then piped back to the parent stdout without color, even though it is capable of displaying it.

This change simply adds an env variable to force color in the child process if the parent supports color.

Test Plan:
----------

Pasted the code into the transpiled watchMode.js file (replacing "chalk" with "_chalk().default" for ES Module interop) in the React Native project the issue was first observed in and re-ran the same operation:

Code_hGwZ4IvBmN

Checklist
----------


thymikee (approved):
Nice find, thank you!

thymikee:
We've made some changes to the watchMode, @SubsequentlySneeds would you mind rebasing? 🙏🏼

SubsequentlySneeds:
Rebased to the latest watchMode.ts

adamTrz (approved):

thymikee:
cc @huntie – as we're freezing `cli-plugin-metro`, could you import this change to [community-cli-plugin](https://github.com/facebook/react-native/tree/main/packages/community-cli-plugin)?

@szymonrybczak would like to test it on Windows as well

szymonrybczak:
Closing this PR, as the code is not longer in this repository - we migrated that part of codebase to React Native repo, and as I can see @huntie [already added](https://github.com/facebook/react-native/pull/39074) this change. Thanks for contribution!

PR #2056, opened 2023-08-14
fix: change request response in `/status` endpoint

szymonrybczak:
Summary:
---------

Recently in https://github.com/react-native-community/cli/pull/2021 I changed the response in `/status` endpoint. It worked out that it was breaking change, and React Native depends internally on this response. In this PR I moved passing this information to header, and that's perfect for us because we don't create breaking change and we can make the same logic as before. 

Test Plan:
----------

Everything described in #2021 should work the same after changes in this PR.

Checklist
----------

### DIFF
--- a/packages/cli-server-api/src/__tests__/statusPageMiddleware.test.ts
+++ b/packages/cli-server-api/src/__tests__/statusPageMiddleware.test.ts
@@ -1,0 +1,20 @@
+import http from 'http';
+import statusPageMiddleware from './../statusPageMiddleware';
+
+describe('statusPageMiddleware', () => {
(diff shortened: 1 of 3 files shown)


szymonrybczak:
cc. @huntie @motiz88

huntie (approved):
LGTM 👍🏻

Would be great to have a unit test against this endpoint, which includes a description/comment documenting that the string response value is widely depended on and should not be modified.

szymonrybczak:
> Would be great to have a unit test against this endpoint, which includes a description/comment documenting that the string response value is widely depended on and should not be modified.

Tests added ✅

PR #2066, opened 2023-09-08
breaking: Remove unreferenced utils from cli-tools

huntie:
## Summary

Removes the `getPidFromPort` and `hookStdout` exports from `cli-tools`. These are no longer referenced in `community-cli-plugin` (formerly `cli-plugin-metro`) since https://github.com/facebook/react-native/pull/39078.

## Test plan

```
tsc
```


szymonrybczak (approved):
Alright, thanks for cleanup! 👍

Issue #2063, opened 2023-09-07
Creating a project without a project name is a bad experience


TheSavior:
![373397563_828752802160927_5015988033141924743_n](https://github.com/react-native-community/cli/assets/249164/7538b126-c4c2-4206-bbe1-d4ee2b931dac)

This experience isn't great, can we improve it? I think the CLI should give a more user friendly error if init is called incorrectly, ideally by printing out an example command showing how to call it.

PR #2071, opened 2023-09-11
fix: improve providing project name in `init` command

szymonrybczak:
Summary:
---------

Closes https://github.com/react-native-community/cli/issues/2063

Test Plan:
----------

2. Run this command:

```sh
```

Should prompt user for the project name. 

 

Checklist
----------

### DIFF
--- a/packages/cli/src/commands/init/init.ts
+++ b/packages/cli/src/commands/init/init.ts
@@ -279,7 +279,7 @@ export default (async function initialize(
     const {projName} = await prompts({
       type: 'text',
       name: 'projName',
-      message: 'What is the project name?',
+      message: 'How would you like to name the app?',
     });
     projectName = projName;
   }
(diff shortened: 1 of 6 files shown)


packages/cli/src/commands/init/init.ts
```diff
@@ -213,6 +214,15 @@ export default (async function initialize(
   [projectName]: Array<string>,
   options: Options,
 ) {
+  if (!projectName) {
+    const {projName} = await prompts({
+      type: 'text',
+      name: 'projName',
+      message: 'What is the project name?',
```
thymikee, line 221: Would that wording make sense?
```suggestion
      message: 'How would you like to name the app?',
```
Also, how about adding a `hint` guiding the user on what is about to happen? Something like:

> New app will be initialized in the directory of the same name. Android and iOS projects will use this name for publishing setup

If this wording works, we can also reflect it in the command description

szymonrybczak: > Also, how about adding a hint guiding the user on what is about to happen? Something like:

Fun fact: when using `type: 'text'` prompt, we cannot show the hint 🙈

> If this wording works, we can also reflect it in the command description

Done ✅

thymikee (approved):

PR #2073, opened 2023-09-12
feat: add Bun to `init` command

TMisiukiewicz:
Summary:
---------
Adding support for Bun package manager with `init` command.
With template and dependencies installed via bun package manager, the JS code execution time went from ~41s to ~2,5s.

In order to make package managers easier to use, this PR deprecates `--npm` flag and introduces `--pm` flag where it's possible to use one of the packagers: `yarn`, `npm` or `bun`.

Test Plan:
----------
Follow the contributing guide
1. Run `init` command as usual. It should install template and dependencies using yarn and you should have `yarn.lock` file in created project.
2. Run `init` command with `--pm yarn` flag.  Once it's done, you should have `yarn.lock` file in created project.
3. Run `init` command with `--pm npm` flag. Once it's done, you should have `package-lock.json` file in created project.
4. Run `init` command with `--pm bun` flag. Once it's done, you should have `bun.lockb` file in created project. 
5. Run `init` command with `--npm` flag. It should show a deprecation warning in the console.
6. Run `bun run ./path-to-cli init ProjectName`. It should use `bun` to install dependencies and template without any additional flag passed. Should also work with `bunx` once deployed.
7. If `bun` is not installed on the machine, it should ~~fallback to `npm`. Created project should have a `package-lock.json` file.~~ throw an error that selected package manager is not available (only when using flag)
8. Run `clean command` in the created project - the `Bun` option should be available. It runs `bun pm cache rm` under the hood.

Checklist
----------

### DIFF
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -97,12 +97,14 @@ module.exports = {
 Skip dependencies installation
 
 #### `--npm`
+> [!WARNING]  
+> `--npm` is deprecated and will be removed in the future. Please use `--pm npm` instead.
 
 Force use of npm during initialization
 
-#### `--bun`
+#### `--pm <string>`
 
-Force use of bun during initialization
+Use specific package manager to initialize the project. Available options: `yarn`, `npm`, `bun`. Default: `yarn`
 
 #### `--package-name <string>`
 
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -100,6 +100,10 @@ Skip dependencies installation
 
 Force use of npm during initialization
 
+#### `--bun`
+
+Force use of bun during initialization
+
 #### `--package-name <string>`
 
 Create project with custom package name for Android and bundle identifier for iOS. The correct package name should:
--- a/packages/cli/src/commands/init/init.ts
+++ b/packages/cli/src/commands/init/init.ts
@@ -27,6 +27,7 @@ const DEFAULT_VERSION = 'latest';
 type Options = {
   template?: string;
   npm?: boolean;
+  bun?: boolean;
   directory?: string;
   displayName?: string;
   title?: string;
@@ -39,6 +40,7 @@ interface TemplateOptions {
   projectName: string;
   templateUri: string;
   npm?: boolean;
+  bun?: boolean;
   directory: string;
   projectTitle?: string;
   skipInstall?: boolean;
@@ -82,6 +84,7 @@ async function createFromTemplate({
   projectName,
   templateUri,
   npm,
+  bun,
   directory,
   projectTitle,
   skipInstall,
@@ -100,7 +103,19 @@ async function createFromTemplate({
   try {
     loader.start();
 
-    await installTemplatePackage(templateUri, templateSourceDir, npm);
+    let packageManager: PackageManager.PackageManager = 'yarn';
+
+    if (npm) {
+      packageManager = 'npm';
+    } else if (bun) {
+      packageManager = 'bun';
+    }
+
+    await installTemplatePackage(
+      templateUri,
+      templateSourceDir,
+      packageManager,
+    );
 
     loader.succeed();
     loader.start('Copying template');
@@ -137,7 +152,7 @@ async function createFromTemplate({
 
     if (!skipInstall) {
       await installDependencies({
-        npm,
+        packageManager,
         loader,
         root: projectDirectory,
       });
@@ -153,18 +168,18 @@ async function createFromTemplate({
 }
 
 async function installDependencies({
-  npm,
+  packageManager,
   loader,
   root,
 }: {
-  npm?: boolean;
+  packageManager?: PackageManager.PackageManager;
   loader: Loader;
   root: string;
 }) {
   loader.start('Installing dependencies');
 
   await PackageManager.installAll({
-    preferYarn: !npm,
+    packageManager,
     silent: true,
     root,
   });
@@ -202,11 +217,30 @@ async function createProject(
     projectName,
     templateUri,
     npm: options.npm,
+    bun: options.bun,
     directory,
     projectTitle: options.title,
     skipInstall: options.skipInstall,
     packageName: options.packageName,
   });
+}
+
+function resolvePackageManager() {
+  const userAgent = process.env.npm_config_user_agent;
+
+  if (userAgent) {
+    if (userAgent.startsWith('yarn')) {
+      return 'yarn';
+    }
+    if (userAgent.startsWith('npm')) {
+      return 'npm';
+    }
+    if (userAgent.startsWith('bun')) {
+      return 'bun';
+    }
+  }
+
+  return 'npm';
 }
 
 export default (async function initialize(
@@ -222,6 +256,12 @@ export default (async function initialize(
   const root = process.cwd();
   const version = options.version || DEFAULT_VERSION;
   const directoryName = path.relative(root, options.directory || projectName);
+
+  const packageManager = resolvePackageManager();
+
+  if (packageManager === 'bun') {
+    options.bun = true;
+  }
 
   await createProject(projectName, directoryName, version, options);
 
--- a/packages/cli/src/commands/init/init.ts
+++ b/packages/cli/src/commands/init/init.ts
@@ -225,7 +225,7 @@ async function createProject(
   });
 }
 
-function resolvePackageManager() {
+function userAgentPackageManager() {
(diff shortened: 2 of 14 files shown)


packages/cli/src/commands/init/init.ts
```diff
@@ -202,13 +217,32 @@ async function createProject(
...
+    bun: options.bun,
     directory,
     projectTitle: options.title,
     skipInstall: options.skipInstall,
     packageName: options.packageName,
   });
 }
 
+function resolvePackageManager() {
+  const userAgent = process.env.npm_config_user_agent;
```
thymikee, line 229: this is more "override package manager based on user config", right? maybe let's rename this fn accordingly

thymikee (approved):

docs/commands.md
```diff
@@ -100,6 +100,10 @@ Skip dependencies installation
 
 Force use of npm during initialization
 
+#### `--bun`
```
satya164, line 103: We should do `--package-manager=bun|npm|yarn` or `-pm=bun|npm|yarn`

jaworek:
> If bun is not installed on the machine, it should fallback to npm. Created project should have a package-lock.json file.

I think it would be safer to give user an error and let them know that `bun` is missing or cannot be found in path. Doing fallback to different tool might be unexpected and not something that user wants to do. As an alternative you could ask a followup question whether they want to use `npm` instead.

docs/commands.md
thymikee: I thought about it, but it's so long to type :D We can have both as well as input and then convert to internal representation that bridges those

TMisiukiewicz: additionally this would mean deprecation of `--npm` flag probably. Anyway people will probably just use `bunx react-native init` without using any additional flag. However if you think creating a flag to specify the package manager would be useful, I can do that. I can see the advantage of that solution once there is any new hot package manager available to use

TMisiukiewicz:
> > If bun is not installed on the machine, it should fallback to npm. Created project should have a package-lock.json file.
> 
> I think it would be safer to give user an error and let them know that `bun` is missing or cannot be found in path. Doing fallback to different tool might be unexpected and not something that user wants to do. As an alternative you could ask a followup question whether they want to use `npm` instead.

thanks, added this kind of a check 🙌

shamilovtim:
Will bun be used as the runtime to run metro, prettier, tsc, eslint and so on with this change?

mani-hipages:
Nice work

Jarred-Sumner:
Exciting. Let me know if you run into any issues with Bun

docs/commands.md
TMisiukiewicz: added `--pm` flag and set `--npm` as deprecated ✅

tido64:
Did you try building the project after initializing a project with Bun? `bun react-native run-android --no-packager` fails for me with

```
error Android project not found. Are you sure this is a React Native project? If your Android files are located in a non-standard location (e.g. not inside 'android' folder), consider setting `project.android.sourceDir` option to point to a new location.
```

Jarred-Sumner:
If post-install of react-native dependencies is necessary, that will need to be worked around until we ship the default allowlist for `trustedDependencies`

tido64:
> If postinstall of react-native dependencies is necessary, that will need to be worked around until we ship the default allowlist for `trustedDependencies`

Never mind, my workaround for https://github.com/oven-sh/bun/issues/5063 didn't work properly. It works once I've fixed it.

TMisiukiewicz:
> Did you try building the project after initializing a project with Bun? `bun react-native run-android --no-packager` fails for me with
> 
> ```
> error Android project not found. Are you sure this is a React Native project? If your Android files are located in a non-standard location (e.g. not inside 'android' folder), consider setting `project.android.sourceDir` option to point to a new location.
> ```

Seems to be working

https://github.com/react-native-community/cli/assets/13985840/7beb315c-8867-4032-a8c8-679cabb9b3c2

szymonrybczak (approved):
Let's ship it :shipit:

docs/commands.md
```diff
@@ -97,12 +97,14 @@ module.exports = {
 Skip dependencies installation
 
 #### `--npm`
+> [!WARNING]  
+> `--npm` is deprecated and will be removed in the future. Please use `--pm npm` instead.
 
 Force use of npm during initialization
 
-#### `--bun`
+#### `--pm <string>`
```
thymikee, line 105: let's be specific what goes in here: npm|yarn|bun

packages/cli/src/commands/init/init.ts
```diff
@@ -176,6 +200,24 @@ async function installDependencies({
...
+
+    return isBunAvailable;
+  } else if (options.pm === 'npm') {
+    const isNpmAvailable = getNpmVersionIfAvailable();
+
+    return isNpmAvailable;
+  } else if (options.pm === 'yarn') {
+    const isYarnAvailable = getYarnVersionIfAvailable();
+
+    return isYarnAvailable;
```
thymikee, line 215:
```suggestion
    return getBunVersionIfAvailable();
  } else if (options.pm === 'npm') {
    return getNpmVersionIfAvailable();
  } else if (options.pm === 'yarn') {
    return getYarnVersionIfAvailable();
```

packages/cli/src/commands/init/init.ts
```diff
@@ -90,6 +96,20 @@ async function createFromTemplate({
...
+      'Flag --npm is deprecated and will be removed soon. In the future, please use --pm npm instead.',
+    );
+
+    packageManager = 'npm';
+  }
+
+  const userAgent = userAgentPackageManager();
+  if (userAgent === 'bun') {
+    packageManager = 'bun';
+  }
```
thymikee, line 111: this favors bun and introduces some inconsistency in picking the preferred package manger. The algorithm should go like this imho:
1. if flag, set pm from flag
2. if no flag, set pm from userAgent
3. if no userAgent, fallback to default (yarn currently)

packages/cli/src/tools/__tests__/packageManager-test.ts
```diff
@@ -83,9 +100,65 @@ describe('npm', () => {
   });
 });
 
+describe('bun', () => {
```
thymikee, line 103: we only test the happy path. what if there's no `bun.lockb` or `getBunVersionIfAvailable` returns false?

packages/cli/src/tools/__tests__/packageManager-test.ts
```diff
@@ -83,9 +100,65 @@ describe('npm', () => {
...
+  it('should install', () => {
+    jest.spyOn(bun, 'getBunVersionIfAvailable').mockImplementation(() => true);
+    jest
+      .spyOn(bun, 'isProjectUsingBun')
+      .mockImplementation(() => './path/to/bun.lockb');
```
thymikee, line 108: We could avoid mocking implementation details by mocking `execa` and setting `mockResolvedValue` manually based on our expectations of what's to be returned from a system call. And setting fixtures including these files with e.g. `__setMockFilesystem`. There are examples of doing this in the repo, so hopefully it's not too much of an ask

packages/cli/src/commands/init/init.ts
```diff
@@ -176,6 +202,22 @@ async function installDependencies({
   loader.succeed();
 }
 
+function checkPackageManagerAvailability(options: Options) {
```
thymikee, line 205: no need to put all Options here, only PackageManager will do

packages/cli/src/commands/init/init.ts
```diff
@@ -176,6 +202,22 @@ async function installDependencies({
...
+  if (!options.pm) {
+    return true;
```
thymikee, line 207: this early return should be moved outside of the fn to the caller:
```js
if (options.pm) {
  checkPackageManagerAvaialability(options.pm);
} else { ... }
```

packages/cli/src/tools/__tests__/packageManager-test.ts
TMisiukiewicz: not sure if we can avoid mocking this implementation details - we need to check if it's using the proper package manager since if it's not available, the `npm` is used as fallback (same way that yarn works). This tests are checking if execa is called with proper package manager rather than it installs/uninstalls dependencies correctly

However in the `init` command we should not ever get into this kind of situation when it's using npm instead of yarn/bun because now we have a check in the beginning of the code execution, throwing an error if package manager is not available to use

PR #2077, opened 2023-09-15
refactor: move pods installation from `init` to ios-specific commands

TMisiukiewicz:
Summary:
---------
Currently Pods are installed during `init` command, what makes project setup much longer than it supposed to be. This change makes pods installation optional during the `init` command by displaying the prompt, and moves the installation process to the `run-ios` and `build-ios` commands. Pods are automatically installed when running ios-specific commands and:
- they were not installed before
- a new dependency/devDependency was introduced in `package.json`
- new flag `--force-pods` was used, which forces to install pods (might be helpful when there are any errors related to pods)

Additionally, new flag `--install-pods` was introduced to `init` command, to avoid displaying prompt. If set to `true`, pods will install, `false` will skip the installation, and when not used, prompt will be displayed.

To install pods during `*-ios` command, it compares md5 hashes of native dependencies names and its versions. To achieve this, I added one new field to dependency config: `version` which holds the version of the library from package.json/podspec file.

## Demo
**1. `init` with prompt**

https://github.com/react-native-community/cli/assets/13985840/1028821d-8a6d-4361-be85-c37f6eca5bbc

**2. `init` with `--install-pods false` - skip prompt and pods installation**

https://github.com/react-native-community/cli/assets/13985840/cde70ccd-a310-42b8-8d54-fa18fc1d36d4

**3. `run-ios` for the first time** (cut the video because it takes too long 😄, app runs after installation)

https://github.com/react-native-community/cli/assets/13985840/a50fabb7-7092-402e-8cb2-3f4b32a9ec6c

Test Plan:
----------

1. Follow the contributing instructions
2. Init new project - it should display prompt about pods installation
3. Press `y` to confirm - pods should be installed
4. Init new project once again - it should display prompt
5. Press 'n' - pods installation should be skipped
6. Init new project with `--install-pods` flag - pods should be installed automatically
7. Init new project with `--install-pods false` - prompt & pods installation should be skipped
8. Run `run-ios` command and verify pods are installed before building the app
9. Run `run-ios` again and verify app is built without installing pods
10. Install any random dependency, e.g. `@react-navigation/native`
11. Run `run-ios` again and verify pods are installed before building the app
12. Run `run-ios` again and verify app is built without installing pods
13. Run `run-ios --force-pods` and verify pods are installed before building the app

Checklist
----------


packages/cli-platform-ios/src/tools/pods.ts
```diff
@@ -0,0 +1,80 @@
+import path from 'path';
+import fs from 'fs-extra';
+import md5 from 'crypto-js/md5';
```
thymikee, line 3: let's use built-in `crypto` module, no need for extra dep here

packages/cli-platform-ios/src/index.ts
```diff
@@ -4,3 +4,4 @@
 
 export {default as commands} from './commands';
 export {projectConfig, dependencyConfig, findPodfilePaths} from './config';
+export {runSudo} from './tools/installPods';
```
thymikee, line 7: `runSudo` is now a public API of `cli-platform-ios`. Is it obvious what command it runs sudo for? Not really. Let's rename it accordingly, `installPodsWithSudo` or something else relevant

packages/cli-platform-ios/src/tools/installPods.ts
```diff
@@ -1,5 +1,6 @@
 import fs from 'fs';
 import execa from 'execa';
+import {Ora} from 'ora';
```
thymikee, line 3:
```suggestion
import type {Ora} from 'ora';
```

packages/cli-tools/src/index.ts
```diff
@@ -14,5 +14,6 @@ export * as link from './doclink';
 export {default as startServerInNewWindow} from './startServerInNewWindow';
 export {default as handlePortUnavailable} from './handlePortUnavailable';
 export * from './port';
+export {default as cacheManager} from './releaseChecker/releaseCacheManager';
```
thymikee, line 17: maybe let's rename `releaseCacheManager` to `cacheManager` then?

packages/cli-platform-ios/src/index.ts
TMisiukiewicz: basically it takes a command as a parameter, so it can be used to run anything. Maybe it's better idea to move it to `cli-tools` then?

packages/cli-platform-ios/src/tools/pods.ts
```diff
@@ -0,0 +1,80 @@
...
+export function compareMd5Hashes(hash1: string, hash2: string) {
+  return hash1 === hash2;
+}
+
+export default async function resolvePods(
+  root: string,
+  options?: ResolvePodsOptions,
+) {
+  const packageJson = getPackageJson(root);
+  const podsPath = path.join(root, 'ios', 'Pods');
```
thymikee, line 43: We have a `findPodfilePath` helper, maybe let's use that?

packages/cli-platform-ios/src/index.ts
thymikee: makes more sense

packages/cli-platform-ios/src/tools/pods.ts
```diff
@@ -0,0 +1,80 @@
...
+  const arePodsInstalled = fs.existsSync(podsPath);
+  const dependencies = normalizeDependencies({
+    ...packageJson.dependencies,
+    ...packageJson.devDependencies,
+  });
```
thymikee, line 48: instead of looping through all dependencies and make any tiny non-ios related dep invalidate our pods cache, we could rather use `config.depdenencies` and filter out those that have `platforms.ios` set

packages/cli-platform-ios/src/config/index.ts
```diff
@@ -66,8 +66,12 @@ export function dependencyConfig(
     return null;
   }
 
+  const packageJson = require(path.join(folder, 'package.json'));
```
thymikee, line 69: missing try catch

packages/cli-platform-ios/src/tools/pods.ts
```diff
@@ -23,13 +23,13 @@ export function getPackageJson(root: string) {
...
-  dependencies: NativeDependencies,
-  dependenciesVersions: Record<string, string>,
-) {
+export function getIosDependencies(dependencies: NativeDependencies) {
   return Object.keys(dependencies)
     .filter((dependency) => dependencies[dependency].platforms.ios)
-    .map((dependency) => `${dependency}@${dependenciesVersions[dependency]}`)
+    .map(
+      (dependency) =>
+        `${dependency}@${dependencies[dependency].platforms.ios?.version}`,
```
thymikee, line 31: `ios?.version` can evaluate to undefined and will produce messy entries

TMisiukiewicz: since it's already filtered by the dependencies where `platforms.ios` exists, this should never happen. Removed `?` and casted it to `IOSDependencyConfig`

packages/cli-platform-ios/README.md
```diff
@@ -204,6 +208,10 @@ Example:
 npx react-native build-ios --extra-params "-jobs 4"
 ```
 
+#### `--force-pods`,
+
+Force running `pod install` before running an app
```
szymonrybczak, line 213:
```suggestion
Force running `pod install` before building an app
```

packages/cli-platform-ios/src/tools/pods.ts
```diff
@@ -0,0 +1,98 @@
...
+interface ResolvePodsOptions {
+  forceInstall?: boolean;
+}
+
+interface NativeDependencies {
+  [key: string]: DependencyConfig;
+}
+
+export function getPackageJson(root: string) {
+  return require(path.join(root, 'package.json'));
```
thymikee, line 26: shouldn't we try/catch here of from the caller? btw, Android introduced similar logic for resolving package.json in this PR: https://github.com/react-native-community/cli/pull/1635 maybe it's worth to share the implementation

thymikee (approved):

adamTrz (approved):

packages/cli/src/commands/init/init.ts
```diff
@@ -170,6 +185,30 @@ async function createFromTemplate({
         loader,
         root: projectDirectory,
       });
+
+      if (process.platform === 'darwin') {
+        const installPodsValue = String(installCocoaPods);
```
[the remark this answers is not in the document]
TMisiukiewicz: basically you can have a multiple values here:
`true` - when you use `--install-pods` 
`"true"`/`"false"` - when `--install-pods true`/ `--install-pods false`
`undefined` - when flag is not used

converting to a string seems like the fastest way to avoid building complicated if statements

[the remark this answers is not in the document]
thymikee: Let's change the type to `string | boolean` then

PR #2078, opened 2023-09-18
refactor(iOS): remove unintended flags from build-ios

thymikee:
Summary:
---------

The `build-ios` command contained some left-over flags from `run-ios` like simulator, uuid, device. Removed the references and moved the code a bit to remove duplication.

As a followup we should focus on bringing interactive behavior as default

Test Plan:
----------

-

Checklist
----------


szymonrybczak (approved):
Great cleanup! 🙌 I still see some flags such us `--binary-path` or `--list-devices` under `build-ios` command instead of `run-ios` in docs ;/ We need to fully sync docs with codebase.

adamTrz (approved):
Thanks for the cleanup, works like a charm 💪

PR #2086, opened 2023-09-26
fix(iOS): restore simulator, device, and UDID args on `run-ios` command

huntie:
Addressing regression from #2078


thymikee (approved):
Thanks!

PR #2088, opened 2023-09-26
feat: create git repository in `init` command

szymonrybczak:
Summary:
---------

Creates Git repository in `init` command. 

Test Plan:
----------

2. Run this command:

```sh
```
3. Git repo should be initalised, and whole project should be commited. 
 
Checklist
----------


packages/cli/src/commands/init/init.ts
```diff
@@ -28,8 +28,10 @@ import {getBunVersionIfAvailable} from '../../tools/bun';
 import {getNpmVersionIfAvailable} from '../../tools/npm';
 import {getYarnVersionIfAvailable} from '../../tools/yarn';
 import {createHash} from 'crypto';
+import execa from 'execa';
 
 const DEFAULT_VERSION = 'latest';
+const packageJson = require('./../../../package.json');
```
szymonrybczak, line 34: Is it safe?

packages/cli/src/commands/init/init.ts
```diff
@@ -307,6 +309,27 @@ function userAgentPackageManager() {
...
+  try {
+    await execa('git', ['init'], {cwd: folder});
+    await execa('git', ['branch', '-M', 'main'], {cwd: folder});
+    await execa('git', ['add', '.'], {cwd: folder});
+    await execa(
+      'git',
+      [
+        'commit',
+        '-m',
+        `Initial commit\n\nGenerated by ${packageJson.name} ${packageJson.version}.`,
```
thymikee, line 322: I’d avoid using package json info. Users don’t know which cli version they use, all they care about is RN version

packages/cli/src/commands/init/init.ts
```diff
@@ -307,6 +309,27 @@ function userAgentPackageManager() {
   return null;
 }
 
+const createGitRepository = async (folder: string) => {
```
thymikee, line 312: This is doing more than creating for repo. Do we really want to make initial commit? What about systems where fit is not available? We’re not making any checks, so this adds extra surface for init errors

packages/cli/src/commands/init/init.ts
szymonrybczak: Agree, CLI version is not very informative. I don't know if we want to declare in codebase which CLI version points to which RN version (I know that there's idea of doing this in this PR https://github.com/react-native-community/cli/pull/1983/files). 🤔

packages/cli/src/commands/init/init.ts
szymonrybczak: If Git won't be available I think it would be good to simply add log to inform that there's a problem with creating Git repo and just continue (and I think that we can log error with `logger.debug` so that advanced users can use check it with `--verbose`).

thymikee: Yeah, let’s just make sure we handle the case

szymonrybczak: done ✅ I also added handling case when somebody tries to init project inside already existing git repo.

TMisiukiewicz (approved):

packages/cli/src/commands/init/init.ts
adamTrz: Maybe sth like this?
https://github.com/react-native-community/cli/blob/dea06427cedcfddd5170509e8dad797f38c356c2/packages/cli-config/src/findDependencies.ts#L11-L13

packages/cli/src/commands/init/init.ts
adamTrz: Maybe let's RN version directly from node_modules? 
Sth like `cat node_modules/react-native/package.json`

szymonrybczak: https://github.com/react-native-community/cli/pull/2088/commits/a725e41e2adc27f1b51f861ffef27bf5ebd4e669 ✅

packages/cli/src/commands/init/init.ts
szymonrybczak: Since we added reading RN version from `node_modules`, we don't need to read `package.json`.

PR #2089, opened 2023-09-27
fix: Creating cache for scoped packages

TMisiukiewicz:
Summary:
---------
If `package.json` had scoped package name, e.g. `@react-native/tester`, it was unable to create CLI cache since `fs.writeFileSync` could not create it in a path that does not exist. With this PR it will recursively create necessary folder for scoped packages before trying to save the cache file.

Test Plan:
----------
1. Follow the contributing guide
2. run `node ./path/to/cli init AppName`
3. Run `yarn ios` and make sure app runs correctly
4. Change `name` in `package.json` to `@react-native/tester`
5. Run `yarn ios` and make sure app runs correctly
6. Open `~/Library/Caches/react-native-cli/@react-native` folder and make sure `tester` file exists

Checklist
----------

### DIFF
--- a/packages/cli-tools/src/cacheManager.ts
+++ b/packages/cli-tools/src/cacheManager.ts
@@ -28,10 +28,7 @@ function loadCache(name: string): Cache | undefined {
 function saveCache(name: string, cache: Cache) {
   const fullPath = path.resolve(getCacheRootPath(), name);
 
-  const dir = path.dirname(fullPath);
-  if (!fs.existsSync(dir)) {
-    fs.mkdirSync(dir);
-  }
+  fs.mkdirSync(path.dirname(fullPath), {recursive: true});
   fs.writeFileSync(fullPath, JSON.stringify(cache, null, 2));
 }
 
--- a/packages/cli-tools/src/cacheManager.ts
+++ b/packages/cli-tools/src/cacheManager.ts
@@ -26,10 +26,13 @@ function loadCache(name: string): Cache | undefined {
 }
 
 function saveCache(name: string, cache: Cache) {
-  fs.writeFileSync(
-    path.resolve(getCacheRootPath(), name),
-    JSON.stringify(cache, null, 2),
-  );
+  const fullPath = path.resolve(getCacheRootPath(), name);
+
+  const dir = path.dirname(fullPath);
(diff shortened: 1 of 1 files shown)


szymonrybczak (approved):

packages/cli-tools/src/cacheManager.ts
```diff
@@ -26,10 +26,13 @@ function loadCache(name: string): Cache | undefined {
...
-  fs.writeFileSync(
-    path.resolve(getCacheRootPath(), name),
-    JSON.stringify(cache, null, 2),
-  );
+  const fullPath = path.resolve(getCacheRootPath(), name);
+
+  const dir = path.dirname(fullPath);
+  if (!fs.existsSync(dir)) {
+    fs.mkdirSync(dir, {recursive: true}); // Create the directory and its parents if necessary
+  }
```
tido64, line 34: If you add `force: true`, you can skip the check entirely:
```suggestion
  fs.mkdirSync(dir, {recursive: true, force: true}); // Create the directory and its parents if necessary
```

TMisiukiewicz: I can't see `force` option in `mkdirSync`, seems like it's available only in `rm`

tido64: Doh! 🤦‍♂️  You're right. I was mixing the two. But still, I don't think you actually need the check here if `recursive: true`.

TMisiukiewicz: yeah you are right, removed!

packages/cli-tools/src/cacheManager.ts
```diff
@@ -26,10 +26,13 @@ function loadCache(name: string): Cache | undefined {
...
-  fs.writeFileSync(
-    path.resolve(getCacheRootPath(), name),
-    JSON.stringify(cache, null, 2),
-  );
+  const fullPath = path.resolve(getCacheRootPath(), name);
+
+  const dir = path.dirname(fullPath);
+  if (!fs.existsSync(dir)) {
+    fs.mkdirSync(dir);
+  }
```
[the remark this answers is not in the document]
TMisiukiewicz: you're right 🤦 thanks for spotting that

PR #2111, opened 2023-10-10
fix: remove hardcoded path to iOS folder

szymonrybczak:
Summary:
---------

We get path to iOS folder by looking for `Podfile`, even if we rename folder containing iOS files it should work, but by hardcoding it will break.

Test Plan:
----------

2. Rename `ios` folder -> `mv ios fios`,
3. Run this command:

```sh
node /path/to/react-native-cli/packages/cli/build/bin.js run-ios
```
4. Install library (to get new hash)
```sh
yarn install react-native-slider
```
6. Should detect change in `package.json` and install install pods in new folder.
```sh
node /path/to/react-native-cli/packages/cli/build/bin.js run-ios
```

Checklist
----------

### DIFF
--- a/packages/cli-platform-ios/src/tools/installPods.ts
+++ b/packages/cli-platform-ios/src/tools/installPods.ts
@@ -110,14 +110,18 @@ async function installCocoaPods(loader: Ora) {
   }
 }
 
-async function installPods(loader?: Ora, options?: PodInstallOptions) {
+async function installPods(
+  loader?: Ora,
+  iosFolderPath?: string,
+  options?: PodInstallOptions,
+) {
   loader = loader || new NoopLoader();
   try {
-    if (!fs.existsSync('ios')) {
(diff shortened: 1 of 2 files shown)


packages/cli-platform-ios/src/tools/installPods.ts
```diff
@@ -110,14 +110,18 @@ async function installCocoaPods(loader: Ora) {
   }
 }
 
-async function installPods(loader?: Ora, options?: PodInstallOptions) {
+async function installPods(
+  loader?: Ora,
+  options?: PodInstallOptions,
+  iosFolderPath?: string,
```
thymikee, line 116: Can we change the order so that options object is the last? This is a convention that config objects are the last argument
```suggestion
  loader?: Ora,
  iosFolderPath?: string,
  options?: PodInstallOptions,
```

szymonrybczak: done ✅

Issue #2105, opened 2023-10-09
run-android with --mode fails without error info (after upgrading to v72.5)


fikkatra:
## Environment
System:
  OS: macOS 14.0
  CPU: (8) arm64 Apple M1 Pro
  Memory: 104.59 MB / 32.00 GB
  Shell:
    version: "5.9"
    path: /bin/zsh
Binaries:
  Node:
    version: 18.15.0
    path: /usr/local/bin/node
  Yarn:
    version: 1.22.19
    path: /usr/local/bin/yarn
  npm:
    version: 9.5.0
    path: /usr/local/bin/npm
  Watchman:
    version: 2023.10.02.00
    path: /opt/homebrew/bin/watchman
Managers:
  CocoaPods:
    version: 1.13.0
    path: /Users/fikkatra/.rbenv/shims/pod
SDKs:
  iOS SDK:
    Platforms:
      - DriverKit 23.0
      - iOS 17.0
      - macOS 14.0
      - tvOS 17.0
      - watchOS 10.0
  Android SDK: Not Found
IDEs:
  Android Studio: 2022.3 AI-223.8836.35.2231.10811636
  Xcode:
    version: 15.0/15A240d
    path: /usr/bin/xcodebuild
Languages:
  Java:
    version: 11.0.20.1
    path: /usr/bin/javac
  Ruby:
    version: 2.7.6
    path: /Users/fikkatra/.rbenv/shims/ruby
npmPackages:
  "@react-native-community/cli": Not Found
  react:
    installed: 18.2.0
    wanted: 18.2.0
  react-native:
    installed: 0.72.5
    wanted: 0.72.5
  react-native-macos: Not Found
npmGlobalPackages:
  "*react-native*": Not Found
Android:
  hermesEnabled: true
  newArchEnabled: false
iOS:
  hermesEnabled: true
  newArchEnabled: false

## Description

After following the guide to upgrade from 71 to 72, we get an error when running Android. 

The command:
`react-native run-android --mode=myAppDevDebug --appId com.myapp.dev`

fails with the following output (and nothing else):

> info JS server already running.
> 
> info 💡 Tip: Make sure that you have set up your development environment correctly, by running react-native doctor. To read more about doctor command visit: https://github.com/react-native-community/cli/blob/main/packages/cli-doctor/README.md#doctor
> 
> error Failed to install the app.

Except from following the steps mentioned in the upgrade guide, and updating library versions, nothing has changed.

We have the following product flavors defined:
```
[...]

PR #2113, opened 2023-10-10
fix(android): properly handle errors

szymonrybczak:
Summary:
---------

Currently we're not displaying errors if they other than we expect, so even we're not showing errors that we return inside CLI - which leads to situations like this one https://github.com/react-native-community/cli/issues/2105, when user don't know what to do because the error is only "**error** Failed to install the app". 

Test Plan:
----------
2. Run this command:

```sh
node /path/to/react-native-cli/packages/cli/build/bin.js run-android --mode=randomMode
```

Checklist
----------


szymonrybczak:
Definitely we need more tests in that part of codebase.

Issue #2115, opened 2023-10-12
`react-native run-ios` fails on 0.73 trying to find Xcode schemes


tido64:
## Environment

```
System:
  OS: macOS 14.0
  CPU: (10) arm64 Apple M1 Max
  Memory: 4.11 GB / 64.00 GB
  Shell:
    version: 3.6.1
    path: /opt/homebrew/bin/fish
Binaries:
  Node:
    version: 18.18.0
    path: ~/.local/bin/node
  Yarn:
    version: 3.6.1
    path: /opt/homebrew/bin/yarn
  npm:
    version: 9.8.1
    path: ~/.local/bin/npm
  Watchman:
    version: 2023.10.09.00
    path: /opt/homebrew/bin/watchman
Managers:
  CocoaPods:
    version: 1.12.1
    path: /Users/tido/.gem/bin/pod
SDKs:
  iOS SDK:
    Platforms:
      - DriverKit 23.0
      - iOS 17.0
      - macOS 14.0
      - tvOS 17.0
      - watchOS 10.0
  Android SDK:
    API Levels:
      - "31"
      - "33"
      - "34"
    Build Tools:
      - 30.0.3
      - 31.0.0
      - 33.0.0
      - 33.0.1
    System Images:
      - android-33 | Google APIs ARM 64 v8a
    Android NDK: Not Found
IDEs:
  Android Studio: 2022.2 AI-222.4459.24.2221.10121639
  Xcode:
    version: 15.0/15A240d
    path: /usr/bin/xcodebuild
Languages:
  Java:
    version: 17.0.8.1
    path: /usr/bin/javac
  Ruby:
    version: 3.2.2
    path: /opt/homebrew/opt/ruby/bin/ruby
npmPackages:
  "@react-native-community/cli":
    installed: 12.0.0-alpha.17
    wanted: 12.0.0-alpha.17
  react:
    installed: 18.2.0
    wanted: 18.2.0
  react-native:
    installed: 0.73.0-rc.2
    wanted: ^0.72.0
  react-native-macos:
    installed: 0.72.0
    wanted: ^0.72.0
npmGlobalPackages:
  "*react-native*": Not Found
Android:
  hermesEnabled: false
  newArchEnabled: true
iOS:
  hermesEnabled: false
  newArchEnabled: false
```

## Description

`react-native run-ios` fails if `.xcodeproj` doesn't live next to `Podfile`:

```
error Cannot read properties of undefined (reading 'schemes').
[...]

PR #2121, opened 2023-10-13
fix: get scheme from `xcodeProject` name

szymonrybczak:
Summary:
---------

Fixes https://github.com/react-native-community/cli/issues/2115
Regression introduced in https://github.com/react-native-community/cli/pull/2078

Reverts this code that was used in 0.72:
https://github.com/react-native-community/cli/blob/9d486c833d5af7f2e6fbf7bec046e54283ba140f/packages/cli-platform-ios/src/commands/runIOS/index.ts#L87-L92

Test Plan:
----------

```
cd react-native-test-app
npm run set-react-version 0.73 -- --core-only
yarn
cd example

# Link CLI locally following CONTRIBUTING.md
# Because of an issue with finding CocoaPods, you may have to remove line 51
# packages/cli-platform-ios/src/commands/runIOS/index.ts

yarn ios
```

Checklist
----------

### DIFF
--- a/packages/cli-platform-ios/src/commands/buildIOS/getConfiguration.ts
+++ b/packages/cli-platform-ios/src/commands/buildIOS/getConfiguration.ts
@@ -2,7 +2,7 @@ import chalk from 'chalk';
 import {IOSProjectInfo} from '@react-native-community/cli-types';
 import {logger} from '@react-native-community/cli-tools';
 import {selectFromInteractiveMode} from '../../tools/selectFromInteractiveMode';
-import {getProjectInfo} from '../../tools/getProjectInfo';
+import {getInfo} from '../../tools/getInfo';
 import {checkIfConfigurationExists} from '../../tools/checkIfConfigurationExists';
 import type {BuildFlags} from './buildOptions';
 import {getBuildConfigurationFromXcScheme} from '../../tools/getBuildConfigurationFromXcScheme';
@@ -13,10 +13,13 @@ export async function getConfiguration(
   sourceDir: string,
   args: BuildFlags,
 ) {
-  const projectInfo = getProjectInfo();
+  const info = getInfo();
 
   if (args.mode) {
-    checkIfConfigurationExists(projectInfo, args.mode);
+    if (!info || !info.configurations) {
+      throw new Error('Cannot determine Xcode project configuration.');
+    }
+    checkIfConfigurationExists(info.configurations, args.mode);
   }
 
   let scheme =
@@ -24,13 +27,13 @@ export async function getConfiguration(
     path.basename(xcodeProject.name, path.extname(xcodeProject.name));
   let mode =
     args.mode ||
-    getBuildConfigurationFromXcScheme(scheme, 'Debug', sourceDir, projectInfo);
+    getBuildConfigurationFromXcScheme(scheme, 'Debug', sourceDir, info);
 
   if (args.interactive) {
     const selection = await selectFromInteractiveMode({
       scheme,
       mode,
-      projectInfo,
+      info,
     });
 
     if (selection.scheme) {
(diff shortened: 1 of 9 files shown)


packages/cli-platform-ios/src/tools/selectFromInteractiveMode.ts
```diff
@@ -20,13 +20,13 @@ export async function selectFromInteractiveMode({
   let newScheme = scheme;
   let newMode = mode;
 
-  if (projectInfo.schemes.length > 1) {
+  if (projectInfo && projectInfo.schemes.length > 1) {
```
thymikee, line 23: knowing that `projectInfo` can be undefined for workspaces (only `schemes` are returned from `xcodebuild -list` then, shall we adjust this to account for a workspace schemes? now it will automatically select `scheme` and `mode` passed to that fn

szymonrybczak:
@tido64 Mind looking at these changes? 🙏 I tested this in RNTA, and it seems to be working correctly.

packages/cli-platform-ios/src/commands/buildIOS/getConfiguration.ts
```diff
@@ -2,32 +2,35 @@ import chalk from 'chalk';
...
   xcodeProject: IOSProjectInfo,
   sourceDir: string,
   args: BuildFlags,
 ) {
-  const projectInfo = getProjectInfo();
+  const info = getInfo();
 
   if (args.mode) {
-    checkIfConfigurationExists(projectInfo, args.mode);
+    checkIfConfigurationExists((info && info.configurations) ?? [], args.mode);
```
tido64, line 19: Nit: Use optional chaining:
```suggestion
    checkIfConfigurationExists(info?.configurations ?? [], args.mode);
```

packages/cli-platform-ios/src/tools/selectFromInteractiveMode.ts
```diff
@@ -9,25 +9,25 @@ import {
...
-  projectInfo,
+  info,
 }: Args): Promise<{scheme?: string; mode?: string}> {
   let newScheme = scheme;
   let newMode = mode;
 
-  if (projectInfo.schemes.length > 1) {
-    newScheme = await promptForSchemeSelection(projectInfo);
+  if (info && info?.schemes && info.schemes.length > 1) {
+    newScheme = await promptForSchemeSelection(info.schemes);
```
tido64, line 24: Nit: `info && info?.schemes` and `info?.schemes` are equivalent. `info &&` is unnecessary. You either write `info && info.schemes` or `info?.schemes`, not mix the two. In this specific case though, I'd probably make a temp so you don't have to do the same lookup multiple times:
```suggestion
  const schemes = info?.schemes;
  if (schemes && schemes.length > 1) {
    newScheme = await promptForSchemeSelection(schemes);
```

packages/cli-platform-ios/src/tools/selectFromInteractiveMode.ts
```diff
@@ -9,25 +9,25 @@ import {
...
   } else {
     logger.info(`Automatically selected ${chalk.bold(scheme)} scheme.`);
   }
 
-  if (projectInfo.configurations.length > 1) {
-    newMode = await promptForConfigurationSelection(projectInfo);
+  if (info && info?.configurations && info?.configurations?.length > 1) {
+    newMode = await promptForConfigurationSelection(info.configurations);
```
tido64, line 30: Same here. Though in this case, you need to be careful about `info?.configurations?.length > 1` since it might return `undefined`, which cannot be compared with numbers. It's fine here because of the previous checks.
```suggestion
  const configurations = info?.configurations;
  if (configurations && configurations.length > 1) {
    newMode = await promptForConfigurationSelection(configurations);
```

tido64 (commented):
Thanks! I had to disable automatic `pod install` (which is fixed in https://github.com/react-native-community/cli/pull/2121), but `react-native run-ios` passes now. Tested against 0.73.0-rc.2.

I left a few nits.

adamTrz (approved):

PR #2122, opened 2023-10-16
fix: install pods on ios-commands with New Arch

TMisiukiewicz:
Summary:
---------
Coming from #2091. Currently, when installed pods with New Architecture manually, added new package and used `run-ios`, it installed pods with old Architecture again. This change checks if the New Architecture is enabled and installs it with/without a flag.

Test Plan:
----------
1. Init new project and install pods during the init
2. Add a library containing native code e.g. `react-native-gesture-handler`
3. Run `run-ios` and verify it installed pods and ran the app
4. Go to `ios` folder and install pods with New Arch manually, `RCT_NEW_ARCH_ENABLED=1 pod install`
5. Add another library, e.g. `react-native-reanimated`
6. Run `run-ios` and verify pods are being installed with a `with New Architecture` message
7. When app is running, verify `fabric: true` is visible in Metro console

Checklist
----------

### DIFF
--- a/packages/cli-platform-ios/src/tools/getArchitecture.ts
+++ b/packages/cli-platform-ios/src/tools/getArchitecture.ts
@@ -1,0 +1,14 @@
+import {readFile} from 'fs-extra';
+import path from 'path';
+
+export default async function getArchitecture(iosSourceDir: string) {
+  try {
+    const project = await readFile(
+      path.join(iosSourceDir, '/Pods/Pods.xcodeproj/project.pbxproj'),
+    );
+
+    return project.includes('-DRCT_NEW_ARCH_ENABLED=1');
+  } catch {
+    return false;
+  }
+}
--- a/packages/cli-platform-ios/src/tools/pods.ts
+++ b/packages/cli-platform-ios/src/tools/pods.ts
@@ -16,6 +16,7 @@ import {
 
 interface ResolvePodsOptions {
   forceInstall?: boolean;
+  newArchEnabled?: boolean;
 }
 
 interface NativeDependencies {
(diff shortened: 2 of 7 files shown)


packages/cli-platform-ios/src/tools/getArchitecture.ts
```diff
@@ -0,0 +1,14 @@
+import {readFile} from 'fs-extra';
+import path from 'path';
+
+export default async function getArchitecture(iosSourceDir: string) {
+  try {
+    const project = await readFile(
+      path.join(iosSourceDir, '/Pods/Pods.xcodeproj/project.pbxproj'),
+    );
+
+    return project.includes('-DRCT_NEW_ARCH_ENABLED=1');
```
thymikee, line 10: @cipolleschi is this fine if we use that to detect new arch?

cipolleschi: yes, this should be enough.

cipolleschi (approved):

thymikee (approved):
@TMisiukiewicz please rebase and merge? :)

packages/cli-platform-ios/src/tools/getArchitecture.ts
```diff
@@ -0,0 +1,14 @@
+import {readFile} from 'fs-extra';
+import path from 'path';
+
+export default async function getArchitecture(iosSourceDir: string) {
```
szymonrybczak, line 4: We have the same logic in `doctor` command:
https://github.com/react-native-community/cli/blob/36d00ea428e52d7d45b0cbf36df707825f185317/packages/cli-doctor/src/commands/info.ts#L55-L64

 mind using this function also there - so we won't have in two places same code? 🙏

szymonrybczak (approved):

packages/cli-platform-ios/src/tools/getArchitecture.ts
TMisiukiewicz: it's already here: https://github.com/react-native-community/cli/blob/ef7a66ab9863ef50ba10b234dc44378d02e0503a/packages/cli-doctor/src/commands/info.ts#L56

TMisiukiewicz:
@szymonrybczak to resolve conflicts i moved `iosFolderPath` to `installPods` options object, would you mind double-check if it's fine?

packages/cli-platform-ios/src/tools/pods.ts
```diff
@@ -114,11 +116,26 @@ export default async function resolvePods(
...
+        currentDependenciesHash,
+      );
+      loader.succeed();
+    } catch {
+      loader.fail();
+      throw new CLIError(
+        `Something when wrong while installing CocoaPods. Please run ${chalk.bold(
+          'pod install',
+        )} manually`,
+      );
```
szymonrybczak, line 138: Hm, I'm wondering should we add here info about flags that we added 🤔

szymonrybczak (approved):
LGTM 🙌

ManasGuptaSprinklr:
@TMisiukiewicz @szymonrybczak  @cortinico  , I am facing an issue. When we run the command **run-ios**, it uses a new arch-enabled flag via the getArchitecture function, which references **Pods/Pods.xcodeproj/project.pbxproj**. However, if someone runs the command **pod-deintegrate**, then running the **run-ios** command will always use the legacy architecture (fallback if file not found). Can you please check this?

RCT_NEW_ARCH_ENABLED=1 pod install
This command can be run before, but it is giving a deprecation message on React Native 0.79. It is suggested to run run-ios directly.

PR #2081, opened 2023-09-19
Add remote debugging deprecation header

huntie:
## Summary

In React Native 0.73 we are deprecating Remote JSC Debugging (the `/debugger-ui` endpoint). This PR adds an informational warning header to this page as a signal to users of this change.

This will be supported by (upcoming) improvements to React Native's debugging docs.

## Test plan

| Light | Dark |
|--------|--------|
| image | image |

PR #2125, opened 2023-10-19
Update remote debugging deprecation notice

huntie:
## Summary

Follows https://github.com/react-native-community/cli/pull/2081.

Tweak details related to 0.73 debugger launch:
- Maintains "Open Debugger workflow" guidance: referring to Flipper or the experimental new debugger.
- Qualifies "pressing `j` in the CLI" with the `--experimental-debugger` flag.
- Adds JSC direct debugging link.

## Test Plan

| Before | After |
|--------|--------|
| image | image |

### DIFF
--- a/packages/cli-debugger-ui/src/ui/index.html
+++ b/packages/cli-debugger-ui/src/ui/index.html
@@ -18,13 +18,23 @@
         <p>
           Remote JavaScript debugging (this workflow) is
           <strong>deprecated</strong> in React Native 0.73 and will be removed
-          in React Native 0.74.
+          in a future release.
         </p>
         <p>
           Please use the
-          <strong>Open Debugger</strong> workflow to debug apps using Hermes
-          directly. This can be accessed from the Dev Menu or by pressing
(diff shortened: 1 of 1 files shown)


liamjones:
@huntie I just saw this in passing from a notification; shouldn't this include some info about what your options are for the JSC as well as Hermes? Or is that irrelevant to this notice?

huntie:
@liamjones That's a fair point and we can do — the only supported JSC debugging flow will be direct JSC debugging via Safari (iOS/Apple platforms only).

szymonrybczak (approved):
👍

packages/cli-debugger-ui/src/ui/index.html
```diff
@@ -23,16 +23,18 @@
         <p>
           Please use the
           <strong>Open Debugger</strong> workflow to debug apps using Hermes
```
huntie, line 25: Note to self: link will be `http://reactnative.dev/docs/next/debugging#opening-the-debugger`.

adamTrz (approved):

PR #2133, opened 2023-10-26
fix: fallback first to `yarn` when `bun` not available

szymonrybczak:
Summary:
---------

When bun is not available, check if yarn is available in fist place. 

Checklist
----------

### DIFF
--- a/packages/cli/src/tools/packageManager.ts
+++ b/packages/cli/src/tools/packageManager.ts
@@ -49,8 +49,10 @@ function configurePackageManager(
   if (options.packageManager === 'bun') {
     if (bunAvailable) {
       pm = 'bun';
+    } else if (yarnAvailable) {
+      pm = 'yarn';
     } else {
-      pm = yarnAvailable ? 'yarn' : 'npm';
+      pm = 'npm';
     }
   }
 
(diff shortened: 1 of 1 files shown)


TMisiukiewicz (approved):

packages/cli/src/tools/packageManager.ts
```diff
@@ -41,9 +41,21 @@ function configurePackageManager(
...
+  let pm: PackageManager = 'npm';
+
   if (options.packageManager === 'bun') {
-    pm = shouldUseBun(options) ? 'bun' : 'npm';
+    if (bunAvailable) {
+      pm = 'bun';
+    } else {
+      pm = yarnAvailable ? 'yarn' : 'npm';
+    }
+  }
```
thymikee, line 55: I somehow feel better about not mixing if/else with ternary
```suggestion
    if (bunAvailable) {
      pm = 'bun';
    } else if (yarnAvailable) {
      pm = 'yarn';
    } else {
      pm = 'npm';
    }
  }
```
switch would make it even more readable maybe. Your call :)

thymikee (approved):
Feel free to accept or ignore my suggestion and merge :)

PR #2134, opened 2023-10-26
feat: use Yarn v3 with `nodeLinker: node-modules` for new projects

szymonrybczak:
Summary:
---------

Closes https://github.com/react-native-community/cli/pull/1931 

When creating new project we'll try to bump Yarn version to v3. We'll only do this if:
- the project is initialised in fresh environment (not in existing Git repo),
- and the Yarn version in the location is Yarn Classic

Test Plan:
----------

2. Run this command:

```sh
```
3. Should use Yarn v3 🎉

Checklist
----------

### DIFF
--- a/jest/helpers.ts
+++ b/jest/helpers.ts
@@ -121,20 +121,14 @@ function getExecaOptions(options: SpawnOptions) {
 
   const cwd = isRelative ? path.resolve(__dirname, options.cwd) : options.cwd;
 
-  let env = Object.assign({}, process.env, {FORCE_COLOR: '0'});
+  let env = Object.assign({}, process.env, {FORCE_COLOR: '0'}, options.env);
 
   if (options.nodeOptions) {
     env.NODE_OPTIONS = options.nodeOptions;
   }
+
   if (options.nodePath) {
     env.NODE_PATH = options.nodePath;
-  }
-
-  if (options.env) {
-    env = {
-      ...env,
-      ...options.env,
-    };
   }
 
   return {
--- a/jest/helpers.ts
+++ b/jest/helpers.ts
@@ -28,6 +28,9 @@ export function runCLI(
   return spawnScript(process.execPath, [CLI_PATH, ...(args || [])], {
     ...options,
     cwd: dir,
+    env: {
+      YARN_ENABLE_IMMUTABLE_INSTALLS: 'false',
+    },
   });
 }
 
@@ -92,6 +95,7 @@ export const getTempDirectory = (name: string) =>
 
 type SpawnOptions = RunOptions & {
   cwd: string;
+  env?: {[key: string]: string | undefined};
 };
 
 type SpawnFunction<T> = (
@@ -117,13 +121,20 @@ function getExecaOptions(options: SpawnOptions) {
 
   const cwd = isRelative ? path.resolve(__dirname, options.cwd) : options.cwd;
 
-  const env = Object.assign({}, process.env, {FORCE_COLOR: '0'});
+  let env = Object.assign({}, process.env, {FORCE_COLOR: '0'});
 
   if (options.nodeOptions) {
     env.NODE_OPTIONS = options.nodeOptions;
   }
   if (options.nodePath) {
     env.NODE_PATH = options.nodePath;
+  }
+
+  if (options.env) {
+    env = {
+      ...env,
+      ...options.env,
+    };
   }
 
   return {
--- a/packages/cli/src/commands/init/init.ts
+++ b/packages/cli/src/commands/init/init.ts
@@ -29,6 +29,8 @@ import {getNpmVersionIfAvailable} from '../../tools/npm';
 import {getYarnVersionIfAvailable} from '../../tools/yarn';
 import {createHash} from 'crypto';
 import createGitRepository from './createGitRepository';
+import semver from 'semver';
+import {executeCommand} from '../../tools/executeCommand';
 
 const DEFAULT_VERSION = 'latest';
 
@@ -63,6 +65,30 @@ interface TemplateOptions {
 interface TemplateReturnType {
   didInstallPods?: boolean;
 }
+
+const YARN_VERSION = '3.6.4';
+
+const bumpYarnVersion = async (silent: boolean, root: string) => {
+  try {
+    let yarnVersion = semver.parse(getYarnVersionIfAvailable());
+    if (yarnVersion && semver.major(yarnVersion) === 1) {
+      await executeCommand('yarn', ['set', 'version', YARN_VERSION], {
+        root,
+        silent,
+      });
+
+      // React Native doesn't support PnP, so we need to set nodeLinker to node-modules. Read more here: https://github.com/react-native-community/cli/issues/27#issuecomment-1772626767
+
+      await executeCommand(
+        'yarn',
+        ['config', 'set', 'nodeLinker', 'node-modules'],
+        {root, silent},
+      );
+    }
+  } catch (e) {
+    logger.debug(e as string);
+  }
+};
 
 function doesDirectoryExist(dir: string) {
   return fs.existsSync(dir);
@@ -182,6 +208,11 @@ async function createFromTemplate({
       packageName,
     });
 
+    if (packageManager === 'yarn') {
+      await bumpYarnVersion(false, projectDirectory);
+    }
+
+    loader.succeed();
     const {postInitScript} = templateConfig;
     if (postInitScript) {
       loader.info('Executing post init script ');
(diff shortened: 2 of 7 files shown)


szymonrybczak:
@tido64 @satya164 I would really appreciate your review  🙏

szymonrybczak:
One note here is that Yarn v4 was released few days ago, and it is working with React Native properly. With this change we're aiming for 0.74 so until that time it will be tested well in various production environments. Also Yarn v4 had 53 releases so we can assume that team tested it well 😅 So let''s wait for some minor fixes that probably will be released, and then will finally bump to Yarn v4 in CLI - with current setup it is just changing number.

packages/cli/src/tools/packageManager.ts
```diff
@@ -18,7 +17,7 @@ const packageManagers = {
     install: ['add'],
     installDev: ['add', '-D'],
     uninstall: ['remove'],
-    installAll: ['install'],
+    installAll: ['install', '--no-immutable'],
```
szymonrybczak, line 20: I know that this line shouldn't be here but I was just testing if CI will install packages when the `--no-immutable` flag, and it will. (Right now it is failing randomly because of some flakiness regarding Android tests). Do you have maybe some ideas how we should workarounds this, because we can't simply pass `--skip-install` when initialising project in CI, because when creating project even with `--skip-install` we install packages in `/tmp` directory get the template from `node_modules`. We need to figure out to somehow pass this flag when installing packages while creating project.

tido64: Does `YARN_ENABLE_IMMUTABLE_INSTALLS=false` work?

image

tido64 (approved):
LGTM

packages/cli/src/tools/packageManager.ts
szymonrybczak: Ah, I missed this. Thanks Tommy! 🙌

jest/helpers.ts
```diff
@@ -180,6 +184,13 @@ function getExecaOptions(options: SpawnOptions) {
     env.NODE_PATH = options.nodePath;
   }
 
+  if (options.env) {
+    env = {
+      ...env,
+      ...options.env,
+    };
+  }
```
robhogan, line 192: Could this be incorporated into the `Object.assign` above? (And perhaps convert that into an object spread)

packages/cli/src/commands/init/init.ts
```diff
@@ -56,6 +58,30 @@ interface TemplateOptions {
   installCocoaPods?: string | boolean;
 }
 
+const YARN_VERSION = '3.6.4';
```
robhogan, line 61: Do we need to pin this to a specific version?

packages/cli/src/commands/init/init.ts
```diff
@@ -56,6 +58,30 @@ interface TemplateOptions {
...
+
+const bumpYarnVersion = async (silent: boolean, root: string) => {
+  try {
+    let yarnVersion = semver.parse(getYarnVersionIfAvailable());
+    if (yarnVersion && semver.major(yarnVersion) === 1) {
+      await executeCommand('yarn', ['set', 'version', YARN_VERSION], {
+        root,
+        silent,
+      });
```
robhogan, line 70: Just to check my understanding here - this will install the requested version via corepack if it isn't currently on the machine? What if corepack isn't enabled?

robhogan (commented):
Couple of questions/suggestions but LGTM in principle

packages/cli/src/commands/init/init.ts
szymonrybczak: Hm, for me to use `3.x` I need to run first `yarn set version berry` and then I can do `yarn set version 3.x`, I don't know why it happens.

satya164: Maybe lets add a comment here as context for future readers

packages/cli/src/commands/init/init.ts
```diff
@@ -58,6 +60,30 @@ interface TemplateOptions {
...
+
+      // React Native doesn't support PnP, so we need to set nodeLinker to node-modules. Read more here: https://github.com/react-native-community/cli/issues/27#issuecomment-1772626767
+
+      await executeCommand(
+        'yarn',
+        ['config', 'set', 'nodeLinker', 'node-modules'],
```
satya164, line 78: Another config option we may want to set by default is `nmHoistingLimits: 'workspaces'`

This is not necessary in regular projects - but maybe necessary if the project is converted to a monorepo later.

packages/cli/src/commands/init/init.ts
```diff
@@ -58,6 +60,30 @@ interface TemplateOptions {
   installCocoaPods?: string | boolean;
 }
 
+const YARN_VERSION = '3.6.4';
+
+const bumpYarnVersion = async (silent: boolean, root: string) => {
+  try {
+    let yarnVersion = semver.parse(getYarnVersionIfAvailable());
+    if (yarnVersion && semver.major(yarnVersion) === 1) {
```
satya164, line 68: Are we checking global yarn version before deciding if we call `yarn set version`? If yes I'd discourage it because the global version is specific to environment. If the project is run on CI or another dev with a different environmnent then it'll result in different behaviour.

e.g. someone has yarn v3 installed globally, creates the project and pushes to GH. then I pull it and run it, but since I have yarn v1 installed globally, it tries to use Yarn v1 and breaks.

To ensure consistent behaviour, we should call this unconditionally.

satya164 (changes requested):

packages/cli/src/commands/init/init.ts
tido64: If we're calling it unconditionally, we should first make sure we are not inside a monorepo. We absolutely cannot break existing setups.

satya164: > If we're calling it unconditionally, we should first make sure we are not inside a monorepo

Sure, tho I don't think the current condition checks anything regarding monorepos so that's a separate thing. With the current check, I could be using a monorepo with Yarn v1 and this will break as well.

szymonrybczak: > it tries to use Yarn v1 and breaks.

Not exactly, because the pulled project has already `.yarnrc.yml` with specified Yarn version so it will use that this one.

> we should first make sure we are not inside a monorepo.

https://github.com/react-native-community/cli/pull/2134/commits/63bec9da3d51e28dc6c3be002a4b70e042edc91a I added logic for checking if we're under git work tree, if yes we're skipping bumping Yarn version.

packages/cli/src/commands/init/init.ts
szymonrybczak: ✅

packages/cli/src/commands/init/init.ts
satya164: > Not exactly, because the pulled project has already .yarnrc.yml with specified Yarn version so it will use that this one.

But how will the Yarn version be there if it's added conditionally? If I have Yarn 3 installed globally then aren't you skipping `yarn set version`? It's only called if the global Yarn version is 1.

szymonrybczak: Ah okay, I didn't get sense of initial comment. Fixed in https://github.com/react-native-community/cli/pull/2134/commits/bc8aaaa134769e2cc970b016bb567e3725122c7b.

jest/helpers.ts
szymonrybczak: done, good catch!

satya164 (approved):

noway:
Sorry if this is not the right forum or if this is too much of a qualitative judgement, but I'd like to share my thoughts.

Yarn Modern is less popular than Yarn Classic. While Yarn Classic is deprecated (and to be fair, subtly broken in different ways), Yarn Modern hasn't shown enough resiliency and it's long term success is far from guaranteed. 

Forcing new projects to start with Yarn v3 may send them on a path where they may end up with an unmaintained package manager few years down the track.

Is it a prudent decision to switch new projects to Yarn Modern, especially for users who are just starting out with React Native? NPM would be a much sure choice.

satya164:
> especially for users who are just starting out with React Native? NPM would be a much sure choice

npm has its issues too, with people not being able to install packages due to misconfigured peer dependencies and being confused due to a lot of not useful vulnerability warnings.

> Forcing new projects to start with Yarn v3 may send them on a path where they may end up with an unmaintained package manager few years down the track

If Yarn ends up being unmaintained, is it that difficult to switch to a different package manager? At least for non monorepos it should be a fairly trivial task.

PR #2144, opened 2023-10-31
perf: remove Gradle task check on run-android

TMisiukiewicz:
Summary:
---------
Related: #2012 
When running `run-android` command, it takes a lot of time to build the app for more complex projects with lots of dependencies and different flavors. It is caused by the code checking if the task for provided mode exists. Under the hood, it's running `./gradlew tasks` which takes a lot of time. A good example of such complex project is [Expensify App](https://github.com/Expensify/App), where delay in building the app takes ~4mins. 

Here is a screenshot from profiling report of `./gradlew tasks --group install` from Expensify :
image

This PR removes this check - Gradlew itself throws an error if the task for provided mode does not exist so we can rely on that:
```

* What went wrong:
Cannot locate tasks that match 'app:installAaa' as task 'installAaa' not found in project ':app'.
```

Additionally, getting task list should still exist in the interactive mode. Unfortunately in more complex projects, it might stil take a lot of time to execute. A note about it was added to the docs. As a DX improvement, I added the loader indicating that the process is currently looking for Gradle tasks.

Test Plan:
----------
1. Run `node path/to/cli run-android`
2. Verify after starting Metro and simulator it moved to building the app immediately
3. Run `node path/to/cli run-android --mode Aaa`
4. Verify it throws an error from Gradle
5. Run `node path/to/cli run-android --interactive`
6. Verify it shows loader "Searching for available gradle tasks..." before prompting to select the task

Checklist
----------

### DIFF
--- a/packages/cli-platform-android/src/commands/runAndroid/listAndroidTasks.ts
+++ b/packages/cli-platform-android/src/commands/runAndroid/listAndroidTasks.ts
@@ -34,12 +34,16 @@ export const getGradleTasks = (
   const loader = getLoader();
   loader.start('Searching for available Gradle tasks...');
   const cmd = process.platform.startsWith('win') ? 'gradlew.bat' : './gradlew';
-
-  const out = execa.sync(cmd, ['tasks', '--group', taskType], {
-    cwd: sourceDir,
-  }).stdout;
-  loader.succeed();
-  return parseTasksFromGradleFile(taskType, out);
+  try {
+    const out = execa.sync(cmd, ['tasks', '--group', taskType], {
+      cwd: sourceDir,
+    }).stdout;
+    loader.succeed();
+    return parseTasksFromGradleFile(taskType, out);
+  } catch {
+    loader.fail();
+    return [];
+  }
 };
 
 export const promptForTaskSelection = async (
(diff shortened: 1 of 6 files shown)


adamTrz (approved):

packages/cli-platform-android/src/commands/runAndroid/listAndroidTasks.ts
```diff
@@ -31,11 +31,14 @@ export const getGradleTasks = (
...
   sourceDir: string,
 ) => {
+  const loader = getLoader();
+  loader.start('Searching for available Gradle tasks...');
   const cmd = process.platform.startsWith('win') ? 'gradlew.bat' : './gradlew';
 
   const out = execa.sync(cmd, ['tasks', '--group', taskType], {
     cwd: sourceDir,
   }).stdout;
+  loader.succeed();
```
thymikee, line 41: `execa.sync` can throw, can we catch the error, logger.fail and re-throw or handle error gracefully?

szymonrybczak:
We should backport this to 0.72 cc. @thymikee

thymikee:
@szymonrybczak here: https://github.com/react-native-community/cli/pull/2145

PR #2148, opened 2023-11-01
fix(e2e): correctly link packages

szymonrybczak:
Summary:
---------

In this PR I changed to link all packages before running testes. Recently there was introduced change to `init` which creates `react-native.config.js` with some flag. But in tests we were only linking `cli-platform-android` and `cli-platform-ios` so creating project was failing since it installs Pods, and during installing Pods we call `use_native_modules` which executes `config` command that was outdated. And it was giving error: 
CleanShot 2023-11-01 at 13 01 54

Test Plan:
----------

CI Green 

Checklist
----------

### DIFF
--- a/__e2e__/root.test.ts
+++ b/__e2e__/root.test.ts
@@ -6,13 +6,17 @@ import {
   getTempDirectory,
   cleanup,
   writeFiles,
+  addRNCPrefix,
+  getAllPackages,
 } from '../jest/helpers';
 
 const cwd = getTempDirectory('test_different_roots');
 
 beforeAll(() => {
+  const packages = getAllPackages();
+
   // Register all packages to be linked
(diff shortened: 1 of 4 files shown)


__e2e__/root.test.ts
```diff
@@ -23,14 +27,11 @@ beforeAll(() => {
   writeFiles(cwd, {});
 
   // Initialise React Native project
-  runCLI(cwd, ['init', 'TestProject', '--install-pods']);
+  runCLI(cwd, ['init', 'TestProject']);
```
szymonrybczak, line 30: Here we don't need to install pods, since this test check only android related things.

szymonrybczak:
We should probably also properly delete cache before linking packages. Linked packages are stored under `~/.config/yarn/link/@react-native-community/`

thymikee (approved):
nice one!

PR #2151, opened 2023-11-01
chore: bump node to 20

szymonrybczak:
Summary:
---------

Bumps node version to 20, to unblock https://github.com/react-native-community/cli/pull/2015. 

**It doesn't mean that to run CLI user needs Node 20**:
CleanShot 2023-11-01 at 13 40 24

Minimal Node version is stored here:

https://github.com/react-native-community/cli/blob/3f206fd01f70706ce17e4a33c3528a60e684d41e/packages/cli-doctor/src/tools/versionRanges.ts#L3

Test Plan:
----------

CI

Checklist
----------

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -10,7 +10,7 @@ module.exports = {
     [
       require.resolve('@babel/preset-env'),
       {
-        targets: {node: '18'},
+        targets: {node: '20'},
         useBuiltIns: 'entry',
         corejs: '2.x',
       },
--- a/package.json
+++ b/package.json
@@ -32,8 +32,9 @@
     "@react-native-community/eslint-config": "^3.2.0",
     "@types/glob": "^7.1.1",
     "@types/jest": "^26.0.15",
-    "@types/node": "^18.0.0",
-    "@types/node-fetch": "^2.3.7",
+    "@types/node": "^20.8.10",
+    "@types/node-fetch": "^2.6.8",
+    "@types/readable-stream": "^4.0.4",
     "babel-jest": "^26.6.2",
     "babel-plugin-module-resolver": "^3.2.0",
     "chalk": "^4.1.2",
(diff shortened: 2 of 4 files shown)


szymonrybczak:
Bumping node version will finally unblock https://github.com/react-native-community/cli/pull/1983 :tada:

thymikee:
rebase needed

thymikee (approved):

adamTrz (approved):

szymonrybczak:
CI Green, good to merge!

babel.config.js
```diff
@@ -10,7 +10,7 @@ module.exports = {
     [
       require.resolve('@babel/preset-env'),
       {
-        targets: {node: '18'},
+        targets: {node: '20'},
```
thymikee, line 13: Will RN 0.74 drop node 18? 0.73 targets node 18 as minimum so we need to keep this as 18, otherwise it’s a breaking change

szymonrybczak: No, please look at the PR description.

robhogan: If the build step targets Node 20, running on Node 18 would be unsafe (even though it might work), so this *looks* like a breaking change to me?

szymonrybczak: Ah, make sense 👍

robhogan: Ah, looks like `babel.config.js` only used for local development, and `tsc` (targeting ES2017) is used for npm builds? If that's the case this particular change isn't user facing.

*But* the bump of `@types/node` would allow the use of runtime features that aren't supported by Node 18. What's the motivation?

szymonrybczak: The initial goal of this PR was to solve problem with CI failures on Node v20, but we solved by upgrading some dependencies, so this bump isn't necessary atm.

PR #2157, opened 2023-11-04
fix(ci): setup global git user

szymonrybczak:
Summary:
---------

Sometimes initialising git repository fails, because machine doesn't have git installed or configured correctly so when running E2E tests we need to take this into account and dynamically assign files that we're expecting.  

Test Plan:
----------

CI

Checklist
----------

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -80,8 +80,8 @@ jobs:
 
       - name: Setup Git user
         run: |
-          git config --global user.name ""
-          git config --global user.email ""
+          git config --global user.name "test"
+          git config --global user.email "test@test.com"
           
       - name: Install dependencies
         run: yarn --frozen-lockfile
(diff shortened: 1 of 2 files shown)


thymikee:
Let's set the user in CI config, should allow us to not modify the test code

szymonrybczak:
> Let's set the user in CI config, should allow us to not modify the test code

Done, CI Green ✅

.github/workflows/test.yml
```diff
@@ -47,6 +47,9 @@ jobs:
         node-version: [18, 20]
         os: [ubuntu-latest, macos-latest, windows-2019]
     runs-on: ${{ matrix.os }}
+    defaults:
+      run:
+        shell: bash
```
szymonrybczak, line 52: It needs to be specified, since by default it launches Powershell on Windows.

PR #2161, opened 2023-11-06
Clean command does not clean metro-cache on win32

acoates-ms:
Summary:
---------

The fs.rm function on win32 does not expand *'s, so the code to delete the metro cache using fs.rm('<tempdir>/metro-*') does not delete anything on win32.

Checklist
----------

### DIFF
--- a/packages/cli-clean/package.json
+++ b/packages/cli-clean/package.json
@@ -10,7 +10,8 @@
   "dependencies": {
     "@react-native-community/cli-tools": "12.0.0",
     "chalk": "^4.1.2",
-    "execa": "^5.0.0"
+    "execa": "^5.0.0",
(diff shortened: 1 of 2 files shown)


thymikee:
Hm, `fs.rmSync` on macOS also doesn't expand (I wouldn't expect that tbh), so this doesn't seem to be Windows related, and it looks like it doesn't work really for these `-*` cases

TMisiukiewicz (approved):

TMisiukiewicz:
> Hm, `fs.rmSync` on macOS also doesn't expand (I wouldn't expect that tbh), so this doesn't seem to be Windows related, and it looks like it doesn't work really for these `-*` cases

I tried it and seems you are right, I'd say we could skip the os check here

PR #2162, opened 2023-11-07
feat(ios): cache last used device 

szymonrybczak:
Summary:
---------

Developers often have a lot of simulators installed on their machines, and if they are running `--list-devices` or `--interactive` they need to go through all of them, but by adding simple caching we'll move last used device to the top of the list. 

Uploading CleanShot 2023-11-07 at 17.06.00.mp4…

Test Plan:
----------

2. Run this command:

```sh
node /path/to/react-native-cli/packages/cli/build/bin.js run-ios --list-devices
```
3. Choose one devices
```sh
node /path/to/react-native-cli/packages/cli/build/bin.js run-ios --list-devices
```
5. Chosen device from third step should be on the top of the list.

Checklist
----------

### DIFF
--- a/packages/cli-platform-ios/src/tools/prompts.ts
+++ b/packages/cli-platform-ios/src/tools/prompts.ts
@@ -35,14 +35,27 @@ export async function promptForConfigurationSelection(
 }
 
 export async function promptForDeviceSelection(
-  availableDevices: Device[],
+  devices: Device[],
+  lastUsedDeviceId?: string,
 ): Promise<Device | undefined> {
+  const sortedDevices = devices;
+  const devicesIds = sortedDevices.map(({udid}) => udid);
+
+  if (lastUsedDeviceId) {
+    const preferredDeviceIndex = devicesIds.indexOf(lastUsedDeviceId);
+
+    if (preferredDeviceIndex > -1) {
+      const [preferredDevice] = sortedDevices.splice(preferredDeviceIndex, 1);
+      sortedDevices.unshift(preferredDevice);
+    }
+  }
+
   const {device} = await prompt({
     type: 'select',
     name: 'device',
     message: 'Select the device you want to use',
-    choices: availableDevices
-      .filter((d) => d.type === 'device' || d.type === 'simulator')
+    choices: sortedDevices
+      .filter(({type}) => type === 'device' || type === 'simulator')
       .map((d) => {
         const version = d.version
           ? ` (${d.version.match(/^(\d+\.\d+)/)?.[1]})`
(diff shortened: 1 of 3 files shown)


packages/cli-tools/src/cacheManager.ts
```diff
@@ -4,7 +4,12 @@ import os from 'os';
 import appDirs from 'appdirsjs';
 import logger from './logger';
 
-type CacheKey = 'eTag' | 'lastChecked' | 'latestVersion' | 'dependencies';
+type CacheKey =
+  | 'eTag'
+  | 'lastChecked'
+  | 'latestVersion'
+  | 'dependencies'
+  | 'lastUsedDeviceId';
```
TMisiukiewicz, line 12: I think it would be good to add the same feature for Android in the future - with that in mind, can we name the cache key accordingly to the plaftorm?

szymonrybczak: ✅

packages/cli-platform-ios/src/tools/prompts.ts
```diff
@@ -35,14 +35,27 @@ export async function promptForConfigurationSelection(
 }
 
 export async function promptForDeviceSelection(
-  availableDevices: Device[],
+  devices: Device[],
+  lastUsedIOSDeviceId?: string,
 ): Promise<Device | undefined> {
+  const sortedDevices = devices;
```
thymikee, line 41: `sortedDevices` still references `devices` and is then mutated. You'd need to copy the `devices` array to avoid mutating it with `unshift` on line 49

szymonrybczak, line 41:
```suggestion
  const sortedDevices = [...devices];
```

packages/cli-platform-ios/src/tools/prompts.ts
```diff
@@ -35,14 +35,27 @@ export async function promptForConfigurationSelection(
 }
 
 export async function promptForDeviceSelection(
-  availableDevices: Device[],
+  devices: Device[],
+  lastUsedIOSDeviceId?: string,
 ): Promise<Device | undefined> {
+  const sortedDevices = [...devices];
```
cipolleschi, line 41: why are these devices sorted? I don't think they are...

cipolleschi (approved):
Just a nit, but overall looks good to me.

packages/cli-platform-ios/src/tools/prompts.ts
szymonrybczak: Not really sorted here, but I needed to create a value to do not mutate params.

PR #2170, opened 2023-11-17
feat: add platform-name option to init for out of tree platforms init

okwasniewski:
Summary:
---------

This PR adds `platform-name` option to allow for out of tree platforms to leverage `init` command without (almost) any changes in their repo. 

### Problem

When initializing project (for OOT platform) with command similar to this: `@callstack/react-native-visionos init MyApp` current approach would be to pass `--template` option to specify custom template for this platform. However this requires to maintain separate template (as RN-TvOS is doing: https://www.npmjs.com/package/react-native-template-typescript-tv).

Also CLI has a hardcoded path to retrieve template URI (in `createTemplateUri`). 

### Solution

Allow OOT platforms to pass it's name to `cli.run()` function call which doesn't require to pass any extra flags by user. 

Inside of `[OutOfTreePlatform]/packages/react-native/cli.js` we pass out of tree platform name from the package.json

```diff
+ const {name} = require('./package.json');

async function main() {
  if (isNpxRuntime && !process.env.SKIP && currentVersion !== HEAD) {
    try {
      const latest = await getLatestVersion();
      if (latest !== currentVersion) {
        const msg = `
  ${chalk.bold.yellow('WARNING:')} You should run ${chalk.white.bold(
    'npx react-native@latest',
  )} to ensure you're always using the most current version of the CLI. NPX has cached version (${chalk.bold.yellow(
    currentVersion,
  )}) != current release (${chalk.bold.green(latest)})
  `;
        console.warn(msg);
      }
    } catch (_) {
      // Ignore errors, since it's a nice to have warning
    }
  }
+  return cli.run(name);
}
```

As an extension of this we could later on detect if we are inside of a project and if so just add new platform to this project (instead of initializing new one)

Test Plan:
----------

- Check if it works 
- Add tests (after the idea is validated)

Checklist
----------

### DIFF
--- a/packages/cli/src/commands/init/index.ts
+++ b/packages/cli/src/commands/init/index.ts
@@ -47,5 +47,10 @@ export default {
       description:
         'Inits a project with a custom package name (Android) and bundle ID (iOS), e.g. com.example.app',
     },
+    {
+      name: '--platform-name <string>',
+      description:
+        'Name of out of tree platform to be used for ex. react-native-macos. This flag is optional as it should be passed automatically by out of tree platform.',
+    },
   ],
 };
--- a/packages/cli/src/commands/init/index.ts
+++ b/packages/cli/src/commands/init/index.ts
@@ -50,7 +50,7 @@ export default {
     {
       name: '--platform-name <string>',
       description:
-        'Name of out of tree platform to be used for ex. react-native-macos. This flag is optional as it should be passed automatically by out of tree platform.',
+        'Name of out of tree platform to be used for ex. react-native-macos. This flag is optional as it should be passed automatically by out of tree platform. It needs to match the name of the platform declared in package.json',
     },
   ],
 };
--- a/__e2e__/init.test.ts
+++ b/__e2e__/init.test.ts
@@ -168,3 +168,29 @@ test('init uses npm as the package manager with --npm', () => {
     expect(fs.existsSync(path.join(initDirPath, file))).toBe(true);
   });
 });
+
+test('init --platform-name should work for out of tree platform', () => {
+  createCustomTemplateFiles();
+  const outOfTreePlatformName = 'react-native-macos';
+
+  const {stdout} = runCLI(DIR, [
+    'init',
+    PROJECT_NAME,
+    '--platform-name',
+    outOfTreePlatformName,
+    '--skip-install',
+    '--verbose',
+  ]);
+
+  expect(stdout).toContain('Run instructions');
+  expect(stdout).toContain(
+    `Installing template from ${outOfTreePlatformName}@latest`,
+  );
+
+  // make sure we don't leave garbage
+  expect(fs.readdirSync(DIR)).toContain('custom');
+
+  let dirFiles = fs.readdirSync(path.join(DIR, PROJECT_NAME));
+
+  expect(dirFiles.length).toBeGreaterThan(0);
+});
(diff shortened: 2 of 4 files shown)


packages/cli/src/commands/init/template.ts
```diff
@@ -82,9 +82,13 @@ export async function copyTemplate(
 
   logger.debug(`Copying template from ${templatePath}`);
   let regexStr = path.resolve(templatePath, 'node_modules');
-  await copyFiles(templatePath, process.cwd(), {
-    exclude: [new RegExp(replacePathSepForRegex(regexStr))],
-  });
+  await copyFiles(
+    '/Users/okwasniewski/workspace/react-native-visionos/packages/react-native/template',
```
szymonrybczak, line 86: 🫣

okwasniewski: Let's pretend you didn't see that

packages/cli/src/commands/init/index.ts
```diff
@@ -47,5 +47,10 @@ export default {
       description:
         'Inits a project with a custom package name (Android) and bundle ID (iOS), e.g. com.example.app',
     },
+    {
+      name: '--platform-name <string>',
+      description:
+        'Name of out of tree platform to be used for ex. react-native-macos. This flag is optional as it should be passed automatically by out of tree platform.',
+    },
```
szymonrybczak, line 54: Let's add this flag to docs.

szymonrybczak (approved):
LGTM

packages/cli/src/commands/init/index.ts
TMisiukiewicz: I think it's worth to be specific in the docs that `platform-name` needs to match the fork name, to avoid trying to use it with e.g. `visionos`

szymonrybczak: +1, we should mention that it should be equal to the `name` property specified in `package.json`

thymikee (approved):
Ok, let's do it. This is explicit enough and doesn't require us to make any assumptions about the platform name based on binary name or some node module resolution.

One extra thing I would love to see would be to extend the `#### Initializing project with custom template` section of `docs/init.md` with an example or instruction on what to do if you're an OOT platform

thymikee:
Thinking about it in the context of your example, if we ported this small diff to the core, we wouldn't need to document it anywhere outside of the code comments

okwasniewski:
@thymikee, @szymonrybczak, @TMisiukiewicz  Thanks for review! I've updated the command description and added tests. Let's keep this as an undocumented flag - I will create PR to core after this will be released 👍🏻 That way normal users won't need to pass it

packages/cli/src/commands/init/index.ts
```diff
@@ -47,5 +47,10 @@ export default {
       description:
         'Inits a project with a custom package name (Android) and bundle ID (iOS), e.g. com.example.app',
     },
+    {
+      name: '--platform-name <string>',
+      description:
+        'Name of out of tree platform to be used for ex. react-native-macos. This flag is optional as it should be passed automatically by out of tree platform. It needs to match the name of the platform declared in package.json',
```
NickGerleman, line 53: FYI @saadnajmi

Note that macOS and Windows both have separate init packages right now (with the windows one in particular growing quite complicated). I worry naming react-native-macOS here might be jumping ahead a bit, until that init flow has changed (unless it already has).

thymikee: The init flow stays the same. While working on a new platform (visionos) we want to make things right and easy for any other platform to follow. Today this particular flag is quite virtual, as it's passed from the programmatic CLI usage in core. It's still in flux and once we're more firm on the design, we'll likely contribute some PoC to macOS and hopefully Windows too.

PR #2171, opened 2023-11-18
fix: platform-name option causing regressions for other commands

okwasniewski:
Summary:
---------

This PR fixes a regression caused by #2170. Now we pass the `platform-name` **only** to init command. Which was the source of error: `Unkown command `--platform-name` as other commands didn't expect this option passed.

I've added a test case to check for this special case

Test Plan:
----------

Green CI

Checklist
----------


szymonrybczak (approved):
LGTM, don't release on Fridays 😃

thymikee (approved):

PR #2174, opened 2023-11-23
fix: check for main manifest if multiple are present

WoLewicki:
Summary:
---------

This PR fixes a newly introduced (https://github.com/react-native-community/cli/pull/1967) problem that having multiple build flavors in app causes the `findManifest` script to take the first found one. We think it seems more reasonable to take the one from `main` folder, if present. It fixes the error in `Expensify` app, where `e2e` manifest is the first found one, and the `MainActivity` is present in the `main` folder's `AndroidManifest.xml`. It throws such error then: 
```
error Main activity not found in /Users/.../android/app/src/e2e/AndroidManifest.xml.
```

Test Plan:
----------

Make a simple app with `activity` tag inside `main` `AndroidManifest.xml` and add a directory e.g. `e2e` with other `AndroidManifest.xml`, which should have the `activity` merged from the `main`. It will cause the error on gradle sync in RN 0.73.x

Checklist
----------

### DIFF
--- a/packages/cli-platform-android/src/config/findManifest.ts
+++ b/packages/cli-platform-android/src/config/findManifest.ts
@@ -10,7 +10,7 @@ import glob from 'glob';
 import path from 'path';
 
 export default function findManifest(folder: string) {
-  const manifestPath = glob.sync(path.join('**', 'AndroidManifest.xml'), {
+  let manifestPaths = glob.sync(path.join('**', 'AndroidManifest.xml'), {
     cwd: folder,
     ignore: [
       'node_modules/**',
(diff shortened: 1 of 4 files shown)


WoLewicki:
Gradle has a mechanism of merging the manifests, so if we could somehow reuse it, then it would most probably be a more general fix for weird package structures.

cortinico (approved):

cortinico (approved):

szymonrybczak (approved):
Thank you! 🙌 Mind adding tests? 🙏

j-piasecki:
I've added a test case resembling a setup that's the reason for this PR. LMK if it's enough or if you want to cover other scenarios as well.

packages/cli-platform-android/src/config/findManifest.ts
```diff
@@ -23,7 +23,16 @@ export default function findManifest(folder: string) {
...
     ],
-  })[0];
+  });
+  if (manifestPaths.length > 1) {
+    // if we have more than one manifest, pick the one in the main folder if present
+    const mainManifest = manifestPaths.filter((manifestPath) =>
+      manifestPath.includes('src/main/'),
+    );
+    if (mainManifest.length === 1) {
+      manifestPaths = mainManifest;
```
thymikee, line 33: nit: how about not mutating `manifestPaths` and extracting the `mainManifest` to a variable to be used later? If we want to make sure only one main manifest is picked, we could use `.find` instead of `.filter`.

thymikee: Ok, let's stick to the current behavior. Unit tests job on Node 20 is flaky, try not to bother it :D

PR #2198, opened 2023-12-08
fix: packager starting when using run-ios and run-android

janicduplessis:
Summary:
---------

When the packager is not running `findDevServerPort` returns `startPackager = false` which is not what we want. Changing the initial value of `startPackager` fixes it (we could also add a branch to check for not_running state but it seemed redundant).

Test Plan:
----------

Tested this patch in an app on RN 0.73 and made sure packager starts when running cli on iOS and Android.

Checklist
----------

### DIFF
--- a/packages/cli-tools/src/findDevServerPort.ts
+++ b/packages/cli-tools/src/findDevServerPort.ts
@@ -10,7 +10,7 @@ const findDevServerPort = async (
   startPackager: boolean;
 }> => {
   let port = initialPort;
-  let startPackager = false;
+  let startPackager = true;
 
   const packagerStatus = await isPackagerRunning(port);
 


robhogan (approved):

robhogan:
I was just about to put up the same fix before I saw this PR. Users are [running into this](https://www.reddit.com/r/reactnative/comments/18eaq0z/comment/kcmeddu/) now that 0.73 is out.

The fix looks sound to me. The [old code](https://github.com/react-native-community/cli/pull/2128/files#diff-7a3b71db0e6ae79e26e5bff1ffe632f8dc3343046a9d2edee7bfcc93d5ff4da8L61) used to mutate the configured value of `packager`.  After https://github.com/react-native-community/cli/pull/2128, `findDevServerPort` is only called if `packager` is `true` anyway, so falling back to `true` matches the previous behaviour.

Let's get this merged and picked into 13.x. CC @thymikee.

cortinico (approved):

szymonrybczak (approved):

szymonrybczak:
Tests failing because major RN version was bumped, fixed in #2200

RayHughes:
Does not appear to be fixed in https://github.com/facebook/react-native/blob/main/CHANGELOG.md#v0732

szymonrybczak:
@RayHughes, will be in 0.73.3, sorry for trouble!

szymonrybczak:
@RayHughes, it should be. 0.73.2 contains CLI v12.3.0, but this fix was released in v12.2.1 🤔

RayHughes:
> @RayHughes, it should be. 0.73.2 contains CLI v12.3.0, but this fix was released in v12.2.1 🤔


I'm on MacOS. This is what I get in native terminal and Jetbrains. 

Screenshot 2024-01-08 at 5 01 20 PM

szymonrybczak:
@RayHughes as the error says, we cannot find terminal to start script in. please try providing `--terminal` flag. Btw the pull request that you're referring to is not related to the error that you're encountering.

PR #2204, opened 2023-12-13
feat: add manual Cocoapods installation info

okwasniewski:
Summary:
---------

This PR aims to: 
1. Add manual cocapods info when installation fails or user chooses "No" when prompted to install
2. Fix CLI exiting the process when cocoapods installation goes wrong

CleanShot 2023-12-13 at 13 02 36@2x

Test Plan:
----------

CI Green

Checklist
----------

### DIFF
--- a/__e2e__/init.test.ts
+++ b/__e2e__/init.test.ts
@@ -193,3 +193,16 @@ test('init --platform-name should work for out of tree platform', () => {
 
   expect(dirFiles.length).toBeGreaterThan(0);
 });
+
+test('init should contain Cocoapods instructions when pods are not installed automatically', () => {
+  createCustomTemplateFiles();
+
+  const {stdout} = runCLI(DIR, ['init', PROJECT_NAME, '--skip-install']);
+  expect(stdout).toContain('Install Cocoapods');
+
+  // make sure we don't leave garbage
+  expect(fs.readdirSync(DIR)).toContain('custom');
+
+  let dirFiles = fs.readdirSync(path.join(DIR, PROJECT_NAME));
+  expect(dirFiles.length).toBeGreaterThan(0);
+});
--- a/__e2e__/init.test.ts
+++ b/__e2e__/init.test.ts
@@ -198,7 +198,9 @@ test('init should contain Cocoapods instructions when pods are not installed aut
   createCustomTemplateFiles();
 
   const {stdout} = runCLI(DIR, ['init', PROJECT_NAME, '--skip-install']);
-  expect(stdout).toContain('Install Cocoapods');
+  if (process.platform === 'darwin') {
+    expect(stdout).toContain('Install Cocoapods');
+  }
 
   // make sure we don't leave garbage
   expect(fs.readdirSync(DIR)).toContain('custom');
--- a/__e2e__/init.test.ts
+++ b/__e2e__/init.test.ts
@@ -119,6 +119,10 @@ test('init skips installation of dependencies with --skip-install', () => {
 
   expect(stdout).toContain('Run instructions');
 
+  if (process.platform === 'darwin') {
+    expect(stdout).toContain('Install Cocoapods');
+  }
+
   // make sure we don't leave garbage
   expect(fs.readdirSync(DIR)).toContain('custom');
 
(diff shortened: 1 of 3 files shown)


packages/cli/src/commands/init/init.ts
```diff
@@ -219,11 +234,15 @@ async function createFromTemplate({
       loader.succeed('Dependencies installation skipped');
     }
   } catch (e) {
+    if (e instanceof Error) {
+      logger.error(e.message);
+    }
     loader.fail();
-    throw e;
```
okwasniewski, line 223: This fixes CLI exiting when cocoapods installation fails

packages/cli/src/commands/init/init.ts
```diff
@@ -207,6 +212,16 @@ async function createFromTemplate({
               'Only needed if you run your project in Xcode directly',
             )}`,
           });
+          automaticPodsInstallation = installCocoapods;
+          logger.log('\n');
+          logger.info(
+            `To enable automatic CocoaPods installation when building for iOS you can create react-native.config.js with automaticPodsInstallation field. \n${chalk.reset.dim(
```
szymonrybczak, line 218:
```suggestion
            `💡To enable automatic CocoaPods installation when building for iOS you can create react-native.config.js with automaticPodsInstallation field. \n${chalk.reset.dim(
```

packages/cli/src/commands/init/init.ts
okwasniewski: Maybe something like: 

Installing pods failed. This doesn't affect project initialization and you can safely proceed. However, you will need to install pods manually when running iOS, follow additional steps in "Run instructions for iOS" section. 

To avoid duplicating the steps

thymikee: 👍🏼

packages/cli/src/commands/init/init.ts
```diff
@@ -112,10 +116,10 @@ async function createFromTemplate({
   skipInstall,
   packageName,
   installCocoaPods,
-}: TemplateOptions) {
+}: TemplateOptions): Promise<TemplateReturnType> {
   logger.debug('Initializing new project');
   logger.log(banner);
-
+  let automaticPodsInstallation = String(installCocoaPods) === 'true';
```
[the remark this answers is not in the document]
okwasniewski: I think they have a slightly different purpose and using `installPodsValue` for this made the code more complex.

packages/cli/src/commands/init/init.ts
```diff
@@ -348,13 +367,20 @@ export default (async function initialize(
     return;
   }
 
-  await createProject(projectName, directoryName, version, options);
+  const {automaticPodsInstallation} = await createProject(
```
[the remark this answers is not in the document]
okwasniewski: I've changed the naming to `didInstallPods`, I think it better suits what this actually does. WDYT?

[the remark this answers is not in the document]
TMisiukiewicz: looks great, thanks 👍

__e2e__/init.test.ts
```diff
@@ -193,3 +193,17 @@ test('init --platform-name should work for out of tree platform', () => {
...
 });
+
+test('init should contain Cocoapods instructions when pods are not installed automatically', () => {
+  createCustomTemplateFiles();
+
+  const {stdout} = runCLI(DIR, [
+    'init',
+    PROJECT_NAME,
+    '--skip-install',
+    '--install-pods',
```
TMisiukiewicz, line 204: i think `--install-pods` is not necessary here as `--skip-install` won't install it anyway

__e2e__/init.test.ts
```diff
@@ -193,3 +193,17 @@ test('init --platform-name should work for out of tree platform', () => {
...
+    'false',
+  ]);
+  console.log(stdout);
```
TMisiukiewicz, line 207: and unnecessary console.log here

__e2e__/init.test.ts
```diff
@@ -198,7 +198,9 @@ test('init should contain Cocoapods instructions when pods are not installed aut
   createCustomTemplateFiles();
 
   const {stdout} = runCLI(DIR, ['init', PROJECT_NAME, '--skip-install']);
-  expect(stdout).toContain('Install Cocoapods');
+  if (process.platform === 'darwin') {
```
szymonrybczak, line 201: I would make this whole test under a condition, because if we don't match condition we're just checking the same thing as in this test:
https://github.com/react-native-community/cli/blob/e1438f714407a9d590d5975ecd3c846b8d91ab21/__e2e__/init.test.ts#L109-L132

okwasniewski: Maybe let's add this check to the test you referenced?

thymikee (approved):

packages/cli/src/commands/init/init.ts
```diff
@@ -196,6 +203,7 @@ async function createFromTemplate({
         const installPodsValue = String(installCocoaPods);
```
cipolleschi, line 203: you already have this variable in `didInstallPods` (line125)

cipolleschi (approved):
Looks good to me. 
I only left a couple of nits

packages/cli/src/commands/init/init.ts
okwasniewski: As discussed here: https://github.com/react-native-community/cli/pull/2204#discussion_r1425289922 

`didInstallPods` is used for a slightly different purpose (also to check if the user didn't specify his choice) and mixing those two made the code more complex.

Issue #2217, opened 2023-12-20
☂️ `cli-platform-apple` - further improvements 


okwasniewski:
## Descripiton

`cli-platform-apple` is a new package introduced for OOT platforms. This issue lists further improvements that can be done to improve the codebase.

- [x] Remove `getFallbackSimulator()` and use first available simulator
- [x] Extract duplicated logic between `createRun` and `createLog` commands: https://github.com/react-native-community/cli/pull/2208#discussion_r1431019199
- [x] Implement app launching for React Native macOS: [PR](https://github.com/react-native-community/cli/pull/2232)
- [x] Move prompt from `createLog` to `prompts.ts`: [PR](https://github.com/react-native-community/cli/pull/2224)
- [x] Create a getter for `buildOptions` / `runOptions` to include proper platformName - [PR](https://github.com/react-native-community/cli/pull/2226)
- [x] Modify warning `Multiple Podfiles were found` to take OOT platforms into account [PR](https://github.com/react-native-community/cli/pulls)
- [x] Make dependencyConfig into a builder - [PR](https://github.com/react-native-community/cli/pull/2226)
- [x] Remove code duplication between `createRun` and `createBuild`. Logic to re-check Xcode project can be moved to `getXcodeProjectAndDir` behind a flag #2220 
- [x] merge `buildOptions` into `runOptions` to avoid confusion when importing/extending

matthargett:
I think the last item in the list can be check off now, right? Are there any additional/new items we want to add here based on @Saadnajmi observations so far on the macOS side?

okwasniewski:
Thanks @matthargett, I checked off the last item from the list 👍🏻 Can we close this @szymonrybczak or is there something more you want to add under this umbrella issue?

PR #2224, opened 2023-12-20
chore: move devices prompt to `prompts.ts`

szymonrybczak:
Summary:
---------

Related https://github.com/react-native-community/cli/issues/2217

Test Plan:
----------

2. Start few simulators
3. Run this command:

```sh
node /path/to/react-native-cli/packages/cli/build/bin.js log-ios --interactive
```
Works as before.

### DIFF
--- a/packages/cli-platform-apple/src/tools/prompts.ts
+++ b/packages/cli-platform-apple/src/tools/prompts.ts
@@ -2,6 +2,10 @@ import chalk from 'chalk';
 import {Device} from '../types';
 import {prompt} from '@react-native-community/cli-tools';
 
+function getVersionFromDevice({version}: Device) {
+  return version ? ` (${version.match(/^(\d+\.\d+)/)?.[1]})` : '';
+}
+
 export async function promptForSchemeSelection(
   schemes: string[],
 ): Promise<string> {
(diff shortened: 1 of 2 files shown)


packages/cli-platform-apple/src/tools/prompts.ts
```diff
@@ -76,3 +76,20 @@ export async function promptForDeviceSelection(
...
+export async function promptForDeviceToTailLogs(
+  platformReadableName: string,
+  simulators: Device[],
+): Promise<string> {
+  const {udid} = await prompt({
+    type: 'select',
+    name: 'udid',
+    message: `Select ${platformReadableName} simulators to tail logs from`,
+    choices: simulators.map((simulator) => ({
+      title: simulator.name,
```
okwasniewski, line 89: Would it make sense to show simulator versions in this prompt, like: `iPhone 14 (17.0)`?

szymonrybczak: good call, added!

TMisiukiewicz (approved):

packages/cli-platform-apple/src/tools/prompts.ts
```diff
@@ -85,10 +85,16 @@ export async function promptForDeviceToTailLogs(
     type: 'select',
     name: 'udid',
     message: `Select ${platformReadableName} simulators to tail logs from`,
-    choices: simulators.map((simulator) => ({
-      title: simulator.name,
-      value: simulator.udid,
-    })),
+    choices: simulators.map((simulator) => {
+      const version = simulator.version
+        ? ` (${simulator.version.match(/^(\d+\.\d+)/)?.[1]})`
```
okwasniewski, line 90: One more nit: maybe let's have a reusable utility to get formattedSimulatorName or something like this so we can reuse this regex and logic in other places and prompts

okwasniewski (approved):

PR #2233, opened 2023-12-22
feat: fallback to OOT scheme

szymonrybczak:
Summary:
---------

When building app for macOS a scheme looks like this `AppName-macOS`. Inside `run-macos` implemention in `react-native-macos` repo [it's hardcoded assumption](https://github.com/microsoft/react-native-macos/blob/164469a0c424b8a6710b9f9b0d6f761ba289cec4/packages/react-native/local-cli/runMacOS/runMacOS.js#L61). I think we can also support this, so if the default one is not accessible, let's give a warning and fallback to one with the suffix. 

Test Plan:
----------

2. Follow steps described [here](https://github.com/react-native-community/cli/blob/main/packages/cli-platform-apple/README.md) to add OOT platform e.g. inside `react-native-macos`
3. Run this command inside app that is using OOT platform:

```sh
node /path/to/react-native-cli/packages/cli/build/bin.js run-macos
```
4. Fallback scheme should be selected, so in this case `ProjectName-macOS`

Checklist
----------

### DIFF
--- a/packages/cli-platform-apple/src/commands/buildCommand/getConfiguration.ts
+++ b/packages/cli-platform-apple/src/commands/buildCommand/getConfiguration.ts
@@ -28,7 +28,7 @@ export async function getConfiguration(
 
   if (!info?.schemes?.includes(scheme)) {
     const {readableName} = getPlatformInfo(platformName);
-    const fallbackScheme = `${scheme}-${readableName}`
+    const fallbackScheme = `${scheme}-${readableName}`;
 
     if (info?.schemes?.includes(fallbackScheme)) {
       logger.warn(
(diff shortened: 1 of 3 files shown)


packages/cli-platform-apple/src/commands/buildCommand/getConfiguration.ts
```diff
@@ -22,6 +25,24 @@ export async function getConfiguration(
   let scheme =
     args.scheme ||
     path.basename(xcodeProject.name, path.extname(xcodeProject.name));
+
+  if (!info?.schemes?.includes(scheme)) {
+    const {readableName} = getPlatformInfo(platformName);
+    const fallbackScheme = scheme + '-' + readableName;
```
okwasniewski, line 31: nit: 
```suggestion
    const fallbackScheme = `${scheme}-${readableName}`
```

okwasniewski (approved):

thymikee (approved):

PR #2239, opened 2024-01-02
refactor: adapt `getSimulators()` for OOT platforms

okwasniewski:
Summary:
---------

This PR removes `getSimulators()` from createRun and adopts them for createLog. I've also added a handful info message from where we are tailing logs from: 

CleanShot 2024-01-03 at 10 58 50@2x

Test Plan:
----------

Check if returned devices are the same.

Checklist
----------

### DIFF
--- a/packages/cli-platform-apple/src/commands/logCommand/createLog.ts
+++ b/packages/cli-platform-apple/src/commands/logCommand/createLog.ts
@@ -4,9 +4,9 @@ import {spawnSync} from 'child_process';
 import os from 'os';
 import path from 'path';
 import getSimulators from '../../tools/getSimulators';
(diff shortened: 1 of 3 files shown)


cipolleschi (approved):
I like that the code is much simpler now!

okwasniewski:
As pointed out by @szymonrybczak (in private message) we can't remove `getSimulators()` for createLog as it contains the necessary info stated in the comment: 

> // Here we're using two command because first command `xcrun simctl list --json devices` outputs `state` but doesn't return `available`. But second command `xcrun xcdevice list` outputs `available` but doesn't output `state`. So we need to connect outputs of both commands.

szymonrybczak (approved):

PR #2250, opened 2024-01-08
docs: projects.md to include type information

atomicpages:
Summary:
---------

While upgrading from 0.67 to 0.7x finding `react-native.config` type information was not obvious. Adding annotations so autocomplete works on most editors.

Test Plan:
----------

Just docs

Checklist
----------


szymonrybczak (approved):
Thanks! That's nice improvement, maybe we could do it everywhere in our docs, I see that there are other places :D

PR #2251, opened 2024-01-08
fix(cli-platform-apple): properly receive simulators

szymonrybczak:
Summary:
---------

Fixes regression introduced in https://github.com/react-native-community/cli/pull/2239, where usage of `getSimulators` was replaced with `listDevices`: 
CleanShot 2024-01-08 at 21 36 15
all simulators were marked as booted. 

In this Pull Request I also removed usage `getSimulators` and unified downloading simulators across CLI into one place (we were doing this in two separate function, and then we were duplicating `.filter`, and `.map` over the codebase). 

Test Plan:
----------

2. Run this command:

```sh
node /path/to/react-native-cli/packages/cli/build/bin.js run-ios 
```
3. Run this command:
```sh
node /path/to/react-native-cli/packages/cli/build/bin.js log-ios
```

Checklist
----------


szymonrybczak:
Idk why units tests are failing, passing locally: 
CleanShot 2024-01-08 at 22 09 25

packages/cli-platform-apple/src/tools/listDevices.ts
```diff
@@ -49,9 +50,49 @@ const parseXcdeviceList = (text: string, sdkNames: string[] = []): Device[] => {
...
+    }
+  });
+
+  parsedSimctlOutput.forEach((item) => {
+    if (!matchedUdids.has(item.udid)) {
+      merged.push({...item});
+    }
+  });
+
+  return merged.filter(({isAvailable}) => isAvailable === true);
```
cipolleschi, line 95: I don't understand: why do we need to run both `xcdevice` and `simctl` commands? Shouldn't they provide the same output?

szymonrybczak: So `xcdevice` provides physical devices, but `simctl` only simulators. Actually I checked, and `simctl` is giving informations about simulator's availability, I don't why we had wrong assumptions earlier. I'll update the comment with current state of things.

packages/cli-platform-apple/src/tools/__tests__/listDevices.test.ts
```diff
@@ -161,9 +180,40 @@ const xcrunOut = `
 ]
 `;
 
+const xcrunSimctlOut = `
+{
+  "devices" : {
+    "com.apple.CoreSimulator.SimRuntime.iOS-16-2" : [
+      {
+        "lastBootedAt" : "2023-05-09T11:08:32Z",
+        "dataPath" : "/Users/szymonrybczak/Library/Developer/CoreSimulator/Devices/54B1D3DE-A943-4867-BA6A-B82BFE3A7904/data",
```
thymikee, line 189: Let's make these paths not exposing your directory structure :) 
```suggestion
        "dataPath" : "<REPLACED_ROOT>/Library/Developer/CoreSimulator/Devices/54B1D3DE-A943-4867-BA6A-B82BFE3A7904/data",
```

packages/cli-platform-apple/src/tools/listDevices.ts
```diff
@@ -49,9 +49,45 @@ const parseXcdeviceList = (text: string, sdkNames: string[] = []): Device[] => {
...
+ */
 async function listDevices(sdkNames: string[]): Promise<Device[]> {
-  const out = execa.sync('xcrun', ['xcdevice', 'list']).stdout;
-  return parseXcdeviceList(out, sdkNames);
+  const xcdeviceOutput = execa.sync('xcrun', ['xcdevice', 'list']).stdout;
+  const parsedXcdeviceOutput = parseXcdeviceList(xcdeviceOutput, sdkNames);
+
+  const simctlOutput = JSON.parse(
+    execa.sync('xcrun', ['simctl', 'list', '--json', 'devices']).stdout,
+  );
```
thymikee, line 63: can you measure whether running these 2 commands in parallel is faster or not?

szymonrybczak: So, I gave it a shot, and used `performance.now()` to measure 5 times execution time in parallel and in sequential run, here are the result:

| Parallel run | Sequential run |
|--------|--------|
| 5601.046166986227 | 5644.935749977827 | 
| 5649.5136660039425 | 5632.401457995176 | 
| 5634.3378329873085 | 5684.918291002512 | 
| 5606.314500004053 | 5615.477707982063 | 
| 5674.328749984503 | 5632.90391600132 | 
| AVG: **5633.108183** | AVG: **5642.127425** |

(Results in milliseconds.)

The difference is really slow, so I think we can leave it as it is 👍

swrobel:
Could this be backported to 12.x so it can ship with RN 0.73.3?

PR #2256, opened 2024-01-10
ci: run tests on every branch

szymonrybczak:
Summary:
---------

In situations like https://github.com/react-native-community/cli/pull/2255, it would be nice to run tests too, to check if everything is fine. In this PR I also included running tests on branches which are following this syntax: `'[0-9]+.x'` (that's how we're naming release branches).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,7 +2,7 @@ name: Node.js CI
 
 on:
   push:
-    branches: [main]
+    branches: [main, '[0-9]+.x']
   pull_request:
     branches: ['**']
 


.github/workflows/test.yml
```diff
@@ -1,15 +1,14 @@
...
-    branches: ['**']
+on: [push, pull_request]
 
 jobs:
   lint:
     name: Lint
     runs-on: ubuntu-latest
+    if: >
+      github.event_name == 'push' ||
+      (github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name != github.repository)
```
szymonrybczak, line 11: I'm wondering if it is possible to add this condition for the whole file 🤔

szymonrybczak:
We should backport this to `12.x`, `11.x`, `10.x` - all the versions that are currently supported, according to this [table](https://github.com/reactwg/react-native-releases?tab=readme-ov-file#which-versions-are-currently-supported).

thymikee:
Backported

PR #2258, opened 2024-01-12
Fix: cleaning metro cache

TMisiukiewicz:
Summary:
---------
`fs.rm` function does not expand, so when using `clean` command, Metro files are not removed.

This is a follow-up PR to #2161 which solves this problem for Windows, but since `fs.rm` does not expand at all, it still doesn't work properly for macOS. Removing OS check in this solution does not solve the issue, probably because `convertPathToPattern` from `fast-glob` modifies the pattern from `metro-*` to `metro-\*`. This PR should work correctly on both operating systems.

Closes #2161

Test Plan:
----------
1. Find your OS temp directory (`os.tmpdir()`)
2. Use `start` command in any CLI app
3. `ls` in the temp directory to verify any `metro-*` are listed
4. Use `clean` command, select `metro` option and confirm
5. `ls` in the temp directory again and verify all `metro-*` are gone

Checklist
----------

### DIFF
--- a/packages/cli-clean/src/clean.ts
+++ b/packages/cli-clean/src/clean.ts
@@ -1,4 +1,4 @@
-import {getLoader, prompt} from '@react-native-community/cli-tools';
+import {getLoader, logger, prompt} from '@react-native-community/cli-tools';
 import type {Config as CLIConfig} from '@react-native-community/cli-types';
 import chalk from 'chalk';
 import execa from 'execa';
@@ -6,6 +6,7 @@ import {existsSync as fileExists, rm} from 'fs';
 import os from 'os';
 import path from 'path';
 import {promisify} from 'util';
+import glob from 'glob';
 
 type Args = {
   include?: string;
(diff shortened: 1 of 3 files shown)


szymonrybczak (approved):
Works great on two platforms, LGTM 🙌

packages/cli-clean/src/clean.ts
```diff
@@ -28,13 +29,37 @@ type CleanGroups = {
...
+  try {
+    if (isDirectoryPattern(directory)) {
+      const directories = await new Promise<string[]>((resolve, reject) => {
+        glob(directory, {}, (err, foundDirectories) => {
+          if (err) {
+            reject(err);
+          } else {
+            resolve(foundDirectories);
+          }
+        });
```
thymikee, line 48: doesn't glob have its promise interface?

thymikee (approved):

packages/cli-clean/src/clean.ts
TMisiukiewicz: it does, but starting from 9.0, and we are currently using 7.x across the project 😢

thymikee: Let's upgrade it in a separate PR :D

PR #2276, opened 2024-01-23
feat: add `link-assets` command 

fabioh8010:
Summary:
---------

This PR aims to implement a feature to help users link their assets in their projects in an automated and efficient way. This PR will basically bring functionality from [react-native-asset](https://github.com/unimonkiez/react-native-asset) package + [this new feature to link Android assets using XML Fonts](https://github.com/unimonkiez/react-native-asset/pull/52). Due to the lack of maintenance in the original repo, we've published a new version under CK organisation [here](https://www.npmjs.com/package/@callstack/react-native-asset), but after some discussions we decided to bring the feature to the CLI.

Relevant links/discussions in chronological order:

1. [Original issue in Expensify repo](https://github.com/Expensify/App/issues/13772)
2. [CK proposal to implement the new feature in react-native-asset library](https://github.com/Expensify/App/issues/13772#issuecomment-1487394471)
3. [PR with the new feature](https://github.com/unimonkiez/react-native-asset/pull/52)
4. [Issue about lack of maintenance and offer to help](https://github.com/unimonkiez/react-native-asset/issues/57)
5. [Decision to create a separate package under CK org, and proposal to move the feature to CLI](https://github.com/Expensify/App/issues/13772#issuecomment-1825545287)
6. [Approval to move the feature to CLI](https://github.com/Expensify/App/issues/13772#issuecomment-1839435348)

Tasks

- [x] Implementation of initial integration and commands
- [x] Migration and refactoring of original codebase to TypeScript
- [x] Support to handle Kotlin files
- [x] Unit tests
- [x] Final documentation

Test Plan:
----------

TODO

Checklist
----------


packages/cli-link-assets/CHANGELOG.md
```diff
@@ -0,0 +1,8 @@
+# Change Log
+
+All notable changes to this project will be documented in this file.
+See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
```
szymonrybczak, line 4: nit: you can link CLI's [CONTRIBUTING.md](https://github.com/react-native-community/cli/blob/main/CONTRIBUTING.md) guide

TMisiukiewicz:
Started code review and left some comments, gonna continue reviewing tomorrow :)

packages/cli-link-assets/src/linkAssets.ts
```diff
@@ -0,0 +1,227 @@
...
+  name: 'link-assets',
+  description: 'TODO',
+  options: [
+    {
+      name: '--project-root <string>',
+      description:
+        'Root path to your React Native project. When not specified, defaults to current working directory.',
+      default: process.cwd(),
+    },
+  ],
```
szymonrybczak, line 226: I don't think this is necessary, project root is stored inside `Config`.

packages/cli-link-assets/src/tools/cleanAssets/android.ts
```diff
@@ -0,0 +1,129 @@
...
+  getXMLFontId,
+  normalizeString,
+  removeLineFromJavaFile,
+  toArrayBuffer,
+  xmlBuilder,
+  xmlParser,
+} from '../helpers/font/androidFontAssetHelpers';
+import {AndroidCleanAssetsOptions, CleanAssets} from './types';
+
+const cleanAssets: CleanAssets = (assetFiles, options) => {
```
TMisiukiewicz, line 21: we could possibly extract some parts of this function and share it with `copyAssets`

packages/cli-link-assets/src/linkAssets.ts
```diff
@@ -0,0 +1,227 @@
...
+    const fileName = files[i];
+    const ext = path.extname(fileName);
+
+    if (ext === '.pbxproj') {
+      return fileName;
+    }
+  }
+
+  return null;
+}
```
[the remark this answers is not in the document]
fabioh8010: `findXcodeProject` doesn't give me the `.pbxproj` file, so we would have to adjust it to return it together with the `IOSProjectInfo` object. WDYT?

packages/cli-link-assets/src/linkAssets.ts
```diff
@@ -0,0 +1,227 @@
...
+    androidAssetsPath = androidAssetsPath.concat(ctx.project.android.assets);
+    androidAppName = ctx.project.android.appName;
+  }
+
+  let iosPath: string = '';
+  let iosAssetsPath: string[] = ctx.assets;
+  let iosPbxprojFilePath: string | null = null;
+  if (ctx.project.ios) {
+    iosPath = ctx.project.ios.sourceDir;
+    iosAssetsPath = androidAssetsPath.concat(ctx.project.ios.assets);
```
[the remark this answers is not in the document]
fabioh8010: good catch, will adjust

packages/cli-link-assets/src/linkAssets.ts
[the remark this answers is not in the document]
TMisiukiewicz: or maybe let's add `path` to return it from `findXcodeProject` and then create a function that takes `IOSProjectInfo` as param and just concatenate path, filename and `.pbxproj`? IIRC `.pbxproj` is a required file so this should do the work. Then this function could live in `cli-platform-apple`

TMisiukiewicz:
Just updated the code to handle Kotlin templates, tested it with freshly built app on 0.73 and seems to work well!

image

packages/cli-link-assets/src/linkAssets.ts
fabioh8010: In the original repo, these were the available options:

```
-p, --path - path to project, defaults to cwd.
-a, --assets - assets paths, for example react-native-asset -a ./src/font ./src/mp3.
-ios-a, --ios-assets - ios assets paths, will disable android linking
-android-a, --android-assets - android assets paths, will disable ios linking.
-n-u, --no-unlink - Not to unlink assets which not longer exists, not recommanded.
```

So maybe we don't need any parameters at all? All of them can be configured in `react-native-config.js` now, I kept `--project-root` to be a replacement for `--path`, but I agree with you that we could remove it too. 

I also removed `--no-unlink`, not sure if it's something really used.

WDYT @szymonrybczak @TMisiukiewicz ?

packages/cli-link-assets/CHANGELOG.md
fabioh8010: Could you elaborate more @szymonrybczak ? I followed the way we've done for other files, e.g [here](https://github.com/react-native-community/cli/blob/main/packages/cli-clean/CHANGELOG.md) and [here](https://github.com/react-native-community/cli/blob/main/packages/cli-config/CHANGELOG.md).

packages/cli-link-assets/src/linkAssets.ts
szymonrybczak: Yeah, let's keep everything in `react-native-config.js`.

packages/cli-link-assets/CHANGELOG.md
szymonrybczak: Ah, I didn't know that we have CHANGELOGs in this repository 😅 Let's leave it as it is then 👍

packages/cli-link-assets/src/linkAssets.ts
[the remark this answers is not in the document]
fabioh8010: Not sure if I understood correctly, I tried this approach for example:

```ts
import {IOSProjectInfo} from '@react-native-community/cli-types';
import path from 'path';

function findPbxprojFile(projectInfo: IOSProjectInfo): string {
  return path.join(projectInfo.path, projectInfo.name, 'project.pbxproj');
}

export default findPbxprojFile;
```

In `findXcodeProject` I'm returning a new `path` property that is basically `path.dirname(fileName);`:

```ts
function findXcodeProject(files: Array<string>): IOSProjectInfo | null {
  const sortedFiles = files.sort();

  for (let i = sortedFiles.length - 1; i >= 0; i--) {
    const fileName = files[i];
    const ext = path.extname(fileName);
    const projectPath = path.dirname(fileName);

    if (ext === '.xcworkspace') {
      return {
        name: fileName,
        path: projectPath,
        isWorkspace: true,
      };
    }
    if (ext === '.xcodeproj') {
      return {
        name: fileName,
        path: projectPath,
        isWorkspace: false,
      };
    }
  }

  return null;
}
```

The issue is, `project.pbxproj` is always under `.xcodeproj` folder, so if I concatenate `path`, `name` and `project.pbxproj` and `name` is e.g. `AwesomeApp.xcworkspace` because we do have a workspace, the logic won't work well.

packages/cli-link-assets/src/tools/cleanAssets/android.ts
fabioh8010: Will do it after finishing the tests, so it will be safer to refactor 🙂

packages/cli-link-assets/src/linkAssets.ts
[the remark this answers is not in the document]
TMisiukiewicz: after internal discussion, let's just replace `.xcworkspace` in `findPbxprojFile` `projectInfo.name`. They are always on the same level so it should be safe

thymikee:
Awesome! @TMisiukiewicz @fabioh8010 is it now ready for getting out of the draft state?

fabioh8010:
@thymikee I will push one more last test today, add documentation and that should be good for opening to review!

thymikee:
Cool, take your time :)

TMisiukiewicz:
did another round of review and tested it on both platforms, looks solid! 👍 some small conflicts need to be resolved before opening

fabioh8010:
@thymikee @TMisiukiewicz I've added the final test and documentation. There are still some improvements in the tests @TMisiukiewicz suggested and I will do, but we are good enough here open the PR while I do the improvements.

packages/cli-link-assets/CHANGELOG.md
fabioh8010, line 1: @thymikee @TMisiukiewicz @szymonrybczak Do you some ideas what I can add to this file? Actually, by looking the `CHANGELOG.md` of other commands it looks to me that they are currently unused and outdated. Should we keep this file here?

thymikee: We don't have changelog files, rather use the Releases tab. Please remove it :)

TMisiukiewicz (commented):
just two more minor comments 👍

fabioh8010:
@TMisiukiewicz @thymikee I've addressed the review comments, simplified some tests, refactored code and improved logic to handle android resource file names

TMisiukiewicz:
@fabioh8010 when running `link-assets` without any linked assets and not having `assets` property in `react-native.config.js` file, there is no output from a command. Can we display a short instruction what to do to in order to use this command in this case?

fabioh8010:
Thanks @TMisiukiewicz I will make these adjustments!

packages/cli-link-assets/src/tools/manifest/index.ts
```diff
@@ -0,0 +1,62 @@
...
+    data,
+    {
+      spaces: 2,
+    },
+  );
+}
+
+const getManifest = (
+  folderPath: string,
+  platform: 'android' | 'ios',
```
szymonrybczak, line 39: nit: I see that `'android' | 'ios'` few times over the changes, maybe we could move this into separate type?

packages/cli-link-assets/src/linkAssets.ts
```diff
@@ -0,0 +1,225 @@
...
+        path: iosPath,
+        pbxprojFilePath: iosPbxprojFilePath!,
+      },
+      cleanAssets: cleanIOSAssets,
+      copyAssets: copyIOSAssets,
+      linkOptionsPerExt: {
+        otf: linkOptionsPerExt?.otf?.ios,
+        ttf: linkOptionsPerExt?.ttf?.ios,
+        mp3: linkOptionsPerExt?.mp3?.ios,
+      },
```
szymonrybczak, line 204: Q: Why are we not specifying image assets here?

fabioh8010: Good catch @szymonrybczak , in practice it doesn't have difference because we will use `customLinkOptions` instead, but I will add the remaining extensions just to be consistent with Android 👍

TMisiukiewicz (approved):

packages/cli/package.json
```diff
@@ -28,6 +28,7 @@
     "@react-native-community/cli-config": "14.0.0-alpha.0",
     "@react-native-community/cli-debugger-ui": "14.0.0-alpha.0",
     "@react-native-community/cli-doctor": "14.0.0-alpha.0",
+    "@react-native-community/cli-link-assets": "14.0.0-alpha.0",
```
thymikee, line 31: Let's not add this command by default for now to the main endpoint

packages/cli/src/commands/index.ts
```diff
@@ -2,12 +2,14 @@ import {Command, DetachedCommand} from '@react-native-community/cli-types';
...
 import {commands as doctorCommands} from '@react-native-community/cli-doctor';
 import {commands as configCommands} from '@react-native-community/cli-config';
+import {commands as linkAssetsCommands} from '@react-native-community/cli-link-assets';
 import init from './init';
 
 export const projectCommands = [
   ...configCommands,
   cleanCommands.clean,
   doctorCommands.info,
+  linkAssetsCommands.linkAssets,
```
thymikee, line 12: remove for now

thymikee:
Ok I got a look at this PR and I have doubts exposing it to users. This commands works in a legacy way of manually changing Xcode and Gradle files – even asking users to do so. I hoped it would leverage the autolinking mechanism so that DX is nicer.

Please take a look at how rnx-kit team achieves that: https://github.com/microsoft/react-native-test-app/pull/1828 in less than 200 lines of code.

If the time is pressure, I'm ok with merging that into the repository, but I'm not ok exposing this command to every user. It will add a lot of size to the package that's already pretty big. Therefore I'd suggest to document how one could use this package (`npx @react-native-community/cli-link-assets` instead of `npx react-native link-assets`) and iterate on a better solution that's aligned with autolinking and doesn't require developers to fiddle with native files.

thymikee (approved):

thymikee:
Waiting for CI to be green and gonna merge it. Thanks so much for porting this code @fabioh8010, it's a great amount of effort to keep this library afloat and should be a good ground to iterate for a better, autolink-enabled solution in the future 👍🏼

fabioh8010:
@thymikee Could you run the E2E workflow tests again? Thanks!

thymikee:
why do I need to approve the pipeline twice, geez GitHub cmon

thymikee:
The CI passed 🎉

PR #2280, opened 2024-01-27
perf: replace glob with fast-glob

abejfehr:
Summary:
---------

Closes https://github.com/react-native-community/cli/pull/2271

I was investigating how `pod install` was so slow in React Native, and it seems like a lot of time is spent in `IO#gets`: 

image

Upon further investigation, this is actually Ruby waiting for the output of `node .../node_modules/@react-native-community/cli/build/bin.js config`.

When investigating what's slow about `node .../node_modules/@react-native-community/cli/build/bin.js config`, I noticed that a lot of time was spent in globs (about ~17 seconds on an M1 Mac):

image

In my testing, I've observed that `glob` can be replaced with `fast-glob` to make performance considerably better with the same output, so I'd like to propose that `glob` be replaced with `fast-glob` in this project.

Test Plan:
----------

The unit tests are great, but one thing I'm encountering is that the FS mock doesn't seem to play well with something that fast-glob is doing internally.

Even though it works in practice, it doesn't seem to work in tests 😢 it's like `fs.__setMockFilesystem` isn't taking effect or something (it is once, but never subsequently. It's like fast-glob doesn't re-scan the disk? idk)

Any help here would be greatly appreciated.

Checklist
----------

- [ ] ~Documentation is up to date to reflect these changes.~


liamjones:
> The unit tests are great, but one thing I'm encountering is that the FS mock doesn't seem to play well with something that fast-glob is doing internally.
> 
> Even though it works in practice, it doesn't seem to work in tests 😢 it's like fs.__setMockFilesystem isn't taking effect or something (it is once, but never subsequently. It's like fast-glob doesn't re-scan the disk? idk)
> 
> Any help here would be greatly appreciated.

There was a previous PR for this which went stale which mentioned that same issue - there might be a solution in there already? https://github.com/react-native-community/cli/pull/2016

abejfehr:
> hey @abejfehr! thanks for contribution, could you please rebase on top of `main`, I've deleted `update-metro.js` file which is no longer anymore used. Also as @liamjones said you can take a look at my attempt on replacing `glob` with `fast-glob` - where I fixed units tests. I see that Windows E2E test is failing there, logs expired so I can't check if this was related, but anyway it may be worth checking 👍

Thanks! I've rebased.

Looking at the prior art helped me resolve the FS mocking issue and it passes tests now locally, so I think it's ready for a review now

szymonrybczak:
@abejfehr thanks! looks like Windows E2E is failing now, same as in my PR 🙈 Could you please look at this? I'll also try to search for solution for this.

abejfehr:
I ran the tests on a Windows machine and noticed that it wasn't building properly, and it looks like that's because I replaced glob with fast-glob in `linkPackages.js` and `build.ts` as well, but there it doesn't do `unixifyPaths` and it wouldn't be available yet for use there anyways.

I opted for adding `glob`/`@types/glob` as devDependencies for the whole workspace instead and using them just for the build/link script.

Now on Windows for me there's ~3 failing tests but they seem to be due to very specific paths and the ones in `main` fail for the same reason too, so I'm hoping they pass in CI 🤞

szymonrybczak (approved):
Thank you @abejfehr! 🙏 CI Green ✅

PR #2286, opened 2024-02-03
refactor: improved type safety of platformConfig in createRun function

hoonjoo-park:
Summary:
---------

I improved type safety of `platformConfig` in `createRun` function. and removed unnecessary optional chaining of the `platformConfig`.
(code is located in cli-platform-apple/src/commands/runCommand/createRun.ts)

before I modified type of `platformConfig`, we couldn't infer that platformConfig could be `undefined`.
so, I added `undefined` type to `platformConfig` as optional.

Additionally, there was unnecessary optional chaining, so I removed it.

``` typescript
// * already checked if platformConfig is 'undefined'.
if (
  platformConfig === undefined ||
  supportedPlatforms[platformName] === undefined
) {
  throw new CLIError(
    `Unable to find ${platformReadableName} platform config`,
  );
}

//  * as far as I see, we don't need optional chaining in here
if (platformConfig?.automaticPodsInstallation || args.forcePods) {
  const isAppRunningNewArchitecture = platformConfig?.sourceDir
    ? await getArchitecture(platformConfig.sourceDir)
    : undefined;

  await resolvePods(ctx.root, ctx.dependencies, platformName, {
    forceInstall: args.forcePods,
    newArchEnabled: isAppRunningNewArchitecture,
  });

  installedPods = true;
}
```

Test Plan:
----------

We need to check that the function is working properly, even if we put a platformName that is invalid.

Checklist
----------

### DIFF
--- a/packages/cli-platform-apple/src/commands/buildCommand/createBuild.ts
+++ b/packages/cli-platform-apple/src/commands/buildCommand/createBuild.ts
@@ -12,9 +12,7 @@ import {supportedPlatforms} from '../../config/supportedPlatforms';
 const createBuild =
   ({platformName}: BuilderCommand) =>
   async (_: Array<string>, ctx: Config, args: BuildFlags) => {
-    const platformConfig = ctx.project[platformName] as
-      | IOSProjectConfig
-      | undefined;
+    const platformConfig = ctx.project[platformName] as IOSProjectConfig;
 
     if (
       platformConfig === undefined ||
--- a/packages/cli-platform-apple/src/commands/buildCommand/createBuild.ts
+++ b/packages/cli-platform-apple/src/commands/buildCommand/createBuild.ts
@@ -12,7 +12,10 @@ import {supportedPlatforms} from '../../config/supportedPlatforms';
 const createBuild =
   ({platformName}: BuilderCommand) =>
   async (_: Array<string>, ctx: Config, args: BuildFlags) => {
-    const platformConfig = ctx.project[platformName] as IOSProjectConfig;
+    const platformConfig = ctx.project[platformName] as
+      | IOSProjectConfig
+      | undefined;
+
     if (
       platformConfig === undefined ||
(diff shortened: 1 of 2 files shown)


packages/cli-platform-apple/src/commands/buildCommand/createBuild.ts
```diff
@@ -12,7 +12,10 @@ import {supportedPlatforms} from '../../config/supportedPlatforms';
 const createBuild =
   ({platformName}: BuilderCommand) =>
   async (_: Array<string>, ctx: Config, args: BuildFlags) => {
-    const platformConfig = ctx.project[platformName] as IOSProjectConfig;
+    const platformConfig = ctx.project[platformName] as
+      | IOSProjectConfig
+      | undefined;
```
szymonrybczak, line 17: I think we can also skip casting this type to `undefined`.

```suggestion
    const platformConfig = ctx.project[platformName] as IOSProjectConfig;
```

packages/cli-platform-apple/src/commands/runCommand/createRun.ts
```diff
@@ -51,7 +51,9 @@ const createRun =
   async (_: Array<string>, ctx: Config, args: FlagsT) => {
     // React Native docs assume platform is always ios/android
     link.setPlatform('ios');
-    const platformConfig = ctx.project[platformName] as IOSProjectConfig;
+    const platformConfig = ctx.project[platformName] as
+      | IOSProjectConfig
+      | undefined;
```
szymonrybczak, line 56: I think we can also skip casting this type to undefined.

```suggestion
    const platformConfig = ctx.project[platformName] as IOSProjectConfig;
```

hoonjoo-park: @szymonrybczak I've just pushed a new commit you suggested

szymonrybczak (approved):
Thanks! 👍

PR #2298, opened 2024-02-14
feat: add `add-platform` command

szymonrybczak:
Summary:
---------

In this Pull Request I've added new command `add-platform` which re-uses code from `init` command to add a new OOT platform to existing React Native app. It gets React Native version from `package.json` and it tries to find a matching platform's version from NPM registry. Then, if successful it reuses `editTemplate` functions to replace placeholders. 

https://github.com/react-native-community/cli/assets/63900941/6529cf35-f27a-4373-b843-1dc4bf66b4c7

For OOT platforms to add a compatibility with the new command:
1. Add `platforms` inside [`template.config.js`](https://github.com/facebook/react-native/blob/main/packages/react-native/template.config.js), which is an array in which OOT platform should provide names of folders containing native code. e.g. `platforms: ["macos"]`
2. Add `postInitScript` (also under `template.config.js`) with e.g. installing Pods and printing run instructions.
3. Add platform's specific native code under `packages/react-native/template${platformName}`.

Checklist
----------

### DIFF
--- a/docs/init.md
+++ b/docs/init.md
@@ -73,6 +73,8 @@ module.exports = {
 
   // Path to script, which will be executed after initialization process, but before installing all the dependencies specified in the template. This script runs as a shell script but you can change that (e.g. to Node) by using a shebang (see example custom template).
   postInitScript: './script.js',
+  // We're also using `template.config.js` when adding new platforms to existing project in `add-platform` command. Thanks to value passed to `platformName` we know which folder we should copy to the project.
+  platformName: 'visionos',
 };
 ```
 
@@ -91,12 +93,16 @@ new Promise((resolve) => {
   spinner.start();
   // do something
   resolve();
-}).then(() => {
-  spinner.succeed();
-}).catch(() => {
-  spinner.fail();
-  throw new Error('Something went wrong during the post init script execution');
-});
+})
+  .then(() => {
+    spinner.succeed();
+  })
+  .catch(() => {
+    spinner.fail();
+    throw new Error(
+      'Something went wrong during the post init script execution',
+    );
+  });
 ```
 
 You can find example custom template [here](https://github.com/Esemesek/react-native-new-template).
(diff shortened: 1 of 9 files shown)


docs/init.md
```diff
@@ -73,6 +73,8 @@ module.exports = {
 
   // Path to script, which will be executed after initialization process, but before installing all the dependencies specified in the template. This script runs as a shell script but you can change that (e.g. to Node) by using a shebang (see example custom template).
   postInitScript: './script.js',
+  // We're also using `template.config.js` when adding new platforms to existing project in `add-platform` command. Thanks to value passed to `platformName` we know which folder we should copy to the project.
+  platformName: 'visionos',
```
okwasniewski, line 77: How about making it an array? In the future let's say react-native-tvos starts to use this approach and wants to place `androidtv` separately to `appletv`

szymonrybczak: good point, changed.

szymonrybczak:
cc @kelset @tido64 @SaadNajmi would love to hear your thoughts about it :pray:

kelset:
looping in also @acoates-ms and @chrisglein who might be interested in this

Saadnajmi:
At its current point, this doesn't work with React Native Windows or macOS, correct? Has this been tested maybe with some local changes to a clone of RNM/RNW to see how well it works? I think it's harder for me to give any feedback (at least macOS specific) without seeing an end to end use case (other than visionOS :P )

docs/init.md
okwasniewski, line 77:
```suggestion
  platformName: ['visionos'],
```

packages/cli/src/commands/addPlatform/addPlatform.ts
```diff
@@ -0,0 +1,259 @@
...
+
+  return name;
+};
+
+const getPackageMatchingVersion = async (
+  packageName: string,
+  version: string,
+) => {
+  const npmResponse = await npmFetch.json(packageName, {
+    registry: NPM_REGISTRY_URL,
```
thymikee, line 75: registry should be configurable or read from npmrc or equivalent

szymonrybczak: agree, it turns out that we have a util for that 👍

Saadnajmi:
Just checking, will this be picked up and/or should macOS go support it?

Saadnajmi:
Just checking, will this be picked up and/or should macOS go support it?

PR #2314, opened 2024-02-28
chore: specify where to run `npx react-native info` for bug reports

mikehardy:
Summary:
---------

the "info" command is not known to the react-native wrapper on npmjs.com so "info" only works if you run it from inside a directory where react-native is installed as a local package

This is what happens if you just run `npx react-native info` outside a project:

```
mike@isabela:~ % npx react-native info
npm WARN cli npm v10.2.4 does not support Node.js v18.15.0. This version of npm supports the following node versions: `^18.17.0 || >=20.5.0`. You can find the latest version at https://nodejs.org/.
Need to install the following packages:
react-native@0.73.5
Ok to proceed? (y) y
error: unknown command 'info'
(Did you mean init?)

```

Test Plan:
----------

Run it inside a working project with react-native installed, and it works

Checklist
----------

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -7,7 +7,8 @@ labels: 'bug report'
 ## Environment
 
 <!-- 
-  Run `npx react-native info` in your terminal and paste its contents here. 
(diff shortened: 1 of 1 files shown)


szymonrybczak (approved):
Thanks!

TMisiukiewicz (approved):

PR #2152, opened 2023-11-02
Revert "chore(e2e): check for undefined code before printing fail"

szymonrybczak:
Reverts react-native-community/cli#2061

PR #2329, opened 2024-03-15
fix: init uses yarn assuming version > 1.x

blakef:
Summary:
---------

Checks that we're using yarn > 1.x when calling `yarn set` which isn't supported [1] but is for later versions [2].  This also updated the E2E test to vary depending on whether yarn 1 or 3 is running.

[1] https://classic.yarnpkg.com/en/docs/cli/set
[2] https://yarnpkg.com/cli/set/version

Test Plan:
----------

Ran against earlier version of yarn where I hit this.

Checklist
----------

### DIFF
--- a/__e2e__/init.test.ts
+++ b/__e2e__/init.test.ts
@@ -1,7 +1,14 @@
 import fs from 'fs';
 import path from 'path';
 import {runCLI, getTempDirectory, cleanup, writeFiles} from '../jest/helpers';
+import {execSync} from 'child_process';
+import semver from 'semver';
 import slash from 'slash';
+
+const yarnVersion = semver.parse(execSync('yarn --version').toString().trim())!;
+
+// .yarnrc -> .yarnrc.yml >= 2.0.0: yarnpkg/berry/issues/239
+const yarnConfigFile = yarnVersion.major > 1 ? '.yarnrc.yml' : '.yarnrc';
 
 const DIR = getTempDirectory('command-init');
 const PROJECT_NAME = 'TestInit';
@@ -23,7 +30,7 @@ function createCustomTemplateFiles() {
 const customTemplateCopiedFiles = [
   '.git',
   '.yarn',
-  '.yarnrc.yml',
+  yarnConfigFile,
   'dir',
   'file',
   'node_modules',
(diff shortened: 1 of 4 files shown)


__e2e__/init.test.ts
```diff
@@ -1,11 +1,20 @@
...
 import slash from 'slash';
 
 const DIR = getTempDirectory('command-init');
 const PROJECT_NAME = 'TestInit';
 
+const yarnVersion = semver.parse(execSync('yarn --version').toString().trim())!;
+
+const yarnConfigFiles: string[] = [];
+if (yarnVersion.major >= 3) {
+  yarnConfigFiles.push('.yarnrc.yml', '.yarn');
```
blakef, line 15: @szymonrybczak these were added in #2134, which seem to cause issues with the test for me locally and on CI.

jest/helpers.ts
```diff
@@ -142,18 +142,18 @@ function getExecaOptions(options: SpawnOptions) {
 function handleTestFailure(
   cmd: string,
   options: SpawnOptions,
-  result: {[key: string]: any},
+  result: execa.ExecaReturnBase<string>,
   args: string[] | undefined,
 ) {
-  if (!options.expectedFailure && result.code !== 0) {
+  if (!options.expectedFailure && result.exitCode !== 0) {
```
blakef, line 148: This was spamming the logs with false positives on the CI E2E tests.

packages/cli-platform-android/src/config/findManifest.ts
```diff
@@ -11,7 +11,7 @@ import path from 'path';
 import {unixifyPaths} from '@react-native-community/cli-tools';
 
 export default function findManifest(folder: string) {
-  let manifestPaths = glob.sync(path.join('**', 'AndroidManifest.xml'), {
+  let manifestPaths = glob.sync('**/AndroidManifest.xml', {
```
blakef, line 14: `fast-glob` [only wants paths expressed as forward slashes](https://github.com/mrmlnc/fast-glob#how-to-write-patterns-on-windows), so we want the opposite behaviour.  This was causing failures on the Windows E2E tests.

jest/helpers.ts
szymonrybczak: I know that some time ago we hit the issue where commands were failing and `execa` where reporting status codes as `undefined`, hopefully this change will catch all situations when there's a failure. 

More details: https://github.com/react-native-community/cli/pull/2152

szymonrybczak (approved):
Thank you!

PR #2379, opened 2024-05-06
perf(autolinking): get platform's specific properties

szymonrybczak:
Summary:
---------

This Pull Request implements fetching only platform's specific properties. It means when passing `--platform` option to `config` command, config logic will only get platform's properties, because it doesn't make sense to get informations about Android platform when running `native_modules.rb`, and same deal with `native_modules.gradle`.

I did some benchmarks in project with 50 linked dependencies, and results are:
| Command | Result |
|--------|--------|
| Standalone `config` | 7.62s |
| `config --platform ios` | 0.66s |
| `config --platform android` | 7.43s | 

So it means that linking dependencies during `pod install` got ~7 seconds faster _in this specific scenario_. If there are more dependencies to be linked, the results are higher.

Test Plan:
----------

2. Run this command:

```sh
```
Same output as before, if there are tools that depends on `config`'s command response they won't be affected by this change.
3. Run this command:

```sh
node /path/to/react-native-cli/packages/cli/build/bin.js config --platform ios
```
Returns only `ios` as in `platforms` field, and also for each entry in `dependencies` only `ios` platform should be presented. And this should be valid for any other platform that is available with linked dependency. 
4. Installing Cocoapods inside `ios/` folder should work in the same way as before.
5. Android app should build correctly and link all native dependencies.

Checklist
----------

### DIFF
--- a/packages/cli-config/src/__tests__/index-test.ts
+++ b/packages/cli-config/src/__tests__/index-test.ts
@@ -66,7 +66,7 @@ test('should have a valid structure by default', () => {
       reactNativePath: "."
     }`,
   });
-  const config = loadConfig(DIR);
+  const config = loadConfig({projectRoot: DIR});
   expect(removeString(config, DIR)).toMatchSnapshot();
 });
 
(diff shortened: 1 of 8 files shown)


szymonrybczak:
CI Blocked by https://github.com/react-native-community/cli/pull/2376

satya164 (approved):

PR #2381, opened 2024-05-07
feat: use native fetch from node

jbroma:
Summary:
---------

When using CLI with Node >= 21, there are deprecation warnings showing up in console, which can be traced to `whatwg-url@5.0.0` which is a dependency of `node-fetch`. I've replaced `node-fetch` with native `fetch` from Node. Since Node18, the fetch is no longer under experimental flag, and since Node21 it's marked as stable. 

Remarks:
----------
- Needed to remove `@types/opentype.js` package as it was polluting global context with DOM types
- `fetchToTemp` used `stream.Readable.fromWeb` which is still marked as experimental but we only use use it inside of `cli-doctor` so the impact is relatively low

Test Plan:
----------
- [ ] -  test by setting up a new project

Checklist
----------


packages/cli-link-assets/package.json
```diff
@@ -26,7 +26,6 @@
   ],
   "devDependencies": {
     "@react-native-community/cli-types": "14.0.0-alpha.4",
-    "@types/opentype.js": "^1.3.8",
```
thymikee, line 29: @fabioh8010 why do we need this? any case we could remove the dep altogether? (as a part of this PR its source was brought to this package and modified)

fabioh8010: `opentype.js` is a core library in `cli-link-assets` used to analyse the font's metadata. I added `@types/opentype.js` because `opentype.js` doesn't have any TS support.

jbroma: doesn't look like we can get rid of `opentype.js`, we could just limit the vendored types to what's actually needed, WDYT?

fabioh8010: Sounds good to me!

szymonrybczak (approved):
LGTM! Still there are some errors in higher Node versions, I'll address them in upcoming PRs 👍

jbroma:
What's the status of this PR, is there anything left to do here?

thymikee:
I'm also interested about these "errors in higher Node versions" – were these present in CI or local testing only?

szymonrybczak:
> were these present in CI or local testing only?

Locally, we don't have tests for this part of codebase 😅
```
~/development/tmp/R74 
❯ node --version
v22.2.0

~/development/tmp/R74                                         
❯ yarn ios
(node:24293) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
(Use `node --trace-deprecation ...` to show where the warning was created)
```

thymikee (approved):

PR #2384, opened 2024-05-08
perf(autolinking)!: prioritise files with `*Package.java|*Package.kt` in `getPackageClassName`

szymonrybczak:
Summary:
---------

This Pull Request changes schema that looks for files to check to find class name, after my tests in various application the match results were included almost in every library in files with `Package` at the end. But our previous logic was going through every file in library with `.kt` and `.java` extension until it found file with `implements ReactPackage`, some libraries can have 100+ files and we were reading them until we matched relevant Regex schema 😅

I also added a fallback which checks files with `*.kt` and `*.java` extensions if it couldn't find `implements ReactPackage` in file with `*Package` at the end, which ensures us that we don't introduce any breaking change. **The output of `config` command stays the same.**

Performance improvement
---------

I did a test in [Expensify](https://github.com/Expensify/App) app, which is an open-source app with 50 libraries containing native code. Before my changes, we were reading 490 files to find the class name. Right now, we only check 79 (still there's a room for improvements).

| | Before | After |
|--------|--------|--------|
| Files read | 490 | 79 |
| Time | ~7.85s | ~**1.15s** | 

Test Plan:
----------

2. Run this command:

```sh
```
The `dependencies` field should be the same as before. 

Checklist
----------

### DIFF
--- a/packages/cli-platform-android/src/config/findPackageClassName.ts
+++ b/packages/cli-platform-android/src/config/findPackageClassName.ts
@@ -11,20 +11,41 @@ import glob from 'fast-glob';
 import path from 'path';
 import {unixifyPaths} from '@react-native-community/cli-tools';
 
-export function getMainActivityFiles(folder: string) {
-  return glob.sync('**/+(*.java|*.kt)', {cwd: unixifyPaths(folder)});
+export function getMainActivityFiles(
+  folder: string,
+  includePackage: boolean = true,
+) {
+  let patternArray = [];
+
+  if (includePackage) {
+    patternArray.push('*Package.java', '*Package.kt');
+  } else {
+    patternArray.push('*.java', '*.kt');
+  }
+
+  return glob.sync(`**/+(${patternArray.join('|')})`, {
+    cwd: unixifyPaths(folder),
+  });
 }
 
 export default function getPackageClassName(folder: string) {
-  const files = getMainActivityFiles(folder);
+  let files = getMainActivityFiles(folder);
+  let packages = getClassNameMatches(files, folder);
 
-  const packages = files
-    .map((filePath) => fs.readFileSync(path.join(folder, filePath), 'utf8'))
-    .map(matchClassName)
-    .filter((match) => match);
+  if (!packages.length) {
+    files = getMainActivityFiles(folder, false);
+    packages = getClassNameMatches(files, folder);
+  }
 
   // @ts-ignore
   return packages.length ? packages[0][1] : null;
+}
+
+function getClassNameMatches(files: string[], folder: string) {
+  return files
+    .map((filePath) => fs.readFileSync(path.join(folder, filePath), 'utf8'))
+    .map(matchClassName)
+    .filter((match) => match);
 }
 
 export function matchClassName(file: string) {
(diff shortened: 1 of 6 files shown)


packages/cli-platform-android/src/config/findPackageClassName.ts
```diff
@@ -11,22 +11,43 @@ import glob from 'fast-glob';
...
+    packages = getClassNameMatches(files, folder);
+  }
 
   // @ts-ignore
   return packages.length ? packages[0][1] : null;
 }
 
+function getClassNameMatches(files: string[], folder: string) {
+  return files
+    .map((filePath) => fs.readFileSync(path.join(folder, filePath), 'utf8'))
```
tido64, line 46: I wonder if streaming is faster than reading the whole file every time: https://tangledeveloper.medium.com/how-to-read-file-line-by-line-efficiently-in-node-js-8125fe5bd185

thymikee: worth to measure definitely, these declarations are usually somewhere at the top of the file so it hints some potential

tido64: Replacing `glob.sync` above with [`glob.stream`](https://github.com/mrmlnc/fast-glob?tab=readme-ov-file#stream) might also be worth looking into, but making everything async might turn this into a very big PR 😛

packages/cli-platform-android/src/config/findPackageClassName.ts
```diff
@@ -11,22 +11,43 @@ import glob from 'fast-glob';
...
-  const files = getMainActivityFiles(folder);
+  let files = getMainActivityFiles(folder);
+  let packages = getClassNameMatches(files, folder);
 
-  const packages = files
-    .map((filePath) => fs.readFileSync(path.join(folder, filePath), 'utf8'))
-    .map(matchClassName)
-    .filter((match) => match);
+  if (!packages.length) {
+    files = getMainActivityFiles(folder, false);
```
satya164, line 36: maybe there's scope to exclude previously matched files from this list not to read same files twice

szymonrybczak: nice! good catch

cortinico (approved):
Great improvement 👍 All good on my end but wait for another CLI reviews before merging this (also because I believe this is not E2E tested)

packages/cli-platform-android/src/config/findPackageClassName.ts
tido64, line 36: A bit late to the game but I wonder if globbing for `**/*.{java,kt}` once and filtering in-memory instead of hitting the disk twice is faster for bigger projects.

thymikee: @szymonrybczak keen to measure it and sending a followup if that helps with perf? :)

szymonrybczak: Hm, I gave it a shot but the improvement won't be big because after my checks the amount of libraries that needs to fallback are very small.

Now: 
```
❯ hyperfine "npx react-native config --platform android"
Benchmark 1: npx react-native config --platform android
  Time (mean ± σ):      2.070 s ±  0.264 s    [User: 1.049 s, System: 0.607 s]
  Range (min … max):    1.916 s …  2.814 s    10 runs
```
With executing glob once and filtering in-memory:
```
❯ hyperfine "npx react-native config --platform android"
Benchmark 1: npx react-native config --platform android
  Time (mean ± σ):      2.096 s ±  0.190 s    [User: 1.078 s, System: 0.620 s]
  Range (min … max):    1.931 s …  2.572 s    10 runs
```

PR #2385, opened 2024-05-09
chore: upgrade parcel to v2

jbroma:
Summary:
---------

builds correctly but no idea if the config is right, needs to be tested

Test Plan:
----------

- [x] - builds without errors
- [x] - works as before 

Checklist
----------

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ build/
 coverage
 !packages/cli-platform-ios/src/config/__fixtures__/native_modules/node_modules
 *.env
+.parcel-cache
--- a/packages/cli-debugger-ui/.babelrc
+++ b/packages/cli-debugger-ui/.babelrc
@@ -1,3 +1,0 @@
-{
-  "plugins": [["@babel/plugin-transform-runtime", {"regenerator": true}]]
-}
--- a/packages/cli-debugger-ui/.parcelrc
+++ b/packages/cli-debugger-ui/.parcelrc
@@ -1,0 +1,8 @@
+{
+  "extends": "@parcel/config-default",
+  "transformers": {
+    "*.js": [
+      "@parcel/transformer-js"
+    ]
+  }
+}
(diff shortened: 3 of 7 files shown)


Piotrfj:
Tested on debugger-ui, works as intended, there are no additional errors in console.

szymonrybczak:
There's something wrong with specifying exports for `browser` and `node` environment, @Piotrfj did you remove previous build?

After running `rm -rf build && yarn build`, `/debugger-ui` route fails with an error:

CleanShot 2024-05-26 at 16 27 08@2x

Piotrfj:
@szymonrybczak I've done clean install/build on the branch, so I don't think there was any issues with that. @jbroma can you also confirm that the problem issued by @szymonrybczak, I will try to test it soon also

szymonrybczak:
> @szymonrybczak I've done clean install/build on the branch, so I don't think there was any issues with that. 

I don't see any changes, are you sure that you can push to @jbroma fork?

thymikee:
Tested this locally and while building on clear and populated cache works, changing the HTML contents and rebuilding causes SIGSEGV when doing `parcel build` or just straight not outputting the files to the filesystem (while the Parcel UI states they were built, which I guess is some kind of race condition). Haven't found anything helpful in how to mitigate that other than removing the `.parcel-cache` folder in project root every time we make a change in HTML. 

Provided that we're not going to change this part of the codebase anytime soon (other than deleting it once Experimental Debugger is not experimental anymore), I'm leaning towards merging it as it shouldn't have impact on our users (we publish the `build/` folder), just on the contributiors wanting to change stuff there or release the new version that may have some files missing.

szymonrybczak (approved):
Alright after @thymikee's investigation and removing `.parcel-cache` in root directory it seems to work! Thank you @jbroma!

CleanShot 2024-06-03 at 22 27 23@2x

PS. I think it'd be reasonable to add `rm -rf .parcel-cache` to `build-clean` script, so whenever someone will be releasing CLI, latest changes will be bundled and published.

jbroma:
perhaps let's just run the build command with `--no-cache` instead of deleting it, that should always overwrite the contents

szymonrybczak:
Yeah, tested and `--no-cache` also works for me! Good idea 👍

thymikee:
Added the `--no-cache` flag, works as expected, thanks! 👍🏼

PR #2386, opened 2024-05-10
ci: skip windows e2e tests in specific scenario

szymonrybczak:
Summary:
---------

When there's a change _only_ in `cli-platform-apple` or `cli-platform-ios` it doesn't make sense to triger also Windows jobs, because there's no scenario that change in these packages would broke Windows builds.

Test Plan:
----------

If there's only a change in `cli-platform-apple` or `cli-platform-ios` Windows jobs shouldn't be triggered. 

Checklist
----------

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -7,6 +7,24 @@ on:
     branches: ['**']
 
 jobs:
+  changes:
+    runs-on: ubuntu-latest
+    outputs:
+      is_specific: ${{ steps.path-check.outputs.specific_dir }}
+      is_others: ${{ steps.path-check.outputs.others }}
+    steps:
+      - id: path-check
+        uses: dorny/paths-filter@v2
+        with:
+          list-files: shell
+          filters: |
+            specific_dir:
+              - 'packages/cli-platform-ios/**'
+              - 'packages/cli-platform-apple/**'
+            others:
+              - '**/**'
+              - '!packages/cli-platform-ios/**'
+              - '!packages/cli-platform-apple/**'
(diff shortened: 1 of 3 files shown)


szymonrybczak:
As discussed with @TMisiukiewicz this Pull Request needs to be first merged to work correctly - Windows shouldn't be triggered when there's _only_ change in `cli-platform-apple` or `cli-platform-ios`. This Pull Request contains a change in `.github/workflows/test.yml`, so in current scenario jobs are triggered correctly.

packages/cli-platform-apple/src/index.ts
```diff
@@ -17,3 +17,4 @@ export {default as installPods} from './tools/installPods';
 
 export {default as findXcodeProject} from './config/findXcodeProject';
 export {default as findPbxprojFile} from './config/findPbxprojFile';
+console.log('hello');
```
thymikee, line 20: Leftover

.github/workflows/test.yml
```diff
@@ -7,6 +7,24 @@ on:
...
+    outputs:
+      is_specific: ${{ steps.path-check.outputs.specific_dir }}
+      is_others: ${{ steps.path-check.outputs.others }}
+    steps:
+      - id: path-check
+        uses: dorny/paths-filter@v2
+        with:
+          list-files: shell
+          filters: |
+            specific_dir:
```
thymikee, line 21: Let’s rename this to something like „apple_dirs” in a follow-up PR and adjust the name of the job. „Changes” is super ambiguous.

PR #2412, opened 2024-06-18
fix: correctly check Yarn version

szymonrybczak:
Summary:
---------

We were checking for Yarn major to be higher than 1, which is okay ✅ but we also checked in every scenario for minor to be higher than 22, which e.g for Yarn version 3.6.4 is not the case and CLI was running `yarn policies set-version x.y.z` for Yarn v3 which was throwing an error, and PnP was used.

Test Plan:
----------

1. Use Yarn v3/v4 globally 
2. Run this command:

```sh
```
3. Packages should be installed with yarn and `nodeLinker: node-modules` should be used.

Checklist
----------

### DIFF
--- a/packages/cli/src/commands/init/init.ts
+++ b/packages/cli/src/commands/init/init.ts
@@ -90,7 +90,7 @@ const bumpYarnVersion = async (silent: boolean, root: string) => {
     if (yarnVersion) {
       // `yarn set` is unsupported until 1.22, however it's a alias (yarnpkg/yarn/pull/7862) calling `policies set-version`.
       let setVersionArgs = ['set', 'version', YARN_VERSION];
-      if (yarnVersion.major === 1 && yarnVersion.minor <= 22) {
+      if (yarnVersion.major === 1 && yarnVersion.minor < 22) {
         setVersionArgs = ['policies', 'set-version', YARN_VERSION];
       }
       await executeCommand('yarn', setVersionArgs, {
--- a/packages/cli/src/commands/init/init.ts
+++ b/packages/cli/src/commands/init/init.ts
@@ -89,12 +89,10 @@ const bumpYarnVersion = async (silent: boolean, root: string) => {
 
     if (yarnVersion) {
       // `yarn set` is unsupported until 1.22, however it's a alias (yarnpkg/yarn/pull/7862) calling `policies set-version`.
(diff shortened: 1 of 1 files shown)


packages/cli/src/commands/init/init.ts
```diff
@@ -90,7 +90,8 @@ const bumpYarnVersion = async (silent: boolean, root: string) => {
     if (yarnVersion) {
       // `yarn set` is unsupported until 1.22, however it's a alias (yarnpkg/yarn/pull/7862) calling `policies set-version`.
       const setVersionArgs =
-        yarnVersion.major > 1 && yarnVersion.minor >= 22
+        (yarnVersion.major > 1 && yarnVersion.minor >= 22) ||
+        yarnVersion.major >= 2
```
thymikee, line 94: Why are we not checking for specifically `yarnVersion.major === 1 && yarnVersion.minor >= 22`? would avoid this extra v2 check

szymonrybczak: Make sense, I guess there's no one using Yarn 0.x 😅

packages/cli/src/commands/init/init.ts
```diff
@@ -89,12 +89,10 @@ const bumpYarnVersion = async (silent: boolean, root: string) => {
...
     if (yarnVersion) {
       // `yarn set` is unsupported until 1.22, however it's a alias (yarnpkg/yarn/pull/7862) calling `policies set-version`.
-      const setVersionArgs =
-        (yarnVersion.major > 1 && yarnVersion.minor >= 22) ||
-        yarnVersion.major >= 2
-          ? ['set', 'version', YARN_VERSION]
-          : ['policies', 'set-version', YARN_VERSION];
-
+      let setVersionArgs = ['set', 'version', YARN_VERSION];
+      if (yarnVersion.major === 1 && yarnVersion.minor <= 22) {
```
szymonrybczak, line 93:
```suggestion
      if (yarnVersion.major === 1 && yarnVersion.minor < 22) {
```

thymikee, line 93:
```suggestion
      if (yarnVersion.major === 1 && yarnVersion.minor < 22) {
```

thymikee (approved):

PR #2417, opened 2024-06-19
fix: update react-native version in the template for versions >= 0.75.

blakef:
The new template handling in @react-native-community/template means that these are shipped with version 1000.0.0 for React Native. The CLI now owns having to update to the correct version.

**NOTE:** This will only apply if the template has a `react-native@1000.0.0`, i.e. is from `@react-native-community/template`.

Summary:
---------

1. Allow using `--template` and `--version` when `--version` is >= 0.75.  Added a warning about doing this, but it's valid if the user gets into trouble or more advanced users want to do fun stuff.
2. `@react-native/` scoped dependencies **must** match the version of `react-native`.  `init` now enforces this.
3. Done some work to normalise tags into concrete versions.  Going forward `@react-native-community/template` will release matching version for each React Native release.  It's important to normalise version at the time of constructing the project otherwise the template will gradually become invalidate as `react-native@latest` (for example) moves.

For example, now running:
```
node ./packages/cli/build/bin.js init Test75RC0 --version next  --skip-install --verbose
```

## Updated `react-native` version on template
CleanShot 2024-06-19 at 15 55 24@2x

## Added some logs for the version update in the template
CleanShot 2024-06-19 at 15 49 03@2x

## Scoped Version Matching
CleanShot 2024-06-19 at 18 04 38@2x

Test Plan:
----------

- additional unit test
- built and run locally: 
```
node ./packages/cli/build/bin.js init Test75RC0 --version next  --skip-install --verbose
node ./packages/cli/build/bin.js init Test75RC0  --skip-install --verbose
```

Checklist
----------

### DIFF
--- a/packages/cli/src/commands/init/__tests__/editTemplate.test.ts
+++ b/packages/cli/src/commands/init/__tests__/editTemplate.test.ts
@@ -11,6 +11,7 @@ import {
   validatePackageName,
   replaceNameInUTF8File,
   updateDependencies,
+  normalizeReactNativeDeps,
 } from '../editTemplate';
 import semver from 'semver';
 
@@ -429,5 +430,27 @@ describe('replaceNameInUTF8File', () => {
 
     expect(beforeReplacement).toEqual(afterReplacement);
     expect(fsWriteFileSpy).toHaveBeenCalledTimes(0);
+  });
+});
+
+describe('normalizeReactNativeDeps', () => {
+  it('returns only @react-native/* dependencies updated to a specific version', () => {
+    const devDependencies = {
+      '@babel/core': '^7.20.0',
+      '@react-native/babel-preset': '0.75.0-main',
+      '@react-native/eslint-config': '0.75.0-main',
+      '@react-native/metro-config': '0.75.0-main',
+      '@react-native/typescript-config': '0.75.0-main',
+      '@types/react': '^18.2.6',
+      '@types/react-test-renderer': '^18.0.0',
+      eslint: '^8.19.0',
+      'react-test-renderer': '19.0.0-rc-fb9a90fa48-20240614',
+    };
+    expect(normalizeReactNativeDeps(devDependencies, '0.75.0')).toMatchObject({
+      '@react-native/babel-preset': '0.75.0',
+      '@react-native/eslint-config': '0.75.0',
+      '@react-native/metro-config': '0.75.0',
+      '@react-native/typescript-config': '0.75.0',
+    });
   });
 });
--- a/packages/cli/src/commands/init/__tests__/editTemplate.test.ts
+++ b/packages/cli/src/commands/init/__tests__/editTemplate.test.ts
@@ -10,6 +10,7 @@ import {
   replacePlaceholderWithPackageName,
   validatePackageName,
   replaceNameInUTF8File,
+  updateDependencies,
 } from '../editTemplate';
 import semver from 'semver';
 
@@ -173,7 +174,7 @@ describe('changePlaceholderInTemplate', () => {
   });
 
   afterEach(() => {
-    jest.resetAllMocks();
+    jest.restoreAllMocks();
   });
 
   skipIfNode20(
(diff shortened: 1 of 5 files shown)


cipolleschi (approved):
LGTM

packages/cli-tools/src/fetch.ts
```diff
@@ -36,7 +36,9 @@ const fetchToTemp = (url: string): Promise<string> => {
         }
 
         const dest = fs.createWriteStream(tmpDir);
-        const body = stream.Readable.fromWeb(result.body);
+        const body: stream.Readable = stream.Readable.fromWeb(
+          result.body as any,
+        );
```
cipolleschi, line 41: with the fix in npm.ts, I think we can remove this, no?

cortinico:
Discussed offline with @blakef:

We also need all the deps inside `devDependencies` to be updated as they can't be `0.75-main`:
Screenshot 2024-06-19 at 17 22 16

packages/cli-tools/src/fetch.ts
blakef: Will remove if this is the case.

blakef:
> Discussed offline with @blakef:
> 
> We also need all the deps inside `devDependencies` to be updated as they can't be `0.75-main`: Screenshot 2024-06-19 at 17 22 16

Done.

packages/cli/src/commands/init/__tests__/editTemplate.test.ts
```diff
@@ -200,6 +201,67 @@ describe('changePlaceholderInTemplate', () => {
...
+
+  afterEach(() => {
+    jest.resetAllMocks();
+  });
+
+  it('updates react-native', () => {
+    updateDependencies({
+      dependencies: {
+        'react-native': '0.75.0',
+      },
```
cortinico, line 256: Let's test also `devDependencies` here?

packages/cli/src/commands/init/__tests__/editTemplate.test.ts
```diff
@@ -200,6 +201,67 @@ describe('changePlaceholderInTemplate', () => {
...
+    });
+    expect(fs.writeFileSync as jest.Mock).toHaveBeenCalledWith(
+      expect.anything(),
+      samplePackageJson.replace('1000.0.0', '0.75.0'),
```
cortinico, line 260: Same here let's also `.replace('0.75.0-main', '0.75.0')`

packages/cli/src/commands/init/init.ts
```diff
@@ -393,37 +446,13 @@ function checkPackageManagerAvailability(
...
-
-async function createTemplateUri(options: Options): Promise<string> {
+async function createTemplateUri(
+  options: Options,
+  version: string,
+): Promise<string> {
   if (options.platformName) {
     logger.debug('User has specified an out-of-tree platform, using it');
-    return `${options.platformName}@${options.version ?? DEFAULT_VERSION}`;
+    return `${options.platformName}@${version}`;
```
cortinico, line 455: Here you're missing the `?? DEFAULT_VERSION` fallback.

packages/cli/src/commands/init/init.ts
```diff
@@ -534,12 +571,24 @@ export default (async function initialize(
...
+  );
+
+  // From 0.75 it actually is useful to be able to specify both the template and react-native version.
+  // This should only be used by people who know what they're doing.
+  if (semver.gte(version, TEMPLATE_COMMUNITY_REACT_NATIVE_VERSION)) {
+    logger.warn(
+      `Use ${chalk.bold('--template')} and ${chalk.bold(
+        '--version',
+      )} only if you know what you're doing. Here be dragons 🐉.`,
+    );
```
cortinico, line 586: How is this going to work if a user specifies:

```
template @react-native-community/template@0.75.1 --version 0.75.2
```

?

I think we should fail here instead

packages/cli/src/commands/init/init.ts
```diff
@@ -490,7 +527,7 @@ async function createProject(
   // +------------------------------------------------------------------+----------+--------------+
   // | --template @react-native-community/template@0.75.1 --version 0.75| 0.75.1   | 0.75.x       |
   // +------------------------------------------------------------------+----------+--------------+
```
cortinico, line 529: Let's update this table with the latest logic also?

cortinico (commented):
LGTM but left a couple of nits

packages/cli/src/commands/init/init.ts
```diff
@@ -281,6 +305,35 @@ async function createFromTemplate({
       packageName,
     });
 
+    // Update the react-native dependency if using the new @react-native-community/template.
+    // We can figure this out as it ships with react-native@1000.0.0 set to a dummy version.
+    const templatePackageJson = JSON.parse(
+      fs.readFileSync(path.join(process.cwd(), 'package.json'), 'utf8'),
+    );
+    if (templatePackageJson.dependencies?.['react-native'] === '1000.0.0') {
```
cortinico, line 313: This needs to be refined.

You would have to check also if:
- Version is set to `nightly`
- Version is set to `0.XX.0-main`

If so, do the replacing

packages/cli/src/commands/init/__tests__/editTemplate.test.ts
blakef: I'll add something that'll covers what we use to do this.

packages/cli/src/commands/init/__tests__/editTemplate.test.ts
blakef: I'll work it in, but have to test this somewhere else.

packages/cli/src/commands/init/init.ts
blakef: I've moved this up the call stack:
```
initialize() -> createProject({ version }) -> createTemplateUri(options, version)
```
In `initialize` I set the `version` that's passed down to:
```typescript
  const version = await npmResolveConcreteVersion(
    'react-native',
    options.version ?? DEFAULT_VERSION,
  );
```

packages/cli/src/commands/init/init.ts
blakef: It'll fail for all version before `0.75.*`, but I think this is a valid workflow that people testing unreleased versions (release team + developers on CLI) would find useful.  We add a warning which is pretty clear.

I think we should keep it, but we can add a hidden argument (something like `--ci` or `--know-better`) to allow the power users I'm targeting to use this combination.

packages/cli/src/commands/init/init.ts
blakef: The template version for `react-native` is always `1000.0.0` to start with.  If it isn't then the template being worked on is one that was shipped with `react-native` (alway in versions < 0.75).

packages/cli-tools/src/fetch.ts
blakef: Done.

packages/cli/src/commands/init/init.ts
blakef: If we keep the option of having `--template` and `--version` (with a warning if you use both), then this table + comments are now accurate.

packages/cli/src/commands/init/init.ts
cortinico: Yup that makes sense. We then need this merged on the template then:
- https://github.com/react-native-community/template/pull/24

cortinico (approved):

PR #2423, opened 2024-06-21
fix(cli-config): include peer dependencies when finding dependencies

### LINKED ISSUE #2419: `config` ignores `peerDependencies` and instead reads `devDependencies`
## Description

`config` ignores `peerDependencies` and instead reads `devDependencies`:

https://github.com/react-native-community/cli/blob/33646ef94a8cb4f9ce10d6db336c7a124144025b/packages/cli-config/src/findDependencies.ts#L18-L21

This leads to errors in `react-native-windows` autolinking (and possibly in other scenarios) where dependencies are only declared under `peerDependencies` and not `devDependencies`, for instance in `@react-native-webapis/web-storage` we have the following:

```json
{
  "peerDependencies": {
    "@callstack/react-native-visionos": ">=0.73",
    "react": ">=18.2.0",
    "react-native": ">=0.72",
    "react-native-macos": ">=0.72",
    "react-native-windows": ">=0.72"
  },
  "peerDependenciesMeta": {
    "@callstack/react-native-visionos": {
      "optional": true
    },
    "react-native-macos": {
      "optional": true
    },
    "react-native-windows": {
      "optional": true
    }
  },
  "devDependencies": {
    "react-native": "^0.73.0"
  }
}
```

In this case, `config` outputs:

```json
  "platforms": {
    "ios": {},
    "android": {}
  },
```

If I add `react-native-windows` to `devDependencies`, `config` outputs:

```json
  "platforms": {
    "ios": {},
    "android": {},
    "windows": {
      "npmPackageName": "react-native-windows"
    }
  },
```

In my opinion, the correct fix should be to replace `devDependencies` with `peerDependencies`, but I don't know if this will break current scenarios so we should probably keep it as is and just add `peerDependencies` to the list in `findDependencies.ts`.

## Reproducible Demo

In any project, you can run the following:

```
yarn add @react-native-webapis/web-storage@0.2.6
cd node_modules/@react-native-webapis/web-storage
yarn react-native config
```

tido64:
Summary:
---------

`config` currently ignores `peerDependencies` and instead reads `devDependencies`, leading to missing dependencies.

Resolves #2419.

Test Plan:
----------

In any project, run the following:

```
yarn add @react-native-webapis/web-storage@0.2.6
cd node_modules/@react-native-webapis/web-storage
yarn react-native config
```

Checklist
----------

### DIFF
--- a/packages/cli-config/src/__tests__/findDependencies-test.ts
+++ b/packages/cli-config/src/__tests__/findDependencies-test.ts
@@ -1,23 +1,35 @@
 import findDependencies from '../findDependencies';
 import {cleanup, writeFiles, getTempDirectory} from '../../../../jest/helpers';
 
-beforeEach(async () => {
-  cleanup(DIR);
-  jest.resetModules();
-});
+describe('findDependencies', () => {
+  const DIR = getTempDirectory('find_dependencies_test');
 
-afterEach(() => cleanup(DIR));
-
-const DIR = getTempDirectory('find_dependencies_test');
-
-test('returns plugins from both dependencies and dev dependencies', () => {
-  writeFiles(DIR, {
-    'package.json': `
-      {
-        "dependencies": {"rnpm-plugin-test": "*"},
-        "devDependencies": {"rnpm-plugin-test-2": "*"}
-      }
-    `,
+  beforeEach(() => {
+    cleanup(DIR);
+    jest.resetModules();
   });
-  expect(findDependencies(DIR)).toHaveLength(2);
+
+  afterEach(() => cleanup(DIR));
+
+  test('returns packages from dependencies, peer and dev dependencies', () => {
+    writeFiles(DIR, {
+      'package.json': JSON.stringify({
+        dependencies: {'rnpm-plugin-test': '*'},
+        peerDependencies: {'rnpm-plugin-test-2': '*'},
+        devDependencies: {'rnpm-plugin-test-3': '*'},
+      }),
+    });
+    expect(findDependencies(DIR)).toHaveLength(3);
+  });
+
+  test('dedupes dependencies', () => {
+    writeFiles(DIR, {
+      'package.json': JSON.stringify({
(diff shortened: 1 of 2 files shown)


tido64:
cc @szymonrybczak @thymikee

szymonrybczak (approved):
Thank you! Maybe let's add some tests? :)

TMisiukiewicz:
I was trying to test it, but couldn't see any change in the `config` output until I installed any of peer dependencies as a dev dependency, am I missing something? Since peer dependencies are not installed by default, might it be the case where a dependency cannot be found in `node_modules`?

tido64:
> I was trying to test it, but couldn't see any change in the `config` output until I installed any of peer dependencies as a dev dependency, am I missing something? Since peer dependencies are not installed by default, might it be the case where a dependency cannot be found in `node_modules`?

The repro steps include running `config` under the `node_modules` folder, so it's definitely not a case of missing dependencies. Alternative repro steps can be found in this PR: https://github.com/microsoft/rnx-kit/issues/3200

tido64:
@TMisiukiewicz: Anything I can help with to move this forward?

TMisiukiewicz:
@tido64 I just want to clarify, what if you use a package manager that does not install peer dependencies automatically e.g. yarn classic? It would probably still not resolve the dependency properly?

tido64:
> @tido64 I just want to clarify, what if you use a package manager that does not install peer dependencies automatically e.g. yarn classic? It would probably still not resolve the dependency properly?

That isn't the issue here though? The issue is that the config algorithm doesn't consider `peerDependencies` at all and are missing out on dependencies that should be included. Whether they resolve is already handled elsewhere.

And to answer your question: IIRC, if a peer dependency is not installed (in case of Yarn classic, or maybe it's optional), CLI fails to resolve it and it is ignored.

szymonrybczak (approved):

PR #2424, opened 2024-06-21
fix: react-native RC2 fixes

blakef:
Summary:
---------

1. Removing ~ npm version range from the template looking for installs > 0.75.0. This effectively forces version pinning against the react native release version.  For example:
```
  npx @react-native-community/cli init Foo --version 0.75.0-rc.2
```
Would install @react-native-community/template@0.75.0-rc.2 template only. We have a plan to do something more comprehensive, but this is affecting the release and had to be paused. There are 2 exceptions:

- nightly - always points to the nightly tag
- 0.75.0-rc.1 - points to 0.75.0-rc.1.1 to unblock the RC1 release

2. Fixed a bug which disabled using --template and --version for installs > 0.75.0, as the release candidates were < 0.75.0.

Test Plan:
----------

1.  ✅ `node ./packages/cli/build/bin.js init RN075 --version 0.75.0-rc.1`, but since 0.75.0-rc.1 hasn't been released yet, when you come to install in `RN075` npm askes you for other versions.  If you install 0.75.0-rc.0 it works, which gives me more confidence for 0.75.0-rc.1's release.
2. ✅ `node ./packages/cli/build/bin.js init RN075 --version nightly`
3. ⌛ `node ./packages/cli/build/bin.js init RN075 --version next` - will only work once we do the react-native RC1 release.
4. I still want to test out various combinations of `--version` and `--template`.

Checklist
----------

### DIFF
--- a/packages/cli/src/commands/init/init.ts
+++ b/packages/cli/src/commands/init/init.ts
@@ -437,29 +437,21 @@ async function createTemplateUri(
   );
 
   if (!useLegacyTemplate) {
-    // This will use ~ to grab the latest patch version.
-    //
-    // Because we allow for many PATCHes to occur for a version of react-native, it's not possible to know
-    // (without inspecting the template's npm metadata) which template works with which react-native version.
-    // We are guaranteed that the latest version of react-native  will work with the latest version of the
-    // template on that version.
-    //
-    // Example:
-    //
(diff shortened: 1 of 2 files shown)


thymikee (approved):
Ok, let's unblock this release then. @blakef please cleanup these comments and version-specific code in a follow-up PR

PR #2434, opened 2024-06-25
feat: add version to welcome screen

okwasniewski:
Summary:
---------

This PR adds a version to the welcome screen when the user uses a custom one. 

This is helpful when testing older versions of React Native (to make sure we are initializing the correct one). It also follows the scheme that Metro does welcoming us with the version. As an extension, we could potentially fetch the latest version from npm and display it but I didn't want to add additional API call. So we display version only if user passed a custom one.

## Without explicit version (latest)

CleanShot 2024-06-25 at 10 13 13@2x

## With explicit version

CleanShot 2024-06-25 at 10 13 36@2x

Test Plan:
----------

Test out init command

Checklist
----------


szymonrybczak (approved):
Thank you!

PR #2436, opened 2024-06-25
fix: issue calling init from 0.75.0-rc.1 release

blakef:
Summary:
---------

When users call the cli through react-native (a deprecated path), it hard-codes the platform name to 'react-native'. This is an error, and doing this bypasses all template selection logic.

This excludes setting the platform name to 'react-native', which is assumed.

Test Plan:
----------

Replicates these changes to the npx cache `~/.npx/_cache/.../index.js` to confirm this fixes the issue holding up the release.  Added some logging to show:

CleanShot 2024-06-25 at 17 41 10@2x

1. The `run()` method is called with `react-native`, which we should ignore.
2. The raw `process.argv` used to call
3. The expected outcome with out change.

Checklist
----------

### DIFF
--- a/packages/cli/src/commands/init/init.ts
+++ b/packages/cli/src/commands/init/init.ts
@@ -400,7 +400,7 @@ async function createTemplateUri(
   options: Options,
   version: string,
 ): Promise<string> {
-  if (options.platformName) {
+  if (options.platformName && options.platformName !== 'react-native') {
     logger.debug('User has specified an out-of-tree platform, using it');
     return `${options.platformName}@${version}`;
   }


packages/cli/src/index.ts
```diff
@@ -234,7 +234,11 @@ async function setupAndRun(platformName?: string) {
   const argv = [...process.argv];
 
   // If out of tree platform specifices custom platform name, we need to pass it to argv array for the init command.
-  if (isCommandPassed('init') && platformName) {
+  if (
+    isCommandPassed('init') &&
+    platformName &&
+    platformName !== 'react-native'
```
blakef, line 240: It's not out of tree if it's `react-native`.

cortinico (approved):

Issue #2479, opened 2024-08-20
Unbounded globbing leads to "freeze" during `pod install`


ephemer:
Hello, and thank you for your work on this project! 🙂

I have been trying to understand a weird bug whereby `pod install` no longer works for me from RN >= 0.74. It just appears to freeze and do nothing. I left it for 30mins and nothing happened.

I finally tracked this down to the `npx @react-native-community/cli config` call performed during install. After some log debugging it turns out the "freeze" occurs during globbing while searching for the iOS Podfile.

We have a submodule containing native code that is very deeply nested (with its own native build system, dependencies, their build artefacts, etc.). `fast-glob` seems to get stuck looking in that deeply nested directory structure (maybe there's even a cycle somewhere due to symlinks – I'm not sure). In any case, it seems like the most likely case by far is that `Podfile` can be found in `${PROJECT_ROOT}/ios/Podfile`. It seems very unlikely that it will be found in a deeply nested directory somewhere – if it will be found anywhere other than that default location in the first place.

The following diff solved my problem, and I believe it'd generally be a good idea to put an upper bound on the depth of the search:

```diff
diff --git a/node_modules/@react-native-community/cli-platform-apple/build/config/findAllPodfilePaths.js b/node_modules/@react-native-community/cli-platform-apple/build/config/findAllPodfilePaths.js
index e82bcee..3d789cb 100644
--- a/node_modules/@react-native-community/cli-platform-apple/build/config/findAllPodfilePaths.js
+++ b/node_modules/@react-native-community/cli-platform-apple/build/config/findAllPodfilePaths.js
@@ -32,7 +32,8 @@ const GLOB_EXCLUDE_PATTERN = ['**/@(Pods|node_modules|Carthage|vendor)/**'];
 function findAllPodfilePaths(cwd) {
   return _fastGlob().default.sync('**/Podfile', {
     cwd: (0, _cliTools().unixifyPaths)(cwd),
-    ignore: GLOB_EXCLUDE_PATTERN
+    ignore: GLOB_EXCLUDE_PATTERN,
+    deep: 10,
   });
 }

szymonrybczak:
Fixed by: https://github.com/react-native-community/cli/pull/2480

PR #2480, opened 2024-08-20
Put an upper bound on "**/Podfile" search depth

ephemer:
Summary:
---------

This PR addresses https://github.com/react-native-community/cli/issues/2479. Basically the unbounded search for `**/Podfile` never finishes on my machine. Normally the required depth should be `2`, since `ios/Podfile` is the default location. So this still provides significantly more depth than most users should need.

Test Plan:
----------

To be honest I'm not sure what causes this issue on my system, I only investigated what fixes it so far. It's clearly an issue with a deeply-nested submodule that we have, but I can't see what about it is causing the issue. If I run `find .` in my project root, all files (including the deeply-nested ones) can be printed within a few seconds, whereas `fast-glob` appears to never finish its search without this `deep: 10` limitation.

Checklist
----------


thymikee:
Thanks! Can you file that in `fast-glob` project as well so they can hopefully fix this?

szymonrybczak (approved):
Thank you!

PR #2487, opened 2024-08-27
perf: skip getting package class name inside expo module

szymonrybczak:
Summary:
---------

Recently [blog post](https://github.com/react-native-community/discussions-and-proposals/discussions/814) regarding performance improvements inside autolinking was published, as it turns between implementations there's a big difference.

After my investigation it is due getting package class name from classes which are implementing `ReactPackage` or `TurboReactPackage`. In past we landed improvements to this logic (read more [here](https://github.com/react-native-community/cli/pull/2384)).

But as mentioned in the blog post the difference is still big and the root cause of this issue is because of reading hundreds of files from expo modules and trying to find classes that implements `ReactPackage` or `TurboReactPackage`, but expo modules don't implement `ReactPackage` or `TurboReactPackage` directly inside their source code, so the return value will be anyway `null`, so it doesn't make sense to read hundreds of files and check for classes that implements `ReactPackage` or `TurboReactPackage`.

Inside this Pull Request I've added a early return if module is expo module, following logic **doesn't affect** output of `config` command because expo modules follows other rules and are included in native build in different way.

Exception is `expo` package itself which contains `expo-module.config.json` and also implements `ReactPackage` or `TurboReactPackage`, I added appropriate condition for this case.  
 
Benchmark:
---------

Tested on the [`react-conf-app`](https://github.com/expo/react-conf-app).

Before:
```
❯ hyperfine 'node ~/development/cli/packages/cli/build/bin.js config'
Benchmark 1: node ~/development/cli/packages/cli/build/bin.js config
  Time (mean ± σ):      2.353 s ±  0.033 s    [User: 2.067 s, System: 0.313 s]
  Range (min … max):    2.313 s …  2.408 s    10 runs
```

After:
```
❯ hyperfine 'node ~/development/cli/packages/cli/build/bin.js config'
Benchmark 1: node ~/development/cli/packages/cli/build/bin.js config
  Time (mean ± σ):     654.7 ms ±  30.7 ms    [User: 453.2 ms, System: 269.3 ms]
  Range (min … max):   620.0 ms … 711.5 ms    10 runs
```

Test Plan:
----------

2. Run this command:

```sh
```
3. Output should be the same. 

Checklist
----------

### DIFF
--- a/packages/cli-platform-android/src/config/findPackageClassName.ts
+++ b/packages/cli-platform-android/src/config/findPackageClassName.ts
@@ -29,9 +29,6 @@ export function getMainActivityFiles(
 }
 
 export default function getPackageClassName(folder: string) {
-  let files = getMainActivityFiles(folder);
-  let packages = getClassNameMatches(files, folder);
-
   /*
     When module contains `expo-module.config.json` we return null
     because expo modules follow other practices and don't implement
@@ -60,6 +57,9 @@ export default function getPackageClassName(folder: string) {
   ) {
     return null;
   }
+
+  let files = getMainActivityFiles(folder);
+  let packages = getClassNameMatches(files, folder);
 
   if (!packages.length) {
     files = getMainActivityFiles(folder, false);
--- a/packages/cli-platform-android/src/config/findPackageClassName.ts
+++ b/packages/cli-platform-android/src/config/findPackageClassName.ts
@@ -29,6 +29,13 @@ export function getMainActivityFiles(
 }
 
 export default function getPackageClassName(folder: string) {
+  let files = getMainActivityFiles(folder);
+  let packages = getClassNameMatches(files, folder);
+
+  if (packages && packages.length > 0 && Array.isArray(packages[0])) {
+    return packages[0][1];
+  }
+
   /*
     When module contains `expo-module.config.json` we return null
     because expo modules follow other practices and don't implement
@@ -36,35 +43,17 @@ export default function getPackageClassName(folder: string) {
     to scan and read hundreds of files to get package class name.
 
     Exception is `expo` package itself which contains `expo-module.config.json`
-    and implements `ReactPackage/TurboReactPackage` inside `ExpoModulesPackage.kt`.
+    and implements `ReactPackage/TurboReactPackage`.
 
     Following logic is done due to performance optimization.
   */
 
-  if (
-    fs.existsSync(path.join(folder, '..', 'expo-module.config.json')) &&
-    !fs.existsSync(
-      path.join(
-        folder,
-        'src',
-        'main',
-        'java',
-        'expo',
-        'modules',
-        'ExpoModulesPackage.kt',
-      ),
-    )
-  ) {
+  if (fs.existsSync(path.join(folder, '..', 'expo-module.config.json'))) {
     return null;
   }
 
-  let files = getMainActivityFiles(folder);
-  let packages = getClassNameMatches(files, folder);
-
-  if (!packages.length) {
-    files = getMainActivityFiles(folder, false);
-    packages = getClassNameMatches(files, folder);
-  }
+  files = getMainActivityFiles(folder, false);
+  packages = getClassNameMatches(files, folder);
 
   // @ts-ignore
   return packages.length ? packages[0][1] : null;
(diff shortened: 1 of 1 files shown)


packages/cli-platform-android/src/config/findPackageClassName.ts
```diff
@@ -32,6 +32,35 @@ export default function getPackageClassName(folder: string) {
...
+        folder,
+        'src',
+        'main',
+        'java',
+        'expo',
+        'modules',
+        'ExpoModulesPackage.kt',
+      ),
+    )
+  ) {
```
thymikee, line 60: we could move this check before querying for `files` and `package`

szymonrybczak: agree 👍 https://github.com/react-native-community/cli/pull/2487/commits/1b27d05d5fc6d9569a267ff5f759535ffe54f69e

cortinico (commented):
LGTM. I'm just wondering if an expo module could actually have both the `'expo-module.config.json'` file and an implementation of a `ReactPackage`

cortinico (approved):

packages/cli-platform-android/src/config/findPackageClassName.ts
```diff
@@ -29,6 +29,35 @@ export function getMainActivityFiles(
...
+    fs.existsSync(path.join(folder, '..', 'expo-module.config.json')) &&
+    !fs.existsSync(
+      path.join(
+        folder,
+        'src',
+        'main',
+        'java',
+        'expo',
+        'modules',
+        'ExpoModulesPackage.kt',
```
Kudo, line 54: looks good to me. we had the similar logic at https://github.com/expo/expo/blob/08cba033007a260318fc438ab1d9f872d4ebac2f/packages/expo-modules-autolinking/src/reactNativeConfig/androidResolver.ts#L113-L125.

not sure if that's possible for not hardcoding `ExpoModulesPackage.kt` here. that's not a public api.

a side note. when implementing autolinking at expo, i feel that on android we waste much work on glob/file parsing. maybe at some point we could let developers to specify java class names or package names in **react-native.config.js**

Kudo (approved):

packages/cli-platform-android/src/config/findPackageClassName.ts
szymonrybczak: > not sure if that's possible for not hardcoding ExpoModulesPackage.kt

Is there anything else that we could check to ensure that module is `expo` package? 

> maybe at some point we could let developers to specify java class names or package names in react-native.config.js

IMO asking every library maintainer do add this value might be hard, we consider it some time ago. https://github.com/react-native-community/cli/pull/2384 this change should be already fine for majority of cases.

Kudo: > Is there anything else that we could check to ensure that package is `expo` package?

in our implementation, we try to match `*Package.{java,kt}` and check whether the file extends `ReactPackage`/`TurboReactPackage` before the `expo-modules.config.json` skip path. the `expo` package will matched by the `*Package.{java,kt}` check.
not sure if implementation helps for rnc-cli though. https://github.com/expo/expo/blob/08cba033007a260318fc438ab1d9f872d4ebac2f/packages/expo-modules-autolinking/src/reactNativeConfig/androidResolver.ts#L113-L125

alternative, to exclude the `expo` package, is it possible to pass packageName into the function? maybe we can also skip when `packageName === 'expo'`
 
> IMO asking every library maintainer do add this value might be hard, we consider it some time ago. #2384 this change should be already fine for majority of cases.

as long as one lib adding this value for optimization, it could help to reduce the whole app autolinking time. but that's fine if's that's being discussed already.

Kudo: > let's remember to always have a fallback across implementations because there's no official guideline to have implments ReactPackage/TurboReactPackage inside file with *Package.kt/java file (and also there are some community libraries that don't follow this rule).

yep we have the fallback as https://github.com/expo/expo/blob/08cba033007a260318fc438ab1d9f872d4ebac2f/packages/expo-modules-autolinking/src/reactNativeConfig/androidResolver.ts#L123-L129. thanks for sharing this!

PR #2510, opened 2024-09-18
fix(macCatalyst): construct correct path for executable

mikehardy:
Summary:
---------

it appears targetBuildDir var now has `-maccatalyst` in it before it gets to this method, so no need to add it again

This was tweaked in #2312 but appears to have regressed again - are macCatalyst builds checked in CI 🤔 ?

Without this change, the path is incorrect (note the duplicate `-maccatalyst` in path):

```

Error: spawn /Users/mike/Library/Developer/Xcode/DerivedData/rnfbdemo-gsuyjvwnbefbzvbfqtovnhthepew/Build/Products/Debug-maccatalyst-maccatalyst/rnfbdemo.app/Contents/MacOS/rnfbdemo ENOENT
    at ChildProcess._handle.onexit (node:internal/child_process:284:19)
    at onErrorNT (node:internal/child_process:477:16)
    at process.processTicksAndRejections (node:internal/process/task_queues:82:21)
Emitted 'error' event on ChildProcess instance at:
    at ChildProcess._handle.onexit (node:internal/child_process:290:12)
    at onErrorNT (node:internal/child_process:477:16)
    at process.processTicksAndRejections (node:internal/process/task_queues:82:21) {
  errno: -2,
  code: 'ENOENT',
  syscall: 'spawn /Users/mike/Library/Developer/Xcode/DerivedData/rnfbdemo-gsuyjvwnbefbzvbfqtovnhthepew/Build/Products/Debug-maccatalyst-maccatalyst/rnfbdemo.app/Contents/MacOS/rnfbdemo',
  path: '/Users/mike/Library/Developer/Xcode/DerivedData/rnfbdemo-gsuyjvwnbefbzvbfqtovnhthepew/Build/Products/Debug-maccatalyst-maccatalyst/rnfbdemo.app/Contents/MacOS/rnfbdemo',
  spawnargs: []
}
```

Highlight:

> /Users/mike/Library/Developer/Xcode/DerivedData/rnfbdemo-gsuyjvwnbefbzvbfqtovnhthepew/Build/Products/Debug-maccatalyst-maccatalyst/rnfbdemo.app/Contents/MacOS/rnfbdemo

Test Plan:
----------

I build and test with macCatalyst all the time: https://github.com/mikehardy/rnfbdemo/blob/main/make-demo.sh

I added a unit test that exercises the method, sending it old (without -maccatalyst) and new (with -maccatalyst) inputs to verify it handles both, and verify I didn't regress the more common ios build code path

Checklist
----------

### DIFF
--- a/packages/cli-platform-apple/src/commands/runCommand/getBuildPath.ts
+++ b/packages/cli-platform-apple/src/commands/runCommand/getBuildPath.ts
@@ -1,5 +1,6 @@
 import {CLIError} from '@react-native-community/cli-tools';
 import path from 'path';
+import fs from 'fs';
 import {BuildSettings} from './getBuildSettings';
 import {ApplePlatform} from '../../types';
 
@@ -8,7 +9,7 @@ export async function getBuildPath(
   platform: ApplePlatform = 'ios',
   isCatalyst: boolean = false,
 ) {
-  const targetBuildDir = buildSettings.TARGET_BUILD_DIR;
+  let targetBuildDir = buildSettings.TARGET_BUILD_DIR;
   const executableFolderPath = buildSettings.EXECUTABLE_FOLDER_PATH;
   const fullProductName = buildSettings.FULL_PRODUCT_NAME;
 
@@ -24,11 +25,31 @@ export async function getBuildPath(
     throw new CLIError('Failed to get product name.');
   }
 
-  if (isCatalyst) {
-    return path.join(`${targetBuildDir}-maccatalyst`, executableFolderPath);
-  } else if (platform === 'macos') {
-    return path.join(targetBuildDir, fullProductName);
-  } else {
-    return path.join(targetBuildDir, executableFolderPath);
+  // Default is platform == ios && isCatalyst == false
+  let buildPath = path.join(targetBuildDir, executableFolderPath);
+
+  // platform == ios && isCatalyst == true needs build path suffix,
+  // but this regresses from time to time with suffix present or not
+  // so check - there may be one already, or we may need to add suffix
+  if (platform === 'ios' && isCatalyst) {
+    // make sure path has one and only one '-maccatalyst' suffix on end
+    if (!targetBuildDir.match(/-maccatalyst$/)) {
+      targetBuildDir = `${targetBuildDir}-maccatalyst`;
+    }
+    buildPath = path.join(targetBuildDir, executableFolderPath);
   }
+
+  // macOS gets the product name, not the executable folder path
+  if (platform === 'macos') {
+    buildPath = path.join(targetBuildDir, fullProductName);
+  }
+
+  // Make sure the directory exists and fail fast vs silently failing
+  if (!fs.existsSync(targetBuildDir)) {
+    throw new CLIError(
+      `target build directory ${targetBuildDir} does not exist`,
+    );
+  }
+
+  return buildPath;
 }
--- a/packages/cli-platform-apple/src/commands/runCommand/__tests__/getBuildPath.test.ts
+++ b/packages/cli-platform-apple/src/commands/runCommand/__tests__/getBuildPath.test.ts
@@ -1,0 +1,97 @@
+import path from 'path';
+import fs from 'fs';
+import {getTempDirectory} from '../../../../../../jest/helpers';
+import {BuildSettings} from '../getBuildSettings';
+import {getBuildPath} from '../getBuildPath';
+
+const targetBuildDirName = 'foo';
+const targetBuildDirNameWithMaccatalyst = `${targetBuildDirName}-maccatalyst`;
+const executableFolderPath = path.join('foo.app', 'Contents', 'MacOS', 'foo');
+
+test('correctly determines macCatalyst build artifact path new style', async () => {
+  // setup:
+  const tmpBuildPath = getTempDirectory('maccatalyst-test-dir');
+  fs.mkdirSync(path.join(tmpBuildPath, targetBuildDirNameWithMaccatalyst), {
+    recursive: true,
+  });
+
+  // - create buildSettings object that represents this to CLI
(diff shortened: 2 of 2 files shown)


mikehardy:
Note: this was against CLI v14.1.0 - I'm currently making sure rn75 works
This may be changed in rn76 RCs - I'll be checking those later

szymonrybczak:
> are macCatalyst builds checked in CI 🤔 ?

Unfortunately no :( 

> Note: this was against CLI v14.1.0 - I'm currently making sure rn75 works
This may be changed in rn76 RCs - I'll be checking those later

Maybe to make this solution more future-proof we'll fallback to other value if initial one fails? 🤔

mikehardy:
I guess it is hard in CI, since you need an Xcode development team to do a macCatalyst build - it was pretty annoying to set up in my build test rig, at least 😅 

Would you like me to alter the patch to actually check existence of the path so it can try both and return the one that works ?

szymonrybczak:
> I guess it is hard in CI, since you need an Xcode development team to do a macCatalyst build - it was pretty annoying to set up in my build test rig, at least 😅

Ah, right 😞

> Would you like me to alter the patch to actually check existence of the path so it can try both and return the one that works ?

Yes, IMO that's the safest solution?

cipolleschi (approved):

mikehardy:
I added a unit test and handle both cases now, force-pushed, and while the idea is roughly the same code is now different so re-requesting review, thanks you all

packages/cli-platform-apple/src/commands/runCommand/getBuildPath.ts
```diff
@@ -24,11 +25,30 @@ export async function getBuildPath(
...
-    return path.join(targetBuildDir, executableFolderPath);
+    // make sure path has one and only one '-maccatalyst' on end
+    if (!targetBuildDir.match(/-maccatalyst$/)) {
+      targetBuildDir = `${targetBuildDir}-maccatalyst`;
+    }
+    buildPath = path.join(targetBuildDir, executableFolderPath);
+  }
+
+  // macOS gets the product name, not the executable folder path
+  if (platform === 'macos') {
```
cipolleschi, line 42: shouldn't this be `if-else if`?

cipolleschi (approved):
left a nit, but lgtm

packages/cli-platform-apple/src/commands/runCommand/getBuildPath.ts
mikehardy: it certainly could be, but...it seemed to me the three cases were all mutually exclusive so simple `if` overrides/clobbers were semantically identical ?

case 1 - we're the default platform == ios && isCatalyst == false:  default case join targetBuildDir + executableFolderPath

case 2 - we're the rare platform == ios && isCatalyst == true: clobber with result of maybe adding `-maccatalyst` to targetBuildDir  -- **note that catalyst is an ios platform subset, it will not go in to the platform == macos conditional as true**

case 3 - we're the platform == macos: clobber join of targetBuildDir + fullProductName

Each of the three cases is unambiguously identifiable and unique so I thought a simple set of ifs + clobbers worked?

cipolleschi: Yes, they were equivalent. But with 
```
if <cond1> {
}
if <cond2> {
}
if <cond3> {
}
```
All three conditions are always evaluated. With 

```
if <cond1> {
} else if <cond2> {
} else if <cond3> {
}
```
cond2 is evaluated if and only if cond1 is false. cond3 is evaluated if and only if both cond1 and cond2 are false.

So, the if-elseif approach is slightly more efficient, and makes it clear that the conditions are mutually exclusive.

cipolleschi (approved):

packages/cli-platform-apple/src/commands/runCommand/getBuildPath.ts
mikehardy: @cipolleschi lots of discussion and we are in agreement with each other :-) - the question is that with the mixed signal I am hearing of "if/elseif/elseif is slightly better" vs "but I have approved the if/if/if version" - do you want me to convert it and re-push or not?

I don't want to swirl endlessly on this - it's a relatively small fix after all, but I'll do what you like (because, it's an easy conversion from if/if/if/ to if/elseif/elsif too).

So just let me know but I won't change+repush unless you ask, so we don't have another approval cycle unless desired

Cheers man

mikehardy: @cipolleschi I'm still carrying this patch around locally - macCatalyst won't work without it - and just saw the stale bot ping it. Just waiting for a final judgement call here:

> I don't want to swirl endlessly on this - it's a relatively small fix after all, but I'll do what you like (because, it's an easy conversion from if/if/if/ to if/elseif/elsif too).
> 
> So just let me know but I won't change+repush unless you ask, so we don't have another approval cycle unless desired

https://github.com/mikehardy/rnfbdemo/blob/7703b147a627d4ae5448863a957508743a6c9100/patches/%40react-native-community%2Bcli-platform-apple%2B15.0.1.patch#L33-L34

I'll also note as a separate thing - I was able to get macos and macCatalyst builds to work for local execution without an Xcode development team! This is a huge advance (for me anyway) and maybe would unblock CI for macCatalyst builds? Trick is to add `CODE_SIGNING_IDENTITY = "";` to the pbxproj file for the build targets:

https://github.com/mikehardy/rnfbdemo/blob/7703b147a627d4ae5448863a957508743a6c9100/make-demo.sh#L293-L297

and 

https://stackoverflow.com/a/59315354/9910298

The development team requirement used to be a big blocker for CI but this gets around it

thymikee:
Thanks @mikehardy!

PR #2521, opened 2024-10-06
refactor: replace strip-ansi with native module

stianjensen:
Inspired by https://github.com/eslint/eslint/pull/18982

This API is built-in since node 16.11. react-native-cli already requires node 18 as of last year.


szymonrybczak (approved):
Thanks!

stianjensen:
Not sure if this windows snapshot diff could be related to this PR or not?

szymonrybczak:
@stianjensen not really, failure on E2E Windows pipeline was caused by other change that landed on `main` branch. This PR should be good to merge 👍

Issue #2221, opened 2023-12-20
React Native Doctor is unable to detect the Android SDK in Windows 10


bogdanbaghiu:
Hello I tried to exec the `npx react-native doctor` and obtain that:
![image](https://github.com/react-native-community/cli/assets/37840937/0e0a5ba0-6cc6-41a0-a509-f9e8385f263a)

I have installed the android-sk 33 and 34, but supported `.b33.0.0`. I need help please!!

The `npx react-native info` get that:
![image](https://github.com/react-native-community/cli/assets/37840937/aede54a3-57cb-4ebf-b630-1df0e7bf6259)
![image](https://github.com/react-native-community/cli/assets/37840937/5272408b-97be-4f86-bbda-7e7470f8856b)

And I installed the latest version (11) from Android SDK Command-line Tools

React native project to test: https://github.com/bogdanbaghiu/test_image_colors

But the check of Android Sdk not pass :(

szymonrybczak:
Hey @bogdanbaghiu, that's because you're using `expo` in your project. So you should use Expo CLI to manage your Expo project, I also see `react-native-windows` inside your `package.json`, if you want to use out of tree platforms you should use prebuild, and use React Native Community CLI.

bogdanbaghiu:
@szymonrybczak Thanks for your answer! You were right that I had the added library in my package.json and after that I have made several changes to my package.json that I invite you to look at and try. I have made some progress but I will tell you what I have been discovering.

I Delete the `node_modules,` `android folder` and `package-lock`. I did an install `npx expo install react-native-image-colors` library, I already had it installed but I did it so that the `yarn.lock` file updates all the dependencies. Then I did an `npm install` that installed all the dependencies in my project. Next I launched the `npx expo run:android` command and it created an android build again. It launched successfully on my device. Finally I told myself that I am going to check if the `npx react-native doctor` command has solved my problem and apparently it still looked the same.

Then I realized that if I delete the android folder from my project I get the following:
![image](https://github.com/react-native-community/cli/assets/37840937/030106ba-4146-4daf-aa69-cd12cc093c97)

I have noticed that the npx expo run:android command generates the android build but for some reason it says that the Android SDK supported has to be .b33.0.0 and the ones I have installed are 33.0.0

Could you check if the same thing happens to you with my project but on your computer? If it doesn't happen to you, could you please try to guide me to know why it tells me that it needs an sdk with .b in front of 33.0.0?

justdan0227:
[...]

PR #2531, opened 2024-10-16
fix: be able to read more complex buildToolsVersion definitions

dloic:
Summary:
---------

As highlighted in https://github.com/react-native-community/cli/issues/2221, the way some projects (like expo ones) define the `buildToolsVersion` is breaking `react-native doctor`. 

```gradle
buildscript {
    ext {
        buildToolsVersion = findProperty('android.buildToolsVersion') ?: '34.0.0'
        minSdkVersion = Integer.parseInt(findProperty('android.minSdkVersion') ?: '23')
   ...
```

leads to `buildToolsVersion => .b34.0.0`.

The PR is slightly tweaking the parsing to account for those cases.

Test Plan:
----------

- A new test has been added to demonstrate code is now able to properly read the `buildToolsVersion`
- Existing tests are already covering the basic ` buildToolsVersion = "a version"` syntax

Checklist
----------

- ~~[ ] Documentation is up to date to reflect these changes.~~ NA

### DIFF
--- a/packages/cli-doctor/src/tools/healthchecks/androidSDK.ts
+++ b/packages/cli-doctor/src/tools/healthchecks/androidSDK.ts
@@ -49,9 +49,9 @@ const getBuildToolsVersion = (projectRoot = ''): string => {
       // Get only the portion of the declaration of `buildToolsVersion`
       .substring(buildToolsVersionIndex)
       .split('\n')[0]
-      // Get only the the value of `buildToolsVersion`
-      .match(/\d|\../g) || []
-  ).join('');
+      // Get only the value of `buildToolsVersion`
+      .match(/\d+\.\d+\.\d+/g) || []
+  ).at(0);
 
   return buildToolsVersion || 'Not Found';
 };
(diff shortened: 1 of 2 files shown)


dloic:
@cipolleschi friendly bump on this one ;) Any feedback?

packages/cli-doctor/src/tools/healthchecks/androidSDK.ts
```diff
@@ -50,8 +50,8 @@ const getBuildToolsVersion = (projectRoot = ''): string => {
       .substring(buildToolsVersionIndex)
       .split('\n')[0]
       // Get only the the value of `buildToolsVersion`
-      .match(/\d|\../g) || []
-  ).join('');
+      .match(/\d+/g) || []
```
cipolleschi, line 53: if the version is always three parts, shouldn't we do `/\d+\.\d+\.\d+/g`? 🤔 
This probably allow to drop the `join` below as well.

cipolleschi (approved):
If it works, it looks ok to me. I suggested a more precise approach, but I'm not super familiar on how the versioning of the Android build tool is implemented.

packages/cli-doctor/src/tools/healthchecks/androidSDK.ts
dloic: Good catch. Build tools versions are `\d+\.\d+\.\d+`. PR updated.

cortinico (approved):
Not a huge fan of us doing those Regex magic but I guess that's at least an improvement from our status quo

thymikee:
Thank you!

PR #2562, opened 2024-12-11
fix: config with --platform flag to work with platforms defined in multiple packages

thymikee:
Summary:
---------

Selecting particular platform wouldn't work for out-of-tree platforms, as extra platforms would be added to the `config.platforms` objects anyway. This PR fixes it and adds tests.

Test Plan:
----------

Added ✅ 

Checklist
----------

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -8,7 +8,7 @@
     "**/build": true
   },
   "editor.codeActionsOnSave": {
-    "source.fixAll.eslint": true
+    "source.fixAll.eslint": "explicit"
   },
   "flow.useNPMPackagedFlow": true,
   "javascript.validate.enable": false,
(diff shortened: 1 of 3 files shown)


.vscode/settings.json
```diff
@@ -8,7 +8,7 @@
     "**/build": true
   },
   "editor.codeActionsOnSave": {
-    "source.fixAll.eslint": true
+    "source.fixAll.eslint": "explicit"
```
szymonrybczak, line 11: nice. vscode was also changing this property for me 👍

szymonrybczak (approved):
solid fix, thank you!

.vscode/settings.json
thymikee: let's keep it then and not fight 😅

PR #2580, opened 2025-01-20
fix: adjust log in install Cocoapods prompt

szymonrybczak:
Summary:
---------

It's also required for running with `run-ios` :)


okwasniewski (approved):
😅

PR #2599, opened 2025-02-18
Improve and clarify error for missing scheme file

friederbluemle:
## Summary

This improves and clarifies the error message printed by React Native CLI when the shared scheme file is missing.

## Problem

I recently attempted to run an app with `yarn ios`, and received an error:

```
error Could not find scheme MyApp. Please make sure the schema you want to run exists. Available schemas are: MyApp'.
```

Subsequently I spent 1h+ digging through the code and local caches etc trying to figure out where that mysterious trailing apostrophe originated from (note, it mentions `MyApp'` is available, but it is looking for `MyApp`). It was nowhere to be found. As it turned out, the issue was with how RN CLI formatted the error.

Additionally, the existing error message makes it sound like the first reference to "MyApp" is directly related to or equivalent to the second reference. That's not the case. The first reference represents the shared scheme **file** it is trying to load, while the second reference is a list of schemes declared in the **project configuration**.

Lastly, the error message (as well as an internal variable in code) contains the term "schema", which should be "scheme". Minor difference, but to be consistent with the Apple/iOS ecosystem only "scheme" should be used.

## Solution

1. Remove misleading `'` (apostrophe) from error message
2. Slightly reword error message to clarify what schemes are declared in project config vs. expected files on file system
3. Replace all usage of "schema" with the proper term "scheme" 

## Test Plan

First, I created a new test file for `getBuildConfigurationFromXcScheme`, ensuring 100% line coverage in the _existing_ file. Then, in two additional commits, I made the changes, ensuring the tests will continue to run. When reviewing the PR commit-by-commit, it will be easiest to see.

### Before

```
error Could not find scheme MyApp. Please make sure the schema you want to run exists. Available schemas are: MyApp'.
```

### After

```
error Could not load the shared scheme for MyApp. Your project configuration includes: MyApp. Please ensure a valid .xcscheme file exists in xcshareddata/xcschemes.
```

### DIFF
--- a/packages/cli-platform-apple/src/tools/__tests__/getBuildConfigurationFromXcScheme.test.ts
+++ b/packages/cli-platform-apple/src/tools/__tests__/getBuildConfigurationFromXcScheme.test.ts
@@ -77,8 +77,8 @@ describe('getBuildConfigurationFromXcScheme', () => {
       );
     } catch (err) {
       const msg = (err as CLIError).message;
-      expect(msg).toContain(`Could not find scheme ${scheme}`);
-      expect(msg).toContain(`Available schemas are: ${scheme}`);
+      expect(msg).toContain(`Could not load the shared scheme for ${scheme}`);
+      expect(msg).toContain(`includes: ${projectInfo.schemes[0]}`);
     }
   });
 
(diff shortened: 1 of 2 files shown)


thymikee (approved):
Thank you!

PR #2600, opened 2025-02-18
feat(0.79): generate codegen before `pod install`

szymonrybczak:
## Summary

Starting from React Native 0.79 `pod install` won't invoke codegen generation logic so we have to do inside CLI.

## Test Plan

`yarn ios` in your project should generate codegen under `ios/build` and install Cocoapods properly.

## Checklist

### DIFF
--- a/packages/cli/src/commands/init/init.ts
+++ b/packages/cli/src/commands/init/init.ts
@@ -281,6 +281,9 @@ async function createFromTemplate({
 
       if (process.platform === 'darwin') {
         const installPodsValue = String(installCocoaPods);
+        const reactNativePath = path.dirname(
+          require.resolve('react-native', {paths: [projectDirectory]}),
+        );
 
         try {
           if (installPodsValue === 'true') {
@@ -288,6 +291,7 @@ async function createFromTemplate({
             await runCodegen({
               root: projectDirectory,
               platform: 'ios',
+              reactNativePath,
             });
             await installPods(loader, {});
             loader.succeed();
@@ -306,6 +310,7 @@ async function createFromTemplate({
               await runCodegen({
                 root: projectDirectory,
                 platform: 'ios',
+                reactNativePath,
               });
               await installPods(loader, {});
               loader.succeed();


packages/cli-config-apple/src/tools/installPods.ts
```diff
@@ -10,27 +10,55 @@ import {
...
 } from '@react-native-community/cli-tools';
 import runBundleInstall from './runBundleInstall';
+import path from 'path';
 
 interface PodInstallOptions {
   skipBundleInstall?: boolean;
   newArchEnabled?: boolean;
   iosFolderPath?: string;
+  platform: string;
+  root: string;
```
cipolleschi, line 20: This should be the location of the `xcodeproj` file. f that's the case, we are good to go.

packages/cli-config-apple/src/tools/installPods.ts
```diff
@@ -10,27 +10,55 @@ import {
...
+    execa.sync('node', [
+      codegenScript,
+      '-p',
+      options.root,
+      '-o',
+      process.cwd(),
+      '-t',
+      options.platform,
+    ]);
+
```
cipolleschi, line 61: i would like to split it in a separate function and not couple it with the runPodInstall function. In this way, when we drop CocoaPods, you'll still have the `runCodegen` function in place and you can just remove the `runPodInstall` function

packages/cli/src/commands/init/init.ts
```diff
@@ -284,7 +284,10 @@ async function createFromTemplate({
         try {
           if (installPodsValue === 'true') {
             didInstallPods = true;
-            await installPods(loader);
+            await installPods(loader, {
+              platform: 'ios',
+              root: projectDirectory,
+            });
```
cipolleschi, line 290: splitting the call will mean that this will become:

```
await runCodegen
await installPods
```

cipolleschi (changes requested):
Left a couple of comments to make the code more maintainable! 
But thank you for implementing this.

packages/cli-config-apple/src/tools/installPods.ts
```diff
@@ -10,27 +10,55 @@ import {
...
   shouldHandleRepoUpdate?: boolean;
   newArchEnabled?: boolean;
 }
 
-async function runPodInstall(loader: Ora, options?: RunPodInstallOptions) {
+async function runPodInstall(loader: Ora, options: RunPodInstallOptions) {
   const shouldHandleRepoUpdate = options?.shouldHandleRepoUpdate || true;
   try {
+    if (fs.existsSync('build')) {
+      fs.rmSync('build', {recursive: true});
```
thymikee, line 34: @cipolleschi can you give us a bit of a context on why we remove build folder?

packages/cli-config-apple/src/tools/installPods.ts
szymonrybczak: I don't think so, `root` is where `package.json` should be - based on this code:

https://github.com/facebook/react-native/blob/e41887e62fcaf5a8d6bfc2817f88298b49d1b7a5/packages/react-native/scripts/codegen/generate-artifacts-executor.js#L995

packages/cli-config-apple/src/tools/installPods.ts
cipolleschi: The build folder is where codegen is generated. There are cases while development where you might end up with a lot of junk in that folder. For example, if a user changes the name of the module in JS and rerun Codegen, they will end up having both the old module and the new one. In most of the cases , that should be negligible. But in some other cases it might cause build issues. 
For example if the user forget to update the include in some file, given that the old file it is still there, the library/app might build but it might not behave as the user expect... and debugging that the problem is in the import is tough.

If we clean and they forget to update the import path, Xcode will tell them that that file does not exist and they can realize that's because codegen changed the filename.

packages/cli/src/commands/init/init.ts
szymonrybczak: ✅

cipolleschi (approved):
thanks for applying the suggestions!

packages/cli-config-apple/src/tools/runCodegen.ts
```diff
@@ -0,0 +1,35 @@
...
+}
+
+async function runCodegen(options: CodegenOptions): Promise<void> {
+  if (fs.existsSync('build')) {
+    fs.rmSync('build', {recursive: true});
+  }
+
+  const reactNativePath = path.dirname(
+    require.resolve('react-native', {paths: [process.cwd()]}),
+  );
```
thymikee, line 17: we could use `reactNativePath` from config

packages/cli-config-apple/src/tools/runCodegen.ts
```diff
@@ -0,0 +1,35 @@
...
+    'scripts',
+    'generate-codegen-artifacts.js',
+  );
+
+  await execa('node', [
+    codegenScript,
+    '-p',
+    options.root,
+    '-o',
+    process.cwd(),
```
thymikee, line 29: is the output always gonna be at cwd?

szymonrybczak: yes, as we're switching to proper directory to run specific commands.

packages/cli-config-apple/src/tools/runCodegen.ts
szymonrybczak: https://github.com/react-native-community/cli/pull/2600/commits/48c4d92593376cdcbf589ac89b652460eea45bca

thymikee (approved):

PR #2609, opened 2025-02-21
fix: remove cocoapods log from `init` command

szymonrybczak:
## Summary

In #2602 we enabled installing Cocoapods by default, so it doesn't make longer sense to log an information how to enable it in `react-native.config.js`


thymikee:
Maybe we could point to how to disable it?

szymonrybczak:
> Maybe we could point to how to disable it?

It's default for now, this logic is validated for quite some time now. FWIW we have informations how to disable it in docs 👍

thymikee (approved):

PR #2629, opened 2025-03-08
fix(apple): read `Gemfile` & fallback to globally installed `pod`

szymonrybczak:
## Summary

Some projects have `Gemfile` in the root directory but for other purposes, so checking if it's presented is not enough. 

This Pull Request adds reading and simple regex to check if `gem cocoapods` is included which should make the workflow more solid.

Also this Pull Requests adds a fallback if for any reason CLI decided to install with `bundle exec pod install` we fail silently and fallback to globally installed `pod`.

## Test Plan

1. `yarn add react-native-svg`
2. Remove `Gemfile` and run `run-ios`
3. Cocoapods should be properly installed.

## Checklist

### DIFF
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -1,0 +1,1 @@
+nodeLinker: node-modules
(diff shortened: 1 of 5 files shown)


szymonrybczak:
cc @tido64 I hope that with this one we'll finally make everything work in your setup 🤞

PR #2642, opened 2025-04-01
fix(macCatalyst): assume a provided UDID is valid

### LINKED ISSUE #2313: Not possible to specify macCatalyst UDID as run target
## Environment

`yarn react-native info` from inside a project (note: `npx react-native info` does not work from outside a project)

```
System:
  OS: macOS 14.3.1
  CPU: (8) arm64 Apple M2
  Memory: 66.75 MB / 16.00 GB
  Shell:
    version: 3.2.57
    path: /bin/bash
Binaries:
  Node:
    version: 18.15.0
    path: /private/var/folders/mm/jkjz89x50k1bgv16lm9wwch80000gn/T/xfs-76f07cda/node
  Yarn:
    version: 4.1.0
    path: /private/var/folders/mm/jkjz89x50k1bgv16lm9wwch80000gn/T/xfs-76f07cda/yarn
  npm:
    version: 10.2.4
    path: ~/.nvm/versions/node/v18.15.0/bin/npm
  Watchman:
    version: 2024.01.22.00
    path: /opt/homebrew/bin/watchman
Managers:
  CocoaPods:
    version: 1.14.3
    path: /Users/mike/.asdf/shims/pod
SDKs:
  iOS SDK:
    Platforms:
      - DriverKit 23.2
      - iOS 17.2
      - macOS 14.2
      - tvOS 17.2
      - visionOS 1.0
      - watchOS 10.2
  Android SDK:
    API Levels:
      - "28"
      - "30"
      - "33"
      - "34"
    Build Tools:
      - 28.0.3
      - 30.0.2
      - 30.0.3
      - 31.0.0
      - 33.0.0
      - 33.0.1
      - 33.0.2
      - 34.0.0
    System Images:
      - android-23 | ARM 64 v8a
      - android-24 | Google Play Intel x86 Atom
      - android-33 | ARM 64 v8a
      - android-33 | Google APIs ATD ARM 64 v8a
      - android-34 | Google APIs ARM 64 v8a
      - android-34 | Google Play ARM 64 v8a
    Android NDK: Not Found
IDEs:
  Android Studio: 2023.1 AI-231.9392.1.2311.11330709
  Xcode:
    version: 15.2/15C500b
    path: /usr/bin/xcodebuild
Languages:
  Java:
    version: 17.0.10
    path: /Library/Java/JavaVirtualMachines/temurin-17.jdk/Contents/Home/bin/javac
  Ruby:
    version: 3.2.2
    path: /Users/mike/.asdf/shims/ruby
npmPackages:
  "@react-native-community/cli": Not Found
  react:
    installed: 18.2.0
    wanted: 18.2.0
  react-native:
    installed: 0.74.0-rc.1
    wanted: 0.74.0-rc.1
  react-native-macos: Not Found
npmGlobalPackages:
  "*react-native*": Not Found
Android:
  hermesEnabled: true
  newArchEnabled: false
iOS:
  hermesEnabled: true
  newArchEnabled: false
```

## Description

As requested [in a comment on a PR](https://github.com/react-native-community/cli/pull/2312#issuecomment-1968349372) I posted here for the CLI - this issue is about the difficulty of specifying macCatalyst as a run target

I have a script that does fully scripted / reproducible build testing for react-native-firebase + react-native, and I want to test the macCatalyst target since we support it.

As far as I know there is a `run-ios` command in the CLI but there is no `run-catalyst` (which makes sense, as it is really a sub-flavor of iOS)

So as far as I know the only way to specify that I want it to be a macCatalyst build is to either specify the device name (Preferred method!) or to specify the UDID of the machine.

Here's the problem: On Apple Silicon machines at least, xsimctl does not return the correct UDID for the available macCatalyst hostname!

This leads to two downstream problems in the CLI: 

- the CLI will refuse to build for the target specified by name since there does not appear to be a device by that name
- even if you find the UDID yourself and request a build to UDID instead of build to name, the CLI will assert that the UDID does not exist and thus will not build

What is the solution? Determine why xsimctl isn't returning the right UDID I suppose but that's upstream, not in our control.

What *I* do in [this patch](https://github.com/mikehardy/rnfbdemo/commit/a4596188a27935e0a521dd52eace2f1b5a4e2343#diff-bc218854a45288de2973d93616e4798ffd0dc0c6056cd6174d44c80d5ed34390R13-R21) is simply accept - with a warning - that if a developer has gone to the trouble to specify a UDID - that the UDID must be valid to try even though it doesn't show up in the device list from xsimctl and just have the CLI build to it anyway.

A little bit ugly but I also set it to catalyst build type at the same time - this part specifically is not great

It works.

And I get the UDID by on purpose requesting a build that will not work from `xcodebuild` so it gives me the list of UDIDs that will work and I parse out the catalyst UDID

> CATALYST_DESTINATION=$(xcodebuild -workspace ios/rnfbdemo.xcworkspace -configuration Debug -scheme rnfbdemo -destination id=7153382A-C92B-5798-BEA3-D82D195F25F8 2>&1|grep macOS|grep Catalyst|head -1 |cut -d':' -f5 |cut -d' ' -f1)

That part is definitely awful but it does work. Why isn't xsimctl working ?

## Reproducible Demo

Run this with a valid XCODE_DEVELOPMENT_TEAM set https://github.com/mikehardy/rnfbdemo/blob/main/make-demo.sh

It should try to build for catalyst. Modify as needed. Remove the patch hunk to see it fail

mikehardy:
## Summary

This patch works around a problem where macCatalyst UDIDs are not correct when fetched from system, so attempting to find them in list of UDIDs always fails and macCatalyst apps will not start

It basically just assumes - but with an informative warning - that if you specify a UDID you know what you're doing and it will try to use the one you specified instead of simply erroring out.

- Fixes #2313

## Test Plan

I've been carrying this patch around forever. It's the only way I can get my build demonstrator for react-native-firebase to correctly build-test and run macCatalyst version of react-native + react-native-firebase

Should be noted this very same build test is the source of a large number of macCatalyst build fixes over the years, it's been quite helpful in upstreaming things to make macCatalyst work, this is just another one

https://github.com/mikehardy/rnfbdemo/blob/0f16f47dc9823a592673d849d981db1fb8a96bf9/patches/%40react-native-community%2Bcli-platform-apple%2B15.0.1.patch#L1-L24

## Checklist

### DIFF
--- a/packages/cli-platform-apple/src/commands/runCommand/createRun.ts
+++ b/packages/cli-platform-apple/src/commands/runCommand/createRun.ts
@@ -272,12 +272,22 @@ const createRun =
     }
 
     if (args.udid) {
-      const device = devices.find((d) => d.udid === args.udid);
+      let device = devices.find((d) => d.udid === args.udid);
       if (!device) {
-        return logger.error(
-          `Could not find a device with udid: "${chalk.bold(
-            args.udid,
-          )}". ${printFoundDevices(devices)}`,
+        // On arm64 machines, the catalyst UDID returned by 'xcrun xctrace list devices' is not correct
+        // xcodebuild will return an error indicating the UDID is unknown and offering a different one
+        // you may obtain it by running xcodebuild with the UDID you think works then parse out the other one from the returned error:
+        // CATALYST_DESTINATION=$(xcodebuild -workspace ios/rnfbdemo.xcworkspace -configuration Debug -scheme rnfbdemo -destination id=7153382A-C92B-5798-BEA3-D82D195F25F8 2>&1|grep macOS|grep Catalyst|head -1 |cut -d':' -f5 |cut -d' ' -f1)
+        //
+        // How to handle the incorrect catalyst UDID?
+        // Assume if a UDID is specified, the user knows what they are doing.
+        // Use the given UDID and force the type, so "catalyst" will launch the app correctly
+        device = {name: 'unknown', udid: args.udid, type: 'catalyst'};
+        logger.warn(
+          `Could not find a device with udid: "${chalk.bold(args.udid)}".`,
+        );
+        logger.warn(
+          'Running with provided udid anyway, and type "catalyst". \'xcodebuild\' command may return error.',
         );
       }
       if (device.type === 'simulator') {


mikehardy:
The CI failure seems unrelated to this PR

cipolleschi (approved):
It makes sense to me.

mikehardy:
@thymikee  Still the last patch I'm carrying on the react-native-firebase build demo and has an approval :-) - curious if it can merge? Only risk is *if* it is a macCatalyst build (already small user population!) *and* they specify a machine UDID (advanced usage, super small user population) *then* if the specified UDID doesn't exist the app start will fail. But who would specify a non-existent UDID *and* not be paying attention to see the failure and understand it?

Seems like an easy win...and allows macCatalyst app starts to actually start...

Cheers

thymikee (commented):
catalyst is only handled in runOnDevice. Can we scope it to that code branch below?

mikehardy:
Hey @thymikee - thanks for the review.

I altered the patch and re-pushed (after rebasing to current main) based on your feedback and confirmed it still works in my demo where I exercise simulator debug+release and macCatalyst

https://github.com/mikehardy/rnfbdemo/commits/main/patches

approach now is to nullish-ly check for device?.type in the main simulator pathway - which should be semantically equivalent to it's unpatched check in effect

...only then do we fall in to the non-simulator branch, and only there do we see if device was undefined and then emit our warning but do the new "trust the user vs returning error" stance and just execute it

```

info A dev server is already running for this project on port 8081.
info Found Xcode workspace "rnfbdemo.xcworkspace"
warn Could not find a device with udid: "00008112-000E45A20C2BC01E".
warn Running with provided udid anyway, and type "catalyst". 'xcodebuild' command may return error.
info Building (using "xcodebuild -workspace rnfbdemo.xcworkspace -configuration Debug -scheme rnfbdemo -destination id=00008112-000E45A20C2BC01E")

```

packages/cli-platform-apple/src/commands/runCommand/createRun.ts
```diff
@@ -310,15 +310,8 @@ const createRun =
     }
 
     if (args.udid) {
-      const device = devices.find((d) => d.udid === args.udid);
-      if (!device) {
-        return logger.error(
-          `Could not find a device with udid: "${chalk.bold(
-            args.udid,
-          )}". ${printFoundDevices(devices)}`,
```
thymikee, line 318: Cna we still print this?

mikehardy: absolutely - thanks again for the review - I think I focused in too much on the macCatalyst case and while I was still emitting a similar warning below, it was without the devices list which I think is really helpful in the general case.

So I put that back and re-pushed. What do you think?

packages/cli-platform-apple/src/commands/runCommand/createRun.ts
```diff
@@ -310,15 +310,17 @@ const createRun =
     }
 
     if (args.udid) {
-      const device = devices.find((d) => d.udid === args.udid);
+      let device = devices.find((d) => d.udid === args.udid);
       if (!device) {
-        return logger.error(
+        // We may be able to recover from this for macCatalyst app starts,
+        // but the general case is this is a problem and device list will help
+        return logger.warn(
```
mikehardy, line 317: "warn" now, not "error" because it may be recoverable (that's my criteria for error vs warn)
But it is still serious and the device list could help people

packages/cli-platform-apple/src/commands/runCommand/createRun.ts
mikehardy: @thymikee re-pushed, hopefully addressing your concern. Cheers 👋

thymikee (approved):
Thank you!

PR #2707, opened 2025-08-27
Fix: Remove hardcoded index when selecting the correct build target

jenskuhrjorgensen:
## Summary

In our app we have some old static lib targets that need to be built before the actual application target. This means that our app target is not the first one in the targets list in the Xcode scheme which in turn means `let selectedTarget = targets[0]` is not sophisticated enough to find the actual app target. 

This PR fixes this by only looking at targets with `WRAPPER_EXTENSION === 'app'` which was also partially done before but just at a later stage.

## Test Plan

You can test this in a vanilla RN project by adding e.g. a "Static" library target:
Screenshot 2025-08-27 at 12 27 16

And then make it the first build target in the default Xcode scheme:
Screenshot 2025-08-27 at 12 26 37

And then running 
```
npx react-native run-ios
```

Before this PR, this would fail because the application target is now put in `targets[1]`.

If you have multiple application targets you can test out the `--target` parameter like this:
```
npx react-native run-ios --target "<YOUR_APP_TARGET>"
```

## Checklist

### DIFF
--- a/packages/cli-platform-apple/src/commands/runCommand/getBuildSettings.ts
+++ b/packages/cli-platform-apple/src/commands/runCommand/getBuildSettings.ts
@@ -36,14 +36,21 @@ export async function getBuildSettings(
 
   const settings = JSON.parse(buildSettings);
 
-  const targets = settings.map(
-    ({target: settingsTarget}: any) => settingsTarget,
-  );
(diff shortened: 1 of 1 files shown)


jenskuhrjorgensen:
Hi @thymikee 
Is there a chance you will pick this up soon? 🙏  Or are you awaiting something from me?

PR #2715, opened 2025-09-16
Return to less intrusive way of starting server in a new window on macOS

### LINKED ISSUE #2696: Metro open in new window instead of tab
## Describe the Feature
On MacOS with iTerm, on the previous version of the CLI, Metro opened in a new tab instead of a new window.
I don't know if I'm alone in this situation, but I found it extremely painful to have so many windows open.

## Possible Implementations
After investigation, this is due to a little change in `@react-native-community/cli-tools/build/startServerInNewWindow.js:103`:
```js
// Old version
return _execa().default.sync('open', ['-a', terminal, launchPackagerScript], procConfig);

// New version
return _execa().default.sync('open', ['-na', terminal, launchPackagerScript], procConfig);
```

The last version have `-na` option instead of `-a` when running `launchPackager.command` that force the new window.

If other people are facing the same problem, we may consider putting an environment variable to override these flags, like `REACT_LAUNCHPACKAGER_OPTIONS="-a"`.
For the moment I use `patch-package` to fix it on my repo.

What do you think about this proposal ?

swrobel:
Fixes #2696. Reverts #2628.

## Summary

A change was made in #2628 to work around a bug in Ghostty by forcing a new _process_ to open in the default terminal app by adding the following option to the macOS `open` command:

`-n  Open a new instance of the application(s) even if one is already running.`

This means that, in iTerm at least (and likely others), an entirely new dock icon will appear, which prevents switching between existing iTerm windows and this new one using the typical `cmd+~` keyboard shortcut.

Furthermore, iTerm has overrides that allow you to open in a tab instead of a window when the command is `open -a`, but it can't do anything to override open `open -na` because of the fact that it launches an entirely separate _process_, not just _window_. (Side note: I would love to understand the rationale behind the fact that metro insists on launching in a new window instead of letting the terminal app in use decide how to handle it per the user's preferences, but I digress).

**This removes the `-n` flag as the Ghostty bug that it added to work around has been [verified fixed](https://github.com/ghostty-org/ghostty/discussions/4839#discussioncomment-14226864) and released in [Ghostty 1.2.0](https://ghostty.org/docs/install/release-notes/1-2-0), out today**

## Test Plan

This reverts to longtime behavior for the CLI when launching the metro server process in a terminal. However, I have tested it with the following, and all seems to behave as expected:

1. `REACT_TERMINAL=iTerm yarn ios`
2. `REACT_TERMINAL=Ghostty yarn ios` (using Ghostty 1.2.0)
3. `REACT_TERMINAL=Terminal yarn ios`

cc @Kamefrede for additional testing

## Checklist

### DIFF
--- a/packages/cli-tools/src/startServerInNewWindow.ts
+++ b/packages/cli-tools/src/startServerInNewWindow.ts
@@ -100,7 +100,7 @@ function startServerInNewWindow(
     try {
       return execa.sync(
         'open',
-        ['-na', terminal, launchPackagerScript],
+        ['-a', terminal, launchPackagerScript],
         procConfig,
       );
     } catch (error) {


robhogan:
> I would love to understand the rationale behind the fact that metro insists on launching in a new window

Metro has no opinion on this, FWIW. For example, Expo doesn't launch Metro in a separate window - it's purely a community CLI choice. Possibly to do with displaying native build progress in the main terminal instead?

(Thanks for this improvement btw)

andidev:
@thymikee any chance to get this released soon? its really messing up the windows/tabs on iterm2 so you cannot switch tabs and work correctly, going crazy here 😂

PR #2727, opened 2025-10-25
feat: autolink native peer dependencies

### LINKED ISSUE #870: RFC: Simplify workflow for native dependencies for libraries

patrickkabwe:
## Summary

Add opt-in support for autolinking native peer dependencies (closes #870). At the moment the CLI only autolinks packages that are listed directly in an app’s `package.json`, so native peers declared by a library remain invisible unless every app developer installs them manually. This change introduces a `dependency.autolinkTransitiveDependencies` flag that a library can set in its `react-native.config.js`. When enabled, the CLI reads the library’s own package.json, discovers its peer dependencies, and links any native peers that are already present in node_modules. 

The approach stays entirely inside the autolinking pipeline unlike PR #2054’s `--dependency-check`, which mutates app dependencies so it avoids surprise installs, CI failures, and extra workflow steps. Once a library opts in, every app automatically benefits. The updates touch the config loader, schema, public types, documentation, and add `loadConfigAsync` test coverage.

## Test Plan

`yarn test packages/cli-config/src/__tests__/index-test.ts
`
## Checklist

### DIFF
--- a/docs/dependencies.md
+++ b/docs/dependencies.md
@@ -23,6 +23,7 @@ The following type describes the configuration of a dependency that can be set u
 
 ```ts
 type DependencyConfig = {
+  autolinkTransitiveDependencies?: boolean;
   platforms: {
     android?: AndroidDependencyParams;
     ios?: IOSDependencyParams;
@@ -36,6 +37,10 @@ type DependencyConfig = {
 ### platforms
 
 A map of specific settings that can be set per platform. The exact shape is always defined by the package that provides given platform.
+
+### autolinkTransitiveDependencies
+
+When set to `true`, the CLI will inspect the dependency's `peerDependencies` and attempt to autolink any peers that are also React Native native modules. The CLI does not install those peers for the user, but they will be linked automatically whenever they are present in `node_modules`. Use this if your library relies on a native peer dependency (for example, `react-native-nitro-text` depending on `react-native-nitro-modules`) and would otherwise require users to manually add that peer.
 
 In most cases, as a library author, you should not need to define any of these.
 
--- a/docs/dependencies.md
+++ b/docs/dependencies.md
@@ -40,7 +40,7 @@ A map of specific settings that can be set per platform. The exact shape is alwa
 
 ### autolinkTransitiveDependencies
 
-When set to `true`, the CLI will inspect the dependency's `peerDependencies` and attempt to autolink any peers that are also React Native native modules. The CLI does not install those peers for the user, but they will be linked automatically whenever they are present in `node_modules`. Use this if your library relies on a native peer dependency (for example, `react-native-nitro-text` depending on `react-native-nitro-modules`) and would otherwise require users to manually add that peer.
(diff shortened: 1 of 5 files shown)


patrickkabwe:
Hi @thymikee any chance for a review on this PR?

thymikee:
Appreciate this, although it will be hard to propagate this changes, especially that [`expo-modules-autolinking`](https://github.com/expo/expo/tree/main/packages/expo-modules-autolinking) exists and supports this use case without requiring libraries to adopt. I'd rather lean towards reusing this package if possible (might require some changes to it, but idk at the moment)

PR #2735, opened 2025-11-12
fix: RCE vulnerability from CVE-2025-11953

thymikee:
## Summary

Continuation of the fix that landed in https://github.com/react-native-community/cli/commit/15089907d1f1301b22c72d7f68846a2ef20df547, that prevents RCE using a spoofed URL with `|` character, such as: https://evil.com?|calc.exe.

cc @633kh4ck @mbaraniak-exodus

### DIFF
--- a/packages/cli-server-api/src/openURLMiddleware.ts
+++ b/packages/cli-server-api/src/openURLMiddleware.ts
@@ -14,7 +14,7 @@ import open from 'open';
 /**
  * Open a URL in the system browser.
  */
-async function openURLMiddleware(
+export async function openURLMiddleware(
   req: IncomingMessage & {
     // Populated by body-parser
     body?: Object;
--- a/packages/cli-server-api/src/__tests__/openURLMiddleware.test.ts
+++ b/packages/cli-server-api/src/__tests__/openURLMiddleware.test.ts
@@ -1,0 +1,145 @@
+import http from 'http';
+import open from 'open';
+import {openURLMiddleware} from '../openURLMiddleware';
+
+jest.mock('open');
+
+describe('openURLMiddleware', () => {
+  let req: http.IncomingMessage & {body?: Object};
+  let res: jest.Mocked<http.ServerResponse>;
+  let next: jest.Mock;
+
+  beforeEach(() => {
+    req = {
+      method: 'POST',
+      body: {},
+    } as any;
+
+    res = {
+      writeHead: jest.fn(),
+      end: jest.fn(),
+    } as any;
+
+    next = jest.fn();
+    jest.clearAllMocks();
+  });
+
+  afterEach(() => {
+    jest.restoreAllMocks();
+  });
+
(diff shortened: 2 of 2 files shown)


mbaraniak-exodus:
@thymikee,
The fix seems reasonable, unless you switch in the future to a new version of `open`, which uses [PS](https://github.com/sindresorhus/open/blob/8ce0f7d081a4f38f8fd67527c8afa25a64ff2914/index.js#L36) underneath. Then you will need escape also `$` `(`, etc.
Be aware of (non-default) delayed expansion, which will make such syntax possible `!VAR!`

packages/cli-server-api/src/openURLMiddleware.ts
```diff
@@ -31,20 +31,37 @@ async function openURLMiddleware(
...
 
-    await open(url);
+    if (parsedUrl.protocol !== 'http:' && parsedUrl.protocol !== 'https:') {
+      res.writeHead(400);
+      res.end('Invalid URL protocol');
+      return;
+    }
+
+    // Reconstruct URL with proper encoding to prevent command injection
+    // The URL constructor doesn't automatically encode special characters like | in query strings,
```
633kh4ck: To be specific, it encodes special characters, but only sets of them in each URL part [^0]. For example, `|` is encoded in [`userinfo`](https://url.spec.whatwg.org/#userinfo-percent-encode-set):
```js
new URL('https://user|:pass@example.com')`
// https://user%7C:pass@example.com/
```

Current implementation double-encodes several characters for that reason; for example, whitespaces:
```js
const parsedUrl = new URL('https://example.com/?#some hash')
// https://example.com/?#some%20hash
const sanitizedUrl = new URL(parsedUrl.origin);
// ...
console.log(sanitizedUrl.href)
// https://example.com/#some%2520hash
```

A simpler approach could be:
```js
const sanitizedUrl = encodeURI(url);
```

[^0]: https://url.spec.whatwg.org/#percent-encoded-bytes

633kh4ck:
For posterity: this is likely still fragile, but better than it was.

On a side note, this can (still) be exploited to exfiltrate some environment variables; possibilities are more limited, though. For example, `https://example.com/?a=%¾TA%` is encoded to `https://example.com/?a=%25%C2%BETA%25` (note `%BETA%`).

m01e-40x:
@thymikee hi, 
may I ask why this fix PR has not been merged into the main branch and the latest release version？

thymikee:
Uh, I wanted to followup with a more robust fix, but then forgot about it. I'll try to prioritize this soon. Maintaining Community CLI is not my primary focus and anyone is free to contribute

huntie:
@thymikee Opened https://github.com/react-native-community/cli/pull/2758 as an alternative. Uses [strict-url-sanitise](https://www.npmjs.com/package/strict-url-sanitise) and continues to cover logic with our own unit tests.

thymikee:
Thank you @huntie, let's move the discussion there!

PR #2778, opened 2026-03-10
fix: always launches fallback simulator

dannyhw:
## Summary

When passing the --simulator flag with either the device name or udid it is ignored and the fallback is used

I believe it is the intended behaviour of the flag to launch with the id you specify but it was not being respected

Heres a video where i first use the current latest version without my change, then I link my local version of the cli and show it working

The udid i specify is for iphone17 so the first time when it launches the iphone16 thats just using the fallback

https://github.com/user-attachments/assets/4f75aea6-9276-41d1-9f1a-69573a0bf4f4

## Test Plan

link to a local project and note that you can pick the simulator now

for example
```
yarn ios --simulator 'iPhone 17'
yarn ios --simulator '32D43F28-6167-488D-A94C-7F988698225A'
yarn ios --udid '32D43F28-6167-488D-A94C-7F988698225A'
```

## Checklist

### DIFF
--- a/packages/cli-platform-apple/src/commands/runCommand/createRun.ts
+++ b/packages/cli-platform-apple/src/commands/runCommand/createRun.ts
@@ -406,11 +406,12 @@ const createRun =
       );
 
       if (!matchedSimulator) {
-        return logger.error(
+        logger.warn(
           `Could not find a simulator with name or UDID: "${pico.bold(
             args.simulator,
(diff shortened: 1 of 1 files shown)


dannyhw:
added a fallback for when the id doesn't match
image

dannyhw:
ci failure seems to be about `react-native-safe-area-context` version changing to 5.7.0

PR #2803, opened 2026-06-04
chore: Replace ora with nanospinner

stianjensen:
## Summary

Swap out ora for nanospinner (^1.2.2) as the terminal spinner library, using nanospinner's API directly without a compatibility wrapper.

ora has 16 transitive deps (one of which being chalk, which was removed as a direct dep of the cli in a previous release). nanospinner is far more modern and lean, with only one transitive dependency.

API changes across call sites:
- .succeed() → .success()
- .fail() → .error()
- loader.text = '...' → loader.update('...')

Add a Jest auto-mock for nanospinner since Jest 26 cannot resolve node: protocol imports that nanospinner uses internally.

## Test Plan

I've ran the entire test suite locally, and made adjustments where needed.

## Checklist

---

This is a breaking change to the doctor plugin API: third-party healthcheck plugins implementing runAutomaticFix receive the loader and call loader.succeed()/loader.fail() on it — those calls will now throw. Same for any external consumer of getLoader/NoopLoader from cli-tools.

I've left it for now, but it would probably either need to be included in a major version change, or I could include a compat layer for a subset, or larger set of function aliases from Ora, as a stepping stone.

### DIFF
--- a/__mocks__/nanospinner.js
+++ b/__mocks__/nanospinner.js
@@ -1,0 +1,21 @@
+function createSpinner() {
+  const spinner = {
+    start: () => spinner,
+    stop: () => spinner,
+    success: () => spinner,
+    error: () => spinner,
+    warn: () => spinner,
+    info: () => spinner,
+    clear: () => spinner,
+    render: () => spinner,
+    update: () => spinner,
+    reset: () => spinner,
+    spin: () => spinner,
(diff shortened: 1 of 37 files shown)


cortinico:
cc @thymikee if you have thoughts about this one

PR #2817, opened 2026-06-26
breaking: Bundle iOS and Android as core platforms

huntie:
## Summary

Bundle `cli-platform-android` and `cli-platform-ios` in the main `@react-native-community/cli` package.

Relates to this proposed change: https://github.com/react-native-community/template/pull/235

**Motivation**

Simplify Community CLI setup for users (see above). iOS and Android are first-class RN platforms.

- While the CLI was historically separated out into granular plugins, I don't believe we need to be so aggressive in splitting the core Android and iOS platforms into separate installs. This optimisation provides little value when RN itself bundles both — match this baseline.
    - Other plugins, including OOT platforms, remain selectively applied via plugin discovery. Again, the redundant weight of ~20 scripts (each way) being omitted for a single platform repo is a marginal win. The convention (in the case of libraries) is that OOT platform support is *additional*.
    - We can always un-bundle later (in a hypothetical project where RN also separates core from platforms).
- **Prior art**: Expo CLI doesn't bother separating platforms — the dev install impact on projects is negligible. No related artifacts leak into the project.

**Notes**

- `loadConfig` and `loadConfigAsync` gain an optional platforms argument that seeds the config. Project and dependency configs still layer on top and can override it, and the `--platform` filter is honoured.

## Test Plan

TODO

## Checklist

### DIFF
--- a/packages/cli-config/src/loadConfig.ts
+++ b/packages/cli-config/src/loadConfig.ts
@@ -85,14 +85,36 @@ const removeDuplicateCommands = <T extends boolean>(commands: Command<T>[]) => {
 };
 
 /**
+ * Returns the built-in platforms to seed the config with, honoring the
+ * `selectedPlatform` filter the same way dependency platforms are filtered. The
+ * project's own and dependencies' platforms are layered on top of these.
+ */
+const getBasePlatforms = (
+  platforms: Config['platforms'] | undefined,
+  selectedPlatform: string | undefined,
+): {[platform: string]: Config['platforms'][string]} => {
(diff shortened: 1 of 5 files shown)


huntie:
@thymikee Ready for review, pending the corresponding changes in core. Aim to land this for the 0.89 cycle (NOTE: CLI should already have branched on 0.88 so this is to `main`→0.89).

thymikee (approved):
Let's explicitly note what's breaking, because it's about observable behavior rather than APIs

PR #2821, opened 2026-07-13
fix(cli-platform-apple): do not re-pass -CurrentDeviceUDID to an already-booted simulator

### LINKED ISSUE #2820: run-ios: "Unable to boot device in current state: Booted" alert when the simulator is already booted
## Environment

- **`@react-native-community/cli`**: `20.2.0` (also reproduces on `main`, latest commit `9013a48` at time of writing)
- **OS**: macOS (Apple Silicon)
- **Xcode**: stable (not Xcode 27 beta) — `Simulator.app` present at `$(xcode-select -p)/Applications/Simulator.app`, so the `run-ios` code path takes the `Simulator.app` branch (not the DeviceHub branch added in #2806)

## Description

Running `run-ios` against a simulator that is **already booted** pops a modal alert from **Simulator.app** reading:

> **Unable to boot device in current state: Booted**

before the build proceeds. The alert must be dismissed manually and it recurs on **every** re-launch until the simulator is shut down. The build itself still succeeds. A cold first launch (no booted device) is unaffected.

### Root cause

In [`packages/cli-platform-apple/src/commands/runCommand/runOnSimulator.ts`](https://github.com/react-native-community/cli/blob/9013a480c66aece0e23cc8082ee23f92274e3cf4/packages/cli-platform-apple/src/commands/runCommand/runOnSimulator.ts#L53-L68), the `open Simulator.app --args -CurrentDeviceUDID <udid>` call runs **unconditionally**, *before* the `simulator.state !== 'Booted'` guard:

```ts
if (fs.existsSync(simulatorApp)) {
  child_process.execFileSync('open', [
    simulatorApp,
    '--args',
    '-CurrentDeviceUDID',
    simulator.udid,
  ]);
} else if (fs.existsSync(deviceHubApp)) {
  ...
}

if (simulator.state !== 'Booted') {
  bootSimulator(simulator);
}
```

Passing `-CurrentDeviceUDID` for an **already-booted** device makes Simulator.app re-attempt a boot of that device, which CoreSimulator rejects with the exact string `Unable to boot device in current state: Booted`. Simulator.app surfaces that rejection as a modal alert.

The later `simulator.state !== 'Booted'` guard correctly skips the explicit `xcrun simctl boot`, but it is too late — the `open ... --args -CurrentDeviceUDID` call has already provoked the dialog. The DeviceHub branch (Xcode 27+) is unaffected, since it never passes `-CurrentDeviceUDID`.

### Fix

Only pass `-CurrentDeviceUDID` when the device is **not** already booted; when it is booted, just bring the running Simulator.app to the foreground. A PR is incoming.

## Reproducible Demo

1. `npx react-native run-ios` — boots a fresh simulator; no alert.
2. Leave the simulator booted (don't shut it down).
3. `npx react-native run-ios` again → the **"Unable to boot device in current state: Booted"** modal appears over Simulator.app.
4. Click OK to dismiss; the build continues normally.
5. Reproduces on every subsequent `run-ios` until the simulator is shut down.

**Expected:** Step 3 should just foreground Simulator.app and install/launch the app — no alert — same as the clean first launch.

bryandent:
## Summary

Fixes #2820.

`runOnSimulator` calls `open Simulator.app --args -CurrentDeviceUDID <udid>` **unconditionally**, before the `simulator.state !== 'Booted'` guard. For a simulator that is **already booted**, passing `-CurrentDeviceUDID` makes Simulator.app re-issue a boot request that CoreSimulator rejects with `Unable to boot device in current state: Booted` — surfaced as a modal alert that must be dismissed on **every** `run-ios` re-launch until the simulator is shut down. A cold first launch is unaffected.

This change only passes `-CurrentDeviceUDID` when the device is **not** already booted. When it is booted, we just bring the already-running Simulator.app to the foreground (`open <Simulator.app>`), which is enough — the target device is already showing. The DeviceHub (Xcode 27+) branch is untouched, since it never passed `-CurrentDeviceUDID`.

```diff
   if (fs.existsSync(simulatorApp)) {
-    child_process.execFileSync('open', [
-      simulatorApp,
-      '--args',
-      '-CurrentDeviceUDID',
-      simulator.udid,
-    ]);
+    child_process.execFileSync(
+      'open',
+      simulator.state === 'Booted'
+        ? [simulatorApp]
+        : [simulatorApp, '--args', '-CurrentDeviceUDID', simulator.udid],
+    );
   } else if (fs.existsSync(deviceHubApp)) {
```

## Test Plan

**Unit tests** (`packages/cli-platform-apple/src/commands/runCommand/__tests__/runOnSimulator.test.ts`):

- The existing test that asserts `-CurrentDeviceUDID` is passed was retargeted to a **not-booted** device (`state: 'Shutdown'`), since that is the case where the flag is still needed (fresh launch selects the right device).
- Added a test for the **already-booted** case: `open` is called with just `[Simulator.app]` and **never** with `-CurrentDeviceUDID`.

```
PASS unit .../runOnSimulator.test.ts
  ✓ opens Simulator.app with the device UDID when it is not already booted
  ✓ opens Simulator.app without -CurrentDeviceUDID when the device is already booted
  ✓ falls back to DeviceHub via devices:// deep link when Simulator.app is absent
  ✓ does not boot the simulator when it is already booted

Tests: 4 passed, 4 total
```

`yarn build` (TS compile of all packages) and `eslint` on the changed files both pass.

**Manual reproduction** (stable Xcode, `Simulator.app` present):

1. `run-ios` — boots a fresh simulator; no alert.
2. Leave it booted; `run-ios` again → **before this change** the "Unable to boot device in current state: Booted" modal appears; **after** it does not.
3. Cold-boot case (`xcrun simctl shutdown all` then `run-ios`) still selects the correct device.

## Checklist

- [x] Documentation is up to date (updated the in-code doc comment explaining why the flag is now conditional; no external docs affected).
- [ ] For functional changes, my test plan has linked these CLI changes into a local `react-native` checkout. (Change is covered by unit tests and is deterministic; happy to run the full local-link smoke test if a maintainer would like it before merge.)


bryandent:
Any update on this?

ahmedu007 (approved):

PR #2830, opened 2026-08-26
fix(android): wait for emulator framework boot

### LINKED ISSUE #2824: run-android fails with "No connected devices!" on a cold-booting emulator (waits for adb, not for sys.boot_completed)
## Environment

`npx react-native info` is not representative here (pnpm monorepo), so the relevant
versions collected by hand:

```
System:
    OS: macOS 26.5.2 (25F84)
    Node: 24.15.0
    Java: openjdk 17.0.20 2026-07-21 LTS
  SDKs:
    Android SDK Platform-Tools: 37.0.0 (adb 1.0.41)
    Android Emulator: 36.6.11.0 (build_id 15507667)
  npmPackages:
    react-native: 0.86.0
    @react-native-community/cli: 20.1.0
    @react-native-community/cli-platform-android: 20.1.0
```

Also reproduces against `main` — `tryLaunchEmulator.ts` is unchanged since
2514405b (Apr 2024), and the two relevant lines are identical in v20.2.0.

## Description

When `run-android` has to launch the emulator itself, and that emulator does a
**genuine cold boot** (no quick-boot snapshot — a fresh AVD, an AVD whose
snapshot was invalidated, `-no-snapshot-load`, or CI), the build fails:

```
error Failed to install the app. Command failed with exit code 1: ./gradlew app:installDevDebug …
com.android.builder.testing.api.DeviceException: No connected devices!
```

with Gradle having logged `Device is still booting` just before.

### Root cause

`launchEmulator` in
[`packages/cli-platform-android/src/commands/runAndroid/tryLaunchEmulator.ts`](https://github.com/react-native-community/cli/blob/main/packages/cli-platform-android/src/commands/runAndroid/tryLaunchEmulator.ts#L43-L52)
treats the emulator as ready the moment `adb devices` lists it:

```ts
const bootCheckInterval = setInterval(async () => {
  const connected = port
    ? devices.find((d) => d.includes(`${port}`))
    : devices.length > 0;
  if (connected) {
    cleanup();
    resolve(true);          // ← too early
  }
}, 1000);
```

`adb.getDevices` only keeps devices whose state is `device`
([adb.ts#L25](https://github.com/react-native-community/cli/blob/main/packages/cli-platform-android/src/commands/runAndroid/adb.ts#L25)),
which is an honest signal — but it is a signal about **adbd**, not about the
Android framework. `adbd` starts early in boot; ActivityManager finishes much
later. `run-android` therefore resolves, hands off to Gradle, and Gradle's
install task refuses a device that is still booting.

### Measurements

Timing one cold boot on the machine above (AVD launched at t=0):

| Event | t |
|---|---|
| `adb devices` first reports state `device` | **9.3 s** |
| `getprop sys.boot_completed` first returns `1` | **20.2 s** |

That is a **~10.9 s window** in which the current check is satisfied and the
device cannot actually be installed to. Anything that fits inside that window —
Gradle configuration being warm, a cached build — loses the race. Anything that
takes longer than 10.9 s (a cold Gradle daemon, a first build) accidentally
survives it, which is why this is so often written off as flaky.

### Suggested fix

Gate on `sys.boot_completed`, which is set once ActivityManager has finished
starting and is therefore a strictly later — and safe — signal:

```
adb -s <serial> shell getprop sys.boot_completed
```

One consequence worth calling out: with that gate in place the existing
`timeout = 30` no longer budgets for "adb answers", it budgets for "the whole
framework is up". On this machine that is 20.2 s of the 30 s, leaving ~10 s of
headroom; slower hardware and CI runners exceed it routinely. The timeout needs
to grow alongside the gate, otherwise the fix converts a wrong-success into a
spurious timeout.

I have this running as a local patch and can open a PR — say the word and I'll
send it. (Happy to make the timeout a flag instead of a constant if you'd
prefer that shape.)

## Reproducible Demo

```sh
npx @react-native-community/cli init BootRace --version 0.86.0
cd BootRace

# make sure no emulator is running, and force a true cold boot
adb devices          # expect: empty
emulator -list-avds  # pick one, then wipe its snapshot:
emulator @<avd> -no-snapshot-save -wipe-data &  # let it boot once, then kill it

npx react-native run-android
```

The failure is timing-sensitive by nature. To see the window directly, without
building anything:

```sh
emulator @<avd> -no-snapshot-load &
START=$(date +%s)
while ! adb devices | grep -q "device$"; do sleep 0.1; done
echo "adb says 'device' after $(( $(date +%s) - START ))s"
while [ "$(adb shell getprop sys.boot_completed 2>/dev/null | tr -d '\r')" != "1" ]; do sleep 0.1; done
echo "sys.boot_completed after $(( $(date +%s) - START ))s"
```

Every second between those two lines is a second in which `run-android` believes
the emulator is ready and Gradle disagrees.

huytdps13400:
## Summary

- keep polling after `adb devices` reports the launched emulator until that serial returns `sys.boot_completed=1`
- use the connected serial selected by the requested emulator port for the framework readiness check
- increase the cold-boot timeout from 30 seconds to 120 seconds so slower machines and CI can finish booting

Fixes #2824

## Test Plan

- RED: the new fake-timer tests resolved as soon as adbd appeared and timed out after only 30 seconds before the implementation change
- `yarn test packages/cli-platform-android/src/commands/runAndroid/__tests__/tryLaunchEmulator.test.ts --runInBand` (2 passed)
- `yarn test packages/cli-platform-android --runInBand` (5 suites, 19 passed)
- `yarn test:ci:unit` (56 suites, 315 passed, 1 pre-existing todo)
- `yarn lint` (0 errors; 2 pre-existing warnings outside this diff)
- `yarn build`
- `git diff --check`

All commands above ran with Node 22.22.2, matching the repository CI matrix.

## Checklist

- [x] Documentation is up to date; this changes the internal readiness signal without changing CLI options.
- [ ] For functional changes, my test plan has linked these CLI changes into a local `react-native` checkout. The cold-boot timing and timeout behavior are covered deterministically with mocked ADB responses and fake timers instead.


cortinico (approved):

huytdps13400:
Thanks for the approval. I checked the failed Node 20 Ubuntu E2E job. It is unrelated to the Android emulator boot change:

- the only asserted failure is `__e2e__/config.test.ts` because the generated config now contains the base branch's `spm [action]` command while the snapshot does not
- the log also contains existing E2E fixture setup failures from `yarn link .../__e2e__ --all` and ESM config fixtures using `require`
- lint and both Node 20/22 unit suites pass

The other E2E matrix jobs were cancelled after this fail-fast result. There is no Android/emulator failure in the run and no PR code change is indicated; a rerun after the base E2E snapshot/fixture is corrected should be sufficient.

PR #2844, opened 2026-08-28
fix(android): reject invalid manifest XML before parsing

OskarEichler:
## Summary

`XMLValidator.validate()` returns either `true` or an error object. The previous truthiness check accepted the error object, so malformed manifests could be parsed as if validation succeeded. Require an actual `true` result.

The existing multi-activity fixture also contained an unmatched `</queries>` tag. This PR corrects that fixture and adds a dedicated malformed-manifest regression.

## Compatibility

Malformed manifests now return `null` instead of occasionally yielding a launcher from partially parsed XML. Valid manifests retain their existing behavior. No dependency or SDK version changes.

## Verification

- `yarn build`
- `yarn test packages/cli-config-android/src/config/__tests__/getMainActivity.test.ts --runInBand` (6 tests passed)
- Focused ESLint on the implementation and test
- Focused Prettier check on the TypeScript files
- `git diff --check`

Type: fix


OskarEichler:
Resolved the draft blocker: I removed the stray unmatched `</queries>` from the shared multi-activity fixture and added an explicit malformed-manifest regression. The focused suite now passes all 6 tests, and the full package build plus focused lint/format checks are clean. I’ve marked the PR ready for review.

PR #2858, opened 2026-09-14
feat: detect Swift Package Manager projects

### LINKED ISSUE #2856: Support existing CLI apple project use cases without relying on podfile existing on the disk (projects using SPM)
## Describe the Feature
Since RN 0.87 there's an experimental SPM support introduced

Right now when project is migrated from Cocoapods to SPM, all the commands that rely on the `getProjectConfig` from `@react-native-community/cli-config-apple` (e.g. `react-native spm`) will fail once the `Podfile` is removed from its default location. Having dead `Podfile` just to keep the `config` command work doesn't make any sense

## Possible Implementations

looking at where `findPodfilePath` or `findPodfilePaths` are used, there're 2 internal use cases

- `https://github.com/react-native-community/cli/blob/9013a480c66aece0e23cc8082ee23f92274e3cf4/packages/cli-config-apple/src/config/index.ts#L36` (`getProjectConfig`)
- `https://github.com/react-native-community/cli/blob/9013a480c66aece0e23cc8082ee23f92274e3cf4/packages/cli-doctor/src/tools/healthchecks/xcodeEnv.ts#L33` (relies on the project config)

One idea would be to at least prioritize user config (`project.ios.sourceDir` from `react-native.config.js`) and use it instead of always returning `null` config when there's no `Podfile`. That way the amount of changes in CLI will be minimal and SPM users will be able to just set the value to e.g. `ios` and to drop the `Podfile`.

Another idea would be to look for `.spm-injected.json` within the candidate directory (it's a file injected via `react-native spm` command) - if it's found, it means that its parent directory is in fact the one that consists `.pbxproj` file and the directory with swift/objc source code.

Last one I can think of is to look for `.xcode.env` file, which is afaik should exist in each RN project next to `Podfile`

The resolving order should be like this: (user-config) -> (.spm-injected.json OR .xcode.env) -> (Podfile) - if none of those exist only then return `null` for the project config value

thymikee:
## Summary

Fixes #2856

React Native 0.87 ships `react-native spm`, which migrates the Xcode project to Swift Package Manager, strips React Native out of the `Podfile` and leaves the file on disk. We treated those projects as broken: with no `Podfile` found, `getProjectConfig` bailed out, so `config` had no `project.ios.sourceDir` and `run-ios`/`build-ios` insisted on installing pods.

- `findSpmProjectDir` looks for the `.xcodeproj` containing `.spm-injected.json` — the marker `react-native spm` writes inside the project bundle, and `react-native spm deinit` removes. It wins over a leftover `Podfile`, since `react-native spm` refuses to migrate a CocoaPods-integrated project anyway.
- `project.ios.buildSystem` is `spm` or `cocoapods`, so commands know where React Native comes from.
- `resolvePods` returns whether pods were installed, instead of both callers setting `installedPods = true` next to it, and skips CocoaPods for `spm` projects. `--force-pods` / `--only-pods` still run it, which is how pods unrelated to React Native keep working next to Swift Package Manager. Forcing pods where there is no `Podfile` now errors with a hint to run `react-native spm`.

Libraries still need a `Package.swift`, which is RN's `react-native spm scaffold` + `patch-package` story. Nothing to configure in the CLI anymore — no `project.ios.sourceDir`, no `automaticPodsInstallation: false`, no stub `Podfile`, e.g. the ones needed in [bottom-tabs' SPM example](https://github.com/callstack/react-native-bottom-tabs/pull/575).

## Test Plan

```
yarn test packages/cli-config-apple
```

Verified against a real 0.87.1 app migrated with `node node_modules/react-native/scripts/setup-apple-spm.js add --deintegrate --yes`, with this build linked into `node_modules`:

```
$ npx react-native config | jq '.project.ios | {sourceDir, buildSystem}'
{
  "sourceDir": "/tmp/spm-proto/SpmApp/ios",
  "buildSystem": "spm"
}

$ npx react-native build-ios
success Successfully built the app

$ npx react-native run-ios --simulator "Apex CI Smoke" --port 8090
success Successfully built the app
success Successfully launched the app
```

Debug and Release both build, install, launch and render the app; no CocoaPods step runs. RN's own `generate-spm-autolinking-config.js` succeeds against this `config` output — before the change it failed with `CLI config did not provide project.ios.sourceDir`. A CocoaPods project's output is identical apart from `buildSystem: "cocoapods"`.

The e2e config snapshot doesn't see the new field: that test `init`s a project, which pulls `@react-native-community/cli-platform-ios` from the React Native template, so the iOS project config comes from the published CLI rather than this checkout. Its only update here is RN's new `spm [action]` command, which landed with 0.87 and is pre-existing drift rather than something from this change.

One unrelated thing I tripped over: `run-ios --port` starts the dev server on that port, but `installApp` launches the app with a bare `simctl launch`, so the app still asks the packager on 8081. I had to launch with `-RCT_jsLocation localhost:8090` to load my bundle. Worth a separate issue.

## Checklist


mateusz1913:
Works for a sample app

```
  "project": {
    "ios": {
      "sourceDir": "/Users/mateusz.medrek/work/rn/TestSample087/ios",
      "xcodeProject": {
        "name": "TestSample087.xcodeproj",
        "path": ".",
        "isWorkspace": false
      },
      "automaticPodsInstallation": false,
      "assets": [],
      "buildSystem": "spm"
    },
    "android": {
      "sourceDir": "/Users/mateusz.medrek/work/rn/TestSample087/android",
      "appName": "app",
      "packageName": "com.testsample087",
      "applicationId": "com.testsample087",
      "mainActivity": ".MainActivity",
      "assets": []
    }
  }
```

